PR #356, opened 2013-12-03
Add PNG support to plBitmap and plMipmap.

Deledrius:
Using the existing PNG library support, this allows PNG-compressed images as plBitmaps and plMipmaps


Hoikas:
:+1: Once this is rebased on top of ST, it will be ready to merge.

PR #376, opened 2013-12-27
NetCore DNS

dpogue:
:boom: It begins...
- Start of a new NetCore system. Currently only handling DNS lookups using C-Ares.

Some things for discussion:
- Since this netcore system isn't going to be ready any time soon, should it be conditionally enabled by a CMake flag?
- Concerns around adding C-Ares as a dependency?
- Error reporting/handling from async network threads?


zrax:
Closed by #835

PR #448, opened 2014-10-11
Store plClient password using platform-specific credential storage.

Deledrius:
Add an abstracted password-storage system, and utilize it in plClient to replace the drive-based obfuscation.


Deledrius:
Updated to include all suggestions, except refactoring the account-name storage.  Do we have a preference on where this should be moved to?

The remaining decision is regarding where to store the preference to save the username and password, and if so, where to store the username.  This information is used as the key to retrieve the password if saved.

I chose to add the extra code storing this information in the registry as it's the preferred location on Windows, the `login.dat` file is deprecated under the new pfPasswordStore, there are no appropriate console commands (adding new ones for this seemed hackish), and the INI parser currently in the code appears to require significant custom code to support a new file.

The reason I'm reluctant to merge the username storage into the pfPasswordStore is because this feels more like generic preference information, something which we might want to expand to include settings currently stored across the various ini files which litter the init folder.

Sources/Plasma/Apps/plClient/winmain.cpp
```diff
@@ -766,114 +765,89 @@ BOOL CALLBACK UruTOSDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM l
...
 
-
-static void LoadUserPass (LoginDialogParam *pLoginParam)
+static void LoadUserPass(LoginDialogParam *pLoginParam)
 {
-    uint32_t cryptKey[4];
-    ZeroMemory(cryptKey, sizeof(cryptKey));
-    GetCryptKey(cryptKey, arrsize(cryptKey));
+    HKEY hKey;
+    char accountName[kMaxAccountNameLength];
```
Hoikas: Random crap from the stack can appear in the login dialog

Deledrius: Whoops!  Thanks.

PR #469, opened 2015-01-04
Fix crash when user tries to add a clothing item which doesn't exist.

Deledrius:
Check whether we got anything back from FindItemByName instead of just handing off an empty item to AddItemsToCloset.


zrax:
:+1:

PR #470, opened 2015-01-04
Fix CMake 3.1 warning regarding quoted arguments in conditionals.

Deledrius:
Fixes the warning encountered when generating build files with CMake 3.1.

See [CMP0054](http://www.cmake.org/cmake/help/v3.1/policy/CMP0054.html) for more information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 project(Plasma)
 cmake_minimum_required(VERSION 2.8.11.2)
+cmake_policy(SET CMP0054 NEW)
 
 # Set up Product Identification parameters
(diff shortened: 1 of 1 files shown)


zrax:
Hmm, if I understand the policy correctly, I think I'd rather see us set CMP0054 to NEW rather than make a (not necessarily correct) hack in order to shut it up.  Either way, I don't believe this policy will have a direct impact on this line (either OLD or NEW), since there's probably no variable named `.*Clang` ;)

Deledrius:
Updated to set the policy behavior instead of changing the `if` statement.

zrax:
:+1:

PR #471, opened 2015-01-10
GCC fixes

zrax:
Some fixes for compiling recent additions against GCC and Clang, as well as older versions of CMake.


dpogue:
:+1: from me

PR #477, opened 2015-06-07
Ensure uuid-dev is present when building on Linux

zrax:


Hoikas:
Wow... there is no libuuid finder in cmake?

dpogue:
:+1: from me

PR #482, opened 2015-06-29
Fix some moar Coverity

Hoikas:
Fixes some issues pointed out by Coverity. The biggest change here is that the Max plugin will now produce lightmaps correctly :smile:.


zrax:
:+1: 
I wonder if older versions of MSVC handled `const` differently in virtual function overrides...

dpogue:
:+1:

Hoikas:
That's what I was thinking, @zrax.

branan:
I would not at all be surprised if old versions of MSVC did not correctly deal with `const` virtual function overrides

branan:
Did you double-check that the constness is the same across all versions of the Max SDK? I also would not be surprised if we need to #ifdef this...

Hoikas:
Zrax uses the 2012 SDK on covering, and I saw a mismatch on max 7,which was
good enough for me

PR #484, opened 2015-07-04
Clean up some incorrect uses of operator delete

zrax:
This fixes some Coverity defects (which Clang also warns about)


dpogue:
:+1: Fixes all the warnings in clang

Hoikas:
Looks good on read-through

PR #485, opened 2015-07-05
Fixes and small cleanups

dpogue:
Mostly removing unused code and fixing clang warnings. The plStringStream fix from @zrax is needed to make plSDL build on Linux.


zrax:
:+1: 
I didn't runtime-test the `plDrawSpanPair` marshalling fix though -- someone who has a better idea of where those are likely to show up may want to do that.

dpogue:
The `plDrawSpanPair` stuff was from my plGLPipeline branch, and I did see that code being called several times while the scene was sorted.

PR #486, opened 2015-07-05
Make hsAssert useful on Linux / Mac

zrax:
Previously, `hsAssert()` would be completely ignored on anything other than Windows, even when `HS_DEBUGGING` was properly defined.  This makes it actually print the error message, and then abort afterwards for `HS_DEBUGGING` mode.  Release mode will still silently ignore asserts as before.

EDIT: Also included @dpogue's `DebugIsDebuggerPresent` fix for Linux, and a `DebugBreak` equivalent for all Unixes.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.cpp
+++ b/Sources/Plasma/CoreLib/HeadSpin.cpp
@@ -46,6 +46,13 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #ifdef _MSC_VER
 #   include <crtdbg.h>
 #endif
+
+#if defined(HS_DEBUGGING) && defined(HS_BUILD_FOR_LINUX)
+#   include <cstring>
+#   include <sys/stat.h>
+#   include <fcntl.h>
+#   include <unistd.h>
+#endif
 #pragma hdrstop
 
 #include "hsTemplates.h"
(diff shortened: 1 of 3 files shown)


Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -46,6 +46,18 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#if defined(HS_DEBUGGING) && defined(HS_BUILD_FOR_LINUX)
+#   include <cstring>
+#   include <sys/stat.h>
+#   include <fcntl.h>
+#   include <unistd.h>
+#endif
+
+#if defined(HS_BUILD_FOR_UNIX)
+#   include <signal.h>
+#endif
```
Hoikas: Any includes above `#pragma hdrstop` should be added to Pch.h

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -136,8 +154,27 @@ void ErrorAssert(int line, const char* file, const char* fmt, ...)
...
+    // From http://google-perftools.googlecode.com/svn/trunk/src/heap-checker.cc
+    char buf[256];   // TracerPid comes relatively earlier in status output
+    int fd = open("/proc/self/status", O_RDONLY);
+    if (fd == -1) {
+        return false;  // Can't tell for sure.
+    }
+    const int len = read(fd, buf, sizeof(buf));
+    bool rc = false;
+    if (len > 0) {
+        const char *const kTracerPid = "TracerPid:\t";
```
Hoikas: inconsistent... `const char* const`

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -136,8 +154,27 @@ void ErrorAssert(int line, const char* file, const char* fmt, ...)
...
+        buf[len - 1] = '\0';
+        const char *p = strstr(buf, kTracerPid);
```
Hoikas: `const char*`

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -136,8 +154,27 @@ void ErrorAssert(int line, const char* file, const char* fmt, ...)
...
+        if (p != NULL) {
```
Hoikas: `nullptr`

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -118,17 +130,23 @@ void ErrorAssert(int line, const char* file, const char* fmt, ...)
...
-#endif // HS_DEBUGGING
-      if (DebugIsDebuggerPresent()) {
+#endif // _MSC_VER
+    {
         char str[] = "-------\nASSERTION FAILED:\nFile: %s   Line: %i\nMessage: %s\n-------";
         DebugMsg(str, file, line, msg);
+        DebugBreakIfDebuggerPresent();
+
+        // In case the debug trap is ignored
+        abort();
```
dpogue: This makes it impossible to step past a failing assertion even with the debugger. Is that intentional? I'm not totally clear on what Visual Studio debugger does in this case.

zrax: That's a good point...  Right now it would behave differently in MSVC vs Unix -- I'll revise it to allow proceeding from a debugger, but continue to abort if no debugger is attached.

dpogue:
:+1: from me

PR #490, opened 2015-07-07
Allow internal clients to start in any age

Hoikas:
Usage: plClient.exe /PlayerId=[kinum] /Age=[agefilename]


zrax:
This should help testing quite a bit :+1:

Deledrius:
:+1: Works great!
Specifying a PlayerId but no Age doesn't bypass the StartUp as I expected.  It would be nice behaviour to have, but that would be a larger change than this (which allows access to existing commands through the shortcut).

Hoikas:
Hmm... It shouldn't be too difficult to go to Personal if the player ID is
specified, but not the age name. I'll look into it tomorrow

Deledrius:
I think the logic for that is currently all handled in the stupStartUp.py script, so it may not be as easy to do (that's where my HomeAge additions were added).

Deledrius:
As discussed in IRC, any new functionality can be added in later work.  This looks good.

PR #492, opened 2015-07-12
String tests

zrax:
This PR:
- Significantly extend the coverage of the plString tests
- Adds tests for plFormat
- Fixes several bugs discovered while writing tests in both plString and plFormat

It currently passes all tests on:
- MSVC 2013
- GCC 5.1 (Arch x86_64)
- Clang 3.6 (Arch x86_64, using libstdc++ since GTest 1.6 doesn't work with libc++)
- (EDIT) Clang 3.6 with libc++ works with GTest 1.7.  I'll submit that as a separate PR


dpogue:
:+1:

Hoikas:
Looks good on a read-through

PR #494, opened 2015-07-12
OSX fixes

zrax:
It doesn't make everything build on Mac yet, but it at least gets the tools and tests fully building, and gets a bit further along with the rest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,14 +69,16 @@ endif()
 
 include(PrecompiledHeader)    #Precompiled Header helper macros
 
-# libCurl isn't smart enough to detect this for us, so we have to configure it ourselves
-option(CURL_IS_STATIC "Using the static version of libcurl?" ON)
-if(CURL_IS_STATIC)
-    add_definitions(-DCURL_STATICLIB)
-endif(CURL_IS_STATIC)
+if(WIN32)
+    # libCurl isn't smart enough to detect this for us, so we have to configure it ourselves
+    option(CURL_IS_STATIC "Using the static version of libcurl?" ON)
+    if(CURL_IS_STATIC)
+        add_definitions(-DCURL_STATICLIB)
+    endif(CURL_IS_STATIC)
 
-# Same story for expat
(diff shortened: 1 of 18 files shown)


CMakeLists.txt
```diff
@@ -69,14 +69,16 @@ endif()
...
 
-# libCurl isn't smart enough to detect this for us, so we have to configure it ourselves
-option(CURL_IS_STATIC "Using the static version of libcurl?" ON)
-if(CURL_IS_STATIC)
-    add_definitions(-DCURL_STATICLIB)
-endif(CURL_IS_STATIC)
-
-# Same story for expat
-option(EXPAT_IS_STATIC "Using the static version of libexpat" ON)
+if(WIN32)
```
Hoikas: ???

Sources/Plasma/FeatureLib/pfLocalizationMgr/CMakeLists.txt
```diff
@@ -3,9 +3,12 @@ include_directories(../../PubUtilLib)
 
 include_directories(${EXPAT_INCLUDE_DIR})
 
-add_definitions(-DWIN32)
-if(EXPAT_IS_STATIC)
-    add_definitions(-DXML_STATIC)
+if(WIN32)
```
Hoikas: ???

CMakeLists.txt
zrax: If you read the commit comment, the CURL_IS_STATIC and EXPAT_IS_STATIC flags are meaningless outside of WIN32 (i.e. they only control whether to use `__declspec(dllimport)` / `__declspec(dllexport)`).  I discovered that I've been leaving them set to the default (ON) for ages even though my system libs are not static.

Sources/Plasma/FeatureLib/pfLocalizationMgr/CMakeLists.txt
zrax: See above

Hoikas:
Seems fine to be :+1:

PR #500, opened 2015-07-27
Convert NetCommPlayer to std::vector

Hoikas:
This fixes a crash observed when creating a new avatar. `ARRAY` and friends do not run constructors when the `New` method is used. `ARRAY` is evil anyway, so here's my fix.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.h
+++ b/Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.h
@@ -76,6 +76,11 @@ struct NetCommPlayer {
     plString    playerName;
     plString    avatarDatasetName;
     unsigned    explorer;
+
+    NetCommPlayer() { }
+    NetCommPlayer(unsigned id, const plString& name, const plString& shape, unsigned ex)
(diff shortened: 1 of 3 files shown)


Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.h
```diff
@@ -76,6 +76,11 @@ struct NetCommPlayer {
     plString    playerName;
     plString    avatarDatasetName;
     unsigned    explorer;
+
+    NetCommPlayer() { }
+    NetCommPlayer(unsigned id, const plString& name, const plString& shape, unsigned ex)
+        : playerInt(id), playerName(name), avatarDatasetName(shape), explorer(ex)
```
zrax: The `emplace_back` probably would work with a braced initializer list without the constructor, I think...  Not that having a constructor is bad, just FYI ;)

Hoikas: I tried doing just that on something in VC++2013 awhile back and it didn't like the initializer list :cry:

zrax: :cry:

PR #503, opened 2015-07-28
Fix Jittering Animations

Hoikas:
Casting the floating point time to an integer causes issues with boundary frame comparisons. So, actually use floats. This makes animations smoother than a baby's arse once again :smile:


Deledrius:
:+1: This is terrific!

PR #507, opened 2015-07-31
Really Fast Client Init

Hoikas:
This moves plClient's init into a thread such that there should no longer be any need to waste time with the "Starting URU... Please wait." dialog under normal usage. Please note that if your logic succeeds before the client is ready, we will still pop up the aforementioned dialog until the client is ready.


dpogue:
Missing plClientLoader.cpp?

Hoikas:
Missing files added

Hoikas:
I believe I have all the corner cases worked out now. You should now be able to use full screen and change graphics settings.

PR #509, opened 2015-08-02
Revert 507

Hoikas:
This pull request reverts #507, which breaks full screen mode.


Deledrius:
:+1: Sad but necessary, for the moment.

PR #510, opened 2015-08-03
Fix sound debug

Hoikas:
Some fixes and additions to the audio console commands that I'm using to debug why the heek table sound is so loud. Previously, the `Audio.NextDebugPlate` command did not work. I also added `Audio.ShowDebugPlate` so that we wouldn't have to step through tons of useless crap before we get to the one we want.


zrax:
:+1:

Deledrius:
:+1:

PR #511, opened 2015-08-03
Exception stuff

zrax:
This makes use of some exception information where it was previously being ignored, and derives HeadSpin exceptions from `std::exception`.  It also adds compiler feature detection for `noexcept`, `override`, and `final`, and makes use of `noexcept` in a few places where it may be useful (mostly the hsExceptions and plString)


Sources/Plasma/PubUtilLib/plSDL/plSDLMgr.cpp
```diff
@@ -191,7 +191,19 @@ int plSDLMgr::Read(hsStream* s, plSDL::DescriptorList* dl)
...
-    catch(...)
+    catch (std::exception &e)
+    {
+        if (fNetApp)
+        {
+            hsLogEntry(fNetApp->DebugMsg("Something bad happened while reading SDLMgr data: %s", e.what()));
+        }
+        else
+        {
+            printf("Something bad happened while reading SDLMgr data: %s", e.what());
```
Hoikas: Where does this message go?

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -130,6 +130,22 @@ typedef int32_t   hsError;
 #   define hsDeprecated(message)
 #endif
 
+#ifdef HAVE_OVERRIDE
+#   define HS_OVERRIDE  override
```
Hoikas: This has been around since MSVC++2012. Do we actually have to test for it?

zrax: We've specified our minimum MSVC compiler, but I don't know what we're targetting for GCC now that it has better support.  IIRC, GCC only added override/final in version 4.7

PR #513, opened 2015-08-07
Fix weird camera angles caused by 9c49b57

Hoikas:
The reason for the previous commit was that coverity indicated view was being passed to up and vice versa. I don't know what they were intending, but the calculation was correct, just the argument naming was wrong. This has all been corrected now.

Test cases: Neighborhood classroom, Palace Alcove Nexus Terminal


dpogue:
:+1:

PR #514, opened 2015-08-07
Allow PXPhysicals to change scene nodes

Hoikas:
If this isn't done, the client will crash on exit if markers have been loaded at some point in the game.


dpogue:
:+1:

Issue #312, opened 2013-02-18
Deprecate the GameMgr


Hoikas:
The game manager is not supported by next-gen FOSS servers. Remove it from the client.

This is achieved by #516.

Hoikas:
All games that use the game manager have been rewritten to not do so -- PRs are now open. A PR to obliterate the game manager from the engine is forthcoming.

PR #515, opened 2015-08-08
Vault marker engine changes

Hoikas:
This is part of #312, it represents all engine changes needed for storage of marker games in the vault. Please note that merging this before the python pull request will break marker games.


Deledrius:
:+1: 
Code looks good, and it functions well on the staging server.

Hoikas:
This PR has been updated to fix compile issues introduced by #519.

PR #517, opened 2015-08-08
Fix loud sound issue

Hoikas:
This pull request fixes problems with certain sounds being entirely too loud, for example, the ayoheek drone sound. This reverts commit b3d6afc1c1436f94c7a1591f9aa156dd49948712.

As best as I can remember, that commit was originally intended to fix some crashing issues present in transfused versions of Doobes' Guild of Messengers Pub. We determined that sounds exported in PyPRP ages would cause issues if they were shorter than 4 seconds in length. The real problem here was PyPRP bugs. For future reference: PyPRP should only be pointed to `.wav` files (not `.ogg`). Anything 4 seconds or shorter MUST be a plWin32StaticSound.


dpogue:
:+1:

PR #522, opened 2016-01-10
Missed some overrides in the creatable macros :(

dpogue:
/cc @zrax


zrax:
Haven't tested, but looks :+1:

PR #526, opened 2016-01-25
Remove redundant in_set helper and replace both with strchr

zrax:


zrax:
Discarding in favor of #529

PR #527, opened 2016-02-21
Fixes circular dependency between pnNucleusInc and pnTimer

Lunanne:
Moved plTimerCallBackManager.h and pnTimerCreatable.h to pnNucleusInc so the dependency is only from pnTimer to pnNucleusInc.

Could still log into Uru so nothing appears to be broken.


Sources/Plasma/NucleusLib/inc/CMakeLists.txt
```diff
@@ -23,8 +23,10 @@ set(pnNucleusInc_HEADERS
     plProfile.h
     plProfileManager.h
     plRefFlags.h
+	plTimerCallbackManager.h
```
Hoikas: illegal tab character

Sources/Plasma/NucleusLib/inc/CMakeLists.txt
```diff
@@ -23,8 +23,10 @@ set(pnNucleusInc_HEADERS
...
     pnAllCreatables.h
     pnNucleusCreatables.h
+	pnTimerCreatable.h
```
Hoikas: see above

Deledrius:
@Lunanne, can you fix the whitespace issue blocking this from being merged?

Lunanne:
Yeah of course, sorry I hadn't noticed Hoikas made comments.

Issue #480, opened 2015-06-10
Update all logging to ST::format/ST::string


dpogue:
There's a lot of LogMsgV and LogMsgF right now that should all be updated to use ST::format.

A start of that work (by making plLoggable raise deprecation warnings) is here:
https://github.com/H-uru/Plasma/compare/master...dpogue:plLoggable

PR #528, opened 2016-04-03
String theory preparation

zrax:
This prepares for an eventual String Theory port by removing the deprecated parts of plString and replacing them with ST-compatible equivalents.  The actual ST port will be larger and more mind-numbing, so it would be good to ensure this part works properly first.


dpogue:
This probably also resolves #480.

Michellacoste:
Is this going to solve the display problem of special foreign language characters (such as é, è, â, etc ...)?

zrax:
Issues with display of non-ASCII characters should be in the font rendering code... This PR shouldn't affect the font rendering

Deledrius:
Finally had some time to read through this monster.  I don't see anything out of order here. :+1:

PR #529, opened 2016-04-11
String theory port

zrax:
Port to string_theory string library.  Depends on #528.


zrax:
Also makes #526 irrelevant

Hoikas:
Did we ever get a fix for the debug vs release library issues on MSVC?

zrax:
No, not yet...  I haven't decided yet whether to fix it by using different naming (like Python or Qt) to allow them to install side-by-side, or just including two different builds on Windows only...  I'd really like to avoid moving everything out of inlines though, since that defeats a lot of the optimization potential.

Hoikas:
Agreed. I think I would go the Python/Qt route since that's a fairly common practice.

zrax:
Ok, string_theory 1.1 should fix the library issue.

Hoikas:
The readme should probably be updated to mention string_theory 1.1

Hoikas:
On play testing, I encountered a bug in which the books on the Relto bookshelf are nonfunctional. This issue does not occur in the latest commit on master.

Hoikas:
This gets a :+1: from me, but I would like to hear from one more play test before we merge.

Hoikas:
It's time to get off this pot. Any OHBjections to this pull request being merged should be voiced within the next 24-36 hours.

PR #532, opened 2016-12-19
Add support for saving optional text chunks to PNG images.

Deledrius:
By default, this addition inserts the client version and a timestamp into every image saved.  Other fields such as Explorer, Current Age, etc. are context-dependent and can be added by the caller.

Localized strings are possible, but as we don't have any translations for them I did not implement it in this changeset.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyImage.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyImage.cpp
@@ -189,10 +189,9 @@ void pyImage::SaveAsJPEG(const plFileName& fileName, uint8_t quality)
     plJPEG::Instance().WriteToFile(fileName, this->GetImage());
 }
 
-void pyImage::SaveAsPNG(const plFileName& fileName)
+void pyImage::SaveAsPNG(const plFileName& fileName, const std::multimap<ST::string, ST::string>& textFields)
 {
-
-    plPNG::Instance().WriteToFile(fileName, this->GetImage());
+    plPNG::Instance().WriteToFile(fileName, this->GetImage(), textFields);
 }
 
 PyObject* pyImage::LoadJPEGFromDisk(const plFileName& filename, uint16_t width, uint16_t height)
--- a/Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
@@ -165,21 +165,33 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
 
 PYTHON_METHOD_DEFINITION(ptImage, saveAsPNG, args)
 {
-    PyObject* filenameObj;
-    if (!PyArg_ParseTuple(args, "O", &filenameObj))
+    PyObject* filenameObj, *optionalText = nullptr;
+    if (!PyArg_ParseTuple(args, "O|O", &filenameObj, &optionalText))
     {
-        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string");
+        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string, and an optional dict of key-value string pairs");
         PYTHON_RETURN_ERROR;
     }
 
     if (PyString_CheckEx(filenameObj))
     {
-        self->fThis->SaveAsPNG(PyString_AsStringEx(filenameObj));
+        if (optionalText && PyDict_Check(optionalText)) {
+            std::multimap<ST::string, ST::string> textFields;
+            PyObject *key, *value;
+            Py_ssize_t pos = 0;
+            while (PyDict_Next(optionalText, &pos, &key, &value)) {
+                if (PyString_CheckEx(key) && PyString_CheckEx(value)) {
+                    textFields.emplace(PyString_AsStringEx(key), PyString_AsStringEx(value));
+                }
+            }
+            self->fThis->SaveAsPNG(PyString_AsStringEx(filenameObj), textFields);
+        }
+        else
+            self->fThis->SaveAsPNG(PyString_AsStringEx(filenameObj));
         PYTHON_RETURN_NONE;
     }
     else
     {
-        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string");
+        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string, and an optional dict of key-value string pairs");
         PYTHON_RETURN_ERROR;
     }
 }
--- a/Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
+++ b/Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
@@ -44,6 +44,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsStream.h"
 #include "hsExceptions.h"
 
+#include "plProduct.h"
+#include <ctime>
+
 #include "plPNG.h"
 #include "plGImage/plMipmap.h"
 
@@ -189,7 +192,7 @@ plMipmap* plPNG::ReadFromFile(const plFileName& fileName)
     return ret;
 }
 
-bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
+bool plPNG::IWrite(plMipmap* source, hsStream* outStream, const std::multimap<ST::string, ST::string>& textFields)
 {
     bool result = true;
 
@@ -226,11 +229,46 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
             row_ptrs[i] = (png_bytep)srcp + (i * stride);
         }
 
+        // Write image data
         png_write_image(png_ptr, row_ptrs);
+
+        // Prepare Textual Metadata
+        char time_string[50]; // RFC-1123: "Day, DD-Mon-YYYY hh:mm:ss";
+        time_t rawtime = time(nullptr);
+        strftime(time_string, sizeof(time_string), "%a, %d-%b-%Y %T", gmtime(&rawtime));
+
+        // Add our standard fields, then combine with the custom fields
+        std::multimap<ST::string, ST::string> all_fields;
+        all_fields.emplace("Software", plProduct::ProductString());
+        all_fields.emplace("Creation Time", time_string);
+        for (auto field : textFields)
+            all_fields.insert(field);
+
+        png_text* text = new png_text[all_fields.size()];
+        size_t num_txtfields = 0;
+        for (auto it = all_fields.begin(); it != all_fields.end(); it++, num_txtfields++) {
+            // The PNG specification requires Latin-1 in the 'key' field
+            text[num_txtfields].key = (png_charp)(strdup(it->first.left(PNG_KEYWORD_MAX_LENGTH).trim().to_latin_1().data()));
+            text[num_txtfields].text = (png_charp)(it->second.c_str());
+            text[num_txtfields].lang = "en-us";  //  Language used in 'text' and 'lang_key'.
+            text[num_txtfields].lang_key = "";   //  Translation of 'key' into 'lang', if needed.
+            text[num_txtfields].compression = PNG_ITXT_COMPRESSION_NONE;
+        }
+
+        // Write Textual Metadata
+        png_set_text(png_ptr, info_ptr, text, num_txtfields);
+
+        // Finish Up
         png_write_end(png_ptr, info_ptr);
+
+        //  Clean up text buffers
+        for (size_t i = 0; i < num_txtfields; i++)
+            free(text[i].key);
+
         //  Clean up allocated structs
         png_destroy_write_struct(&png_ptr, &info_ptr);
         delete [] row_ptrs;
+        delete [] text;
     } catch (...) {
         result = false;
     }
@@ -238,7 +276,7 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
     return result;
 }
 
-bool plPNG::WriteToFile(const plFileName& fileName, plMipmap* sourceData)
+bool plPNG::WriteToFile(const plFileName& fileName, plMipmap* sourceData, const std::multimap<ST::string, ST::string>& textFields)
 {
     hsUNIXStream out;
 
@@ -246,7 +284,7 @@ bool plPNG::WriteToFile(const plFileName& fileName, plMipmap* sourceData)
         return false;
     }
 
-    bool ret = IWrite(sourceData, &out);
+    bool ret = IWrite(sourceData, &out, textFields);
     out.Close();
     return ret;
 }
--- a/Sources/Plasma/FeatureLib/pfPython/pyImage.h
+++ b/Sources/Plasma/FeatureLib/pfPython/pyImage.h
@@ -51,6 +51,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "pyGlueHelpers.h"
 #include <vector>
+#include <map>
 
 #include "pnKeyedObject/plKey.h"
 #include "pyColor.h"
@@ -144,7 +145,7 @@ public:
     uint32_t GetWidth(); // returns the width of the image
     uint32_t GetHeight(); // returns the height of the image
     void SaveAsJPEG(const plFileName& fileName, uint8_t quality = 75);
-    void SaveAsPNG(const plFileName& fileName);
+    void SaveAsPNG(const plFileName& fileName, const std::multimap<ST::string, ST::string>& textFields = std::multimap<ST::string, ST::string>());
     static PyObject* LoadJPEGFromDisk(const plFileName& filename, uint16_t width, uint16_t height); // returns pyImage
     static PyObject* LoadPNGFromDisk(const plFileName& filename, uint16_t width, uint16_t height); // returns pyImage
 #endif
--- a/Sources/Plasma/PubUtilLib/plGImage/plPNG.h
+++ b/Sources/Plasma/PubUtilLib/plGImage/plPNG.h
@@ -43,6 +43,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #ifndef _plPNG_h
 #define _plPNG_h
 
+#include <map>
 
 //// Class Definition /////////////////////////////////////////////////////////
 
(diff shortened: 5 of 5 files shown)


Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,35 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
...
-        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string");
+        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string, and an optional dict of key-value pairs");
         PYTHON_RETURN_ERROR;
     }
 
     if (PyString_CheckEx(filenameObj))
     {
-        self->fThis->SaveAsPNG(PyString_AsStringEx(filenameObj));
+        if (PyDict_Check(optionalText)) {
+            auto textFields = new std::multimap<plString, plString>();
```
Hoikas: Considering that this is a temporary object that does not persist outside this block, can we allocate the multimap on the stack?

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,35 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
...
+            PyObject *key, *value;
+            Py_ssize_t pos = 0;
+            while (PyDict_Next(optionalText, &pos, &key, &value)) {
+                if (PyString_Check(key) && PyString_Check(value)) {
```
Hoikas: This will die hard if a unicode object is passed in. There is a way to make Python convert both `str` and unicode objects to UTF-8 `char*`. I can't think of where that's done in the code currently, but I know it happens somewhere.

Sources/Plasma/FeatureLib/pfPython/pyImage.cpp
```diff
@@ -189,10 +189,10 @@ void pyImage::SaveAsJPEG(const plFileName& fileName, uint8_t quality)
     plJPEG::Instance().WriteToFile(fileName, this->GetImage());
 }
 
-void pyImage::SaveAsPNG(const plFileName& fileName)
+void pyImage::SaveAsPNG(const plFileName& fileName, std::multimap<plString, plString>* textFields)
```
Hoikas: Would be better to accept a const reference.

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
Deledrius: The PNG specification prohibits Unicode in these fields:

> Keywords and text are in the ISO 8859-1 (Latin-1) character set (a superset of regular ASCII) and can not contain NUL characters, and should not contain control or other unprintable characters.  To make the comments widely readable, stick with basic ASCII, and avoid machine specific character set extensions like the IBM-PC character set.

- http://www.libpng.org/pub/png/libpng-manual.txt

Hoikas: That's kind of concerning... At the moment, we are converting things to try to deal with UTF8 everywhere. Maybe we should ensure the accepted object is in latin-1? (latin-1 counts as "unicode" in Python 2...)

Deledrius: You mean accept & use Unicode coming in from Python, and then convert to the safe charset when sending it along?

Hoikas: Indeed

Sources/Plasma/FeatureLib/pfPython/pyImage.cpp
Deledrius: Is there a way to do that and still allow it to be an optional argument?

zrax: Since the code this eventually gets passed to checks for `nullptr` and then defaults it with:
```c++
auto textFields = new std::multimap<plString, plString>();
```
It seems reasonable to just make this a const reference and set the default to an empty multimap (rather than `nullptr`)

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,59 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
             row_ptrs[i] = (png_bytep)srcp + (i * stride);
         }
 
+        // Write image data
         png_write_image(png_ptr, row_ptrs);
+
+        // Prepare Textual Metadata
+        bool owned_map = false;
+        if (!textFields) {
+            auto textFields = new std::multimap<plString, plString>();
```
zrax: Come to think of it, this is also a bug, since the inner `textFields` declaration hides the outer one, and them promptly leaks.

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
zrax: For reference, `ST::string`'s `to_latin1()` converter allows you to replace invalid characters during conversion (instead of asserting):
https://github.com/zrax/string_theory/blob/master/doc/st_string.md#to_latin_1

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,35 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
...
+                    textFields->insert(std::pair<plString, plString>(PyString_AsString(key), PyString_AsString(value)));
```
zrax: Simpler to use `std::make_pair(...)`

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,59 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+            owned_map = true;
+        }
+
+        char time_string[26]; // RFC-1123: "Day, DD-Mon-YYYY hh:mm:ss";
+        time_t rawtime;
+        time(&rawtime);
```
zrax: Simpler to use:
```c++
time_t rawtime = time(nullptr);
```

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,59 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+        strftime(time_string, sizeof(time_string), "%a, %d-%b-%Y %T", gmtime(&rawtime));
+        textFields->insert(std::pair<plString, plString>("Software", plProduct::ProductString()));
+        textFields->insert(std::pair<plString, plString>("Creation Time", time_string));
```
zrax: `std::make_pair(...)`

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
Deledrius: Already switched to using `to_latin1()` about an hour before your comment here in my working edits. 😄

Deledrius:
I'm currently working on a rebased version that takes into account the suggestions provided.  Please ignore the commit before this comment. 😉

Sources/Plasma/FeatureLib/pfPython/pyImage.cpp
Deledrius: I'll give this a try.  Thanks for the suggestions.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
Deledrius: Are you sure?  I made sure to clean it up at the end, and only if we created the inner one.  A bit messy, and the one part I wasn't happy about, but it all goes away now anyway.

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,34 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
 
 PYTHON_METHOD_DEFINITION(ptImage, saveAsPNG, args)
 {
-    PyObject* filenameObj;
-    if (!PyArg_ParseTuple(args, "O", &filenameObj))
+    PyObject* filenameObj, *optionalText;
```
zrax: Optional parameters do need to be pre-initialized.  If the caller doesn't pass that parameter, `PyArg_ParseTuple` will leave the passed parameter in an undefined state.

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,34 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
...
     {
-        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string");
+        PyErr_SetString(PyExc_TypeError, "saveAsPNG expects a string, and an optional dict of key-value pairs");
         PYTHON_RETURN_ERROR;
     }
 
     if (PyString_CheckEx(filenameObj))
     {
-        self->fThis->SaveAsPNG(PyString_AsStringEx(filenameObj));
+        if (PyDict_Check(optionalText)) {
```
zrax: Likewise, optional parameters will need to be checked for presence (e.g. NULL).

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
```diff
@@ -165,21 +165,34 @@ PYTHON_METHOD_DEFINITION(ptImage, saveAsJPEG, args)
...
+            std::multimap<ST::string, ST::string> textFields;
+            PyObject *key, *value;
+            Py_ssize_t pos = 0;
+            while (PyDict_Next(optionalText, &pos, &key, &value)) {
+                if (PyString_CheckEx(key) && PyString_CheckEx(value)) {
+                    textFields.emplace(PyString_AsStringEx(key), PyString_AsStringEx(value));
+                }
```
zrax: Want to check for errors?  Otherwise, this will just silently ignore any entries in the dict that don't have the correct types.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,57 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+        std::multimap<ST::string, ST::string> standard_fields;    
+        standard_fields.emplace("Software", plProduct::ProductString());
+        standard_fields.emplace("Creation Time", time_string);
+
+        png_text* text = new png_text[standard_fields.size() + textFields.size()];
+        int num_txtfields = 0;
+        for (auto it = standard_fields.begin(); it != standard_fields.end(); it++, num_txtfields++) {
+            // The PNG specification requires Latin-1 in these fields
+            text[num_txtfields].key = const_cast<png_charp>(it->first.left(79).to_latin_1().data());
+            text[num_txtfields].text = const_cast<png_charp>(it->second.to_latin_1().data());
```
zrax: These are storing a temporary (`to_latin_1()` returns a `ST::char_buffer` which has its own data, converted to Latin-1).  You'll need to store off these temporary buffers somewhere, or else copy (`strdup`) the string and free it later.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,57 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+            text[num_txtfields].compression = PNG_TEXT_COMPRESSION_NONE;
+        }
+        for (auto it = textFields.begin(); it != textFields.end(); it++, num_txtfields++) {
+            // The PNG specification requires Latin-1 in these fields
+            text[num_txtfields].key = const_cast<png_charp>(it->first.left(79).to_latin_1().data());
+            text[num_txtfields].text = const_cast<png_charp>(it->second.to_latin_1().data());
```
zrax: Same as above

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
Deledrius: Thanks, I had come to guess that's what I was doing wrong here, and was hoping I could avoid the extra mess of storing and freeing them all in a third container.

Hoikas: you could write a macro block or a helper function to make it cleaner

Sources/Plasma/FeatureLib/pfPython/pyImageGlue.cpp
Deledrius: We could auto-convert to string values.  I don't think failing with an error would be preferable to skipping bad values and continuing to complete the function's main purpose.  Reporting a warning to the log would be nice as a compromise.  Can we do that from inside the glue functions?

I think the better option would be to use the string representation of the values provided.

Deledrius:
Updated to address requested changes.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,72 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+        for (auto it = standard_fields.begin(); it != standard_fields.end(); it++, num_txtfields++) {
+            // The PNG specification requires Latin-1 in these fields
+            auto key_text = strdup(it->first.left(79).to_latin_1().data());
+            auto val_text = strdup(it->second.to_latin_1());
+            text[num_txtfields].key = const_cast<png_charp>(key_text);
+            text[num_txtfields].text = const_cast<png_charp>(val_text);
+            text[num_txtfields].compression = PNG_TEXT_COMPRESSION_NONE;
+            text_buffer->push_back(key_text);
+            text_buffer->push_back(val_text);
+        }
```
zrax: This could all be simplified by creating a single multimap which copies `textFields` and merges in the two standard fields.  That way, you only have one loop to convert.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,72 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
             row_ptrs[i] = (png_bytep)srcp + (i * stride);
         }
 
+        // Write image data
         png_write_image(png_ptr, row_ptrs);
+
+        // Prepare Textual Metadata
+        char time_string[26]; // RFC-1123: "Day, DD-Mon-YYYY hh:mm:ss";
+        time_t rawtime = time(nullptr);
+        strftime(time_string, sizeof(time_string), "%a, %d-%b-%Y %T", gmtime(&rawtime));
```
zrax: Is this limited to 25 chars in all locales, or only "C" / "en_*"?  If the latter, we need to either force the locale to "C" or expand the string length...

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,72 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+        std::multimap<ST::string, ST::string> standard_fields;    
+        standard_fields.emplace("Software", plProduct::ProductString());
+        standard_fields.emplace("Creation Time", time_string);
+
+        png_text* text = new png_text[standard_fields.size() + textFields.size()];
+        auto text_buffer = new std::vector<char*>();
```
zrax: This vector does not need to be on the heap.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,72 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+
+        // Write Textual Metadata
+        png_set_text(png_ptr, info_ptr, text, num_txtfields);
+
+        // Finish Up
         png_write_end(png_ptr, info_ptr);
+
+        //  Clean up text buffers
+        for (auto buffer : *text_buffer)
+            free(buffer);
```
zrax: Why not walk the `text` variable instead of storing these separately?

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
Deledrius: Yeah, especially since I forgot to keep it from leaking away at the end!

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
Deledrius: 😢 That's how I had it originally, before the rewrites...  

I'll gladly simplify this again.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
Deledrius: There doesn't appear to be a limit to the field content length.  It simply needs to be NULL-terminated.

What would you recommend for determining the temporary string size here?

While looking into this to answer, I think I discovered something about the specs that might save some headaches.  It appears we can use an International Text (iTXT) chunk for any of the fields, and that allows Unicode.  Testing confirms this works as expected.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -226,27 +229,63 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream)
...
+        all_fields.emplace("Software", plProduct::ProductString());
+        all_fields.emplace("Creation Time", time_string);
+        for (auto field : textFields)
+            all_fields.insert(field);
+
+        png_text* text = new png_text[all_fields.size()];
+        size_t num_txtfields = 0;
+        for (auto it = all_fields.begin(); it != all_fields.end(); it++, num_txtfields++) {
+            text[num_txtfields].key = (png_charp)(strdup(it->first.left(79).to_utf8().data()));
+            text[num_txtfields].text = (png_charp)(strdup(it->second.to_utf8().data()));
```
Hoikas: Maybe @zrax will know better, but I feel like we don't need to `strdup` at minimum `it->second` because we're just using the ST::string's UTF-8 buffer... Isn't the ST::string already internally UTF-8, so this buffer is guaranteed to not go out of scope? I'm a little uncomfortable with `strdup`'ing `it->first`, but I don't see a reasonable way around it.

Deledrius: I think you're correct.  I've simplified this, pending further corrections from @zrax.

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
zrax: `strftime` is still fine to use (and it looks like we'll have to wait for C++20 to get an STL equivalent with `std::chrono`).  I was just suggesting a larger buffer, or a format (or locale) that's not likely to change based on the user's language setting.

Deledrius: That format is suggested (but not strictly required) by the specifications, so it seemed the best choice.

PR #533, opened 2016-12-28
Implement a generic hull/trimesh collider format

Hoikas:
PhysX requires all of its actors to have cooked mesh data. This cooked data is unfortunately not compatible with other versions of PhysX outside of the minor branch. The format is also largely unknown. Therefore, it is relevant to allow CWE prps to include an engine agnostic physical format to reduce our dependencies on proprietary libraries.

All PhysX blobs thankfully begin with the magic string NXS\x01, so we now abuse that by adding our own format that begins with the magic string HSP\x01. These blobs are cooked on the fly while already cooked blobs are passed directly on to PhysX.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPhysX/plSimulationMgr.cpp
+++ b/Sources/Plasma/PubUtilLib/plPhysX/plSimulationMgr.cpp
@@ -42,6 +42,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plSimulationMgr.h"
 
 #include <NxPhysics.h>
+#include <NxCooking.h>
 
 #include <algorithm>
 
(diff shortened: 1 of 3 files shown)


Hoikas:
Changed the magic string to HSP\x01 ("**H**ead**S**pin**P**hysical) after discussion in IRC

Sources/Plasma/PubUtilLib/plPhysX/plSimulationMgr.cpp
```diff
@@ -292,11 +293,20 @@ plSimulationMgr::plSimulationMgr()
...
+    if (!fSDK) {
+        fLog->AddLine("Phailed to init PhysX SDK");
         return false; // client will handle this and ask user to install
+    }
 
     fLog = plStatusLogMgr::GetInstance().CreateStatusLog(40, "Simulation.log", plStatusLog::kFilledBackground | plStatusLog::kAlignToTop);
-    fLog->AddLine("Initialized simulation mgr");
+    fLog->AddLine("Initialized PhysX SDK");
+
+    if (!NxInitCooking(nullptr, &gErrorStream)) {
```
zrax: Does the normal PhysX client software include the cooking SDK libraries?  If not, can they be redistributed?

Hoikas: You have to redistribute NxCooking.dll with your app, which is what I've been doing with Korman. It works as long as the user has the legacy PhysX package installed :smile: I have not seen any prohibition on redistributing this file. I could be mistaken, however.

dpogue: NxCooking.dll is distributed with MOUL

PR #534, opened 2017-01-12
GTest 1.8

zrax:
Update to Google Test 1.8.  This brings in some fixes for MSVC2015/MSVC2017 and matches the version string_theory is now using.


Deledrius:
👍

PR #536, opened 2017-01-16
Implement reward string storage on marker nodes

Hoikas:


Hoikas:
Verified in Gehn 25 beta

PR #537, opened 2017-01-19
Cleanup VolumeSensorConditionalObject

Hoikas:
The goal behind this changeset is to fix the intermittent issues seen in ages such as Gahreesen and Teledahn in which the avatar falls through the floor when leaving a subworld. This generally happens when the player's ping time is unsatisfactorily high. `plVolumeSensorConditionalObject` acquires an exclusive server lock before triggering the transition, and if this lock takes too long... the player has fallen through the floor. At some point, Cyan introduced the special case `plVolumeSensorConditionalObjectNoArbitration` exactly for this purpose, but it appears to never have been used.

This changeset refactors both VolumeSensor classes and moves the logic into the base class. The NoArbitration variant just sets a flag on the base class. This enables us to toggle which behavior we want in Python, allowing us to silently upgrade Cyan's VSCOs to VSCONAs and therefore fix the falling through the floor issue. A console command `Logic.DelayArbitration` was introduced to model both the problem and solution. This can be removed from the changeset if it is deemed inappropriate.

--code_duplication
--bugs
++code_quality

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
+++ b/Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
@@ -49,8 +49,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pnNetCommon/plNetApp.h"
 #include "plAvatar/plArmatureMod.h"
 #include "pnSceneObject/plSceneObject.h"
-
-bool plVolumeSensorConditionalObject::makeBriceHappyVar = true;
+#include "pnMessage/plPlayerPageMsg.h"
+#include "plgDispatch.h"
 
 plObjectInBoxConditionalObject::plObjectInBoxConditionalObject() :
 fCurrentTrigger(nil)
@@ -79,7 +79,7 @@ bool plObjectInBoxConditionalObject::MsgReceive(plMessage* msg)
                     if (pActivateMsg->fHitterObj == fCurrentTrigger && fCurrentTrigger && fLogicMod->HasFlag(plLogicModBase::kTriggered) && !IsToggle())
                     {
                         fCurrentTrigger = nil;
-                        fLogicMod->GetNotify()->AddContainerEvent( pActivateMsg->fHiteeObj, pActivateMsg->fHitterObj, false );
+                        fLogicMod->GetNotify()->AddContainerEvent(pActivateMsg->fHiteeObj, pActivateMsg->fHitterObj, false);
                         fLogicMod->RequestUnTrigger();
                     }
                 }
@@ -96,12 +96,12 @@ bool plObjectInBoxConditionalObject::Verify(plMessage* msg)
 {
     plActivatorMsg* pActivateMsg = plActivatorMsg::ConvertNoRef(msg);
     if (pActivateMsg)
-    {   
+    {
         for (int i = 0; i < fInside.Count(); i++)
         {
             if (pActivateMsg->fHitterObj == fInside[i])
             {
-                fLogicMod->GetNotify()->AddContainerEvent( pActivateMsg->fHiteeObj, pActivateMsg->fHitterObj, true );
+                fLogicMod->GetNotify()->AddContainerEvent(pActivateMsg->fHiteeObj, pActivateMsg->fHitterObj, true);
                 fCurrentTrigger = pActivateMsg->fHiteeObj;
                 return true;
             }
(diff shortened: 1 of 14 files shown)


Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
```diff
@@ -125,155 +124,160 @@ bool plObjectInBoxConditionalObject::Verify(plMessage* msg)
...
+}
+
+void plVolumeSensorConditionalObject::NoServerArbitration(bool arbitration)
+{
+    if (arbitration) {
+        plgDispatch::Dispatch()->RegisterForExactType(plPlayerPageMsg::Index(), GetKey());
+        fFlags |= kNoServerArbitration;
+    } else {
+        plgDispatch::Dispatch()->UnRegisterForExactType(plPlayerPageMsg::Index(), GetKey());
+        fFlags &= kNoServerArbitration;
```
Deledrius: I suspect you meant `~kNoServerArbitration` here.

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
```diff
@@ -286,163 +290,19 @@ void plVolumeSensorConditionalObject::Read(hsStream* stream, hsResMgr* mgr)
...
-        }
-    }
-        return plConditionalObject::MsgReceive(msg);
-}
 void plVolumeSensorConditionalObjectNoArbitration::Read(hsStream* stream, hsResMgr* mgr)
 {
     plVolumeSensorConditionalObject::Read(stream, mgr);
-    plgDispatch::Dispatch()->RegisterForExactType(plPlayerPageMsg::Index(), GetKey());
+
+    // We must have a valid fpKey before we do this, hence why this is not in the constructor
```
Deledrius: fpKey?

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
Hoikas: :trollface: I'll fix this tonight

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
Hoikas: hsKeyedObject::fpKey

Deledrius: Ah. :trollface: I thought it was a local member that I wasn't seeing... or a typo.

Deledrius:
I have two questions about this:

1. What is the general purpose (if any) intended by requiring server arbitration on these types of conditional events?  
2. Are there any foreseeable negative effects by turning it off at the specific places you've done so in H-uru/moul-scripts/pull/105 ?

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.h
```diff
@@ -76,56 +77,61 @@ class plVolumeSensorConditionalObject : public plConditionalObject
...
-    GETINTERFACE_ANY( plVolumeSensorConditionalObject, plConditionalObject );
-    
-    virtual bool MsgReceive(plMessage* msg);
+    virtual bool MsgReceive(plMessage* msg) HS_OVERRIDE;
 
-    void Evaluate(){;}
+    void Evaluate() { }
     void Reset() { SetSatisfied(false); }
-    virtual bool Satisfied();
+    virtual bool Satisfied() HS_OVERRIDE;
```
zrax: `virtual` is not necessary (and is IMO redundant) with `HS_OVERRIDE`

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
```diff
@@ -125,155 +124,160 @@ bool plObjectInBoxConditionalObject::Verify(plMessage* msg)
...
 
+void plVolumeSensorConditionalObject::IgnoreExtraEnters(bool ignore)
+{
+    if (ignore)
+        fFlags |= kIgnoreExtraEnters;
+    else
+        fFlags &= ~kIgnoreExtraEnters;
+}
+
+void plVolumeSensorConditionalObject::NoServerArbitration(bool arbitration)
```
zrax: If the flag to set is `kNoServerArbitration`, the variable name should reflect that instead of the opposite.

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.h
```diff
@@ -76,56 +77,61 @@ class plVolumeSensorConditionalObject : public plConditionalObject
...
     void    SetType(int i) { fType = i; }
 
     void SetTrigNum(int i) { fTrigNum = i; }
     void SetFirst(bool b) { fFirst = b; }
 
-    void IgnoreExtraEnters(bool ignore = true) {fIgnoreExtraEnters = ignore;}
+    void IgnoreExtraEnters(bool ignore = true);
+    void NoServerArbitration(bool arbitration = true);
```
zrax: See comment in implementation

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.h
```diff
@@ -76,56 +77,61 @@ class plVolumeSensorConditionalObject : public plConditionalObject
...
 
-    virtual void Read(hsStream* stream, hsResMgr* mgr); 
-    virtual void Write(hsStream* stream, hsResMgr* mgr);
+    virtual void Read(hsStream* stream, hsResMgr* mgr) HS_OVERRIDE;
+    virtual void Write(hsStream* stream, hsResMgr* mgr) HS_OVERRIDE;
```
zrax: `virtual` is not necessary (and is IMO redundant) with `HS_OVERRIDE`

Sources/Plasma/NucleusLib/pnModifier/plLogicModBase.cpp
```diff
@@ -185,6 +169,28 @@ bool plLogicModBase::MsgReceive(plMessage* msg)
     return plSingleModifier::MsgReceive(msg);
 }
 
+void plLogicModBase::IHandleArbitration(plServerReplyMsg* pSMsg)
+{
+    hsAssert(pSMsg->GetType() != plServerReplyMsg::kUnInit, "uninit server reply msg");
+    plNetClientApp::GetInstance()->DebugMsg("LM: LogicModifier {} recvd trigger request reply:{}, wasRequesting={}, t={f}\n",
+                                            GetKeyName().c_str(),
```
zrax: `.c_str()` unnecessary for ST::format

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.h
```diff
@@ -76,56 +77,61 @@ class plVolumeSensorConditionalObject : public plConditionalObject
...
 
 };
 class plVolumeSensorConditionalObjectNoArbitration : public plVolumeSensorConditionalObject
```
zrax: Should `plVolumeSensorConditionalObjectNoArbitration` set the `kNoServerArbitration` in its constructor?  Otherwise, it defaults to unset like `plVolumeSensorConditionalObject`

Hoikas: No, see comment in the NoArbitration reader--the fpKey must be valid so we can register for `plPlayerPageMsg`.

Sources/Plasma/NucleusLib/pnModifier/plLogicModBase.cpp
Hoikas: Copypasta code. Will fix :smile:

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
```diff
@@ -67,25 +67,24 @@ bool plObjectInBoxConditionalObject::MsgReceive(plMessage* msg)
         {
             fInside.Append(pActivateMsg->fHitterObj);
             return true;
-        }
-        else
-        if (pActivateMsg->fTriggerType == plActivatorMsg::kVolumeExit)
-        {
-            for (int i = 0; i < fInside.Count(); i++)
+        } else
+            if (pActivateMsg->fTriggerType == plActivatorMsg::kVolumeExit)
```
Hoikas: It looks like Visual Studio auto styled this for me. I have no idea how it came up with this crap nor do I remember adding it to the commit. Will fix

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
Hoikas: Indeed. I meant to fix that here but forgot -- will fix

Sources/Plasma/FeatureLib/pfConditional/plObjectInBoxConditionalObject.cpp
Hoikas: :trollface:

Hoikas:
I've fixed the issues reported in the review.

@Deledrius 
To recap the discussion in IRC, server arbitration acts much like a mutex in that the logic modifier requests a lock from the server before it triggers. The server will deny any lock request that comes in while the logic mod is already locked.

Note that I say the logic modifier is locked, not the conditional. In the case of trivial enter/exit volume logic modifiers, the change from arbitration to no arbitration is, in the grand scheme of things, a no-op because plVolumeSensorConditionalObject will only trigger on the hitting object owner's client and the action on those specific cases only happens to the local avatar. I did not change all volume sensors to no arbitration types because if you have a more esoteric setup with a volume sensor AND something like a clickable, you need the server's help with ensuring the same logic modifier doesn't get fired by multiple clients at once (imagine if everyone clicks a button at the same time, trying to troll -- all avatars would be overlapping pressing the button, and the responder would get fired many times. Yikes!)

dpogue:
Just to confirm for my own interest: the correct way to avoid this problem is still to use VSCONA when building Ages, but we're adding the Python API so that we can fix existing cases without modifying the data files?

Hoikas:
Correct @dpogue. I hope that we can get Korman to handle all of that detail for us (stay tuned...)

PR #539, opened 2017-01-21
Fix 0-byte patch files

Hoikas:
This changeset fixes the issues with 0-byte files being downloaded from MOULa via the Gehn repair tool. It turned out that we were opening too many concurrent files. Oops! Bonus: a few code cleanups such as `HS_OVERRIDE` and `emplace_back`

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -164,30 +164,40 @@ class pfPatcherStream : public plZlibStream
 
 public:
     pfPatcherStream(pfPatcherWorker* parent, const plFileName& filename, uint64_t size)
-        : fParent(parent), fFilename(filename), fFlags(0), fBytesWritten(0), fDLStartTime(0.f)
+        : fParent(parent), fFilename(filename), fFlags(0), fBytesWritten(0), fDLStartTime(0.f), plZlibStream()
     {
         fParent->fTotalBytes += size;
         fOutput = new hsRAMStream;
     }
 
-    pfPatcherStream(pfPatcherWorker* parent, const plFileName& filename, const NetCliFileManifestEntry& entry)
-        : fParent(parent), fFlags(entry.flags), fBytesWritten(0)
+    pfPatcherStream(pfPatcherWorker* parent, const plFileName& reqName, const plFileName& cliName, const NetCliFileManifestEntry& entry)
+        : fParent(parent), fFilename(cliName.Normalize()), fFlags(entry.flags), fBytesWritten(0), plZlibStream()
     {
         // ugh. eap removed the compressed flag in his fail manifests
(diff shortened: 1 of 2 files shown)


Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -199,16 +208,14 @@ class pfPatcherStream : public plZlibStream
...
-    virtual uint32_t GetEOF() { return fOutput->GetEOF(); }
-    virtual uint32_t GetPosition() const { return fOutput->GetPosition(); }
-    virtual uint32_t GetSizeLeft() const { return fOutput->GetSizeLeft(); }
-    virtual uint32_t Read(uint32_t count, void* buf) { return fOutput->Read(count, buf); }
-    virtual void Rewind() { fOutput->Rewind(); }
-    virtual void SetPosition(uint32_t pos) { fOutput->SetPosition(pos); }
-    virtual void Skip(uint32_t deltaByteCount) { fOutput->Skip(deltaByteCount); }
+    bool AtEnd() HS_OVERRIDE { return fOutput->AtEnd(); }
+    uint32_t GetEOF() HS_OVERRIDE { return fOutput->GetEOF(); }
+    uint32_t GetPosition() const HS_OVERRIDE { return fOutput->GetPosition(); }
```
zrax: Was the removal of `GetSizeLeft` intentional?

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -170,24 +170,33 @@ class pfPatcherStream : public plZlibStream
...
     {
-        fFilename = filename.Normalize();
-        return plZlibStream::Open(fFilename, mode);
+        fDLStartTime = hsTimer::GetSysSeconds();
+        if (!fOutput)
+            Open(fFilename, "wb");
     }
 
-    virtual uint32_t Write(uint32_t count, const void* buf)
+    bool Open(const plFileName& filename, const char* mode) HS_OVERRIDE
```
zrax: Unused parameter `filename` (we only use the stored `fFilename` below).

Hoikas: Required to override base implementation

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
Hoikas: Yes, there is nothing to override, and it was unused

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
Hoikas: I'll update it to use the parameter instead however. Could be useful in the future

Hoikas:
Fixed unused parameter issue

PR #542, opened 2017-01-22
Fix not being able to edit names of markers not in the current age

Hoikas:


Hoikas:
Verified on Gehn.25 beta

PR #543, opened 2017-01-22
Fix PageInAge ignoring the .age file

Hoikas:
Previously, the client loaded every PRP in the global age specified. Now, we test to ensure that the requested page is listed in the age file. This is relevant considering all of the clothing PRPs going into MOULa. We don't want folks able to bring random clothing files onto Gehn. Please note this code path is only used for global ages.

### DIFF
--- a/Sources/Plasma/Apps/plPageInfo/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plPageInfo/CMakeLists.txt
@@ -13,7 +13,7 @@ set(plPageInfo_SOURCES
 )
 
 add_executable(plPageInfo ${plPageInfo_SOURCES})
-target_link_libraries(plPageInfo CoreLib plResMgr plAudioCore pnUUID)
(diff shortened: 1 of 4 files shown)


Hoikas:
Updated to fix some linker errors in tools as noticed in Gehn 25 beta

Hoikas:
Any ohbjections to merging this? It's currently live on Gehn.25 beta...

Deledrius:
This looks good to me.  Do you have any idea if there was a purpose behind letting it be so permissive?

Hoikas:
I think it was an oversight, honestly. Previously, plRegistry only loaded RegistryPageNodes for files listed in the .age file. It's the RPN loading that has become extremely permissive. I did try to fix that, but it was a huge headache, so I felt this would be a decent solution.

Deledrius:
This broke plFontConverter and plLocalizationEditor due to the change in function signatures.

PR #544, opened 2017-01-23
Patcher File Deletion

Hoikas:
Implements the ability for the FileSrv to instruct the patcher to delete files. Also, we ensure that pfPatcher can only operate inside the client directory to prevent mischief. This will be useful considering we don't want the stupid clothing files from MOULa.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -94,9 +94,12 @@ struct pfPatcherWorker : public hsThread
         // Executable flags
         kRedistUpdate               = 1<<4,
 
+        // Deleted file
+        kFileDeleted                = 1<<5,
+
         // Begin internal flags
-        kLastManifestFlag           = 1<<5,
-        kSelfPatch                  = 1<<6,
+        kLastManifestFlag           = 1<<6,
+        kSelfPatch                  = 1<<7,
     };
 
     std::deque<Request> fRequests;
@@ -108,6 +111,7 @@ struct pfPatcherWorker : public hsThread
 
     pfPatcher::CompletionFunc fOnComplete;
     pfPatcher::FileDownloadFunc fFileBeginDownload;
+    pfPatcher::FileDownloadFunc fFileDeleted;
     pfPatcher::FileDesiredFunc fFileDownloadDesired;
     pfPatcher::FileDownloadFunc fFileDownloaded;
     pfPatcher::GameCodeDiscoverFunc fGameCodeDiscovered;
@@ -171,7 +175,7 @@ public:
     }
 
     pfPatcherStream(pfPatcherWorker* parent, const plFileName& reqName, const plFileName& cliName, const NetCliFileManifestEntry& entry)
-        : fParent(parent), fFilename(cliName.Normalize()), fFlags(entry.flags), fBytesWritten(0), plZlibStream()
+        : fParent(parent), fFilename(cliName), fFlags(entry.flags), fBytesWritten(0), plZlibStream()
     {
         // ugh. eap removed the compressed flag in his fail manifests
         if (reqName.GetFileExt().compare_i("gz") == 0) {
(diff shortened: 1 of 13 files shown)


Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -476,55 +480,78 @@ void pfPatcherWorker::Run()
...
+        fQueuedFiles.pop_front();
 
         // eap sucks
-        plFileName clName = ST::string::from_wchar(entry.clientName);
+        plFileName clName = plFileName(ST::string::from_wchar(entry.clientName)).Normalize();
         ST::string dlName = ST::string::from_wchar(entry.downloadName);
 
+        // Only operate on files if they are contained within the client directory...
+        plFileName clAbsPath = clName.AbsolutePath();
+        if (!clAbsPath.AsString().starts_with(clientDir.AsString())) {
```
Deledrius: My instinct is telling me that a whitelist approach is better than a blacklist approach.  Something along the lines of "Is the manifest requesting we delete a file we have in our local store?  Then delete the file indicated from the local store."  This way we never use the manifest's supplied name for anything other than the comparison, keeping it at the lowest-possible level of trust.

Hoikas: The only issue is that there are two stores that the client has, plResManager (PRP) and plStreamSource (pak, SDL). What if we want to delete ogg, bik, webm, etc?

Deledrius: The ostensible reason for this fix was to prevent potential conflicts in errant PRPs due to the way that they are loaded.  I see no such cause for concern with the asset files (except, possibly, the .pak archives).

It's a separate issue to be culling old or unused files simply to reduce clutter, and that's not one I think is worth the added security risk.

Hoikas: The reasoning for this one is to cull old files. #543 fixes the loading issue

Deledrius: Then perhaps I misunderstood.  I thought these two were both addressing aspects of the same issue.

Deledrius: Might a better solution then to be removing any files in the local store that aren't specified in the manifest(s)?  It wouldn't rely on trusting a remote source at all, and would achieve the desired result.

Hoikas: I considered that solution as well but eventually decided I preferred outright deletion. I know it's sketchy to allow deletion, but we're already downloading arbtirary Python code in python.pak, which is significantly more dangerous IMO. This changeset is not essential for Gehn 25, so we can defer until later if needed

Deledrius: That's a good point regarding the Python, and it did cross my mind.  I would definitely like to get @zrax and @dpogue to weigh in on this concept before merging it.

Hoikas:
NOTE: This changeset appears to introduce a potential crash in the external client.

Sources/Plasma/NucleusLib/pnMessage/plClientMsg.h
```diff
@@ -113,9 +117,9 @@ class plClientMsg : public plMessage
 
     void AddRoomLoc(plLocation loc);
 
-    // Used for kLoadAgeKeys, kLetGoOfAgeKeys only
-    ST::string  GetAgeName() const { return fAgeName; }
-    void        SetAgeName(const ST::string& age) { fAgeName = age; }
+    // Used for kLoadAgeKeys, kLetGoOfAgeKeys, kAddPage, and kDropPage
+    const ST::string& GetPath() const { return fPath; }
```
zrax: This should return a `ST::string` copy, not a reference.  In practice, it shouldn't matter much, but there are some cases where it does, and it's better to be consistent to avoid those cases (IMO).

PR #546, opened 2017-01-29
Server Capabilties

Hoikas:
Introduces the server capabilities message, as discussed for the Gahreesen Wall and H-uru/dirtsand#112. Currently, I have gated auth score leaderboard generation behind server capabilties so that our clients aren't kicked off of MOULa when we send that unimplemented message :smile: If the server does not send the broadcast, this is effectively a no-op.


Hoikas:
Updated to use an hsBitVector

Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
```diff
@@ -2387,6 +2388,20 @@ static bool Recv_ScoreGetHighScoresReply(
...
+//============================================================================
+static bool Recv_ServerCaps(
+    const uint8_t   msg[],
+    unsigned        bytes,
+    void *          param
+    ) {
+    const Auth2Cli_ServerCaps & caps = *(const Auth2Cli_ServerCaps *)msg;
+
+    hsReadOnlyStream stream(caps.byteCount, reinterpret_cast<const void*>(caps.buffer));
+    ((CliAuConn*)param)->caps.Read(&stream);
```
zrax: Note that this technically has the size twice, since `hsBitVector::Read` also includes a 32-bit dword count.  I don't see any reasonable way around this though, and this message should only be transferred once per client connection.

Hoikas:
Agreed. There are plenty of places in the protocol where a duped size
parameter is not unusual. Although a lot of those are actually unneeded :D

Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
Hoikas: Agreed. There are plenty of places in the protocol where a duped size parameter is not unusual. Although a lot of those are actually unneeded :D

Hoikas:
This has been rebased and corrected to play nicely with some of the more recent modernization commits.

PR #547, opened 2017-01-30
Add missing linkage for Tools.

Deledrius:
Adds missing libraries for plFontConverter and plLocalizationEditor that were overlooked in #543.


Hoikas:
:trollface:

PR #552, opened 2017-12-05
OpenSSL 1.1 / Linux compile fixes

zrax:
Using OpenSSL 1.1 for the client is still not an option with this change, but this allows us to once again build the tools on Linux systems with newer system OpenSSL libraries.


zrax:
It's the other way around, actually...  plResMgr depends on plAgeDescription, and because of the way Linux links static libs, that means plAgeDescription needs to be after anything that depends on it.

dpogue:
hmm, weird, okay... so plResMgr should have plAgeDescription in its target_link_libraries? (It currently does not)

PR #560, opened 2018-03-26
Add workaround for mouse handling bug in Win10 FCU

Hoikas:
Quoth @Deledrius 

Windows 10 Fall Creators Update changed the behavior of SetCursorPos, which no longer sends the expected WM_MOUSEMOVE message when recentering the cursor.  This adds a function to detect affected versions of Windows and flags the necessity to manually send our own message.

This workaround currently only detects the known affected version of Win10, in the hope that it will be fixed and no further attention will be required.  If this regression is not fixed, additional Win10 versions can be added.


Deledrius:
👍 Looks great to me.

PR #561, opened 2018-03-30
Update Windows version identification.

Deledrius:
This switches over to using the recently-added `hsGetWindowsVersion` for `DisplaySystemVersion`.  I also modified `hsGetWindowsVersion` to use `RTL_OSVERSIONINFOEXW` instead of `RTL_OSVERSIONINFOW` as it contains necessary information.

`DisplaySystemVersion` no longer parses information for Win95/Win98/WinME as those are not supported and are not needed.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.cpp
+++ b/Sources/Plasma/CoreLib/HeadSpin.cpp
@@ -487,165 +487,98 @@ char    *hsWStringToString( const wchar_t *str )
 //
 std::vector<ST::string> DisplaySystemVersion()
 {
+    std::vector<ST::string> versionStrs;
 #if HS_BUILD_FOR_WIN32
 #ifndef VER_SUITE_PERSONAL
 #define VER_SUITE_PERSONAL 0x200
 #endif
-    std::vector<ST::string> versionStrs;
-    OSVERSIONINFOEX osvi;
-    BOOL bOsVersionInfoEx;
+    const RTL_OSVERSIONINFOEXW& version = hsGetWindowsVersion();
 
-    // Try calling GetVersionEx using the OSVERSIONINFOEX structure.
-    //
-    // If that fails, try using the OSVERSIONINFO structure.
-
-    ZeroMemory(&osvi, sizeof(OSVERSIONINFOEX));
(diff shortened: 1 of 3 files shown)


Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -487,165 +487,98 @@ char    *hsWStringToString( const wchar_t *str )
...
-
-        if (osvi.dwMajorVersion == 4 && osvi.dwMinorVersion == 90)
-        {
-            versionStrs.push_back("Microsoft Windows Me ");
-        }
-        break;
-
-    case VER_PLATFORM_WIN32s:
-
-        versionStrs.push_back("Microsoft Win32s ");
```
zrax: :trollface:

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -487,165 +487,98 @@ char    *hsWStringToString( const wchar_t *str )
...
-    switch (osvi.dwPlatformId)
+    switch (version.dwPlatformId)
     {
     case VER_PLATFORM_WIN32_NT:
 
         // Test for the product.
 
-        if ( osvi.dwMajorVersion <= 4 )
+        if ( version.dwMajorVersion <= 4 )
             versionStrs.push_back("Microsoft Windows NT ");
```
Hoikas: :trollface:

Sources/Plasma/CoreLib/HeadSpin.cpp
```diff
@@ -487,165 +487,98 @@ char    *hsWStringToString( const wchar_t *str )
...
 
-        if ( osvi.dwMajorVersion == 5 && osvi.dwMinorVersion == 0 )
+        if ( version.dwMajorVersion == 5 && version.dwMinorVersion == 0 )
             versionStrs.push_back("Microsoft Windows 2000 ");
```
Hoikas: :trollface:

Sources/Plasma/CoreLib/HeadSpin.cpp
Deledrius: Pretty much.  I don't think we had a pressing need to keep that around.

PR #562, opened 2018-03-31
Patcher fixes

Hoikas:
Fixes an issue in which the `-NoSelfPatch` flag is ignored if a patcher is found in the client manifest. Also addresses the issue of the patcher not exiting correctly in some situations.

### DIFF
--- a/Sources/Plasma/Apps/plUruLauncher/plClientLauncher.cpp
+++ b/Sources/Plasma/Apps/plUruLauncher/plClientLauncher.cpp
@@ -258,8 +258,20 @@ bool plClientLauncher::IApproveDownload(const plFileName& file)
 {
     // So, for a repair, what we want to do is quite simple.
     // That is: download everything that is NOT in the root directory.
-    plFileName path = file.StripFileName();
-    return !path.AsString().is_empty();
+    if (hsCheckBits(fFlags, kGameDataOnly)) {
+        plFileName path = file.StripFileName();
+        return !path.AsString().is_empty();
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/Plasma/Apps/plUruLauncher/plClientLauncher.cpp
```diff
@@ -256,8 +257,20 @@ bool plClientLauncher::IApproveDownload(const plFileName& file)
 {
     // So, for a repair, what we want to do is quite simple.
     // That is: download everything that is NOT in the root directory.
-    plFileName path = file.StripFileName();
-    return !path.AsString().is_empty();
+    if (hsCheckBits(fFlags, kGameDataOnly)) {
+        plFileName path = file.StripFileName();
+        return !path.AsString().is_empty();
```
zrax: `!path.AsString().is_empty()` can be simplified as `path.IsValid()`...

Sources/Plasma/Apps/plUruLauncher/plClientLauncher.h
```diff
@@ -70,8 +70,16 @@ class plClientLauncher
...
+    enum NetCoreState
+    {
+        kNetCoreInactive,
+        kNetCoreActive,
+        kNetCoreShutdown,
+    };
+
     uint32_t   fFlags;
     plFileName fServerIni;
+    uint8_t    fNetCoreState;
```
zrax: Why not just declare it `NetCoreState` and let the compiler warn on funny values?

Hoikas:
Rebased and updated based on feedback from review

PR #563, opened 2018-03-31
Fix typo for GCC/Clang align attribute

dpogue:


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -392,7 +392,7 @@ inline float hsRadiansToDegrees(float rad) { return float(rad * (180 / M_PI)); }
 #ifdef _MSC_VER
 #   define ALIGN(n) __declspec(align(n))
 #else
-#   define ALIGN(n) __atribute__(aligned(n))
+#   define ALIGN(n) __attribute__(aligned(n))
```
zrax: I should not have merged this -- it's still incorrect.  The correct syntax should be `__attribute__((aligned(n)))`

PR #567, opened 2018-04-05
Follow redirects when fetching server status.

branan:
cURL does not follow redirects by default, meaning the Gehn launcher
is currently printing the raw HTML of a 301 response.

### DIFF
--- a/Sources/Plasma/Apps/plClient/winmain.cpp
+++ b/Sources/Plasma/Apps/plClient/winmain.cpp
@@ -731,8 +731,8 @@ BOOL CALLBACK UruLoginDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM
                 // For reporting errors
                 char curlError[CURL_ERROR_SIZE];
                 curl_easy_setopt(hCurl, CURLOPT_ERRORBUFFER, curlError);
-                curl_easy_setopt(curl.get(), CURLOPT_FOLLOWLOCATION, 1);
-                curl_easy_setopt(curl.get(), CURLOPT_MAXREDIRS, 5);
+                curl_easy_setopt(hCurl, CURLOPT_FOLLOWLOCATION, 1);
+                curl_easy_setopt(hCurl, CURLOPT_MAXREDIRS, 5);
 
                 while (s_loginDlgRunning) {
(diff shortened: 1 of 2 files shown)


Sources/Plasma/Apps/plUruLauncher/plClientLauncher.cpp
```diff
@@ -109,6 +109,7 @@ void plShardStatus::Run()
     curl_easy_setopt(curl.get(), CURLOPT_ERRORBUFFER, fCurlError);
     curl_easy_setopt(curl.get(), CURLOPT_USERAGENT, "UruClient/1.0");
     curl_easy_setopt(curl.get(), CURLOPT_URL, url.c_str());
+    curl_easy_setopt(curl.get(), CURLOPT_FOLLOWLOCATION, 1);
```
zrax: Should probably also set `CURLOPT_MAXREDIRS` so we don't risk spinning forever in a redirect loop.

branan: Ugh, of course that has an unsafe default

branan:
Updated to set MAXREDIRS to an appropriate value, and duplicated in winmain.cpp for the plClient.

I'm gonna try to get a windows VM spun up tonight after work, so I should be able to actually verify this is building correctly soon.

Sources/Plasma/Apps/plClient/winmain.cpp
```diff
@@ -732,6 +732,8 @@ BOOL CALLBACK UruLoginDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM
                 // For reporting errors
                 char curlError[CURL_ERROR_SIZE];
                 curl_easy_setopt(hCurl, CURLOPT_ERRORBUFFER, curlError);
+                curl_easy_setopt(curl.get(), CURLOPT_FOLLOWLOCATION, 1);
+                curl_easy_setopt(curl.get(), CURLOPT_MAXREDIRS, 5);
```
Hoikas: Should be `hCurl` instead of `curl.get()`

Deledrius:
Any luck on finishing this, @branan?

Hoikas:
Thanks for fixing this @zrax

PR #569, opened 2018-05-12
Windows 10 Cursor (Part II)

Hoikas:
Cursor bug is still present in Windows 10 April 2018 Update 😢


Deledrius:
👍 
I'm not happy with needing to leave this open-ended, but I think that it's probably the best strategy given that we have no idea if a fix is incoming.  This requires only a single commit to end it, rather than constantly adding new builds as their regression is carried forward without a fix.  I just would rather be explicit about the builds which need it.

OTOH, this method automatically includes Insider testing builds too, so it's actually more complete a fix (now that the regression exists in multiple builds).

Hoikas:
If you want to start talking about insider builds, then we should probably be talking about where in the heck this bug appeared between Anniversary and the 2017 FCU. Ugh.

Deledrius:
True.  I'm not overly concerned about that, so I don't think this is necessary.  I was just noting that it's a side-benefit to include them.

PR #570, opened 2018-05-23
Don't error on a clean CMake build

AbigailBuccaneer:
`set_property(CACHE ...)` must refer to an existing cache variable. This
means it must be created before the properties are set on it. Without
this change, running `cmake` the first time fails on my machine, and
then succeeds when run the second time.


Deledrius:
Good ~~cache~~ catch. ;)

I can confirm this error, and the fix works.

PR #571, opened 2018-05-23
Support libpng with no iTXt chunk support

AbigailBuccaneer:
libpng's support for the `iTXt` chunk ('international textual data') is
optional, and off-by-default before libpng 1.4, which causes a compile
error. This writes an iTXt chunk only if supported, and writes a tEXt
chunk otherwise. (A tEXt chunk is identical but specified to contain
Latin-1 text instead of UTF-8, and has no language tag.)

### DIFF
--- a/Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
+++ b/Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
@@ -249,10 +249,15 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream, const std::multimap<ST
         for (auto it = all_fields.begin(); it != all_fields.end(); it++, num_txtfields++) {
             // The PNG specification requires Latin-1 in the 'key' field
             text[num_txtfields].key = (png_charp)(strdup(it->first.left(PNG_KEYWORD_MAX_LENGTH).trim().to_latin_1().data()));
-            text[num_txtfields].text = (png_charp)(it->second.c_str());
+#ifdef PNG_WRITE_iTXT_SUPPORTED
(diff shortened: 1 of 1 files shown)


Deledrius:
Oh, I was looking in the wrong place the whole time.  😳

Sources/Plasma/PubUtilLib/plGImage/plPNG.cpp
```diff
@@ -249,10 +249,15 @@ bool plPNG::IWrite(plMipmap* source, hsStream* outStream, const std::multimap<ST
         for (auto it = all_fields.begin(); it != all_fields.end(); it++, num_txtfields++) {
             // The PNG specification requires Latin-1 in the 'key' field
             text[num_txtfields].key = (png_charp)(strdup(it->first.left(PNG_KEYWORD_MAX_LENGTH).trim().to_latin_1().data()));
-            text[num_txtfields].text = (png_charp)(it->second.c_str());
+#ifdef PNG_WRITE_iTXT_SUPPORTED
             text[num_txtfields].lang = "en-us";  //  Language used in 'text' and 'lang_key'.
             text[num_txtfields].lang_key = "";   //  Translation of 'key' into 'lang', if needed.
             text[num_txtfields].compression = PNG_ITXT_COMPRESSION_NONE;
+#else
+            it->second = it->second.to_latin_1();
```
zrax: This can throw, since `ST::string` expects to hold UTF-8 data.  Instead, the `char_buffer` should be stored off separately, and should point to either the UTF-8 data or Latin-1 data to be written by libpng.

AbigailBuccaneer:
@zrax LGTY?

Deledrius:
Looks great.  👍

PR #572, opened 2018-05-23
Fix compiling with OpenGL

AbigailBuccaneer:
This adds the necessary declarations needed for successful compilation
with OpenGL. It doesn't define them - none of the declarations in this
class are defined, it looks like this is all unfinished stub code.


Deledrius:
You are correct, this is mostly still unfinished code.  @dpogue has been working on getting the OpenGL pipeline added, and we merged this recently to make sure the structural changes don't get too far out of sync with the rest of the codebase while this work continues.

dpogue:
Hi @AbigailBuccaneer :)
The OpenGL stubs here were a starting point for my bigger experiments with adding a full OpenGL rendering pipeline. That ongoing work is on the [`gl-client` branch in my repo](https://github.com/dpogue/Plasma/tree/gl-client) if you're curious. (Caution: that branch gets rebased against master fairly often)

If this PR fixes compilation errors, then I'm :+1: on merging, although it still won't do anything useful until the full GL-pipeline is finished.

PR #574, opened 2018-07-08
Fix build for Appveyor environment.

Deledrius:
This is a collection of fixes to make building under Appveyor's environment possible/easier, and some general updates to the library usages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
+cmake_minimum_required(VERSION 3.5.1)
 project(Plasma)
-cmake_minimum_required(VERSION 2.8.11.2)
 
 include(FeatureSummary)
 
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Plasma currently requires the following third-party libraries:
 The following libraries are optional:
 
 - (for building resource.dat) PyGTK - http://www.pygtk.org/downloads.html
-- (for building resource.dat) PIL - http://www.pythonware.com/products/pil/
+- (for building resource.dat) Pillow - https://python-pillow.org/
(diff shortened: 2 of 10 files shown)


Deledrius:
Note: This does *not* include the appveyor.yaml, as I had intended, because there are a few aspects that need discussion.  I opened this PR to hash that out with the team, after which I can add it as a final commit if desired.

The current configuration is fully functional and builds the entire client, all the tools, and the `resource.dat`, only currently excluding the Max plugin.  In order to accomplish building the client, I supply a zipped archive containing the minimal necessary portions of the PhysX and OpenAL SDKs.  What are our options for doing this as a regular build?  

Connected to the above question is the Max SDKs: is it possible to use them in our public build environment?

zrax:
I have no idea about the proprietary SDKs...  In fact I've been avoiding the Plasma repo for now for exactly that reason.  As for the other libs, I'm working on https://github.com/H-uru/Plasma/PlasmaPrefix to try to make it easier to build those and keep them up-to-date for our supported compilers.

Hoikas:
Is it possible to run the installers for PhysX and OpenAL on AppVeyor?

zrax:
As long as it can be run without user interaction, yes, it should be possible.  I was briefly running the nasm installer like that, and it worked as expected.

Deledrius:
I wasn't able to get them to run that way, unfortunately.  It might be possible, but I did not find a way to do so.  That was the first thing I tried.  They both use weird installers, IIRC.

Hoikas:
That seems like a good compromise solution to me :smile:

Deledrius:
If so, do we also want to build the Max plugin that way too?  We could theoretically build each supported version this way.

PR #577, opened 2018-10-27
Remove references to visitors

Lunanne:
I'm fairly certain these changes are safe. Considering it seems I managed to break my Dirtsand & Plasma installation while trying to remove a bit too many references, I couldn't test it.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyAccountManagement.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyAccountManagement.cpp
@@ -128,11 +128,6 @@ bool cyAccountManagement::IsActivePlayerSet()
     return NetCommGetPlayer()->playerInt != 0;
 }
 
-void cyAccountManagement::UpgradeVisitorToExplorer(unsigned playerId)
-{
-    NetCommUpgradeVisitorToExplorer(playerId, nil);
-}
-
 void cyAccountManagement::ChangePassword(const ST::string& password)
 {
     NetCommChangeMyPassword(password);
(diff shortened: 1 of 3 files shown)


Hoikas:
It looks like you missed the second function? I think it was PtUpgradeVisitorToSubscribor.

Sources/Plasma/FeatureLib/pfPython/cyAccountManagementGlue.cpp
```diff
@@ -241,7 +221,6 @@ void cyAccountManagement::AddPlasmaConstantsClasses(PyObject *m)
     PYTHON_ENUM_START(PtAccountUpdateType);
     PYTHON_ENUM_ELEMENT(PtAccountUpdateType, kCreatePlayer,     plAccountUpdateMsg::kCreatePlayer);
     PYTHON_ENUM_ELEMENT(PtAccountUpdateType, kDeletePlayer,     plAccountUpdateMsg::kDeletePlayer);
-    PYTHON_ENUM_ELEMENT(PtAccountUpdateType, kUpgradePlayer,    plAccountUpdateMsg::kUpgradePlayer);
```
Hoikas: I think I would avoid removing enum items for now

Sources/Plasma/FeatureLib/pfPython/cyAccountManagement.cpp
```diff
@@ -134,11 +128,6 @@ bool cyAccountManagement::IsActivePlayerSet()
     return NetCommGetPlayer()->playerInt != 0;
 }
 
-void cyAccountManagement::UpgradeVisitorToExplorer(unsigned playerId)
-{
-    NetCommUpgradeVisitorToExplorer(playerId, nil);
```
Hoikas: This deletion orphans the function `NetCommUpgradeVisitorToExplorer`. I can't help but wonder if we want to remove that as well? Of course, if we do that, we then orphan `NetCliAuthUpgradeVisitorRequest` and its transaction. Comments from the gallery, @Deledrius, @zrax?

PR #579, opened 2018-12-11
Ensure OpenSSL EVP digests are initialized

Hoikas:
If we don't do this, `EVP_get_digestbyname` returns null and the game crashes on startup.


Hoikas:
Considering that there are so many projects with (Win)Main, I thought it would be a pretty poor showing. I'm not happy about this solution either though. I also didn't like the idea of polluting the applications with direct calls into OpenSSL. Maybe a more generalized plChecksum::Init()?

PR #582, opened 2019-01-21
Several fixes for finding and linking against newer versions of 3rd party libs

zrax:
(as built by H-uru/PlasmaPrefix)


Deledrius:
This appears to have some crossover with #574.

PR #584, opened 2019-06-06
plPythonPack fixes.

Deledrius:
Any `path` other than "." was failing to check the correct location for sub-directory statistics, returning incorrect results.

Also includes a minor fix for using the full local system path.

### DIFF
--- a/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
@@ -95,7 +95,7 @@ void PythonInterface::initPython(const plFileName& rootDir)
             PyList_Append(path_list, more_path3);
             temp = plFileName::Join(rootDir, "system");
             printf("%s\n\n", temp.AsString().c_str());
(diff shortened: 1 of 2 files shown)


Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
```diff
@@ -95,7 +95,7 @@ void PythonInterface::initPython(const plFileName& rootDir)
             PyList_Append(path_list, more_path3);
             temp = plFileName::Join(rootDir, "system");
             printf("%s\n\n", temp.AsString().c_str());
-            PyObject* more_path2 = PyString_FromString("system");
+            PyObject* more_path2 = PyString_FromString(temp.AsString().c_str());
```
Hoikas: If the path contains non-ASCII characters, bad things may happen. But, the rest of the code shares this issue, so do we care?

Deledrius: As long as we're using Python 2.x, I feel like this will be a headache (and one I'm not qualified to solve).

PR #585, opened 2019-06-13
Add Quiet option to silence output from plPythonPack.

Deledrius:
Adds a new option to silence non-error output messages.  A new option parser was added to accommodate this, while being backward-compatible with existing option allowances.

Also removes a few vestigial functions which have been left unused for some time.

### DIFF
--- a/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
@@ -77,9 +77,10 @@ void PythonInterface::initPython(const plFileName& rootDir, FILE* outstream, FIL
         {
             // get the sys's dictionary to find the stdout and stderr
             PyObject* sys_dict = PyModule_GetDict(sysmod);
-            if (sys_dict != nullptr && stdOut != nullptr && stdErr != nullptr)
-            {
+            if (sys_dict != nullptr && stdOut != nullptr) {
                 PyDict_SetItemString(sys_dict, "stdout", stdOut);
+            }
+            if (sys_dict != nullptr && stdErr != nullptr) {
                 PyDict_SetItemString(sys_dict, "stderr", stdErr);
             }
             // NOTE: we will reset the path to not include paths
--- a/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
@@ -47,12 +47,16 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "cStringIO.h"
 #include "plFileSystem.h"
 
-static PyObject* stdFile;   // python object of the stdout and err file
+#include <string_theory/stdio>
+#include <string_theory/string>
 
-void PythonInterface::initPython(const plFileName& rootDir)
+static PyObject* stdOut;    // python object of the stdout file
+static PyObject* stdErr;    // python object of the stderr file
+
+void PythonInterface::initPython(const plFileName& rootDir, FILE* outstream, FILE* errstream)
 {
     // if haven't been initialized then do it
-    if ( Py_IsInitialized() == 0 )
+    if (Py_IsInitialized() == 0)
     {
         // initialize the Python stuff
         // let Python do some intialization...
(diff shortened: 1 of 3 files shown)


Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -426,61 +341,54 @@ void PackDirectory(const plFileName& dir, const plFileName& rootPath, const plFi
...
-            || (temp == "-h") || (temp == "/h"))
-        {
-            PrintUsage();
-            return -1;
+    // Parse arguments
+    ST::string packDir = ".";
+    for (int i = 1; i < argc; i++) {
+        ST::string arg = argv[i];
+
+        if (arg.starts_with("?") || arg.starts_with("-") || arg.starts_with("/")) {
```
dpogue: It might/is probably overkill, but worth remembering that we have [plCmdParser](https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/CoreLib/plCmdParser.h) available. See [the tests](https://github.com/H-uru/Plasma/blob/master/Sources/Tests/CoreTests/test_plCmdParser.cpp) for example usage.

Deledrius: ::sigh:: I *thought* we had something, and I went looking and completely missed that.  Thanks, I should probably use it here. 🤦‍♂

Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
```diff
@@ -61,61 +65,61 @@ void PythonInterface::initPython(const plFileName& rootDir)
...
         {
             // get the sys's dictionary to find the stdout and stderr
             PyObject* sys_dict = PyModule_GetDict(sysmod);
-            if (stdFile != nil)
+            if (sys_dict != nullptr && stdOut != nullptr && stdErr != nullptr)
             {
-                PyDict_SetItemString(sys_dict, "stdout", stdFile);
-                PyDict_SetItemString(sys_dict, "stderr", stdFile);
+                PyDict_SetItemString(sys_dict, "stdout", stdOut);
+                PyDict_SetItemString(sys_dict, "stderr", stdErr);
```
Hoikas: This is really nit-picky, but if `stdOut` is null but `stdErr` is non-null, then `stdErr` is not properly redirected. These should probably be split up behind individual checks since `stdFile` was split.

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -169,21 +160,20 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
...
+                ST::printf(out, "an import file ");
             }
             else
-                fputs("a PythonFile modifier(tm) ", stdout);
+                ST::printf(out, "a PythonFile modifier(tm) ");
         }
         else
         {
-            fputs("......blast! Error during run-code!\n", stdout);
+            ST::printf(stderr, "......blast! Error during run-code in {}!\n", fileName);
```
Hoikas: blast!

Sources/Plasma/Apps/plPythonPack/main.cpp
Hoikas: See comment by @dpogue 👅

Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
Deledrius: Good idea.

Sources/Plasma/Apps/plPythonPack/main.cpp
Deledrius: 🎆

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -169,21 +161,20 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
...
-            int chars_read = PythonInterface::getOutputAndReset(&errmsg);
-            if (chars_read > 0)
-            {
-                puts(errmsg);
-            }
+            ST::string msg;
+            PythonInterface::getOutputAndReset(msg);
+            ST::printf(out, "{}", msg);
+            PythonInterface::getErrorAndReset(msg);
+            ST::printf(stderr, "{}", msg);
```
zrax: `puts()` (unlike `fputs()`) writes a terminating newline after its parameter.  Therefore, we still need to check for non-empty, and add a newline to the `printf` format...

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -194,31 +185,31 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
...
-        int chars_read = PythonInterface::getOutputAndReset(&errmsg);
-        if (chars_read > 0)
-        {
-            puts(errmsg);
-        }
+        ST::string msg;
+        PythonInterface::getOutputAndReset(msg);
+        ST::printf(out, "{}", msg);
+        PythonInterface::getErrorAndReset(msg);
+        ST::printf(stderr, "{}", msg);
```
zrax: See above

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -194,31 +185,31 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
...
```
zrax: See above

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -426,61 +342,69 @@ void PackDirectory(const plFileName& dir, const plFileName& rootPath, const plFi
...
-    fputs("plPythonPack [directory to pack...]\n", stdout);
-    fputs("NOTE: the directory to pack must have full system and plasma dirs and\n", stdout);
-    fputs("      must be a relative path to the current working directory\n", stdout);
+    ST::printf(out, "The Python Pack Utility\n");
+    ST::printf(out, "Usage:\n");
+    ST::printf(out, "plPythonPack [options] <directory to pack...>\n");
+    ST::printf(out, "NOTE: the directory to pack must have full system and plasma dirs.\n");
+    ST::printf(out, "\nAvailable options:\n");
+    ST::printf(out, "\t-q\tQuiet  - Only print errors\n");
+    ST::printf(out, "\t-h\tHelp   - Print this help\n");
```
zrax: Is there any reason `PrintUsage()` would want to be redirected?

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -426,61 +342,69 @@ void PackDirectory(const plFileName& dir, const plFileName& rootPath, const plFi
...
+        { kCmdArgOptional | kCmdTypeString, "path",  kArgPath},
+        { kCmdArgFlagged  | kCmdTypeBool,   "quiet", kArgQuiet},
+        { kCmdArgFlagged  | kCmdTypeBool,   "help",  kArgHelp1},
+        { kCmdArgFlagged  | kCmdTypeBool,   "?",     kArgHelp2},
+    };
+
+    std::vector<ST::string> args;
+    args.reserve(argc);
+    for (size_t i = 0; i < argc; i++) {
+        args.push_back(ST::string::from_utf8(argv[i]));
```
zrax: `emplace_back` to avoid copying a temporary

Sources/Plasma/Apps/plPythonPack/main.cpp
Deledrius: Other than copy & paste, I can't think of anything.  In fact it feels like it would be pretty nonsensical to be asked for help and then keep quiet about it.

Sources/Plasma/Apps/plPythonPack/main.cpp
Deledrius: :trollface: 
https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/Apps/plUruLauncher/plClientLauncher.cpp#L458-L462

Sources/Plasma/Apps/plPythonPack/main.cpp
Deledrius: 👍

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -367,22 +368,37 @@ int main(int argc, char *argv[])
...
+
+    plCmdParser cmdParser(cmdLineArgs, arrsize(cmdLineArgs));
+    if (cmdParser.Parse(args)) {
+        if (cmdParser.GetBool(kArgHelp1) || cmdParser.GetBool(kArgHelp2)) {
+            PrintUsage();
+            return 0;
         }
+
+        if (cmdParser.GetBool(kArgQuiet))
+            out = fopen(NULL_DEVICE, "w");
```
Hoikas: This appears to be leaked at program exit. Not sure how much we care about that though?

Deledrius: Amateur mistake.  Who wrote this code?

Deledrius: I have added code to `fclose` what I have `fopen`ed.

Deledrius: @zrax, do you have an opinion on this matter?

zrax: I don't think it's super important to `fclose()` a handle to `/dev/null`...  But it's fine do to so, and your implementation has some symmetry to it.  If you want to get really fancy, you could automatically close it with a `std::unique_ptr<FILE, decltype(&fclose)>`, but then you either have to fclose `stdout` (should be safe since you only do it after the application exits) or scope your unique_ptr differently from the `out` handle.

tl;dr:  It looks fine the way you currently implemented it ;)

PR #587, opened 2019-07-03
Require string_theory 2.0 and update use of deprecated 1.x APIs.

zrax:


Sources/Plasma/FeatureLib/pfPython/plPythonParameter.h
```diff
@@ -377,9 +377,9 @@ typedef struct plPythonParameter
                 if ( count != 0 )
                 {
                     ST::char_buffer str;
-                    char *buffer = str.create_writable_buffer(count-1);
-                    stream->ReadLE(count, buffer);
-                    buffer[count-1] = 0;
+                    str.allocate(count - 1);
+                    stream->ReadLE(count - 1, str.data());
+                    (void)stream->ReadByte();
```
Hoikas: Well, there goes my evil corrupted PRPs with non-null terminated plPythonParameter strings...

Sources/Plasma/FeatureLib/pfPython/pyVault.cpp
```diff
@@ -590,8 +590,7 @@ void pyVault::CreateNeighborhood()
     ST::string title;
     ST::string desc;
 
-    unsigned nameLen = nc->GetPlayerName().size();
-    if (nc->GetPlayerName().char_at(nameLen - 1) == 's' || nc->GetPlayerName().char_at(nameLen - 1) == 'S')
+    if (nc->GetPlayerName().back() == 's' || nc->GetPlayerName().back() == 'S')
```
Hoikas: Well, that's certainly less verbose.

Sources/Plasma/NucleusLib/pnNetProtocol/Private/pnNpCommon.cpp
```diff
@@ -499,9 +499,9 @@ inline void IRead<ST::string>(const uint8_t*& buf, ST::string& dest)
     buf += sizeof(uint32_t);
 
     ST::utf16_buffer str;
-    char16_t* theStrBuffer = str.create_writable_buffer(nChars);
-    memcpy(theStrBuffer, buf, size);
-    theStrBuffer[nChars] = 0;
+    str.allocate(nChars);
+    memcpy(str.data(), buf, size);
+    str[nChars] = 0;
```
Hoikas: This is playing with my mind a little bit since we're juggling both the buffer size and the string length. Perhaps it would be more approachable if we only memcpy'd nChars and removed line 504 altogether?

Sources/Plasma/FeatureLib/pfPython/plPythonParameter.h
zrax: It was actually protected in the old version as well, with the `buffer[count-1] = 0` line...

Sources/Plasma/NucleusLib/pnNetProtocol/Private/pnNpCommon.cpp
zrax: Agreed, and it would match the other updates better too.
+Fixed

Sources/Plasma/FeatureLib/pfPython/plPythonParameter.h
Hoikas: Well I'll be dawg.

PR #590, opened 2019-09-07
Add AppVeyor configuration file.

Deledrius:
This configuration enables a complete build for Internal/External and Debug/Release of the client, tools, tests, and `resource.dat`.  Only the Max plugin is currently excluded.

It uses vcpkg to provide most of the required libraries, and fills in the rest with an externally-hosted archive containing the minimal necessary portions of the PhysX and OpenAL SDKs due to a lack of automation support in their installers.  This configuration uses a custom triplet for vcpkg: `x86-windows-static-dyncrt`.  The PyGTK installer is patched to automatically enable the RSVG option, used to render `resource.dat`.  

Finally, any necessary Qt DLLs are included in the produced artifacts.


appveyor.yml
```diff
@@ -0,0 +1,91 @@
+version: 1.0-{build}
+image:
+- Visual Studio 2015
+configuration:
+- Debug
+- Release
+clone_depth: 1
+environment:
+  QT_VER: 5.11
```
zrax: Why not 5.12 (LTS)?

appveyor.yml
```diff
@@ -0,0 +1,91 @@
+version: 1.0-{build}
+image:
+- Visual Studio 2015
```
zrax: MSVC 2015 libs are compatible with MSVC 2017 and 2019 (see https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2019)...  2019 is only at preview support in AppVeyor, but I feel like we should at least target 2017 if we can...

appveyor.yml
```diff
@@ -0,0 +1,91 @@
...
+
+    Set-Location ..
+
+    refreshenv
+cache: C:\tools\vcpkg\installed
+before_build:
+- ps: |
+    Set-Location build
+
+    cmake -Wno-dev -DCMAKE_TOOLCHAIN_FILE=C:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x86-windows-static-dyncrt -DCMAKE_INSTALL_PREFIX=devlibs -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=bin -DPHYSX_SDK_PATH=devlibs\PhysX -DQt5_DIR=C:\Qt\$Env:QT_VER\msvc2015\lib\cmake\Qt5 -DPLASMA_EXTERNAL_RELEASE="$Env:PLASMA_EXTERNAL_RELEASE" -DPLASMA_BUILD_TESTS=ON -DPLASMA_BUILD_TOOLS=ON -DPLASMA_BUILD_RESOURCE_DAT=ON ..
```
zrax: Some line breaks would be desirable here :)

appveyor.yml
Deledrius: 5.12 is not available for VS2015 in a 32-bit version: https://www.appveyor.com/docs/windows-images-software/#qt

Also, I wasn't aware of the LTS versioning for Qt.

appveyor.yml
Deledrius: I can bump it to VS2017 if desired.  It's been tested and does appear to work.  That is also the last version with support for the XP Toolset.

appveyor.yml
Deledrius: Sure, I suppose if you're the sort who desires readability! ;)

appveyor.yml
```diff
@@ -0,0 +1,93 @@
...
+
+    refreshenv
+cache: C:\tools\vcpkg\installed
+before_build:
+- ps: |
+    Set-Location build
+
+    cmake -Wno-dev -DCMAKE_TOOLCHAIN_FILE=C:/tools/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x86-windows-static-dyncrt `
+    -DCMAKE_INSTALL_PREFIX=devlibs -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=bin -DPHYSX_SDK_PATH=devlibs\PhysX -DQt5_DIR=C:\Qt\$Env:QT_VER\msvc2015\lib\cmake\Qt5 `
+    -DPLASMA_EXTERNAL_RELEASE="$Env:PLASMA_EXTERNAL_RELEASE" -DPLASMA_BUILD_TESTS=ON -DPLASMA_BUILD_TOOLS=ON -DPLASMA_BUILD_RESOURCE_DAT=ON -T v141_xp ..
```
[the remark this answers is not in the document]
Deledrius: I believe I would need to also include the `VCPKG_PLATFORM_TOOLSET` setting in the triplet to be fully correct.  

Either way, I've removed this.

PR #591, opened 2019-09-09
Opus Voice Chat

Hoikas:
This effectively obsoletes the need for speex in the client by adding support for opus voice chat packets. The opus data handler has a small workaround built in to ensure old (read: MOULa) clients do not crash when receiving new-fangled opus audio packets. In my limited testing, the generated opus packets are approximately 10 bytes shorter than equivalent speex packets. You can toggle the encoder by using the new `Audio.SetVoiceCodec` console commands. Valid options are `speex`, `opus`, and `none`.

This pull request depends on H-uru/Plasma#589.

Things needed before this PR is ready to go:
- [x] Verify the library cmake support makes sense.
- [x] Allow opus to decode audio at 48,000 Hz, allowing higher quality audio streams that may not be voice data.
- [x] Allow the voice recorder to capture voice data at any sampling rate.
- [x] Experiment with increasing the opus encoder's sampling rate to, say, 16000 Hz and check bandwidth requirements.


Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
```diff
@@ -987,9 +984,78 @@ bool plAudioSystem::MsgReceive(plMessage* msg)
...
+    if (!fCaptureDevice)
+        return false;
+
+    alcCaptureStop(fCaptureDevice);
+
+    // discard any samples not processed
+    ALCsizei samples;
+    alcGetIntegerv(fCaptureDevice, ALC_CAPTURE_SAMPLES, sizeof(samples), &samples);
+    if (samples) {
+        std::unique_ptr<int16_t> buf{ new int16_t[samples] };
```
zrax: I believe we require C++14 by now, so this would be better as `std::make_unique<int16_t[]>(samples)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
```diff
@@ -139,251 +167,227 @@ void plVoiceRecorder::SetQuality(int quality)
...
     fSound.SetConeOrientation(pos.fX, pos.fY, pos.fZ);
 }
 
+plVoiceDecoder* plVoicePlayer::GetDecoder(uint8_t voiceFlags) const
+{
+    if (voiceFlags & plVoiceFlags::kEncodedOpus)
+        return fOpusDecoder;
+    if (!(voiceFlags & plVoiceFlags::kEncoded))
+        return nullptr;
+    return plVoiceDecoder::GetSpeex();
```
zrax: You have a flag specifically for Speex, why not check it, and leave `nullptr` as the default/fallback case?

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+    return true;
+}
+
+bool plSpeex::Encode(const short* data, int numFrames, int& packedLength, void* out, int outsz)
+{
+    packedLength = 0;
+
+    const short* pData = data;                               // pointer to input data
+    std::unique_ptr<float> input{ new float[fFrameSize] };   // input to speex - used as am intermediate array since speex requires float data
```
zrax: `auto input = make_unique<float[]>(fFrameSize)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+    uint8_t frameLength;                                     // number of bytes speex compressed frame to
+    std::unique_ptr<char> frameData{ new char[fFrameSize] }; // holds one frame of encoded data
```
zrax: `auto frameData = make_unique<char[]>(fFrameSize)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+    // encode data
+    hsRAMStream stream;
+    for (int i = 0; i < numFrames; i++) {
+        // convert input data to floats
+        for (int j = 0; j < fFrameSize; j++)
+            input.get()[j] = pData[j];
```
zrax: When using `unique_ptr<T[]>`, an `operator[]` is provided so you don't need to use `.get()`

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+    hsAssert(stream.GetEOF() <= outsz, "Speex packet will be truncated");
+    stream.Rewind();
+    stream.Read(std::min(packedLength, outsz), out);
+
+    return true;
+}
+
+bool plSpeex::Decode(const void* data, int size, int numFrames, int& numOutputBytes, short* out)
+{
+    if (!fInitialized) return false;
```
zrax: Newline

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+    numOutputBytes = 0;
+
+    hsReadOnlyStream stream(size, data);
+    std::unique_ptr<float> speexOutput{ new float[fFrameSize] };        // holds output from speex
```
zrax: `auto speexOutput = std::make_unique<float[]>(fFrameSize)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+    size_t outputCount = (numFrames * fFrameSize);                      // length of decoded voice data(out) in samples
+    short* pOut = out;                                                  // pointer to output short buffer
+
+    // create buffer for input data
+    std::unique_ptr<char> frameData{ new char[fFrameSize] };            // holds the current frames data to be decoded
```
zrax: `auto frameData = std::make_unique<char[]>(fFrameSize)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
```diff
@@ -0,0 +1,518 @@
...
+    for (int i = 0; i < numFrames; i++) {
+        stream.ReadLE(&frameLen);                                       // read the length of the current frame to be decoded
+        stream.Read(frameLen, frameData.get());                         // read the data
+
+        memset(speexOutput.get(), 0, fFrameSize * sizeof(float));
+        speex_bits_read_from(fBits.get(), frameData.get(), frameLen);   // give data to speex
+        speex_decode(fDecoderState, fBits.get(), speexOutput.get());    // decode data
+
+        for (int j = 0; j < fFrameSize; j++)
+            pOut[j] = (short)(speexOutput.get()[j]);                    // convert floats to shorts
```
zrax: Remove `.get()`

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
Hoikas: Is `std::make_unique` available in VC++2013?

Sources/Plasma/PubUtilLib/plAudible/plWinAudible.cpp
```diff
@@ -664,7 +664,7 @@ bool pl2WayWinAudible::MsgReceive(plMessage* msg)
     if (pCtrlMsg)
     {
         if (pCtrlMsg->GetControlCode() == S_PUSH_TO_TALK && fVoiceRecorder)
-            fVoiceRecorder->SetMikeOpen(pCtrlMsg->ControlActivated());
+            fVoiceRecorder->SetMicOpen(pCtrlMsg->ControlActivated());
```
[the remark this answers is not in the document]
Hoikas: I'm not sure WHO Mike is, but he must like being opened and closed...

Sources/Plasma/PubUtilLib/plAudio/plVoiceCodec.cpp
Hoikas: Kewl

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
Hoikas: The Cyan standard client sends speex data flagged only as `kEncoded`. I wanted to be able to handle this data gracefully.

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
```diff
@@ -139,251 +167,227 @@ void plVoiceRecorder::SetQuality(int quality)
...
             // cap uncompressed data
-            if(totalSamples > MAX_DATA_SIZE)
-                totalSamples = MAX_DATA_SIZE;
+            totalSamples = std::min(totalSamples, MAX_DATA_SIZE);
 
             // convert to correct units:
-            short *buffer = new short[totalSamples];
-
-            alcCaptureSamples(captureDevice, buffer, totalSamples);
+            std::unique_ptr<int16_t> buffer{ new int16_t[totalSamples] };
```
zrax: `auto buffer = std::make_unique<int16_t[]>(totalSamples)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
```diff
@@ -139,251 +167,227 @@ void plVoiceRecorder::SetQuality(int quality)
...
-                plSpeex::GetInstance()->Encode(buffer, numFrames, &packedLength, &ram);
-
-                if (packedLength)
-                {
-                    // extract data from ram stream into packet
-                    ram.Rewind();
-                    ram.Read(packedLength, packet);
+                bytesSent = totalSamples * sizeof(int16_t);
+            } else {
+                std::unique_ptr<uint8_t> packet{ new uint8_t[totalSamples] }; // packet to send encoded data in
```
zrax: `auto packet = std::make_unique<uint8_t[]>(totalSamples)`

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
```diff
@@ -139,251 +167,227 @@ void plVoiceRecorder::SetQuality(int quality)
...
-        if(!plSpeex::GetInstance()->Decode((uint8_t *)data, size, numFramesInBuffer, &numBytes, nBuff))
-        {
-            delete[] nBuff;
-            return;
+    if (flags & plVoiceFlags::kEncoded) {
+        plVoiceDecoder* decoder = GetDecoder(flags);
+        if (decoder) {
+            int numBytes;
+            int bufferSize = numFramesInBuffer * decoder->GetFrameSize();
+            std::unique_ptr<short> nBuff{ new short[bufferSize] };
```
zrax: `make_unique<short[]>(bufferSize)`

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
zrax: Hmm, good question.  Is there a reason we still support VC++2013 (6 years old now)?  And can we... not?

zrax:
This should probably also be rebased so the CMake changes get CI tested in both MSVC and Linux

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
Hoikas: I just looked it up in a huge list on the MS website, and it appears make_unique was added in VC++ 2013. Bumping the compiler requirement is probably not a bad idea at this stage, I just preferred for my PR to not depend on that 😄

Sources/Plasma/PubUtilLib/plAudio/plVoiceChat.cpp
Hoikas: Seriously? How did I miss these....

Hoikas:
Uncompressed audio is still handled. `plVoicePlayer::PlaybackVoiceMessage` calls `PlaybackUncompressedVoiceMessage` if the `kEncoded` flag is not set. You were probably seeing the case when `GetDecoder()` returns `nullptr`--that happens when someone sends us data encoded in a format the client does not have library support for.

zrax:
Got it, thanks.  I was actually looking at the "removal" of `PlaybackUncompressedVoiceMessage` in one context, but I see you just moved it down to a lower level.

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
zrax: Ah, good enough for me :)

Hoikas:
Fixed some linker problems. What fun!

PR #576, opened 2018-10-24
Coverity fixes for uninitialized members.

Deledrius:
Fix uninitialized members reported by Coverity analysis.  Minor formatting changes included where relevant to match our coding style when possible.

PR #595, opened 2019-09-14
Assorted Warning Cleanup

Hoikas:
This changeset attempts to silence and/or fix some of the warnings emitted during compilation. When I began, 406 warnings were being emitted in my build. It is now down to 313 warnings. I avoided fixing uninitialized members to prevent conflicts with #576.

### DIFF
--- a/Sources/Plasma/Apps/plClient/plClient.cpp
+++ b/Sources/Plasma/Apps/plClient/plClient.cpp
@@ -517,8 +517,8 @@ bool plClient::InitPipeline()
 
     float   yon = 500.0f;
 
-    pipe->SetFOV( 60.f, int32_t( 60.f * pipe->Height() / pipe->Width() ) );
-    pipe->SetDepth( 0.3f, yon );
+    pipe->SetFOV(60.f, 60.f * (float)pipe->Height() / (float)pipe->Width());
+    pipe->SetDepth(0.3f, yon);
 
     hsMatrix44 id;
     id.Reset();
--- a/Sources/Plasma/CoreLib/hsBitVector.h
+++ b/Sources/Plasma/CoreLib/hsBitVector.h
@@ -116,23 +116,19 @@ public:
 
 inline hsBitVector::hsBitVector(const hsBitVector& other)
 {
-    if( 0 != (fNumBitVectors = other.fNumBitVectors) )
-    {       
+    if ((fNumBitVectors = other.fNumBitVectors) != 0) {
         fBitVectors = new uint32_t[fNumBitVectors];
-        int i;
-        for( i = 0; i < fNumBitVectors; i++ )
+        for (uint32_t i = 0; i < fNumBitVectors; i++)
             fBitVectors[i] = other.fBitVectors[i];
+    } else {
+        fBitVectors = nullptr;
     }
-    else
-        fBitVectors = nil;
 }
 
 inline bool hsBitVector::Empty() const
 {
-    int i;
-    for( i = 0; i < fNumBitVectors; i++ )
-    {
-        if( fBitVectors[i] )
+    for (uint32_t i = 0; i < fNumBitVectors; i++ ) {
+        if (fBitVectors[i])
             return false;
     }
     return true;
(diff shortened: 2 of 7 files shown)


Deledrius:
🎖

Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -517,8 +517,8 @@ bool plClient::InitPipeline()
 
     float   yon = 500.0f;
 
-    pipe->SetFOV( 60.f, int32_t( 60.f * pipe->Height() / pipe->Width() ) );
-    pipe->SetDepth( 0.3f, yon );
+    pipe->SetFOV(60.f, 60.f * (pipe->Height() / pipe->Width()));
```
zrax: this is `float * (int / int)`, which will cause truncation in the divide.  It therefore needs to be either without parens, or with parens around the multiplication.

Sources/Plasma/CoreLib/hsBitVector.h
```diff
@@ -116,137 +116,114 @@ class hsBitVector {
 
 inline hsBitVector::hsBitVector(const hsBitVector& other)
 {
-    if( 0 != (fNumBitVectors = other.fNumBitVectors) )
-    {       
+    if ((fNumBitVectors = other.fNumBitVectors)) {
```
zrax: Personally I prefer to keep the extra comparison when an assignment is in an if statement...  Or else to separate the assignment out of the comparison, so it's clear the assignment is not a typo.

Sources/Plasma/CoreLib/hsBitVector.h
```diff
@@ -116,137 +116,114 @@ class hsBitVector {
...
 {
-    if( fNumBitVectors < other.fNumBitVectors )
+    if (fNumBitVectors < other.fNumBitVectors)
         return other.operator==(*this);
-    int i;
-    for( i = 0; i < other.fNumBitVectors; i++ )
-        if( fBitVectors[i] ^ other.fBitVectors[i] )
+    uint32_t i;
+    for (i = 0; i < other.fNumBitVectors; i++)
+        if (fBitVectors[i] ^ other.fBitVectors[i])
```
zrax: AFAICT, this is an attempted (and unnecessary) optimization of `!=`  >.>

Sources/Plasma/CoreLib/hsBitVector.h
```diff
@@ -116,137 +116,114 @@ class hsBitVector {
...
-    {
+    if (this == &other) {
         Clear();
         return *this;
     }
 
-    int minNum = fNumBitVectors < other.fNumBitVectors ? fNumBitVectors : other.fNumBitVectors;
-    int i;
-    for( i = 0; i < minNum; i++ )
+    uint32_t minNum = fNumBitVectors < other.fNumBitVectors ? fNumBitVectors : other.fNumBitVectors;
```
zrax: `std::min`

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -1809,68 +1809,50 @@ static uint32_t   IConvertHex( const wchar_t *str )
...
-
-                // Special case to remove any last trailing carriage return
-//              if( count > 1 && lastParChunk->fText[ count - 1 ] == '\n' )
-//                  lastParChunk->fText[ count - 1 ] = 0;
-
-                fHTMLSource.Append( lastParChunk );
+                lastParChunk = nullptr;
+            } else if(lastParChunk) {
+                size_t count = ((uintptr_t)c - (uintptr_t)start) / sizeof(wchar_t); // wchar_t is 2 bytes
+                lastParChunk->fText.assign(start, 0, count);
```
zrax: 🥇

Sources/Plasma/CoreLib/hsBitVector.h
Hoikas: I was wondering what in the devil was going on here. Will fix.

Sources/Plasma/CoreLib/hsBitVector.h
Hoikas: I found this line to be confusing to begin with, so that makes sense. Will fix.

Sources/Plasma/CoreLib/hsBitVector.h
Hoikas: Do we want to include the algorithm header in this header?

Hoikas:
Pushed fixes for all, save the one question re `std::min`.

Sources/Plasma/CoreLib/hsBitVector.h
zrax: Nah, probably not worth it here...

PR #596, opened 2019-09-15
Audio System Improvements

Hoikas:
This adds flexibility to the audio system by allowing any endpoint available to OpenAL for both playback and capture. This should fix several assumptions that prevented OpenAL Soft from working properly as well. Additionally, the broken "My Voice" slider in the audio options dialog has been fixed by rewriting the `plWinMicLevel` class to target Windows Core Audio.

In my testing, MOULa ships with an oddball OpenAL32.dll that does not support the `ALC_ENUMERATE_ALL_EXT`, so only the default playback endpoint is available. Also, there appears to be a bug in the Creative OpenAL implementation that causes an access violation when accessing the `ITaskbarList3` pointer after changing the playback device. OpenAL Soft does not display this bug. ~~Finally, OpenAL Soft 1.19.1 appears to have an issue in its WASAPI backend that causes audio capturing at any sampling rate *other than* 44.1kHz to have discernible distortion. This issue is not present in OpenAL Soft master and can be worked around in 1.19.1 by disabling the WASAPI backend in favor of the DirectSound or WinMM backend. Since Windows Vista, DirectSound and WinMM are just layers on top of Windows Core Audio WASAPI, so this is effectively bypassing an AL Soft issue.~~

~~This depends on H-uru/moul-scripts#111.~~


Sources/Plasma/FeatureLib/pfPython/pyAudioControlGlue.cpp
```diff
@@ -429,6 +455,9 @@ PYTHON_START_METHODS_TABLE(ptAudioControl)
     PYTHON_METHOD_NOARGS(ptAudioControl, getPriorityCutoff, "Returns current sound priority"),
     PYTHON_METHOD(ptAudioControl, setPriorityCutoff, "Params: priority\nSets the sound priority"),
     PYTHON_METHOD(ptAudioControl, enableVoiceChat, "Params: state\nEnables or disables voice chat."),
+    PYTHON_METHOD(ptAudioControl, setCaptureDevice, "Sets the name for the capture device being used by the audio system"),
```
Deledrius: This is confusingly worded.  It might be better to say "Sets the capture device by name [...]"

Sources/Plasma/FeatureLib/pfPython/pyAudioControlGlue.cpp
```diff
@@ -479,13 +403,16 @@ PYTHON_START_METHODS_TABLE(ptAudioControl)
...
                 "...Less of a performance hit, harder on memory."),
-    PYTHON_METHOD(ptAudioControl, useHardwareAcceleration, "Params: state\nEnables or disables audio hardware acceleration."),
-    PYTHON_METHOD_NOARGS(ptAudioControl, isHardwareAccelerated, "Is audio hardware acceleration enabled? Returns 1 if true otherwise returns 0."),
     PYTHON_METHOD(ptAudioControl, useEAXAcceleration, "Params: state\nEnables or disables EAX sound acceleration (requires hardware acceleration)."),
     PYTHON_METHOD_NOARGS(ptAudioControl, isUsingEAXAcceleration, "Is EAX sound acceleration enabled? Returns 1 if true otherwise returns 0."),
+    PYTHON_METHOD_NOARGS(ptAudioControl, isEAXSupported, "Is EAX acceleration supported by the current device?"),
     PYTHON_BASIC_METHOD(ptAudioControl, muteAll, "Mutes all sounds."),
     PYTHON_BASIC_METHOD(ptAudioControl, unmuteAll, "Unmutes all sounds."),
     PYTHON_METHOD_NOARGS(ptAudioControl, isMuted, "Are all sounds muted? Returns 1 if true otherwise returns 0."),
+    PYTHON_METHOD(ptAudioControl, setPlaybackDevice, "Params: devicename,restart\nSets the device name for the audio system, and optionally restarts it"),
```
Deledrius: Same as above.  Perhaps "Sets the audio system output device by name [...]"

Hoikas:
This is ready for playtime!

Hoikas:
This has been rebased to facilitate review.

Hoikas:
Updated the PR text due to the release of OpenAL Soft 1.20.

Hoikas:
Rebased after merge of #612.

zrax:
Looks like the rebased version needs a `#include <set>`

Hoikas:
I seem to have messed up the rebase and gained an extra definition for `plAudioSystem`. That should be fixed now.

Hoikas:
rebased to fix chaos from astyling and variadic plStatusLog

Hoikas:
Looks like the linux CI failure is a github bug, not a necessarily a problem with this PR?

Hoikas:
Rebased to fix the linux-ci issue and bring in the scripts portion.

Deledrius:
It might be a good idea to add `openal-soft` to the vcpkg install step in our appveyor.yml.

Hoikas:
As I mentioned in IRC the other day, openal-soft is now a C++ library, and CMake's FindOpenAL does not handle separate debug/release libs. This causes an error when mixing debug and release builds of the client an alsoft. We could fix this by using the `OpenAL::OpenAL` thingo exposed by vcpkg though. Thoughts?

zrax:
> We could fix this by using the `OpenAL::OpenAL` thingo exposed by vcpkg though. Thoughts?

Does the CMake import target actually fix it?  If so, I'm in favor of using it (and therefore dropping support for legacy OpenAL), making openal-soft a hard requirement.

Hoikas:
> Does the CMake import target actually fix it?

I had to test to make sure. From what I can tell, we have to explicitly opt-out of the cmake finder module by adding `NO_MODULE` to the `find_package()` line. Adding `target_link_libraries(plAudio OpenAL::OpenAL)` fixes building in Release and Debug mode, however it breaks the RelWithDebInfo target. I suspect that can be fixed (somehow) though. Also, I'm not sure if vcpkg or alsoft itself is generating this cmake confiig. *shrug*

Deledrius:
> I'm in favor of using it (and therefore dropping support for legacy OpenAL), making openal-soft a hard requirement.

That's acceptable to me.

vcpkg.txt
```diff
@@ -0,0 +1,15 @@
...
+libogg
+libpng
+libvorbis
+libvpx
+openal-soft
+openssl
+opus
+speex
+string-theory
+zlib
```
zrax, line 15: It seems this file isn't used?

Sources/Plasma/PubUtilLib/plAudio/plEAXEffects.cpp
```diff
@@ -326,31 +326,31 @@ void    plEAXListener::ProcessMods( hsTArray<plEAXListenerMod *> &modArray )
...
 
         // Accumulate settings from all the active listener regions (shouldn't be too many, we hope)
         totalStrength = 0.f;
         firstOne = true;
-        for( i = 0; i < modArray.GetCount(); i++ )
+        for (auto mod : modArray)
         {
             float strength = modArray[ i ]->GetStrength();
-            DebugLog( "{4.2f} - {}", strength, modArray[ i ]->GetKey()->GetUoid().GetObjectName() );
+            DebugLog( "{4.2f} - {}", strength, modArray->GetKey()->GetUoid().GetObjectName() );
```
zrax, line 353: `mod->GetKey()...`

Hoikas:
Removed unused file and fixed a few would-be compile errors in ifdef'd code as pointed out in review.

Hoikas:
:boom:

PR #597, opened 2019-09-15
Simplify CMake header configuration a bit

zrax:
* Remove two configuration values no longer used
* Merge hsCompilerSpecific.h into hsConfig.h
* Use modern CMake mechanism for setting C++ standard


Hoikas:
I had intended hsConfig to be separate from the compiler specific stuff. I wanted to limit hsConfig to being included in cpp files in so that changes to cmake configuration items would not retrigger a rebuild of nearly all of the code due to our horrible header dependencies. I was specifically think about moving the project level defines into hsConfig.

zrax:
Are you expecting the items in that file to be changed often?  What you have so far I would expect to be very rare to change within the lifetime of a given build configuration.

Hoikas:
Indeed, the items currently in there shouldn't change much at all. I was thinking that a good candidate for the file would be the `PLASMA_EXTERNAL_RELEASE` definition. While that does not change often either, any time it gets toggled presently causes a complete rebuild of the entire codebase. I was hoping to have a place to put definitions that, while they may not change often, changing them would not trigger rebuilding all the units.

zrax:
IMO it makes much more sense to have two build directories for Internal vs. External than toggle it in CMake anyway...  At least, that's what I tend to do.

Hoikas:
That makes sense. I guess I'm just being overzealous about headers.

PR #598, opened 2019-09-16
Fix forward declarations for ALSoft

Hoikas:
In OpenAL Soft master, the delcarations of `ALdevice` and `ALcontext` do not match the declarations found in the Creative OpenAL 1.1 SDK, resulting in a compile error. This should fix our forward declarations to match both.


Hoikas:
This doesn't work for Creative AL. I'll just hack my ALSoft headers for now.

PR #600, opened 2019-09-18
Use ST formatting for console printing, and clean up some unnecessary sprintf

zrax:
This is the first step towards cleaning up the console with string_theory


Sources/Plasma/FeatureLib/pfConsole/pfCCRConsoleCommands.cpp
```diff
@@ -62,8 +62,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 PF_CONSOLE_FILE_DUMMY(CCR)
 //// DO NOT REMOVE!!!!
```
Hoikas: *finger hovers over the delete key*

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.h
```diff
@@ -282,4 +284,10 @@ class pfConsoleCmdIterator
 #define PF_CONSOLE_FILE_DUMMY( name ) \
     void _console_##name##_file_dummy() { }
 
+template <typename... Args>
+void pfConsolePrintF(void pfun(const char *), const char *fmt, Args &&...args)
```
Hoikas: Style nit (pointer binds to the type, not the argument)

zrax: I tend to follow the prevailing style in the modified file for pointer/reference alignment...  In this case, the vast majority of the file binds to the name, not the type.

Hoikas: I guess we can save the formatting fixes for later in that case.

PR #602, opened 2019-10-17
Fix newly uncovered issue with string_theory 3.0

zrax:
Additional uses of std::min run into conflicts with Windows headers in plDXEnumerate.cpp.  This excludes the braindead Windows macro.


zrax:
Updated to use `hsWindows.h` instead of defining `NOMINMAX` directly.

PR #606, opened 2019-10-18
Don't rely on the Max SDK being in a specific directory name.

zrax:
This changes 3dsm_PATH to point to the actual SDK directory, rather than a directory containing the SDK directory.


Hoikas:
Makes sense, considering the 64-bit SDK is in a subdirectory of maxsdk.

PR #607, opened 2019-10-18
VC++2013 does not support constexpr scalars.

Hoikas:
Sad. We can always change it back if/when we upgrade to VC++2015...


zrax:
I've actually started working on the VC++2015 stuff...  I may open an issue to discuss the various tasks and decision points.  However, I don't have any objections to this being merged before we're ready to roll out the compiler upgrade.

Hoikas:
That sounds good to me. I would prefer for us to upgrade deliberately with the benefits of vcpkg :smile:

PR #610, opened 2019-11-11
Fix PBML parsing errors introduced in #595.

Hoikas:
Caused Yessha Book pages to be blank https://cdn.discordapp.com/attachments/282045216221822978/643491894743007242/unknown.png


zrax:
The build failure appears to be an AppVeyor issue uploading the artifact.  I'd call it a pass.

PR #612, opened 2020-01-06
Fix compatibility with OpenAL Soft 1.20

zrax:
OpenAL Soft 1.20 changed the ALCdevice and ALCcontext typedefs.

Since the `<alc.h>` header is very small and does not include any other headers, we can just include it directly where necessary instead of trying to forward-declare the types.


zrax:
Note: This will conflict with #596.

Hoikas:
The conflict with #596 can be resolved easily, I think. I'd say go ahead and merge this.

Issue #608, opened 2019-10-18
Upgrade minimum compiler requirement to VC++ 2015 for Windows.


zrax:
This was discussed briefly in IRC.  The main advantages are the ability to use vcpkg for dependencies and full C++11 / almost-complete C++14 support.  The main disadvantages are that it requires a build system change for anyone using the current recommendations, including potentially downloading and installing a new version of Visual Studio.

Below are the tasks I know of; feel free to add and modify as necessary:
- [X] Fix any issues building with vcpkg dependencies and VC++ 2015
- [x] Update Quick-Start process (prepare_env.ps1) to use VC++ 2015 and vcpkg.  Some options for consideration:
  1. Require the user to install vcpkg and make it available in PATH (as we do with CMake), then configure our required x86-static-dyncrt triplet and build the necessary packages as part of bootstrapping.  This needs to not conflict with other things the user already has done or configured with their vcpkg...
  2. Clone our own vcpkg, bootstrap it, build the required libs.
  3. Skip vcpkg altogether and provide a pre-built lib bundle (closest to what we do today).
- [x] Update README.md with the new requirements and build instructions
- [x] Update any wiki pages or pinned forum topics with specifics about the new build requirements.

zrax:
I feel like this is about ready to be marked complete...  However, I'll leave it open for a while in case there are issues or pieces I missed.

Hoikas:
Looks complete to me.

Hoikas:
@zrax We haven't seen any issues on Gehn 27 (VC++2019), and there haven't been any complaints. Think it's time to close this?

PR #614, opened 2020-01-14
Update build instructions and devlibs for VS 2015

zrax:
This implements parts of #608


Deledrius:
Looks good.  The page at http://h-uru.github.io/Plasma/ will need to be updated to use the correct links as well.

PR #616, opened 2020-01-16
Remove unused plDiffBuffer and plBSDiffBuffer

zrax:


Deledrius:
Do we know what were these used for?

dpogue:
These were used for the .pat patch files back in the Prologue (or earlier days).

.pat (patch) files were a special type of PRP that contained binary object diffs instead of KeyedObjects. All of the code for handling that was thrown out in the plRegistry refactor between PotS and MOUL though, so I think it's probably fine to trash these.

PR #617, opened 2020-01-16
Clean up some GCC warnings

zrax:
This is far from all the warnings exposed by `-Wall -Wextra`, but it does catch several "low-hanging fruit" legitimate issues.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
@@ -273,17 +273,6 @@ PYTHON_METHOD_DEFINITION(ptVaultNode, getNode, args)
     return self->fThis->GetNode2(id);
 }
 
-PYTHON_METHOD_DEFINITION(ptVaultNode, getChildNode, args)
-{
-    unsigned long id;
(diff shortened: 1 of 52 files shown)


Hoikas:
It looks like this should be merged before #618 to prevent conflicts?

zrax:
Yep, I knew going in that one or the other would have to be rebased...  But I didn't want to bundle them together for review either...

Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
```diff
@@ -273,17 +273,6 @@ PYTHON_METHOD_DEFINITION(ptVaultNode, getNode, args)
...
 
-PYTHON_METHOD_DEFINITION(ptVaultNode, getChildNode, args)
-{
-    unsigned long id;
-    if (!PyArg_ParseTuple(args, "l", &id))
-    {
-        PyErr_SetString(PyExc_TypeError, "getNode expects an unsigned long");
-        PYTHON_RETURN_ERROR;
-    }
-    return self->fThis->GetChildNode(id);
```
Hoikas: It appears that this deletion obsoletes `pyVaultNode::GetChildNode()`.

PR #619, opened 2020-01-28
Fix potential whitespace prepend to age descriptions

Hoikas:
This fixes an observed issue in the Nexus where the entry for the GreatTreePub was listed as " Watcher's Pub". It also corrects a potential ugly description generation issue when instances are generated by the server with no user name. Previously this would be " (1) Ae'gura" -- this has been changed to "Ae'gura(1)". This issue should never be triggered.


Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
```diff
@@ -46,6 +46,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 //////////////////////////////////////////////////////////////////////
 
 #include <Python.h>
+#include <string_theory/st_stringstream.h>
```
zrax: The public header should be `<string_theory/string_stream>`

PR #622, opened 2020-01-29
Restore sequence conflict verification.

Hoikas:
This was removed when the page list was converted to an `std::map`. However, not checking for conflicting sequences causes a deadlock on loading an age that had a PRP renamed without changing its sequence suffix.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plResMgr/plResManager.cpp
+++ b/Sources/Plasma/PubUtilLib/plResMgr/plResManager.cpp
@@ -126,16 +126,21 @@ bool plResManager::IInit()
 
     kResMgrLog(1, ILog(1, "Initializing resManager..."));
 
-    if (plResMgrSettings::Get().GetLoadPagesOnInit())
-    {
+    if (plResMgrSettings::Get().GetLoadPagesOnInit()) {
         // We want to go through all the data files in our data path and add new
         // plRegistryPageNodes to the regTree for each
         std::vector<plFileName> prpFiles = plFileSystem::ListDir(fDataPath, "*.prp");
-        for (auto iter = prpFiles.begin(); iter != prpFiles.end(); ++iter)
-        {
+        for (auto iter = prpFiles.begin(); iter != prpFiles.end(); ++iter) {
             plRegistryPageNode* node = new plRegistryPageNode(*iter);
-            plPageInfo pi = node->GetPageInfo();
-            fAllPages[pi.GetLocation()] = node;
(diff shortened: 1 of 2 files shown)


Sources/Plasma/PubUtilLib/plResMgr/plResManager.cpp
```diff
@@ -126,16 +126,21 @@ bool plResManager::IInit()
...
-        {
+        for (auto iter = prpFiles.begin(); iter != prpFiles.end(); ++iter) {
             plRegistryPageNode* node = new plRegistryPageNode(*iter);
-            plPageInfo pi = node->GetPageInfo();
-            fAllPages[pi.GetLocation()] = node;
+            const plPageInfo& pi = node->GetPageInfo();
+
+            // If a page is already added with this location, add both the already known page
+            // and the newly discovered page to a set of conflicts.
+            auto pageResult = fAllPages.try_emplace(pi.GetLocation(), node);
```
zrax: `try_emplace` is C++17

Hoikas: I just noticed our standard is C++14 :cry:

Sources/Plasma/PubUtilLib/plResMgr/plResManager.cpp
```diff
@@ -1330,25 +1335,11 @@ bool plResManager::VerifyPages()
...
-            {
-                invalidPages.Append(upPage);
-                fAllPages.erase(itUp);
-                break;
-            }
-        }
-    }*/
+    for (auto badPage : fConflictingPages) {
+        fAllPages.erase(badPage->GetPageInfo().GetLocation());
+        invalidPages.Append(badPage);
```
zrax: If I'm reading the old code correctly, it only removed (and added to invalidPages) the second (and later) copies of the page; your change here appears to remove and track both the duplicate and the original...

Hoikas: That seems to be correct. Removing both (or more) was a purposeful design decision on my part however. It is better to wipe away all of the potentially invalid pages and let the client download the correct one when linking to the Age. Otherwise, you can potentially get stuck in a spin-wash cycle where the old logic would delete the new page and not touch the old one, requiring a redownload on each link. The inverse could happen as well, but that's a best-case scenario...

zrax: Makes sense.

PR #623, opened 2020-02-02
ASan-directed Destructor Fixes

Hoikas:
Visual C++ ASan points out these destructors should be virtual. There is still an issue with an out of scope access around `pfGUITextBoxMod::IUpdate` and `plLocalization::StringToLocal` that I wasn't sure how to fix.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetClient/plNetLinkingMgr.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetClient/plNetLinkingMgr.cpp
@@ -84,6 +84,8 @@ struct NlmOp {
     NlmOp (const ENlmOp & op)
     : opcode(op)
     { }
+
+    virtual ~NlmOp() = default;
 };
 
 struct NlmNoOpOp : NlmOp {


Sources/Plasma/PubUtilLib/plNetClient/plNetLinkingMgr.cpp
```diff
@@ -84,18 +84,24 @@ struct NlmOp {
...
+
+    virtual ~NlmOp() = default;
 };
 
 struct NlmNoOpOp : NlmOp {
     NlmNoOpOp ()
     : NlmOp(kNlmOpNoOp)
     { }
+
+    virtual ~NlmNoOpOp() = default;
```
zrax: This doesn't need to be declared -- the base is already virtual, and this override is a no-op.

Sources/Plasma/PubUtilLib/plNetClient/plNetLinkingMgr.cpp
```diff
@@ -84,18 +84,24 @@ struct NlmOp {
...
 };
 
 struct NlmOpWaitOp : NlmOp {
     NlmOpWaitOp ()
     : NlmOp(kNlmOpWaitOp)
     { }
+
+    virtual ~NlmOpWaitOp() = default;
```
zrax: This doesn't need to be declared -- the base is already virtual, and this override is a no-op.

Sources/Plasma/PubUtilLib/plNetClient/plNetLinkingMgr.cpp
```diff
@@ -112,6 +120,8 @@ struct NlmLeaveAgeOp : NlmOp {
     NlmLeaveAgeOp ()
     : NlmOp(kNlmOpLeaveAgeOp), quitting(false)
     { }
+
+    virtual ~NlmLeaveAgeOp() = default;
```
zrax: This doesn't need to be declared -- the base is already virtual, and this override is a no-op.

PR #625, opened 2020-02-09
Fix memory leaks.

Hoikas:
This changeset fixes a major source of memory leaks in the vault client API. Any time `hsRef<NetVaultNode> foo = new NetVaultNode;` was called resulted in a leak due to a change in behavior when the old AtomicRef was merged with hsRefCnt.

### DIFF
--- a/Sources/Plasma/CoreLib/hsRefCnt.h
+++ b/Sources/Plasma/CoreLib/hsRefCnt.h
@@ -132,6 +132,15 @@ public:
     _Ref *operator->() const { return fObj; }
     operator _Ref *() const { return fObj; }
 
+    /** Creates a new object of type _Ref. Use this instead of operator new. */
+    template<typename... _Args>
+    static hsRef<_Ref> New(_Args&&... args)
+    {
+        hsRef<_Ref> obj;
+        obj.fObj = new _Ref(std::forward<_Args>(args)...);
+        return obj;
+    }
+
 private:
     _Ref *fObj;
 };
--- a/Sources/Plasma/CoreLib/hsRefCnt.h
+++ b/Sources/Plasma/CoreLib/hsRefCnt.h
@@ -78,6 +78,8 @@ public:
             dst = src;                  \
         } while (0)
 
+struct hsStealRef_Type {};
+constexpr hsStealRef_Type hsStealRef;
 
 template <class _Ref>
 class hsRef
@@ -86,6 +88,7 @@ public:
     hsRef() : fObj(nullptr) { }
     hsRef(std::nullptr_t) : fObj(nullptr) { }
     hsRef(_Ref *obj) : fObj(obj) { if (fObj) fObj->Ref(); }
+    hsRef(_Ref *obj, hsStealRef_Type) : fObj(obj) { }
     hsRef(const hsRef<_Ref> &copy) : fObj(copy.fObj) { if (fObj) fObj->Ref(); }
     hsRef(hsRef<_Ref> &&move) : fObj(move.fObj) { move.fObj = nullptr; }
 
@@ -132,13 +135,11 @@ public:
     _Ref *operator->() const { return fObj; }
     operator _Ref *() const { return fObj; }
 
-    /** Creates a new object of type _Ref. Use this instead of operator new. */
-    template<typename... _Args>
-    static hsRef<_Ref> New(_Args&&... args)
+    void Steal(_Ref *obj)
     {
-        hsRef<_Ref> obj;
-        obj.fObj = new _Ref(std::forward<_Args>(args)...);
-        return obj;
+        if (fObj)
+            fObj->UnRef();
+        fObj = obj;
     }
 
 private:
--- a/Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore.cpp
+++ b/Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore.cpp
@@ -49,21 +49,17 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 /* Get the pfPasswordStore instance */
 pfPasswordStore* pfPasswordStore::Instance()
 {
-    static pfPasswordStore* store = nullptr;
-
-    if (store == nullptr) {
 #if defined(HS_BUILD_FOR_WIN32)
-        store = new pfWin32PasswordStore();
+    static pfWin32PasswordStore store;
 #elif defined(HS_BUILD_FOR_APPLE)
-        store = new pfApplePasswordStore();
+    static pfApplePasswordStore store;
 #elif defined(HAVE_LIBSECRET)
-        store = new pfUnixPasswordStore();
+    static pfUnixPasswordStore store;
 #else
-        store = new pfFilePasswordStore();
+    static pfFilePasswordStore store;
 #endif
-    }
 
-    return store;
+    return &store;
 }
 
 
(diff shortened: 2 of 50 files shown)


Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore.cpp
```diff
@@ -49,21 +51,22 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        store = std::make_unique<pfUnixPasswordStore>();
 #else
-        store = new pfFilePasswordStore();
+        store = std::make_unique<pfFilePasswordStore>();
 #endif
     }
 
-    return store;
+    return store.get();
 }
```
zrax: This whole function could be done without dynamic allocation...

```c++
pfPasswordStore* pfPasswordStore::Instance()
{
#if defined(HS_BUILD_FOR_WIN32)
    static pfWin32PasswordStore store;
#elif defined(HS_BUILD_FOR_OSX)
    static pfMacPasswordStore store;
#elif defined(HAVE_LIBSECRET)
    static pfUnixPasswordStore store;
#else
    static pfFilePasswordStore store;
#endif

    return &store;
}
```

Now that we depend on MSVC2015, this is guaranteed to be thread safe.

Hoikas: Good point!

Hoikas:
For some background reading...

The onus of this is that on Gehn Beta, downloading a new copy of @DoobesURU Age Veelay Tsahvahn can cause the client to crash shortly after linking in with an `std::bad_alloc`. I started linking around and noticing that the client's memory usage seems to grow for each link. For example, standing in Relto for the first time, the client uses ~600mb. Linking to Gahreesen and back to Relto causes the client memory usage to be around ~900mb. That's an extreme example--smaller Ages like the Nexus have less of an impact. While I am fixing leaks reported by VLD, I'm really not seeing anything that corresponds to the observed behavior.

Hoikas:
I recently received a complaint in IRC that boiled down to folks with large buddy lists saw `std::bad_alloc` rather frequently, so I decided to revisit this changeset due to its fixing of leaked vault nodes. It has been updated to merge cleanly. However, due to an issue introduced in the upgrade to Python 3, we depend on #720.

DamnBriggsy:
Prior to these changes, the client would crash on MOULa in an hour or less. After these changes, it stayed up for over 10 hours with no problems.

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore.cpp
```diff
@@ -49,21 +49,17 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 {
-    static pfPasswordStore* store = nullptr;
-
-    if (store == nullptr) {
 #if defined(HS_BUILD_FOR_WIN32)
-        store = new pfWin32PasswordStore();
+    static pfWin32PasswordStore store;
 #elif defined(HS_BUILD_FOR_APPLE)
-        store = new pfApplePasswordStore();
+    static pfMacPasswordStore store;
```
dpogue, line 55: `pfMacPasswordStore` -> `pfApplePasswordStore`

This was renamed in https://github.com/H-uru/Plasma/pull/675

Sources/Plasma/FeatureLib/pfPython/pyVaultPlayerInfoListNode.cpp
```diff
@@ -79,14 +73,13 @@ bool pyVaultPlayerInfoListNode::HasPlayer( uint32_t playerID )
...
+    templateNode.SetNodeType(plVault::kNodeType_PlayerInfo);
+    VaultPlayerInfoNode access(&templateNode);
     access.SetPlayerId(playerID);
-    
-    hsRef<RelVaultNode> rvn = fNode->GetChildNode(templateNode, 1);
-    
-    return (rvn != nil);
+
+    hsRef<RelVaultNode> rvn = fNode->GetChildNode(&templateNode, 1);
+    return rvn;
```
zrax, line 82: This implicit cast to bool is misleading (it looks like you're returning the node)...

Sources/Plasma/CoreLib/hsRefCnt.cpp
```diff
@@ -123,7 +123,7 @@ hsRefCnt::hsRefCnt(int initRefs)
 hsRefCnt::~hsRefCnt()
 {
 #ifdef HS_DEBUGGING
-    hsThrowIfFalse(fRefCnt == 1);
+    hsThrowIfFalse(fRefCnt <= 1);
```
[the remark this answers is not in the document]
Hoikas: `NetVaultNode`'s starting reference count is zero. An earlier refactor changed its base class from `AtomicRef` to `hsRefCnt` but neglected to address the difference in starting ref count, introducing all of these leaks. I initially wanted to add a template parameter for the starting ref count to `hsRefCnt`, but I wasn't able to come up with a clean solution that did expose implementation details to the rest of the engine

Hoikas:
@zrax RE our discussion in IRC about this PR... Maybe the easiest path forward is to back out 80e00cd? That restores the starting ref count to 1, but uses a weird hack for initializing an `hsRef` from a newly allocated object. I'm not sure if we're going to find anything much cleaner. :/

zrax:
Part of the problem here is the mix of managed (`hsRef<T>`) and unmanaged objects all over the code base...  If we could assume all were managed, this could be simplified and made safer.  Unfortunately, the code base is far from that goal, so we have these ugly hacks instead.

Perhaps instead of the `New()` template, we could add a more generic mechanism for adopting/stealing a ref.  Something like (completely untested):

```c++
struct hsAdoptRef_Type {};
constexpr hsAdoptRef_Type hsAdoptRef;

template <class _Ref>
class hsRef
{
public:
    /* ... */

    hsRef(_Ref *obj, hsAdoptRef_Type) : fObj(obj) { }

    /* ... */

    void Adopt(_Ref *obj)
    {
        if (fObj)
            fObj->UnRef();
        fObj = obj;
        return *this;
    }

    /* ... */
};
```

Then in usage:

```c++
hsRef<Whatever> myref(new Whatever, hsAdoptRef);
// or
hsRef<Whatever> myref;
myref.Adopt(new Whatever);
```

In the long run, we'd want to eventually ensure that promotion from a "raw" `hsRefCnt` to an `hsRef` managed object is always one-way, which would mean that the adoption behavior becomes the default (and this workaround can be removed).  We may actually be close already, but that would require more investigation and testing than I think you're willing to add in this PR ;)

Hoikas:
Agreed. I'll look into this, although I may change `hsAdoptRef` to `hsStealRef` - I think the naming is clearer in that case 😛

Hoikas:
It compiles! :shipit:

Sources/Plasma/CoreLib/hsRefCnt.h
```diff
@@ -132,6 +135,13 @@ class hsRef
...
+    void Steal(_Ref *obj)
+    {
+        if (fObj)
+            fObj->UnRef();
+        fObj = obj;
+    }
+
     /** Creates a new object of type _Ref. Use this instead of operator new. */
     template<typename... _Args>
     static hsRef<_Ref> New(_Args&&... args)
```
zrax, line 147: > Remove `hsRef::New()` in favor of clearer reference stealing operations. 

And yet, here it is....  ;)

Hoikas: Rofl

Hoikas:
`hsRef<_Ref>::New()` is really gone now 🤣

zrax (approved):

Hoikas:
💥

PR #627, opened 2020-02-11
Fix age description spacing.

Hoikas:
This fixes the lack of a space between the age's user name ("Adam's") and the age instance name ("Teledahn") if the sequence number is zero introduced in #619.


Deledrius:
I saw this last night and thought it was an optical illusion.  Glad I wasn't seeing things!

Hoikas:
:trollface:

PR #628, opened 2020-02-11
Remove hard-coded framecount for linking progress animation.

Deledrius:
This fixes a long-standing crash which could occur if using a different `resource.dat` than one the client was hard-coded to expect.  Previously, the number of frames was dictated at compile-time.  This change allows the plProgressMgr to query for the list of resources, find the ones matching the given pattern, and subsequently use that sorted list as the IDs from which it builds and cycles the animation.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
+++ b/Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
@@ -134,3 +134,15 @@ plMipmap* plClientResMgr::getResource(const ST::string& resname)
 
     return resmipmap;
 }
+
+
+std::vector<ST::string> plClientResMgr::getResourceNames()
+{
+    std::vector<ST::string> names;
+    names.reserve(ClientResources.size());
+
+    for (const auto& resource : ClientResources)
+        names.push_back(resource.first);
+
+    return names;
+}
--- a/Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
+++ b/Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
@@ -54,6 +54,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsTimer.h"
 
 #include "plPipeline/plPlates.h"
+#include "plClientResMgr/plClientResMgr.h"
+
+#include <regex>
 
 
 //////////////////////////////////////////////////////////////////////////////
@@ -62,9 +65,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 plProgressMgr* plProgressMgr::fManager = nullptr;
 
-#define LOADING_RES_COUNT   18
-
-ST::string plProgressMgr::fImageRotation[LOADING_RES_COUNT];
+std::vector<ST::string> plProgressMgr::fImageRotation;
 
 const ST::string plProgressMgr::fStaticTextIDs[] = {
     ST_LITERAL("xLoading_Linking_Text.png"),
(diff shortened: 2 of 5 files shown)


Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
```diff
@@ -134,3 +134,14 @@ plMipmap* plClientResMgr::getResource(const ST::string& resname)
 
     return resmipmap;
 }
+
+
+std::vector<ST::string> plClientResMgr::getResourceNames()
+{
+    std::vector<ST::string> names;
+
+    for (auto resource : ClientResources)
```
Hoikas: I believe that this makes an unnecessary copy of the string. We may prefer `const auto&`

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
```diff
@@ -80,9 +81,13 @@ plProgressMgr::plProgressMgr()
     fCallbackProc = nullptr;
     fCurrentStaticText = kNone;
 
-    // Fill array with pre-computed loading frame IDs
-    for (int i=0; i < LOADING_RES_COUNT; i++)
-        fImageRotation[i] = ST::format("xLoading_Linking.{02}.png", i);
+    // Find linking-animation frame IDs and store the sorted list
+    std::regex re("xLoading_Linking\\.[0-9]+?\\.png");
+    for (auto name : plClientResMgr::Instance().getResourceNames()) {
```
Hoikas: I believe that this makes an unnecessary copy of the string. We may prefer `const auto&`

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
```diff
@@ -80,9 +81,13 @@ plProgressMgr::plProgressMgr()
...
+        if (std::regex_match(name.to_std_string(), re))
```
zrax: You can avoid the conversion to `std::string` here by using `std::regex_match(name.begin(), name.end(), re)` instead

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
zrax: `const auto& name` to avoid copying...

Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
zrax: `const auto& resource` to avoid copies (if applicable)

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
```diff
@@ -230,15 +235,12 @@ void    plProgressMgr::CancelAllOps( void )
 
 const ST::string plProgressMgr::GetLoadingFrameID(int index)
 {
-    if (index < LOADING_RES_COUNT)
-        return fImageRotation[index];
-    else
-        return fImageRotation[0];
+    return fImageRotation[index % (fImageRotation.size() - 1)];
```
zrax: `%` doesn't protect against negative indexes...  I'd suggest either making this unsigned or keeping the old behavior (with the non-hard-coded size)

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
Deledrius: Excellent, I didn't see that option when I was looking at the regex matches.  I was hoping to write something a bit better here.

Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
Deledrius: I wish I were better at seeing these myself.

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
Deledrius: I wish I were better at seeing these myself.

Sources/Plasma/PubUtilLib/plProgressMgr/plProgressMgr.cpp
Deledrius: I must have been thinking of this backwards.  Thanks.

Deledrius: The old method doesn't protect against negatives, either.  Yikes.

Deledrius: Is std::clamp available for our usage?  Would that be appropriate?

Hoikas: `std::clamp` was added in C++17--we are still using C++14.

Deledrius: Ah, that's out then. :)

zrax: > The old method doesn't protect against negatives, either. Yikes.

Ah, good point...  Maybe not a big deal then.

PR #633, opened 2020-02-19
Decompress audio files on download.

Hoikas:
This fixes the issue where Zandi might still be listening to Peter Gabriel instead of Mozart. In other words, this fixes the initially downloaded version of the sound playing instead of the updated version.


Hoikas:
This has been updated to merge cleanly.

dpogue (approved):

PR #642, opened 2020-03-17
Fix ptVaultNode's richcompare function.

Hoikas:
The Cyan standard function did not properly handle comparing vault nodes with incompatible types, which is expected to return False. Instead, they tried to return an exception. Unfortunately, they did this incorrectly by returning a value of -1, which was interpreted to be a result of True.

This resulted in interesting things like `node == "Ahnonay"` evaluating to True. See also H-uru/moul-scripts#120

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
@@ -57,49 +57,52 @@ PYTHON_NO_INIT_DEFINITION(ptVaultNode)
 
 PYTHON_RICH_COMPARE_DEFINITION(ptVaultNode, obj1, obj2, compareType)
 {
-    if ((obj1 == Py_None) || (obj2 == Py_None))
-    {
-        bool isEqual = (obj1 == obj2);
-
-        if (compareType == Py_EQ)
-        {
-            if (isEqual)
-                PYTHON_RCOMPARE_TRUE;
-            else
-                PYTHON_RCOMPARE_FALSE;
-        }
-        else if (compareType == Py_NE)
-        {
-            if (isEqual)
-                PYTHON_RCOMPARE_FALSE;
-            else
-                PYTHON_RCOMPARE_TRUE;
-        }
-        PyErr_SetString(PyExc_NotImplementedError, "invalid comparison for a ptVaultNode object");
-        PYTHON_RCOMPARE_ERROR;
+    // Only allow == or !=
+    if (compareType != Py_EQ && compareType != Py_NE) {
+        PyErr_SetString(PyExc_TypeError, "ptVaultNode comparison not supported");
+        return nullptr;
     }
 
-    if (!pyVaultNode::Check(obj1) || !pyVaultNode::Check(obj2))
-    {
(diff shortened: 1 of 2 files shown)


Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
```diff
@@ -57,49 +57,52 @@ PYTHON_NO_INIT_DEFINITION(ptVaultNode)
...
+    if (compareType == Py_EQ) {
+        PYTHON_RETURN_BOOL((*vnObj1) == (*vnObj2));
+    } else if (compareType == Py_NE) {
+        PYTHON_RETURN_BOOL((*vnObj1) != (*vnObj2));
     }
-    PyErr_SetString(PyExc_NotImplementedError, "invalid comparison for a ptVaultNode object");
-    PYTHON_RCOMPARE_ERROR;
+
+    // Should never happen.
+    return Py_NotImplemented;
```
zrax: `Py_RETURN_NOTIMPLEMENTED ` (which increments the `Py_NotImplemented` singleton's ref count)

Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
```diff
@@ -57,49 +57,52 @@ PYTHON_NO_INIT_DEFINITION(ptVaultNode)
...
-    if (compareType == Py_EQ)
-    {
-        if ((*vnObj1) == (*vnObj2))
-            PYTHON_RCOMPARE_TRUE;
-        PYTHON_RCOMPARE_FALSE;
+
+    // Vault nodes can only be equal to vault nodes
+    if (!pyVaultNode::Check(obj1) || !pyVaultNode::Check(obj2)) {
+        if (compareType == Py_EQ) {
+            PYTHON_RETURN_BOOL(false);
```
zrax: `Py_RETURN_FALSE`

Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
```diff
@@ -57,49 +57,52 @@ PYTHON_NO_INIT_DEFINITION(ptVaultNode)
...
+        } else if (compareType == Py_NE) {
+            PYTHON_RETURN_BOOL(true);
```
zrax: `Py_RETURN_TRUE`

zrax:
Also, the `PYTHON_RCOMPARE_*` macros are all clearly wrong.  We should probably remove them entirely from `pyGlueHelpers.h` and fix up the rest of the rich comparisons...

Sources/Plasma/FeatureLib/pfPython/pyVaultNodeGlue.cpp
zrax: ... didn't get updated?

Hoikas: whoops, missed this one

Hoikas: undeclared identifier

zrax: Ah, looks like the macro was introduced in Python 3.3.

PR #648, opened 2020-05-05
Updates for C++17

zrax:
* Compatibility fixes with changed/removed features
* Use C++17's built-in aligned `operator new` instead of platform-specific memory allocation in `hsAlignedAllocator`

TODO: Python 2.7's headers now spit out a bunch of warnings due to the deprecation of the `register` keyword in C++17

### DIFF
--- a/Sources/Tools/MaxMain/plMaxNode.cpp
+++ b/Sources/Tools/MaxMain/plMaxNode.cpp
@@ -2075,8 +2075,6 @@ bool plMaxNode::ConvertToOccluder(plErrorMsg* pErrMsg, bool twoSided, bool isHol
     plLocation nodeLoc = GetLocation();
 
     bool moving = IsMovable();
-    if( moving )
-        moving++;
 
     Matrix3 tmp(true);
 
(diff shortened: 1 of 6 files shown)


Sources/Tools/MaxMain/plMaxNode.cpp
```diff
@@ -2075,8 +2075,6 @@ bool plMaxNode::ConvertToOccluder(plErrorMsg* pErrMsg, bool twoSided, bool isHol
     plLocation nodeLoc = GetLocation();
 
     bool moving = IsMovable();
-    if( moving )
-        moving++;
```
Hoikas, line 2079: What in the world 😱

dpogue: Extra truthy...

zrax: Sometimes the truth just isn't `true` enough

zrax:
I think the string_theory issue only affects VS 2019...  But yeah, I should do a new release soon with that fix.

zrax:
The crash should be fixed now

PR #649, opened 2020-05-06
Change `PtDebugPrint` to behave more like Py3 `print`.

Hoikas:
This is to facilitate H-uru/moul-scripts#119.

`PtDebugPrint()` now outputs to a single line and accepts the keyword argument `sep`, which functions in the same manner as Python 3's sep keyword argument. ~~The `end` keyword argument was not added.~~ Known limitation: the `end` argument cannot be used to elide newlines.

I have also fixed a memory leak in which the repr of a PyObject was not decref'd.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
@@ -402,37 +402,69 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
     PYTHON_RETURN_NONE;
 }
 
-PYTHON_GLOBAL_METHOD_DEFINITION_WKEY(PtDebugPrint, args, kwargs, "Params: *msgs, **kwargs\nPrints msgs to the Python log given the message's level")
+PYTHON_GLOBAL_METHOD_DEFINITION_WKEY(PtDebugPrint, args, kwargs, "Params: *msgs, level, sep, end\n"
+                                     "Prints msgs to the Python log given the message's level, "
+                                     "optionally separated and terminated by the given strings")
 {
     uint32_t  level = cyMisc::kErrorLevel;
+    ST::string sep = ST_LITERAL(" ");
+    ST::string end = ST_LITERAL("\n");
 
     do {
-        // Grabbin' levelz
         if (kwargs && PyDict_Check(kwargs)) {
-            PyObject* value = PyDict_GetItem(kwargs, PyString_FromString("level"));
+            PyObject* value = PyDict_GetItemString(kwargs, "level");
             if (value) {
                 if (PyInt_Check(value))
                     level = PyInt_AsLong(value);
                 else
                     break;
             }
+
+            value = PyDict_GetItemString(kwargs, "sep");
+            if (value) {
+                if (PyString_CheckEx(value))
+                    sep = PyString_AsStringEx(value);
+                else
+                    break;
+            }
+
+            value = PyDict_GetItemString(kwargs, "end");
+            if (value) {
+                if (PyString_CheckEx(value))
+                    end = PyString_AsStringEx(value);
+                else
+                    break;
+            }
         }
 
+        ST::string_stream ss;
         for (size_t i = 0; i < PySequence_Fast_GET_SIZE(args); ++i) {
             PyObject* theMsg = PySequence_Fast_GET_ITEM(args, i);
-            if (!PyString_CheckEx(theMsg))
-                theMsg = PyObject_Repr(theMsg);
-
-            if (theMsg)
-                cyMisc::DebugPrint(PyString_AsStringEx(theMsg), level);
+            if (PyString_CheckEx(theMsg))
+                Py_XINCREF(theMsg);
             else
-                break;
+                theMsg = PyObject_Str(theMsg);
+
+            if (i != 0)
+                ss << sep;
+            if (theMsg) {
+                ss << PyString_AsStringEx(theMsg);
+                Py_DECREF(theMsg);
+            } else {
+                PyErr_Format(PyExc_RuntimeError, "Failed to `str()` argument index %n", i);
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -405,34 +405,50 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
...
+                theMsg = PyObject_Str(theMsg);
 
+            if (i != 0)
+                ss << sep;
             if (theMsg)
-                cyMisc::DebugPrint(PyString_AsStringEx(theMsg), level);
+                ss << PyString_AsStringEx(theMsg);
             else
-                break;
+                ss << "<ERROR>";
```
zrax, line 443: I feel like the previous use of `break` was intended to cause us to get to the `PyError_SetString` code outside of the `do ... while (false)` block (however, due to the inner loop, it didn't actually do that)...  Perhaps we should fix that rather than dumping `"<ERROR>"` to the log?  Especially since a null `theMsg` should indicate that an error was already set by another Python API...

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -405,34 +405,50 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
...
+            Py_XDECREF(theMsg);
         }
+        cyMisc::DebugPrint(ss.to_string(), level);
         PYTHON_RETURN_NONE;
     } while (false);
 
     // fell through to the type error case
-    PyErr_SetString(PyExc_TypeError, "PtDebugPrint expects a sequence of strings and an optional int");
+    PyErr_SetString(PyExc_TypeError, "PtDebugPrint expects a sequence of strings, an optional int, and an optional string");
```
zrax, line 451: `kwargs` are not "optional parameters", since they need to be explicitly keyed...  Also, since we're `str()`ifying parameters (to match `print`), I feel like this message is misleading in multiple ways.

Hoikas: :trollface:

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -405,34 +405,50 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
...
-            PyObject* value = PyDict_GetItem(kwargs, PyString_FromString("level"));
+            PyObject* value = PyDict_GetItemString(kwargs, "level");
             if (value) {
                 if (PyInt_Check(value))
                     level = PyInt_AsLong(value);
                 else
                     break;
             }
+
+            value = PyDict_GetItemString(kwargs, "sep");
```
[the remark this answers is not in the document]
Hoikas: Unfortunately, `plStatusLog` unconditionally treats each call to `AddLine` as at least one line. Fixing that is outside the scope of this PR, but I will add an `end` argument that allows terminating with an arbitrary string.

Hoikas:
Updated to fix issues from review.

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -402,37 +402,71 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
...
+            if (value) {
+                value = PyObject_Str(value);
```
zrax, line 425: Nice addition, but if you want to be strict about compatibility with `print()`, Python3 does actually require that `sep` and `end` are strings (it even rejects bytes)

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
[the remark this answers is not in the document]
zrax: Fair enough, but I'd rather not see `end` default to `'\n'` if we're going to end up writing two newlines (one from `AddLine` and one from the default `end`)...  Maybe it would be better to just leave out the `end` parameter until it's needed (and then update the `DebugPrint` to use it properly at the same time)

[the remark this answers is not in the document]
Hoikas: `plStatusLog` only adds the trailing `\n` if one is not present. I decided to default `end` to `\n` on the off chance that some later work allows eliding newlines in `plStatusLog` :smile:

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
Hoikas: I didn't realize that. Do you have a preference either way?

zrax: I'd probably prefer to keep it like `print()`, to make future changes easier

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -402,37 +402,69 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtDebugAssert, args, "Params: cond, msg\nDebug o
...
             if (value) {
                 if (PyInt_Check(value))
                     level = PyInt_AsLong(value);
                 else
                     break;
             }
+
+            value = PyDict_GetItemString(kwargs, "sep");
+            if (value) {
+                if (PyString_CheckEx)
```
zrax, line 425: I think you're missing something here :trollface:

Hoikas: Just making sure you're paying attention :trollface:

PR #650, opened 2020-05-11
Python 3.8

Hoikas:
This updates the game engine and scripts to support Python 3. Note that the minimum required Python version is ~3.7~3.8 ~due to the `PyUnicode` API changing to properly return `const char*`~ because Python.pak support requires the new initialization API from 3.8. From my limited testing, the game plays correctly, however a full playthough was not done.

This changeset deletes the in-repo copy of the Python stdlib. We ought to not be versioning external code that changes based off the Python version. The stdlib and the rest of the scripts are now copied to the runtime output directory at build time. Hopefully this will make generating working game clients easier.

This DOES NOT:
- ~update the resource.dat generation scripts to Python 3. Per @Deledrius and my experimentation, these rely on fragile modules that are nearly impossible to trivally `pip install` on Python 3.~
- attempt to deal with the chaos of the duplicated `get[thing]W()` methods in the Python API

~Note this extends the work done in #596 because  it seems silly to replicate these changes on top of the other API changes found therein.~


Hoikas:
:shipit:

Hoikas:
I have pushed a fix for the issue in which junked python.pak files are created by plPythonPack.exe. Using python.pak is still broken however due to changes in Python 3's import machinery... Work will continue on this.

Scripts/Python/plasma/Plasma.py
```diff
@@ -1468,10 +1448,6 @@ def useEAXAcceleration(self,state):
         """Enables or disables EAX sound acceleration (requires hardware acceleration)."""
         pass
 
-    def useHardwareAcceleration(self,state):
-        """Enables or disables audio hardware acceleration."""
-        pass
```
zrax, line 1473: Some of these changes don't look related to Python3-ification -- possibly from your audio PR?

Scripts/Python/plasma/PlasmaTypes.py
```diff
@@ -675,7 +675,7 @@ def __init__(self):
     def getParentKeys(self):
         # if we haven't got the parent keys yet, then add them to the dict
         if not self.gotParentKeys:
-            for anim in dict.values(self):
+            for anim in list(dict.values(self)):
```
zrax, line 678: `list` unnecessary for iteration

Scripts/Python/xEventTrigger.py
```diff
@@ -85,7 +84,7 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
                     #~ PtDebugPrint("Event %s is true!" % (VARname))
                     # are we paging things in?
                     if PageNames.value:
-                        names = string.split(PageNames.value,",")
+                        names = PageNames.values.split(",")
```
zrax, line 87: `value`

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue4.cpp
```diff
@@ -713,19 +693,15 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtSetBehaviorNetFlags, args, "Params: behKey, ne
 
 PYTHON_GLOBAL_METHOD_DEFINITION(PtSendFriendInvite, args, "Params: emailAddress, toName = \"Friend\"\nSends an email with invite code")
 {
-    char* emailIn = nullptr;
-    char* nameIn = nullptr;
-    if (!PyArg_ParseTuple(args, "es|es", "utf8", &emailIn, "utf8", &nameIn))
+    ST::string email;
+    ST::string name;
```
zrax, line 697: `name` defaults to `"Friend"` in the removed code below...

Scripts/Python/plasma/Plasma.py
Hoikas: Yes, I wrote this on top of that

Scripts/Python/plasma/PlasmaTypes.py
Hoikas: This has been observed to occur while the dict is changing.

CMakeLists.txt
```diff
@@ -67,9 +67,8 @@ set_package_properties(OpenAL PROPERTIES
     TYPE REQUIRED
 )
 
-set(Python_ADDITIONAL_VERSIONS 2.7)
-find_package(PythonLibs REQUIRED)
-set_package_properties(PythonLibs PROPERTIES
+find_package(Python3 REQUIRED Interpreter Development)
```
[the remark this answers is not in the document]
Hoikas: I can do that when I get the python.pak issues fixed. There is one case where I was being cute and used f-strings to fix a string issue, but that's not strictly needed.

Scripts/Python/ki/__init__.py
```diff
@@ -2483,9 +2483,9 @@ def ChangeFontSize(self, new):
 
         size = self.GetFontSize()
         if new == 1:
-            fontRange = range(len(kChat.FontSizeList) - 1)
+            fontRange = list(range(len(kChat.FontSizeList) - 1))
         elif new == -1:
-            fontRange = range(len(kChat.FontSizeList) - 1, 0, -1)
+            fontRange = list(range(len(kChat.FontSizeList) - 1, 0, -1))
```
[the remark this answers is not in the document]
Hoikas: Since it's a konstant, sounds right...

Scripts/Python/plasma/Plasma.py
zrax: Hmm, that does make it harder to review then...  I may hold off until that one is merged.

CMakeLists.txt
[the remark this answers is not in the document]
Hoikas: ^^^ was intended to be a response to the overall review comment

Hoikas:
Updated to fix issues from review and add proper PEP 451 compliant import machinery for python.pak.

Hoikas:
Rebased onto master, which now includes two commits that were previously part of this PR. See #651.

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -18,15 +18,15 @@ if(MOVIE_AVAILABLE)
...
+find_package(Python2)
+set_package_properties(Python2 PROPERTIES
     URL "http://www.python.org"
     DESCRIPTION "Python Scripting language interpreter"
     TYPE OPTIONAL
     PURPOSE "Used to (re-)build some client resource files"
 )
 
-if(PYTHONINTERP_FOUND)
+if(Python2_Interpreter_FOUND)
```
Deledrius, line 29: This no longer matches the endif on L#37, causing a CMake Warning.

Scripts/Python/xIniAudio.py
```diff
@@ -323,9 +323,9 @@ def SetAudioMode(init, device, eax):
 
         PtDebugPrint(device)
         if entryDev:
-            entryDev.setValue(0, "\"" + device + "\"")
+            entryDev.setValue(0, "\"" + device.encode("utf-8") + "\"")
         else:
-            gIniFile.addEntry("Audio.SetDeviceName \"" + device + "\"")
+            gIniFile.addEntry("Audio.SetDeviceName \"" + device.encode("utf-8") + "\"")
```
Deledrius, line 328: `TypeError: can only concatenate str (not "bytes") to str`

Hoikas:
9882e88  will be stripped out when #653 is merged.

Hoikas:
Pushed fixes for: Relto Smokers (#654), Relto Trees, and Kadish traceback on link-in (#655).

Hoikas:
Rebased and removed @Deledrius pinger.

Sources/Plasma/FeatureLib/pfPython/pyGlueHelpers.cpp
```diff
@@ -45,31 +45,41 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
     return ST::null;
 }
 
-bool PyString_CheckEx(PyObject* obj)
+int PyUnicode_STStringConverter(PyObject* obj, void* str)
 {
-    return (PyString_Check(obj) || PyUnicode_Check(obj));
+    if (PyUnicode_Check(obj)) {
+        *((ST::string*)str) = PyUnicode_AsSTString(obj);
```
zrax, line 65: I'd probably prefer `reinterpret_cast` here, but that's just a style preference...

Sources/Plasma/FeatureLib/pfPython/pyVaultMarkerGameNodeGlue.cpp
```diff
@@ -130,17 +130,13 @@ PYTHON_METHOD_DEFINITION(ptVaultMarkerGameNode, setMarkers, args)
...
         }
 
-        VaultMarker marker;
-        marker.id = PyLong_AsUnsignedLong(id);
-        marker.age = PyString_AsStringEx(age);
-        marker.pos = pyPoint3::ConvertFrom(pos)->fPoint;
-        marker.description = PyString_AsStringEx(desc);
-        collector.push_back(marker);
+        collector.push_back({ PyLong_AsUnsignedLong(id), PyUnicode_AsSTString(age),
+                              pyPoint3::ConvertFrom(pos)->fPoint, PyUnicode_AsSTString(desc) });
```
zrax, line 139: `emplace_back`?

Hoikas: I tried that, and it resulted in a compile error :sad:

Hoikas: ... It helps if `VaultMarkers` has a constructor, doesn't it? Sigh.

zrax: Hmm, it looks like you can do it as `vector.emplace_back(StructType { ... })`...  However, a quick test shows that that version isn't any more optimal than the `push_back` with optimization turned on :(.

It looks like this will get fixed in C++20, but for now I think it's fine to leave the `push_back`.

zrax: Or just add a constructor like you did -- that works too :)

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -17,15 +17,7 @@ if(MOVIE_AVAILABLE)
...
-find_package(PythonInterp)
-set_package_properties(PythonInterp PROPERTIES
-    URL "http://www.python.org"
-    DESCRIPTION "Python Scripting language interpreter"
-    TYPE OPTIONAL
-    PURPOSE "Used to (re-)build some client resource files"
-)
-
-if(PYTHONINTERP_FOUND)
+if(Python3_Interpreter_FOUND)
```
Deledrius, line 20: This no longer matches the endif on L#30, causing a CMake Warning.

Deledrius:
Missing is the change from `PYTHON_EXECUTABLE` to `Python3_EXECUTABLE` at https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/Apps/plClient/external/CMakeLists.txt#L17-L20.  Without this, VS annoyingly launches the script using whatever the Windows association default action is (typically a text editor) rather than running the script.  Feel free to cherry-pick https://github.com/H-uru/Plasma/commit/0a8972ffc21dc003e50afaca60e880a2c7c29d96. ;)

Hoikas:
Fixed resource generation issues.

Hoikas:
Updated to fix an issue with the KI displaying player folders and editing marker quests.

Hoikas:
Yes, I saw those after @Deledrius pointed them out. I decided to not fix those issues in this pull request. I want this PR to be limited to the set of changes required for the game to function as intended on Python 3.7 and Python 3.8--mostly to help facilitate review.

More specifically, `PyUnicode_GetSize` seems to (mostly) be used in the "W" variants of the old, duplicated string/unicode setters in the API. Those are no longer needed, but, as mentioned above, I was hoping to tackle unifying all that and fixing the deprecation in another PR Further, I believe I have eradicated all uses script-side of `PyObject_AsReadBuffer` since `buffer()` is now dead, but again, I wanted to defer removal of those methods from the API itself to another PR.

zrax:
Ah, makes sense...  I didn't look closely at where they were used, I just saw them in the compiler output while checking if Python 3.7 fixed the `register` issue (spoiler alert: it does)

Hoikas:
Pushed a fix to _not_ generate a `__pycache__.pak` file in plPythonPack, and enabled the `Py_IsolatedFlag` in external releases, preventing the importing of modules from the `sys.path`.

Hoikas:
Bumped to Python 3.8. I have changed this PR to a draft until vcpkg is updated to Py 3.8 and the python.pak machinery is tested again. Rationale copy-pasta'd from the appropriate commit.

This is required due to reports from multiple testers that the client exits on loading StartUp. The issue is that Python 3 loads modules during its initialization--before we have installed our PEP 451 import machinery for python.pak. The previous code worked for us because we have Python 3 installed on our systems, which is not a reasonable assumption to have for our players. Python 3.8 adds APIs that allow us to initialize just the "core" and use the API, then initialize the interpreter itself once we have added the import machinery.

Hoikas:
Rerequesting review specifically for b82d8d3.

Hoikas:
Reopening due to the merge of python 3.8 into vcpkg and successful testing of python.pak.

README.md
```diff
@@ -24,7 +24,7 @@ Plasma currently requires the following third-party libraries:
 
 - nVidia PhysX 2.6.4 - https://web.archive.org/web/20170501130625/http://developer.download.nvidia.com/PhysX/2.6.4/PhysX_2.6.4_SDK_Core.exe
 - Microsoft DirectX SDK - https://www.microsoft.com/en-us/download/details.aspx?id=6812
-- Python 2.7 - http://www.python.org/download/releases/2.7/
+- Python 3.7 (or higher) - https://www.python.org/downloads/
```
Deledrius, line 27: This should be bumped in 05b4deda4f1eb3da4186116d15be79d2486ffb3b.

CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ set_package_properties(OpenSSL PROPERTIES
     TYPE REQUIRED
 )
 
-set(Python_ADDITIONAL_VERSIONS 2.7)
+set(Python_ADDITIONAL_VERSIONS 3.7)
```
Deledrius: This should be bumped in 05b4ded.

Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -464,16 +464,16 @@ std::vector<std::pair<std::string, std::string> > DebuggerCallback::GenerateLoca
                 if (PyObject_Compare((PyObject*)&PyType_Type, PyObject_Type(value)) == 0)
                     continue;
 
-                std::string keyStr = PyString_AsString(PyObject_Str(key));
+                std::string keyStr = PyUnicode_AsUTF8(PyObject_Str(key));
                 if (keyStr == "__builtins__")
```
Deledrius, line 472: Should this be `"builtins"`?

CMakeLists.txt
Hoikas: This diff is outdated--the Python version is no longer supplied to the CMake finder.

Hoikas:
Updated to fix issues from review.

CMakeLists.txt
Deledrius: Ah, good.  I didn't see it updated but also didn't see it was gone!

Hoikas:
Blocking on H-uru/PlasmaPrefix#4.

Hoikas:
@Zrax, @dpogue Is there anything else I need to do here?

Deledrius:
This looks ready to merge.  Any objections?

Hoikas:
:boom: :boom: :boom:

PR #651, opened 2020-05-15
Scripts Changes for Py3 Types

Hoikas:
These changes were split from #650 - they are required for compatibility with Python 3 due to type changes and were apparently missed by 2to3. They are compatible with 2.7, so here they are separately as requested by @zrax.

### DIFF
--- a/Scripts/Python/kdshTreeRingsSolution.py
+++ b/Scripts/Python/kdshTreeRingsSolution.py
@@ -53,7 +53,6 @@ from Plasma import *
 from PlasmaTypes import *
 from PlasmaKITypes import *
 import PlasmaControlKeys
-import string
 
 # If there were another way to have done this, believe me, I would have. 
 # In the mean time, the following 122 items are dedicated to the memory of Pete Gage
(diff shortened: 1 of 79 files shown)


Scripts/Python/kdshTreeRingsSolution.py
```diff
@@ -300,9 +299,9 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
             #~ PtDebugPrint("TRS: Nothing to ff. VARname = ", VARname)
             pass
         else:
-            GUIcode = "GUI" + string.join(string.split(VARname, "Ring"), "") + "_0" + str(newbearing) + ".animation.play()"
-            #~ PtDebugPrint("GUIcode = ", GUIcode)
-            exec(GUIcode) # this runs the animation on the "fake" ring in front of the GUI
+            # this runs the animation on the "fake" ring in front of the GUI
+            animname= "GUI_{id}_0_{bearing}".format(id="".join(VARname.split("Ring")), bearing=newbearing)
```
zrax, line 303: There seem to be a couple of extra `_` here...  Shouldn't it be `GUI{id}_0{bearing}`?

Hoikas: indeed!

PR #653, opened 2020-05-16
Fix another deprecated list.sort().

Hoikas:
This was missed in #651.

### DIFF
--- a/Scripts/Python/xOptionsMenu.py
+++ b/Scripts/Python/xOptionsMenu.py
@@ -54,6 +54,7 @@ This is the python handler for the Options Menu
 MaxVersionNumber = 8
 MinorVersionNumber = 4
 
+import functools
 import os
 
 from Plasma import *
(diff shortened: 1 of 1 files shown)


Scripts/Python/xOptionsMenu.py
```diff
@@ -1954,8 +1955,11 @@ def res_comp(elem1, elem2):
     elem2w = int(elem2[:elem2.find("x")])
     elem2h = int(elem2[elem2.find("x") + 1:])
 
-    result = cmp(elem1w, elem2w)
+    # cmp is gone in Python 3
+    _cmp = lambda a, b: (a > b) - (a - b)
```
zrax, line 1959: Should be `(a > b) - (a < b)`

Hoikas: (insert swearing emoji here)

PR #655, opened 2020-05-17
Kill `exec(code)`

Hoikas:
This removes all uses of the `exec()` function from the Python scripts. This fixes an observed traceback in Python 3 related to exec statements with side effects. The code is also much cleaner as a result.

cPythScopeSolution - Traceback (most recent call last):
  File ".\python\kdshTreeRingsSolution.py", line 241, in OnServerInitComplete
    self.InitRings()
  File ".\python\kdshTreeRingsSolution.py", line 276, in InitRings
    ffcode2 = i + "Ring0" + j + "_0" + str(InitState) + ".animation.skipToEnd()"
NameError: name 'InitState' is not defined

### DIFF
--- a/Scripts/Python/kdshPillarRoom.py
+++ b/Scripts/Python/kdshPillarRoom.py
@@ -218,8 +218,7 @@ class kdshPillarRoom(ptResponder):
         #~ PtDebugPrint("kdshPilllarRoom:OnNotify  state=%f id=%d events=" % (state,id),events)
 
         if id in [1,2,3,4,5]: # One of the four levers or the reset ring were clicked by a player
-            code = "respLever0" + str(id) + ".run(self.key, events=events)"
-            exec(code)
+            globals()["respLever0{}".format(id)].run(self.key, events=events)
             return
 
         elif id in [6,7,8,9] and OnlyOneOwner.sceneobject.isLocallyOwned(): # One of the four levers actually pulled by an avatar
@@ -276,13 +275,12 @@ class kdshPillarRoom(ptResponder):
             return
 
         elif id >= 18 and id <=28 and PtWasLocallyNotified(self.key):
-            PtDebugPrint("Ladderbox %s entered." % ( id- 17))
+            ladderbox = id - 17
+            PtDebugPrint("Ladderbox %s entered." % ladderbox)
             
             LocalAvatar = PtFindAvatar(events)
 
-            code = "MultiStage" + str(id-17) + ".run(LocalAvatar)"
-            #~ PtDebugPrint("code = ", code)
-            exec(code) 
+            globals()["MultiStage{}".format(ladderbox)].run(LocalAvatar)
             
             return
         
@@ -343,8 +341,7 @@ class kdshPillarRoom(ptResponder):
                 
                 #this loop disables all ladder boxes, so the ladders can't be climbed.                
                 for count in range(1,12): 
-                    code = "Ladderbox" + str(count) + ".disable()"
-                    exec(code)
+                    globals()["Ladderbox{}".format(count)].disable()
                     
                 #determine the height of the highest pillar so we can reset them at the same speed as the pillars reset
                 highest = 0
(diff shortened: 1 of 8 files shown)


Scripts/Python/kdshPillarRoom.py
```diff
@@ -531,13 +525,23 @@ def DisableAppropriateLadders(self,pillar): # only the ladders boxes affected by
...
-            exec(code)
+            globals()["Ladderbox{}".format(each)].disable()
             #~ PtDebugPrint("Immediately disabling ladderbox", each)
 
     def EnableAppropriateLadders(self,id): # when pillar is done raising, enable appropriate ladder boxes with new lengths
         ageSDL = PtGetAgeSDL()
-        
+
+        def _ConfigLadder(ladderId, enable=None, deltaH=None):
+            assert enable is not None
```
zrax, line 535: Then why is it defaulted to `None` in the function signature?  ;)

Scripts/Python/kdshTreeRingsSolution.py
```diff
@@ -241,45 +241,15 @@ def OnServerInitComplete(self):
...
-            
-                #~ PtDebugPrint("Fastforwarding: Set = ",j," Ring = ",i," Position = ",InitState)
-                PtDebugPrint("\t",i,"Ring0",j," = ", InitState)
+        ageSDL = PtGetAgeSDL()
+        for i in ("Outer", "Middle", "Inner"):
+            for j in ("1", "2", "3"):
+                ring = "{location}Ring0{idx}".format(location=i, idx=j)
+                ring_attrib = "{ring}_0{bearing}".format(ring=ring, bearing=ageSDL[ring][0])
+                state = globals()[ring_attrib].animation.skipToEnd()
+                PtDebugPrint("{} = {}".format(ring_attrib, state), level=kWarningLevel)
```
zrax, line 252: The old code printed the InitState (i.e. the value read from ageSDL[ring][0]), but this prints the result of `skipToEnd()`...

Also, missing leading tab (if unintentional)

Scripts/Python/tldnShroomieBrain.py
```diff
@@ -234,21 +234,10 @@ def ShroomieSurfaces(self,spawn):
...
-        elif NearOrFar == "Far":
-            code = "target = SpawnFar0" + str(whichspawnpoint) + ".sceneobject.getKey()"
-        PtDebugPrint("target code:", code)
-        exec(code)
-        ShroomieMaster.sceneobject.physics.warpObj(target)
-        
+        target = globals()["Spawn{NearOrFar}0{SpawnPoint}".format(NearOrFar=NearOrFar, SpawnPoint=whichspawnpoint)]
+        ShroomieMaster.sceneobject.physics.warpObj(target.sceneobject.getKey())
+        globals()["respTrick0{}".format(whichbehavior)].run(self.key)
         code = "respTrick0" + str(whichbehavior) + ".run(self.key)"
```
zrax, line 240: Leftover variable

Scripts/Python/kdshPillarRoom.py
Hoikas: whoops. leftover from an old rev

Scripts/Python/kdshTreeRingsSolution.py
Hoikas: Good catch. I had a time figuring out what this particular mess of exec statements was doing.

Scripts/Python/kdshTreeRingsSolution.py
```diff
@@ -241,45 +241,15 @@ def OnServerInitComplete(self):
...
-            
-                #~ PtDebugPrint("Fastforwarding: Set = ",j," Ring = ",i," Position = ",InitState)
-                PtDebugPrint("\t",i,"Ring0",j," = ", InitState)
+        ageSDL = PtGetAgeSDL()
+        for i in ("Outer", "Middle", "Inner"):
+            for j in ("1", "2", "3"):
+                ring = "{location}Ring0{idx}".format(location=i, idx=j)
+                ring_attrib = "{ring}_0{bearing}".format(ring=ring, bearing=ageSDL[ring][0])
+                globals()[ring_attrib].animation.skipToEnd()
+                PtDebugPrint("\t{} = {}".format(ring, ageSDL[ring][0]), level=kWarningLevel)
```
zrax, line 252: Can `ageSDL[ring][0]` change as a result of `animation.skipToEnd()`?

Hoikas: It's possible that some message callback way down the line could, but within the confines of this method, no.

zrax: I have no other concerns then.

PR #656, opened 2020-05-18
Update resource-building scripts for future compatibility.

Deledrius:
This updates the resource.dat creation to render the SVG assets using a newer, maintained Python module "CairoSVG" instead of relying on the Pycairo/rsvg pair which is no longer actively maintained for Python on Windows.  Also included are a few Python 3.x+ fixes which should allow the scripts and build system to function with the newer interpreter.

This is a foundation for fixes coming in #650.

### DIFF
--- a/.github/workflows/linux-ci.yml
+++ b/.github/workflows/linux-ci.yml
@@ -12,6 +12,7 @@ jobs:
         sudo apt install cmake libcurl4-openssl-dev libopenal-dev libssl-dev \
           libogg-dev libopus-dev libspeex-dev uuid-dev libvpx-dev \
           libvorbis-dev qtbase5-dev
+        sudo pip install -r ${GITHUB_WORKSPACE}/requirements.txt
     - name: Build string_theory
       run: |
         mkdir -p build_deps && cd build_deps
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+# - Python Module Requirements
+
+# Required for building plClient's resource.dat assets
+-r ./Sources/Plasma/Apps/plClient/external/requirements.txt
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Plasma currently requires the following third-party libraries:
 
 The following libraries are optional:
 
-- (for building resource.dat) PyGTK - http://www.pygtk.org/downloads.html
+- (for building resource.dat) CairoSVG - https://cairosvg.org/
 - (for building resource.dat) Pillow - https://python-pillow.org/
 - (for plFontConverter) Freetype - http://freetype.org/
 - (for the GUI tools) Qt5 - http://www.qt.io/download-open-source/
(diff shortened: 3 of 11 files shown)


Deledrius:
Yes!  I knew I forgot something.  I'll add that.

Deledrius:
We aren't currently using that method.  Are you requesting that it be added to this PR?

Hoikas:
Yes, we want want to expose on master the packages needed for these scripts to work on Py2.7.

Deledrius:
Alright, will do.

requirements.txt
```diff
@@ -0,0 +1,7 @@
+# - Python Module Requirements
+
+# Required for building plClient's resource.dat assets
+https://download.lfd.uci.edu/pythonlibs/s2jqpv5t/cp27/pycairo-1.18.2-cp27-cp27m-win32.whl; sys_platform == "win32" and python_version ~= "2.7"
```
Hoikas, line 4: ???

.github/workflows/linux-ci.yml
```diff
@@ -12,6 +12,7 @@ jobs:
         sudo apt install cmake libcurl4-openssl-dev libopenal-dev libssl-dev \
           libogg-dev libopus-dev libspeex-dev uuid-dev libvpx-dev \
           libvorbis-dev qtbase5-dev
+        sudo pip install -r ${GITHUB_WORKSPACE}/requirements.txt
```
Hoikas, line 15: Too bad the resource.dat is never built on linux-ci since it's part of the plClient build :trollface:

Deledrius: Not currently, but hopefully that will change.  I know the scripts run on there because I've done so. :trollface:

requirements.txt
Deledrius: The old (Py2.7) version of CairoSVG tries to import cairo, which fails.  This manually installs a pre-compiled version of the module, and only does so on Windows with Python 2.7.  If/when we deprecate using Py2.7, this can be removed.  It's necessary for our 2.7 Windows builds on AppVeyor (and presumably any clean system that doesn't have cairo installed through the old installer on which we used to depend through PyGTK).

.github/workflows/linux-ci.yml
Deledrius: Since the requirements file allows specifications based on environment (such as OS), it makes sense to add this to both build systems for consistency (and to watch for accidental breakages).

Deledrius:
>I feel like the requirements.txt should live in the same directory as the resource generation script.
I considered this, but then the question is when is this step executed?

I think this is redundant.  We currently install all dependencies in the apt-get/vcpkg steps, and let CMake work out what's enabled and available.  This is consistent with that approach.  If these Python modules are unavailable, or not requested in the build, CMake will respond appropriately.

Also, a dependency-installation step that requires a file nested deep inside the project tree seems prone to breakage.  Unless we want to have CMake handle running PIP as a custom command inside the step that sets CAN_BUILD_RESOURCE_DAT on, I don't see a better way to do this (and installing dependencies during configuration seems like an equally bad idea).

PR #661, opened 2020-05-25
Fix Ahnoying Cathedral Seams

Hoikas:
This fixes white seams visible in the Ahnonay Cathedral by setting the render clear color to black.

## Comparison Screenshots
Enlarge to see the difference better!

### Before
ahny_before

### After
ahny_after


Deledrius:
Does this problem also manifest on MOULa?

Hoikas:
Yes:
Screenshot 2020-05-25 19 58 29

PR #663, opened 2020-05-25
Disable artifacts on AppVeyor build.

zrax:
AppVeyor has a maximum total artifact storage limit, and apparently no way to manually clean out old builds (though they are pruned automatically after 6 months)...

This can be re-enabled when some older builds expire, or if we switch to GH actions

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -86,6 +86,3 @@ before_package:
       ls
     }
 test: off
-artifacts:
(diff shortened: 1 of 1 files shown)


Deledrius:
That's unfortunate, but until we start using this to bundle releases they're mostly just a waste anyway.  Useful to check if things are being packaged correctly, but aside from that...

PR #666, opened 2020-06-02
C++17 cleanup

zrax:
Cleans up some macros and feature detection for C++11 keywords (we now require C++17), and replaces our custom `arrsize` utility with C++17's `std::size()` for static arrays.


Hoikas:
😈 😈 😈 😈 😈

PR #667, opened 2020-06-02
Download and install the PhysX SDK from archive.org

zrax:
This removes the final dependency on the non-redist devlibs for CI.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
   - CLIENT_TYPE: External
 install:
 - ps: |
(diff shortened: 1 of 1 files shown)


Hoikas:
O.o it's exploding on OpenAL...

Hoikas:
This is failing because the `vcpkg install [...]` command does not include the `--recurse` option, apparently 🙄

zrax:
I'm not too excited about fixing it with #690 on the horizon ;)

zrax:
Closing this in favor of #690.

PR #671, opened 2020-06-08
prepare_env script improvements

dpogue:
Check PowerShell version and attempt to use more modern commands if they are available. Also don't assume Windows path separators.

This runs (to the point of failing to detect CMake) on Powershell for Linux.

Also ensure CMake generates projects that target Win32 instead of Win64.

### DIFF
--- a/prepare_env.ps1
+++ b/prepare_env.ps1
@@ -1,9 +1,12 @@
+#!/usr/bin/env pwsh
+#Requires –Version 5.0
+
 param([string]$builddir='build')
 
 $devlibs_url = "https://github.com/H-uru/PlasmaPrefix/releases/download/2020.05.01/devlibs.zip"
 
 if (!(Test-Path -PathType Container $builddir)) {
-    Write-Host "Creating build folder at .\$builddir... " -noNewLine
+    Write-Host "Creating build folder at $builddir... " -noNewLine
     New-Item -ItemType directory $builddir | Out-Null
     Write-Host "OK" -foregroundColor Green
 }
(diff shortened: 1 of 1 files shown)


prepare_env.ps1
```diff
@@ -3,34 +3,48 @@ param([string]$builddir='build')
...
+    $zippath = Join-Path -Path $path -ChildPath "devlibs.zip"
+    $libpath = Join-Path -Path $path -ChildPath "devlibs"
+
     Write-Host "Downloading development libraries... " -noNewLine
-    ## This only works in PS3+ so we'll download the file directly
-    # Invoke-WebRequest $devlibs_url -OutFile devlibs.zip
-    $client = New-Object System.Net.WebClient
-    $client.DownloadFile($devlibs_url, $path + "\devlibs.zip")
+    if ($PSVersionTable.PSVersion.Major -ge 3) {
+        Invoke-WebRequest $devlibs_url -OutFile $zippath
```
Hoikas, line 22: PowerShell 3.0 was made available with Windows 8. Given that Windows 7 is no longer supported by Microsoft, I think we can safely dump the old code.

prepare_env.ps1
```diff
@@ -3,34 +3,48 @@ param([string]$builddir='build')
...
     Write-Host "OK" -foregroundColor Green
 
     Write-Host "Extracting development libraries... " -noNewLine
     New-Item -ItemType directory devlibs | Out-Null
-
-    $shell_app = New-Object -com shell.application
-    $zip  = $shell_app.namespace($path + "\devlibs.zip")
-    $dest = $shell_app.namespace($path + "\devlibs")
-    $dest.CopyHere($zip.items(), 0x14)      # 0x14 == "Yes to All" | "No progress dialog"
+    if ($PSVersionTable.PSVersion.Major -ge 5) {
```
Hoikas, line 31: PowerShell 5.0 is 4 years old - maybe we can assume that's available?

dpogue:
For what it's worth, the fact that the script *runs* on Linux doesn't mean that it is *useful* on Linux. Those Win32 devlibs aren't going to do much 😛 

PowerShell 5.0 seems like a nice choice. It's the minimum version that supports everything we want to do here, and is old enough to seemingly have widespread installation.

dpogue:
Updated to set PowerShell 5.0 as the minimum required version

PR #672, opened 2020-06-22
Whitespace cleanup for *Creatable.h files

dpogue:
This is partly to make them all more consistent, and partly because as I'm working on GL Pipeline stuff and need to ifdef out various classes, it's much easier to do if these files are cleanly organized.


dpogue:
> Maybe alphabetize them while you're at it?

Sure, let me create a million more conflicts for myself 😛

dpogue:
Fixed the cases where classes had been under the wrong headers and the reordering broke them.

PR #678, opened 2020-07-02
Clean up InputCore a bit for cross-platform work

dpogue:
* Got rid of plOSMsg, since the only thing using it was Win32-specific and can just use the `WM_` constants directly
* Moved the DInput stuff into its own file, and wrapped all references to it in `HS_BUILD_FOR_WIN32` guards
* Filled in the plKeyDef values for *nix based on the USB HID values (from [depp/keycode](https://github.com/depp/keycode))

Confirmed that both pnInputCore and plInputCore compile on both Windows (VS2019) and Linux (clang).

Note: Does not actually do any input handling for non-Windows.


dpogue:
* Well I guess I should look into that...
* plOSMsg was actually just an enum, not a Plasma Message, so no Creatable stuff to deal with.
* I'm happy to remove it if that's the direction we want to go (it would also solve the CI failures 😛)

dpogue:
CI failures have been addressed

PR #681, opened 2020-07-08
Add dice commands to KI chat.

Deledrius:
Per the feature request submitted to OpenUru (http://forums.openuru.org/viewtopic.php?f=92&p=9831), this implements two dice commands, one generic that expects a common dice notation string, and another that rolls a set of four fate/fudge dice.

The two recommended dice libraries, while promising, require a newer version of Python than the engine currently supports.  Instead, a naïve implementation that only supports a single notation is provided.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -43,6 +43,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 import re
 import time
+import random
 
 # Plasma Engine.
 from Plasma import *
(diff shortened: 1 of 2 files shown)


Deledrius:
Two questions:
1. Should this use PtGetLocalizedString for the strings to make it easier to localize them in the future?
2. Should this perhaps use `/roll` instead of `/dice` to make the command more semantically readable and consistent with most other dice-rolling chat commands in other games?

dpogue:
> Should this use PtGetLocalizedString for the strings to make it easier to localize them in the future?

Probably yes,

> Should this perhaps use `/roll` instead of `/dice` to make the command more semantically readable and consistent with most other dice-rolling chat commands in other games?

I think `/roll 4d6` is probably more consistent with conventions in other apps. We might want to change `/fate` to `/rollfate` or something to better namespace it to dice rolling.

We do not support `/roll 4dF` notation?

Deledrius:
I was thinking at that point, I could allow the `/roll` command to accept other "shorthands", for instance `/roll fate` would perform the same as the `/fate` command currently.  I wasn't aware of the dF notation.  I suppose I could add that.  Is it common?  The quick research I did suggested that Fate commonly rolls in sets of four, so I went with that.

The [dice library suggested](https://pypi.org/project/dice/) handles *a lot* of formats and it would be lovely to just toss the string at that, but alas.

dpogue:
`/roll fate` seems like a good option. The `dF` notation probably isn't very common.

Deledrius:
Alright, I'll roll :trollface: them together and push an update. :)

Deledrius:
Alas, until we version the `.loc` files there isn't an easy way for me to add the strings for PtGetLocalizedString here.

Scripts/Python/ki/xKIChat.py
```diff
@@ -1435,3 +1436,42 @@ def MarkerGameReward(self, reward):
             self.chatMgr.AddChatLine(None, "Current Reward: '{}'".format(markerMgr.reward), 0)
         else:
             self.chatMgr.AddChatLine(None, "This game has no associated reward.", 0)
+
+    ## Rolls the specified dice and displays the results
+    def RollDice(self, dice_str):
+        if not dice_str:
+            self.chatMgr.AddChatLine(None, "Please provide a number of dice and number of faces in common dice notation (e.g. 2d6), or 'fate' to roll a set of fate dice.", kChat.SystemMessage)
```
Hoikas, line 1443: Why not simply roll one cuboid die by default?

Deledrius: I've been debating that.  I weighed it against the utility of a self-documenting help response.  Not sure which is best here.  Error message or automatic 1d6.

Hoikas: From the perspective of someone who does not play tabletop games, I would have had no idea how to "provide a number of dice and number of faces" (until yesterday's research) and would expect rolling a die to yield a single number 1-6. :wink:

PR #682, opened 2020-07-08
KI Easter Egg Updates/Age Constants

DoobesURU:
Adds new Easter Eggs (feathers) and Constants for Fan Ages.

- Adds yellow, raven and dove feathers for GoMePubNew, Chiso, and Veelay respectively.

- Adds proper name constants for Fehnir's House, GoMePubNew, Chiso and Veelay.

- Adds /look comments for GoMePubNew and Chiso.

- Adds GoMePubNew, Chiso and Veelay to "No Invite" list.

- (proposed) Takes out unneeded "The" from some of the Age Name constants to help them display better in areas of the KI. While it does appear to cut off the latter part of the text in lists, this should eliminate instances where the Age name only displays as "The...".

### DIFF
--- a/Scripts/Python/ki/xKIConstants.py
+++ b/Scripts/Python/ki/xKIConstants.py
@@ -98,6 +98,11 @@ kEasterEggs = {
                "Cleft" : {"see" : "You see sand for as far as the eye can see. Gonna need a vehicle of some sort.",
                           "exits" : "... well, I don't know. Maybe you can ask the old man (if he ever stops listening to that music!).",
                           "people" : "an old man. Ok, maybe he's not standing. BTW, wasn't he on M*A*S*H?"},
+               "GoMePubNew" : {"see" : "You are in a festively yet dimly lit Pub. The echoes of Messenger work and play echo along the walls of the massive main chamber.",
+                               "exits" : "... wait, where ARE the exits? Who built this place anyway?",
+                               "people" : "You see many people...unless you are here by yourself...in which case, you don't."},
+               "ChisoPreniv" : {"see" : "All around you are bookshelves upon bookshelves but not a book to read. To the south is a large set of windows that reveal a vast landscape that you'll never be able to access.",
+                                "exits" : " ... I guess in this library, the only exit is through the window, but let's not do that"},
 }
 
 ## Constants for Age display names.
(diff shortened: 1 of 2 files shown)


Scripts/Python/ki/xKIConstants.py
```diff
@@ -130,14 +135,19 @@ class kAges:
                "Shaft" : "D'ni-Tiwah",
                "Spy Room" : "D'ni-Ae'gura",
                "spyroom" : "D'ni-Ae'gura",
+##             Fan Ages
                "Trebivdil" : "Tre'bivdil",
                "trebivdil" : "Tre'bivdil",
-               "vothol" : "Vothol Gallery"}
+               "vothol" : "Vothol Gallery",
+               "FehnirHouse" : "Fehnir's House",
```
dpogue, line 142: Vothol Gallery & Fehnir's House are both Cavern locations, no? They should be `D'ni-<something>`?

Scripts/Python/ki/xKIConstants.py
```diff
@@ -98,6 +98,11 @@
                "Cleft" : {"see" : "You see sand for as far as the eye can see. Gonna need a vehicle of some sort.",
                           "exits" : "... well, I don't know. Maybe you can ask the old man (if he ever stops listening to that music!).",
                           "people" : "an old man. Ok, maybe he's not standing. BTW, wasn't he on M*A*S*H?"},
+               "GoMePubNew" : {"see" : "You are in a festively yet dimly lit Pub. The echoes of Messenger work and play echo along the walls of the massive main chamber.",
+                               "exits" : "... wait, where ARE the exits? Who built this place anyway?",
+                               "people" : "many people...unless you are here by yourself...in which case, you don't."},
+               "ChisoPreniv" : {"see" : "All around you are bookshelves upon bookshelves but not a book to read. To the south is a large set of windows that reveal a vast landscape that you'll never be able to access.",
+                                "exits" : " ... I guess in this library, the only exit is through the window, but let's not do that"},
```
Hoikas, line 105: :trollface:

Scripts/Python/ki/xKIConstants.py
Hoikas: Would this convention also extend to the various pubs?

DoobesURU: That's a good point dpogue. Let me test how that looks in Destiny.

cwalther: Don’t listen to my bikeshedding, but I don’t think this change is a proper continuation of the previous pattern. In the precedent, the “D'ni-” prefix is followed by a D’ni area name, not by an English description. So the pubs would all become “D'ni-J'taeri” (or whatever it was). The ambiguity would make it less useful though.
(In other words, the change makes things no less inconsistent than before. It does make them marginally more useful, so I guess it’s OK.)

PR #684, opened 2020-07-16
Remove pfVirtualCam1 Effect Plate.

Hoikas:
The effect plate was effectively unused until #679. Fixing it to not be gated behind the status log introduced a regression in which the spinning dots and intro video were hidden.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfCamera/plVirtualCamNeu.cpp
+++ b/Sources/Plasma/FeatureLib/pfCamera/plVirtualCamNeu.cpp
@@ -152,7 +152,6 @@ plVirtualCam1::plVirtualCam1()
     // set initial view position
     fOutputPos.Set(100,100,100);
     fOutputPOA.Set(0,0,0);
-    fEffectPlate = nil;
     fFreezeCounter = 0;
     fFadeCounter = 0;
     fX = fY = 0.5f;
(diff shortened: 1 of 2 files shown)


Deledrius:
Is this the effects plate mentioned in the Transition Manager as the replacement for the old hack?

dpogue:
> Is this the effects plate mentioned in the Transition Manager as the replacement for the old hack?

I don't see any mention of this in the Transition Manager. Were you thinking of this comment? https://github.com/H-uru/Plasma/blob/c835484d36c5a6996a0782eb3f11ce0b91de9bd5/Sources/Plasma/PubUtilLib/plPipeline/plTransitionMgr.cpp#L260-L265

My guess is that the Transition Manager was added later as an improved attempt to hide the screen while the data was being loaded into the renderer, because the implementation (and naming) is essentially identical to this.

Deledrius:
I was thinking of this part:

https://github.com/H-uru/Plasma/blob/c835484d36c5a6996a0782eb3f11ce0b91de9bd5/Sources/Plasma/PubUtilLib/plPipeline/plTransitionMgr.cpp#L46-L51

dpogue:
Ahh, I think that sort of control over the plate contents is still yet to be implemented (alas, because there were some cool effects like that in the Descent demo)

PR #685, opened 2020-07-22
Handle font characters with negative baselines

dpogue:
Some font have characters with a negative baseline (usually the underscore character). When multiplied with the (unsigned) stride, this was turning into a value in the vicinity of UINT32_MAX and that gets subtracted from the pointer into the texture image data. In this case, it was going out of bounds and causing the client to segfault.

In reality, `fDestStride` is generally the width of the texture × 4 (bpp), so it's very unlikely that this will ever exceed the capacity of a signed 32-bit int.


zrax:
Are you building for 64-bit?  32-bit pointers should wrap even with the implied promotion of `signed * unsigned => unsigned`

dpogue:
Yes, building for 64-bit on Linux.

PR #688, opened 2020-07-22
Fix `hsMatrix44::DecompRigid()` quat decomposition.

Hoikas:
At some point, `hsQuat::FromMatrix44()` was changed to a static method, but `hsMatrix44::DecompRigid()` was not updated and continued to use it as a member method. This resulted in the quaternion being discarded immediately.

In the interest of preventing these sorts of bugs in the future, I added C++17's `[[nodiscard]]` attribute where it made sense in `hsQuat`.

Required by PhysX 4.1.


Hoikas:
:trollface:

PR #695, opened 2020-08-06
Fix GoMePubNew look command text

DoobesURU:
Fixes Ae'gura GoMe Pub's /look text to better flow with the existing script.


Hoikas:
For future reference, here is the formatter the /look command uses:
```python
"{}: {} Standing near you is {}. There are exits to the{}".format(GetAgeName(), see, people, exits)
```

PR #696, opened 2020-08-10
replace Cyan's xEnum by Python's enum

aqua-uru:


Scripts/Python/clftNpcZandi.py
```diff
@@ -86,9 +86,9 @@
 IgnoreTime = 120
 PageTurnInterval = 120
 
-TimerID = xEnum.Enum("IgnoreFinished = 10, TurnPage")
+TimerID = enum.IntEnum("TimerID", {"IgnoreFinished": 10, "TurnPage": 11})
```
Hoikas, line 89: Maybe `start=10` instead?

aqua-uru: I considered it but I thought it could be too easily read over as "start=0" since that occurs in some places, though indeed it would be more consistent...

Scripts/Python/grsnPowerOn.py
```diff
@@ -132,7 +132,9 @@
 gearStartingDown = False
 weightEngageDisabled = False
 
-TimerID = xEnum.Enum("UpElevatorSwitch, DownElevatorSwitch")
+class TimerID(enum.IntEnum):
+    UpElevatorSwitch = 1
```
Hoikas, line 136: I think this will be more Pythonic in cases where we don't need to change the start value:
```suggestion
    UpElevatorSwitch = enum.auto()
```

aqua-uru: Changed for all of the ones that start at 1!

Would be nice if there was a way to have auto but specify the starting value... Maybe I should take it up to python-dev ;)

PR #697, opened 2020-08-14
Make the Python Plasma files into a proper module, allowing static code analysis

aqua-uru:
I did the following:

- Create an empty `plasma/__init__.py` to make the folder into a module.
- Rename `plasma/Plasma.py` to `plasma/PlasmaStubs.py`.
- Create a new `plasma/Plasma.py` which fills itself with the contents of the engine's builtin Plasma module if available, else the contents of `plasma/PlasmaStubs.py`.
- Change all imports to actually point to the files in `plasma` (see regexes below).
- Change the contents of all functions in `plasma/PlasmaStubs.py` from `pass` to `raise NotImplementedError()`, further improving static code analysis (otherwise errors appear like 'function called that does not have a return value').

Regexes used:

**PATTERN**
`from (PlasmaConstants|PlasmaControlKeys|PlasmaKITypes|PlasmaNetConstants|PlasmaTypes|PlasmaVaultConstants|Plasma) import \*`
**REPLACEMENT**
`from plasma.$1 import *`

**PATTERN**
`import (PlasmaConstants|PlasmaControlKeys|PlasmaKITypes|PlasmaNetConstants|PlasmaTypes|PlasmaVaultConstants|Plasma)`
**REPLACEMENT**
`from plasma import $1`


aqua-uru:
This seems to work well at the moment, both for achieving static code analysis and for not breaking the game, except for the fact that **I can't open my Relto book** (not by clicking the icon or by pressing F3). No Python errors appear while running the game. I looked around a bit but couldn't find a probable cause yet.

aqua-uru:
> rename Plasma.py to Plasma.pyi and change the method contents to `...`

Oh yes I forgot about that! I'll look into it again.

> all of the `from Plasma<whatever>` statements being reduced to a single `from plasma import *`

I thought about that too (it's what we were looking at earlier) but I wanted to leave it for later, first getting the concept working properly. We'd also need to be sure there are no name conflicts between the different parts, do you know that from the top of your head?

Hoikas:
That's a good question regarding name conflicts. The Plasma.py file already has conflicts within itself, but I don't think there should be any between the files themselves. We'll want to look carefully to make sure, of course.

aqua-uru:
It seems that the `.pyi` files aren't quite what we thought. I think they are a way to add typing information to existing code which already lives in a separate `.py` file. But it doesn't seem possible to use them to have just a `.pyi` file. So in our case it wouldn't make much sense, I think. This is also what I'm seeing from a few experiments. But I'd be happy to hear if there is a different way to use them!

Some references:
- https://stackoverflow.com/a/59066738
- https://www.python.org/dev/peps/pep-0484/#stub-files

aqua-uru:
I found out the problem with the Relto book: `Plasma` is not the only module that has a built-in version in the engine: I figured out [this also happens for](https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.h#L72) `PlasmaConstants`, `PlasmaNetConstants` and `PlasmaVaultConstants`, the first of which was causing the problem with the Relto Book. So these should also get the IDE-vs-engine treatment (although I haven't encountered problems caused by the other two, probably because they only define constants and no functions or classes that will be instanced).

I felt making four 'Plasma...Stubs.py`files was ugly, so I've changed to a different method for doing the 'injection'. The steps in my first original comment can be replaced by the following:

- Create an empty `plasma/__init__.py` to make the folder into a module.
- At the bottom of each of `plasma/Plasma.py`, `plasma/PlasmaConstants.py`, `plasma/PlasmaNetConstants.py`, and `plasma/PlasmaVaultConstants.py`, append the code shown below.
- Change all imports to actually point to the files in `plasma` (see regexes in original comment).
- Change the contents of all functions in `plasma/Plasma.py` from `pass` to `raise NotImplementedError()`, further improving static code analysis (otherwise errors appear like 'function called that does not have a return value'). The other tree files have no methods that are expected to return anything, so this is not relevant there.

The code appended at the bottom of the four files, with `Plasma` replaced by the relevant module name:
```python
try:
    from Plasma import *  # builtin exists if if we are in the engine
except ImportError:
    pass  # fallback to stubs for code analysis
```

So this now seems to work fairly well:
- The behavior of the game seems unaffected.
- The IDE does static code analysis. However on my own computer and IDE (VS code), it only seems sporadically able to resolve symbols (not dependent on which symbol, just sometimes it works and sometimes doesn't), or takes a very long time. I'm not sure if this is simply down to the amount of symbols, or what. I would be interested to know how it goes on someone else's system.

aqua-uru:
It could be the speed issue is the same as [this](https://github.com/microsoft/vscode-python/issues/11232), which is apparently [getting addressed](https://github.com/microsoft/vscode-python/issues/11232#issuecomment-634946017).

Scripts/Python/plasma/Plasma.py
```diff
@@ -5120,4806 +5120,4811 @@ class ptGUIPopUpMenu:
...
     def setWindDir(self,s, secs = 0):
         """Sets the attribute to s over secs time"""
-        pass
+        raise NotImplementedError()
 
+
+try:
+    from Plasma import *  # builtin exists if if we are in the engine
+except ImportError:
+    pass  # fallback to stubs for code analysis
```
Hoikas, line 9930: I think I have a slight preference for this handling to be done in __init__.py by:
```python
try:
    from Plasma import *
except ImportError:
    from .Plasma import *
```

I am assuming that Python can correctly tell the difference between Plasma and .Plasma however :/

MareinK: I made a new branch using this idea, see [here](https://github.com/aqua-uru/Plasma/blob/plasma-module-single/Scripts/Python/plasma/__init__.py). It seems to work fine in-game and the code analysis works too (at least as well as the first branch does on my machine), except for the file `PlasmaTypes.py`. The problem there is that it imports the `plasma` package because it uses classes from `Plasma.py` while it is also itself part of the `plasma` package now. This causes pylint to show a 'function already defined' error (though it seems no problem in-game).

Any ideas on how to deal with that? I thought about having it import `Plasma.py` specifically to avoid the recurrence, but that would require a try-catch to make sure it uses the right version (in-engine or stubs). Maybe have a `PlasmaStubsLoader.py` or something that does the try-catch for all maybe-stubs files, and have `from .PlasmaStubsLoader import *` in both `PlasmaTypes.py` and `__init__.py`?

There's also this line in `pch.py`:

```python
if not hasattr(Plasma,name) and not hasattr(PlasmaTypes,name):
```

which should probably be changed to `if not hasattr(plasma, name)`.

aqua-uru: I added the `PlasmaLoader.py` to the `plasma-module-single` branch, and made the change in `pch.py`. All seems to be working well! Do you prefer the `single` branch to this one (using a single `plasma` import rather than the original separate ones? If so, I will use it to replace this one.

Hoikas: I can't find the "go to definition" button. But, when I hover over `ptAgeVault()` in Ahnonay.py, it helpfully displays the docstring. I'm liking the `plasma-module-single` branch--of course, other opinions are always welcome.

Before we merge any of this, we need to be sure to test that running this PR through plPythonPack works as expected.

aqua-uru:
Changed this branch to the former `plasma-module-single` (now deleted), the old contents of this branch are at `plasma-module-multi`.

aqua-uru:
> Before we merge any of this, we need to be sure to test that running this PR through plPythonPack works as expected.

Is this something that someone with build-skills would like to check?

Hoikas:
From what I can tell, a small change will be need to be made to plPythonPack and any other utilities that generate python.pak files. By default, it packages the "plasma" directory into the root level instead of allowing module support. I think a better way for that to work is to just consider any subdirectories of "Python" without __init__.py as root level content. Does this make sense, @aqua-uru?

With the change to plasma becoming a module, the files plasma/pch.py and plasma/ptWordFilter.py will need to me moved from the plasma subdirectory into the root level module. I think? Unless you've fixed up those imports.

Deledrius (changes requested):
Due to condensing the constant modules into the `plasma` module, many existing imports were merged, leaving a redundant `import plasma` alongside a `from plasma import *` in many scripts.  If possible, an additional commit on this PR condensing those would be appreciated.  Per PEP-8, we should prefer the `import plasma` format between the two.

Scripts/Python/xCheat.py
```diff
@@ -48,9 +48,9 @@
 
 
 def ListYeeshaPages(args):
-    import Plasma
+    import plasma
```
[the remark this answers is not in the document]
Hoikas: :trollface:

Hoikas:

PR #698, opened 2020-08-14
Use 'black' to format all Python files

aqua-uru:
I first fixed the weird non-aligned `#~` comments, which otherwise got messed up by black, using repeated applications of the below regex (until there were no more matches). Then I replaced all `#~` by `#` (taking some care not to get false positives and affect some other formatting). And then I ran black.

**PATTERN**
`^( *)#~(.*?)\n\1( +)#~(.*?)\n`
**REPLACEMENT**
`$1#~$2\n$1#~$3$4\n`


Hoikas:
@Deledrius and others who like tinkering with the Python... Do we have a preference on black when compared to others such as autopep8? I'm not really a fan of how black handles multiline statements, especially this particular gem:
```python
            if (
                behavior == PtBehaviorTypes.kBehaviorTypeStandingJump
                or behavior == PtBehaviorTypes.kBehaviorTypeWalkingJump
            ):
```

Deledrius:
I'm not in love with that formatting choice.  What does autopep8 do?

Hoikas:
[Autopep8](https://pypi.org/project/autopep8/) is similar to black, except it uses the PEP8 styling rules. There is a slightly different paradigm between the two formatters though. Black has one right way(TM) but autopep8 has some configurable options, which can be both a good and bad thing.

Deledrius:
We've been trying to use PEP8 for everything, and I prefer something we can configure, as long as we can provide those options easily for all contributors.

Hoikas:
Maybe something like this [action](https://github.com/marketplace/actions/autopep8)?

Hoikas:
Another consideration is whether we should be using GitHub's [super linter](https://github.com/github/super-linter).

aqua-uru:
I don't think my preference should weigh very strongly because I have only been working on the code for a short time and there's no telling whether I will be working on it much in the future. I do think it would be great to use any autoformatter rather than none. Having said that I'll make the case for black here.

There are I think two reasons why I like black:
- No matter how a file is formatted beforehand, as long as it has the 'same code' it will be identical after blackening. "black doesn't take previous formatting into account." It means you don't need to think about formatting _at all_ while coding, because it will have no impact after running the autoformatter. Compare to autopep8 where a lot of existing formatting is retained, so you still need to make decisions about it. (note: I think there is at least one small exception situation in black)
- There is (almost) no configuration. I think this is a good thing, because again, it means you don't need to think about it.
- Extra reason: it's very popular.

> I'm not in love with that formatting choice. What does autopep8 do?

This is the original code:

```python
            if (behavior == PtBehaviorTypes.kBehaviorTypeStandingJump or behavior == PtBehaviorTypes.kBehaviorTypeWalkingJump):
```

Here's black's result again:

```python
            if (
                behavior == PtBehaviorTypes.kBehaviorTypeStandingJump
                or behavior == PtBehaviorTypes.kBehaviorTypeWalkingJump
            ):
```

And running it through autopep8 produces the following:

```python
            if behavior == PtBehaviorTypes.kBehaviorTypeStandingJump or behavior == PtBehaviorTypes.kBehaviorTypeWalkingJump:
```

In my opinion that's not great either, the line stays 114 characters long. autopep8 just doesn't have a way to deal with it. I guess that besides these two options there's really not much else a formatter could do here (without resorting to using `\` :anguished:).

Could it be that this is actually a symptom of bad code? Maybe that's going on a tangent, but writing instead for example `behavior in (PtBehaviorTypes.kBehaviorTypeStandingJump, PtBehaviorTypes.kBehaviorTypeWalkingJump)` results in the following using black:

```python
            if behavior in (
                PtBehaviorTypes.kBehaviorTypeStandingJump,
                PtBehaviorTypes.kBehaviorTypeWalkingJump,
            ):
```

This seems much clearer to me semantically and in terms of formatting. So in this way the resulting formatting could also point at issues with the code itself, rather than with the formatter. autopep8 still keeps it as one line when using the `in` code.

If the if-condition could not be written using `in` (i.e. `a == b and c == d`), then perhaps the solution would be to make two boolean variables in the lines above the `if` and combine those in the condition, resulting in fine formatting regardless of the formatter. But now I'm definitely going on a tangent.

> Autopep8 is similar to black, except it uses the PEP8 styling rules

Note that black is PEP8-compliant. So the output uses PEP8 styling rules, and any PEP8 formatting that was already used in the code may be retained (though it may instead be differently formatted of course). On that note, it seems from the example above that autopep8 is not PEP8-compliant, since it produces too long lines :)

Links:
- [black on github](https://github.com/psf/black)
- [The Black code style](https://github.com/psf/black/blob/master/docs/the_black_code_style.md)

Deledrius:
Great points, thank you so much for writing that all out with examples.

I'm sold.  Using the Black code formatter is 👍 from me.

Hoikas:
I think it's a good point that the code in that particular case is just bad. Is there a way that we can tell black what the maximum line length should be? 80 columns is rather small these days, and it would be nice to use a more modern de-facto standard, such as the width of the GitHub code viewer.

Deledrius:
That would be a nice compromise.

aqua-uru:
Yes, one of the few configuration arguments for black is the line length (`-l`). It's 88 by default.

By the way, this might be of interest: [Migrating your code style without ruining git blame](https://github.com/psf/black#migrating-your-code-style-without-ruining-git-blame).

Hoikas:
@aqua-uru I just read that very page about git-blame. Maybe you could introduce the .git-blame-ignore-revs file for us? There are a few revisions in the history that we can use that for, aside from this pull request :smile:

Hoikas:

PR #702, opened 2020-08-21
Update packages in vcpkg cache

zrax:
(Moved from #701 due to being in the wrong fork and thus duplicating builds)

The vcpkg cache contains old, stale versions of several packages, which do not get updated automatically by vcpkg install. This ensures they get updated before installing new packages.

This now also removes the old manually-created x86-windows-static-dyncrt triplet in favor of the upstream-supported x86-windows-static-md triplet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,13 @@ set_package_properties(Python3 PROPERTIES
     TYPE REQUIRED
 )
 
+find_package(EXPAT REQUIRED)
+set_package_properties(EXPAT PROPERTIES
+    URL "http://expat.sourceforge.net/"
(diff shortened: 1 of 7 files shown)


Hoikas:
I RDP'd into an AppVeyor instance that was failing to build zlib, but I couldn't figure out what exactly the problem was. So, I opened Microsoft/vcpkg#13105. Maybe one of their pkgconfig people can figure it out.

Hoikas:
Now we are trying to link against debug libopus 😞

Hoikas:
Ugh, MSYS2 broke between the first and second build configuration. *sigh*

Hoikas:
@Zrax I've fixed all the weird failures -- are these changes OK?

PR #706, opened 2020-08-23
Create .git-blame-ignore-revs

aqua-uru:
I created the file and entered a handful of my own 2to3 commits. Not sure which others are good candidates. I will add the upcoming 'black' commit when it happens.

It seems to work well after running `git config blame.ignoreRevsFile .git-blame-ignore-revs`.


aqua-uru:
Added!

PR #709, opened 2020-08-26
Fix fallout from upstream libexpat library rename.

Hoikas:
As of libexpat 2.2.8, the library was renamed from libexpat to expat. In 2.2.9, this was supposedly reverted back to libexpat; however, in practice, I find that vcpkg builds libexpatMD. Therefore, we now use libexpat's config. However, this is not present in the most recent devlibs bundle with expat 2.2.7, so we fall back on the out of date cmake find module in that case.

This should fix the latest CI failure in #702.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfLocalizationMgr/CMakeLists.txt
+++ b/Sources/Plasma/FeatureLib/pfLocalizationMgr/CMakeLists.txt
@@ -2,7 +2,24 @@ include_directories(../../CoreLib)
 include_directories(../../NucleusLib/inc)
 include_directories(../../PubUtilLib)
 
-include_directories(${EXPAT_INCLUDE_DIR})
+# Skip CMake's built in FindExpat module and use libexpat's CMake config, if possible,
+# due to library rename in libexpat 2.2.8 and higher.
+find_package(EXPAT CONFIG QUIET)
+if(NOT TARGET expat::libexpat)
+    message(DEBUG "libexpat CMake Config not found, using FindExpat module.")
+    find_package(EXPAT REQUIRED)
+    add_library(expat::libexpat STATIC IMPORTED)
+    set_target_properties(expat::libexpat PROPERTIES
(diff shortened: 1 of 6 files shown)


Sources/Tools/MaxMain/CMakeLists.txt
```diff
@@ -93,7 +93,6 @@ target_link_libraries(MaxMain MaxExport)
 target_link_libraries(MaxMain MaxPlasmaMtls)
 target_link_libraries(MaxMain ${3dsm_LIBRARIES})
 
-target_link_libraries(MaxMain ${EXPAT_LIBRARY})
```
zrax, line 96: There are other references to ${EXPAT_LIBRARY} in plClient, pfPythonTest and plLocalizationEditor

Sources/Plasma/FeatureLib/pfLocalizationMgr/CMakeLists.txt
```diff
@@ -2,14 +2,27 @@ include_directories(../../CoreLib)
...
+    URL "https://github.com/libexpat/libexpat"
+    DESCRIPTION "Expat XML Parser for C"
+    TYPE REQUIRED
+)
 
 if(WIN32)
     add_definitions(-DWIN32)
-
-    if(EXPAT_IS_STATIC)
-        add_definitions(-DXML_STATIC)
```
zrax, line 11: It looks like the "fix" for this is vcpkg-specific, so if we want to support non-vcpkg versions of expat, we should keep this :(

Hoikas: 😢

Sources/Tools/MaxMain/CMakeLists.txt
Hoikas: I think I already got the one in plClient, and I'll get the other two shortly.

PR #717, opened 2020-08-30
Python Error Handling Improvements

Hoikas:
This fixes some issues observed with the Python Interface's error handling that were exacerbated by the move to Python 3. It adds the ability to log [unraisable exceptions](https://vstinner.github.io/sys-unraisablehook-python38.html) and detects null bytes being sent to Python as UTF-8 in debug mode.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
@@ -736,7 +736,7 @@ public:
         delete [] strData;
     }
 
-    void ExceptHook(PyObject* except, PyObject* val, PyObject* tb)
+    void UncaughtHook(PyObject* except, PyObject* val, PyObject* tb)
     {
         PyErr_Display(except, val, tb);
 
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -784,14 +800,33 @@ PYTHON_METHOD_DEFINITION(ptErrorRedirector, excepthook, args)
...
+PYTHON_METHOD_DEFINITION_NOARGS(ptErrorRedirector, flush)
+{
+    // no-op
+    PYTHON_RETURN_NONE;
+}
+
 PYTHON_START_METHODS_TABLE(ptErrorRedirector)
     PYTHON_METHOD(ptErrorRedirector, write, "Adds text to the output object"),
-    PYTHON_METHOD(ptErrorRedirector, excepthook, "Handles exceptions"),
+    PYTHON_METHOD(ptErrorRedirector, excepthook, "Handles uncaught exceptions"),
```
Deledrius, line 827: Keeping this as `excepthook` for compatibility?

Hoikas: It's still called `sys.excepthook`

Hoikas:
Note I removed the `std::string` null check variant due to chaos from trying to forward declare `std::string` on gcc.

PR #726, opened 2020-10-01
Button Puzzle Scripts

Hoikas:
There was a [request on the forum](https://forum.guildofwriters.org/viewtopic.php?f=9&t=7177) for Python scripts for making puzzles that use a set of buttons in a combination to open doors. This script is generalized to take a combination, set of buttons, and some SDL variables. When the combination is input correctly, the solved SDL variable is set, which can trigger any arbitrary reaction with xAgeSDLBoolRespond.py

### DIFF
--- a/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
+++ b/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
@@ -1,0 +1,226 @@
+# -*- coding: utf-8 -*-
+""" *==LICENSE==*
+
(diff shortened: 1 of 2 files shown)


Hoikas:
Fixed an issue by which avatar animations may not be synchronized.

Hoikas:
Given that this is in the offline KI and an Age is in testing on Minkata-alpha using this script, are the any objections to merging this?

Hazado:
ive used it several times in that age and never had issues

Deledrius (approved):

PR #727, opened 2020-10-02
Add source groups for game scripts and associated files.

Deledrius:
This makes it easier to find, browse, and edit the files now included in the recently-added Scripts project.

### DIFF
--- a/Scripts/CMakeLists.txt
+++ b/Scripts/CMakeLists.txt
@@ -1,3 +1,11 @@
+file(GLOB_RECURSE GameScripts "${CMAKE_CURRENT_SOURCE_DIR}/Python/*.py")
+file(GLOB_RECURSE GameSDL "${CMAKE_CURRENT_SOURCE_DIR}/SDL/*.sdl")
+file(GLOB_RECURSE GameData "${CMAKE_CURRENT_SOURCE_DIR}/dat/*.age" "${CMAKE_CURRENT_SOURCE_DIR}/dat/*.fni")
+
+source_group("Python" FILES ${GameScripts})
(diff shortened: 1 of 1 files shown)


Hoikas (approved):

Scripts/CMakeLists.txt
```diff
@@ -1,10 +1,20 @@
+file(GLOB_RECURSE GameScripts "${CMAKE_CURRENT_SOURCE_DIR}/Python/*.py")
+file(GLOB_RECURSE GameSDL "${CMAKE_CURRENT_SOURCE_DIR}/SDL/*.sdl")
+file(GLOB_RECURSE GameData "${CMAKE_CURRENT_SOURCE_DIR}/dat/*.age" "${CMAKE_CURRENT_SOURCE_DIR}/dat/*.fni")
```
zrax, line 3: I'm not suggesting you change anything, but please note that glob lists will only get updated when CMake is re-run...  They will never get automatically updated when files are added or removed.

Deledrius: I'm aware, and considered this.  Since they aren't used as part of the build, I felt it was okay for that loophole to exist.  The copy commands all will continue to work fully, so this won't affect any functionality, just requires a re-run of CMake for a convenience function.  As far as drawbacks go, it seemed acceptable.

Hoikas: It's probably better to do any dedicated hacking on these files in a more filesystem driven editor, such as VS Code.

Deledrius: I agree, but I've found myself wanting to pull them up several times for reference while working on engine code, and this makes it easier to do so.

PR #729, opened 2020-10-02
Fix Er'cana Harvester position (#668)

### LINKED ISSUE #668: Er'cana Harvester
Per @plancsar:
> Weird bug in Er'cana:
> I started the train, then linked out and back in. Ran to the factory, pulled the lever to detach the wagon, but nothing happened.
> I then got back to the train controls, pulled the right lever, and suddenly the train was warped back to its starting point, running all the way to the factory (with the gangways to the wagon down), and only then I could detach the wagon normally.

This occurs in #650 and on Cyan's MOULa Shard.

Hoikas:
The harvester's position was not being properly fast-forwarded if the Age emptied while it was in-transit.


Hoikas:
This script change has been tested on Minkata and will be delivered to the MOULa pipeline Monday. Merging.

PR #731, opened 2020-11-06
Clean up uses of deprecated CPython APIs

zrax:


Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
```diff
@@ -79,7 +79,7 @@ void PythonInterface::initPython(const plFileName& rootDir, const std::vector<pl
         config.optimization_level = 2;
         config.write_bytecode = 0;
         config.user_site_directory = 0;
-        config.program_name = L"plasma";
+        PyConfig_SetString(&config, &config.program_name, L"plasma");
```
Hoikas, line 82: Related: I learned later that PyConfig can leak memory and it needs to be freed with `PyConfig_Clear` 😢

Sources/Plasma/FeatureLib/pfPython/pyGUIControlMultiLineEdit.cpp
```diff
@@ -203,9 +203,12 @@ void pyGUIControlMultiLineEdit::SetEncodedBuffer( PyObject* buffer_object )
...
             // something to do here... later
-            uint8_t* daBuffer = nil;
-            Py_ssize_t length;
-            PyObject_AsReadBuffer( buffer_object, (const void**)&daBuffer, &length);
+            Py_buffer view;
+            PyObject_GetBuffer(buffer_object, &view, PyBUF_SIMPLE);
+            uint8_t* daBuffer = (uint8_t*)view.buf;
+            Py_ssize_t length = view.len;
+            PyBuffer_Release(&view);
+
```
Hoikas, line 211: As part of the scripts update, I *think* I completely removed all uses of buffers because of this deprecation. Maybe we should just remove this binding?

Sources/Plasma/FeatureLib/pfPython/pyGUIControlMultiLineEdit.cpp
```diff
@@ -239,9 +242,12 @@ void pyGUIControlMultiLineEdit::SetEncodedBufferW( PyObject* buffer_object )
...
             // something to do here... later
-            uint16_t* daBuffer = nil;
-            Py_ssize_t length;
-            PyObject_AsReadBuffer( buffer_object, (const void**)&daBuffer, &length);
+            Py_buffer view;
+            PyObject_GetBuffer(buffer_object, &view, PyBUF_SIMPLE);
+            uint16_t* daBuffer = (uint16_t*)view.buf;
+            Py_ssize_t length = view.len;
+            PyBuffer_Release(&view);
+
```
Hoikas, line 250: Same as above.

Sources/Plasma/FeatureLib/pfPython/pyVaultImageNode.cpp
```diff
@@ -194,9 +194,12 @@ void pyVaultImageNode::SetImageFromBuf( PyObject * pybuf )
...
-    PyObject_AsReadBuffer(pybuf, (const void **)&buffer, &bytes);
+    Py_buffer view;
+    PyObject_GetBuffer(pybuf, &view, PyBUF_SIMPLE);
+    uint8_t* buffer = (uint8_t*)view.buf;
+    Py_ssize_t bytes = view.len;
+    PyBuffer_Release(&view);
+
     if (buffer) {
         VaultImageNode access(fNode);
         access.SetImageData(buffer, bytes);
```
Hoikas, line 205: I wonder if `buffer` may no longer be valid since the view has been released.

Hoikas (commented):
Ideally, we should remove all these *W bindings, but I think this is for the best. Many of these deprecated APIs will be removed in Python 3.10.

Sources/Plasma/FeatureLib/pfPython/pyVaultImageNode.cpp
zrax: I used the actual source of `PyObject_AsReadBuffer` as a reference here (which does the same thing), so it *should* be safe...

```c
static int
as_read_buffer(PyObject *obj, const void **buffer, Py_ssize_t *buffer_len)
{
    Py_buffer view;

    if (obj == NULL || buffer == NULL || buffer_len == NULL) {
        null_error();
        return -1;
    }
    if (PyObject_GetBuffer(obj, &view, PyBUF_SIMPLE) != 0)
        return -1;

    *buffer = view.buf;
    *buffer_len = view.len;
    PyBuffer_Release(&view);
    return 0;
}

int PyObject_AsReadBuffer(PyObject *obj,
                          const void **buffer,
                          Py_ssize_t *buffer_len)
{
    return as_read_buffer(obj, buffer, buffer_len);
}
```

NOTE: There was no error checking before, so I didn't fix that part :/

Sources/Plasma/Apps/plPythonPack/PythonInterface.cpp
zrax: Ah, I was wondering how this didn't crash, since Python expects `program_name` to be dynamically allocated...  Not calling `PyConfig_Clear` would certainly explain that :)

Sources/Plasma/FeatureLib/pfPython/pyVaultImageNode.cpp
Hoikas: That's only mildly terrifying. Oh well.

Hoikas:
Let's go ahead and merge this and deal with leaks later.

PR #733, opened 2020-12-04
Kill Usage of DirectX SDK (June 2010)

Hoikas:
This kills off usage of the DirectX SDK from June 2010 in favor of the DirectX support built into the Windows 8 SDK, which has come with Visual Studio since VS 2013. This change will allow us to avoid requiring yet another dependency and allow building on GitHub Actions, where the DXSDK is not preinstalled (unlike AppVeyor). This change means that the following components are no longer available:
- DxErr
- DirectDraw (obsolete)
- DirectSound (obsolete)
- DirectInput (obsolete)
- Direct3D9X

The DxErr deprecation was handled by converting to the standard Win32 `HRESULT` handling mechanisms. The DirectX `HRESULT`s are now in the Windows SDK itself, so no special handler is needed.

The Direct3D9X deprecation is a bit more problematic ~~and why this pull request is still a WIP draf~~t. The nice matrix class was fairly easy to replace with the base matrix and with `XMMATRIX`(in one particular place). The problem is with `plDXVertexShader`, `plDXPixelShader`, and with plShaderAssembler. The Plasma shaders are all written in assembly and built with `D3DXAssembleShaderFromFile`, which is now gone. The closest analogue is `D3DCompileFromFile`, which, per the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/d3dcompiler/nf-d3dcompiler-d3dcompilefromfile) expects HLSL. I have yet to test compiling Plasma's shaders with this replacement method, but I suspect we will have to either rewrite the shaders in HLSL or find some other method forward. ~~For now, this problem is blocking the PR.~~

---

After a bit of reverse engineering, I've arrived at a solution to finish killing off the dependency on D3DX9. When assembling shaders, D3DX9 effectively forwards the entire shader to a function in D3Dcompiler_[0-9][0-9].dll named `D3DAssemble`. This function is not publicly documented, but it is exported from the June 2010 DX SDK compiler through modern Windows SDKs. I refactored plShaderAssembler.exe to load the pointer to this method and assemble the shaders with this function. The resulting assembled shader bytecode appears identical to what is currently checked in.

I also removed support for assembling shaders at runtime from `plDXVertexShader` and `plDXPixelShader`. I'm sure this was great when Cyan was developing the engine, but we've never changed the shaders. I think if we do, we'll want to operate at a level higher than assembly (eg HLSL) and therefore would need to rewrite the code anyway.

Now that the final dependency on D3DX9 has been removed, I turned the DirectX SDK finder into a dumb list of libraries: d3d9 and D3Dcompiler. I'm not sure how clean that is--commentary welcome. The code now compiles without referencing the standalone DirectX SDK at all and appears to work correctly. As a bonus, I fixed an `ST::bad_format` error in plShaderAssembler.


dpogue (approved):

Hoikas:
Manually merged master to fix conflicts.

Hoikas:
Manually merged master to fix conflicts. (again)

Hoikas:
Note for posterity: we'll clean-up the stubbed FindDirectX cmake module in an upcoming cmake pass.

PR #738, opened 2020-12-12
Run Linux CI on both 32- and 64-bit

dpogue:


Hoikas (changes requested):
Python architecture needs to be specified.

PR #739, opened 2020-12-12
Fix some x64 warnings

dpogue:


Sources/Plasma/CoreLib/hsStringTokenizer.cpp
```diff
@@ -169,12 +166,9 @@ void hsStringTokenizer::ParseQuotes(bool qAsTok)
...
 // String Tokenizer routines
-hsWStringTokenizer::hsWStringTokenizer(const wchar_t *string, const wchar_t *seps) :
-    fQAsTok(true),
-    fInQuote(false),
-    fString(nil),
-    fSeps(nil),
-    fLastTerminator(nil)
+hsWStringTokenizer::hsWStringTokenizer(const wchar_t *string, const wchar_t *seps)
+    : fSeps(nullptr), fLastTerminator(nullptr), fQAsTok(true), fInQuote(false),
+      fString(nullptr)
```
Hoikas, line 171:
```suggestion
      fString()
```

Sources/Plasma/PubUtilLib/plCompression/plZlibStream.cpp
```diff
@@ -52,7 +52,7 @@ void ZlibFree(voidpf opaque, voidpf address)
     free(address);
 }
 
-plZlibStream::plZlibStream() : fOutput(nil), fZStream(nil), fHeader(kNeedMoreData), fDecompressedOk(false)
+plZlibStream::plZlibStream() : fOutput(nil), fZStream(nil), fDecompressedOk(false), fHeader(kNeedMoreData)
```
Hoikas, line 55:
```suggestion
plZlibStream::plZlibStream() : fOutput(), fZStream(), fDecompressedOk(), fHeader(kNeedMoreData)
```

Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -193,19 +193,17 @@ bool plEncryptedStream::Close()
     return rtn;
 }
 
-uint32_t plEncryptedStream::IRead(uint32_t bytes, void* buffer)
+size_t plEncryptedStream::IRead(size_t bytes, void* buffer)
 {
     if (!fRef)
         return 0;
-    int numItems = (int)(::fread(buffer, 1 /*size*/, bytes /*count*/, fRef));
+    size_t numItems = ::fread(buffer, 1 /*size*/, bytes /*count*/, fRef);
```
Hoikas, line 200:
```suggestion
    size_t numItems = fread(buffer, 1, bytes, fRef);
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -84,22 +84,22 @@ enum {
...
-    virtual uint32_t  Write(uint32_t byteCount, const void* buffer) = 0;
-    virtual void      Skip(uint32_t deltaByteCount) = 0;
+    virtual size_t    Read(size_t byteCount, void * buffer) = 0;
+    virtual size_t    Write(size_t byteCount, const void* buffer) = 0;
+    virtual void      Skip(size_t deltaByteCount) = 0;
     virtual void      Rewind() = 0;
     virtual void      FastFwd();
-    virtual uint32_t  GetPosition() const;
-    virtual void      SetPosition(uint32_t position);
+    virtual size_t    GetPosition() const;
```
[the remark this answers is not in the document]
dpogue: Yeah, it's going to be a bit of a mess, but it also shouldn't impact 32-bit builds (since `size_t` is a `uint32_t` on 32-bit compilers).

My hope is that we can catch most of the cases where this is a problem for 64-bit through compiler warnings about integer downsizing.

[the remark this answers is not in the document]
Hoikas: I'm talking specifically about cases where we won't receive warnings due to templated code. Eg something like:
```C++
template<typename T>
void WriteThingy(hsStream* S, const T& value)
{
    S->Write(sizeof(T), reinterpret_cast<const void*>(&value));
}
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -703,26 +708,26 @@ void hsUNIXStream::Rewind()
...
         return 0;
 
-    long oldPos = ftell( fRef );
-    (void)::fseek( fRef, 0, SEEK_END );
-    uint32_t end = (uint32_t)ftell( fRef );
-    (void)::fseek( fRef, oldPos, SEEK_SET );
+    size_t oldPos = ftell(fRef);
+    fseek(fRef, 0, SEEK_END);
+    size_t end = ftell(fRef);
+    fseek(fRef, oldPos, SEEK_SET);
```
zrax, line 730: `ftell` and `fseek` actually use `long` values...  This can be cast in the return statement.

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -536,6 +536,7 @@ plFileName plFileSystem::GetCurrentAppPath()
         return appPath;
 
     FATAL("Your OS doesn't make life easy, does it?");
+    return appPath;
```
zrax, line 539: Better to return `"."`, I think...

Sources/Plasma/CoreLib/hsStream.cpp
dpogue: So more generally, how do we want to handle position and EOF?

On *nix 32-bit, `long` and `size_t` are the same size (4-bytes)
On Windows 32-bit, `long` and `size_t` are the same size (4-bytes)
On *nix 64-bit, `long` and `size_t` are the same size (8-bytes)
On Windows 64-bit, `long` and `size_t` are **not** the same size (4 bytes and 8 bytes)

If we type `fPosition` as `long`, then we maybe avoid some of the concerns Hoikas identified on Windows 64-bit, but those concerns would apply on Linux 64-bit.

If we type `fPosition` as `size_t` (which this PR currently does), then we open the can of worms of potentially serializing 8-byte positions instead of 4-byte positions, as well as needing to use `_fseeki64`/`_ftelli64` on Windows.

zrax: My comment was more about the signedness than the size (`size_t` is unsigned), but the size is a valid concern too...  Although in this particular function, it only matters once we start dealing with files >2GiB.

Sources/Plasma/CoreLib/hsStream.h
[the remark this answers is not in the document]
Hoikas: @dpogue I just looked through all the references to these functions, and I don't see anything that would obviously break serialized files.  Have fun fixing all the type warnings this change will create 😛

Sources/Plasma/CoreLib/hsStream.cpp
Hoikas: I'd prefer for us to avoid using types that are different sizes on different platforms.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -109,33 +115,32 @@ void hsStream::CopyToMem(void* mem)
 
 //////////////////////////////////////////////////////////////////////////////////
 
-uint32_t hsStream::WriteSafeStringLong(const ST::string &string)
+size_t hsStream::WriteSafeStringLong(const ST::string &string)
 {
-    uint32_t len = string.size();
-    WriteLE32(len);
+    size_t len = string.size();
+    WriteLE32(uint32_t(len));
```
zrax, line 121: This may be overly pedantic, but...

Although it's extremely unlikely we'd ever have a string >2**32 characters long, this change would break it.  If we can only write a 32-bit value, we should ensure the written string length is limited to the same range, either with a 32-bit type (the previous `uint32_t`) or with an explicit range check (as is done below in `WriteSafeString`).

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -109,33 +115,32 @@ void hsStream::CopyToMem(void* mem)
...
 }
 
-uint32_t hsStream::WriteSafeWStringLong(const ST::string &string)
+size_t hsStream::WriteSafeWStringLong(const ST::string &string)
 {
     ST::utf16_buffer wbuff = string.to_utf16();
-    uint32_t len = wbuff.size();
-    WriteLE32(len);
+    size_t len = wbuff.size();
+    WriteLE32(uint32_t(len));
```
zrax, line 139: Same as above

Sources/Plasma/CoreLib/hsStream.cpp
zrax: There's no perfect way to handle this, but if we want to fully support large files (even on 32-bit systems), you'll need to extend the workarounds a bit further.  For example:

```c++
#ifdef _MSC_VER
    int64_t oldPos = _ftelli64(fRef);
    _fseeki64(fRef, 0, SEEK_END);
    int64_t end = _ftelli64(fRef);
    _fseeki64(fRef, oldPos, SEEK_SET);
#else
    // With -D_FILE_OFFSET_BITS=64
    static_assert(sizeof(off_t) >= sizeof(int64_t));
    off_t oldPos = ftello(fRef);
    fseeko(fRef, 0, SEEK_END);
    off_t end = ftello(fRef);
    fseeko(fRef, oldPos, SEEK_SET);
#endif
    return static_cast<int64_t>(end);
```

PR #740, opened 2020-12-13
Fix issues with resource.dat generation on Windows.

Hoikas:
This fixes our cairosvg dependency such that `pip install -r requirements.txt` works without undocumented dependencies. Further, some invalid assumptions have been corrected in makeres.py. Namely that: the python executable is named exactly `python` and that `pngcrush` is available when building an external client. This should also fix an issue where CMake would error if `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is a generator expression.

### DIFF
--- a/Sources/Plasma/Apps/plClient/external/makeres.py
+++ b/Sources/Plasma/Apps/plClient/external/makeres.py
@@ -53,7 +53,8 @@ if __name__ == '__main__':
 	parser.add_option("-q", "--quiet", dest="verbose", default=True, action="store_false", help="Don't print status messages")
 	parser.add_option("-r", "--render", dest="render", default=False, action="store_true", help="Perform SVG Render to images")
 	parser.add_option("-p", "--package", dest="package", default=False, action="store_true", help="Perform packaging into resource container")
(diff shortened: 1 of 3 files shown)


dpogue (approved):
Looks good to me

Sources/Plasma/Apps/plClient/external/makeres.py
```diff
@@ -72,16 +73,18 @@
 
 	## Do the work!
 	if options.render:
-		ret = subprocess.call(["python", os.path.join(inpath, "render_svg.py"), "-i", inpath, "-o", os.path.join(workpath, "render")], stdout=sys.stdout, stderr=sys.stderr)
+		ret = subprocess.call([sys.executable, os.path.join(inpath, "render_svg.py"), "-i", inpath, "-o", os.path.join(workpath, "render")], stdout=sys.stdout, stderr=sys.stderr)
```
Deledrius, line 76: We should probably use `sys.executable` for the subprocess call to `create_resource_dat.py` as well.

Deledrius (changes requested):
Works here for me.  Just one minor edit seems necessary.

Deledrius (approved):

PR #741, opened 2020-12-13
GoogleTest update

dpogue:
Updates from gtest 1.8 to gtest 1.10.

The additional options in the CMakeLists.txt were taken [from string_theory](https://github.com/zrax/string_theory/blob/master/test/CMakeLists.txt#L21-L35).


Hoikas:
Is there any way we could *not* vendor gtest?

dpogue:
git submodule? Maybe something through vcpkg?

They sort of encourage vendoring it: https://github.com/dpogue/Plasma/tree/6f69d0ec7d79ebffdfa9ee8346eca3f48d2b2ae8/Sources/Tests/gtest-1.10.0#incorporating-into-an-existing-cmake-project

Hoikas:
A submodule would work nicely - though vcpkg *does* have a gtest port that's at version 1.10. I just cringe at all of this non-Plasma code churn in our history :/

zrax:
:+1: for submodule; I think that would work well for this particular use case.

dpogue:
Okay, I'll close this and open a new PR to include gtest 1.10 as a git submodule

PR #742, opened 2020-12-13
pfPasswordStore test

dpogue:
I had a test app lying around from back when we added pfPasswordStore, but it turned out to be easy to slightly restructure it as a unit test.

My only concern here is whether this will cause prompts on Windows or macOS when it tries to access the keychain, which would cause the test to hang/fail.

There are (unfortunately) lots of hoops to jump through to get the keychain to work on Linux CI 😞

### DIFF
--- a/Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
+++ b/Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
@@ -1,0 +1,22 @@
+set(pfPasswordStoreTest_SOURCES
+    test_pfPasswordStore.cpp
+)
+
+add_executable(test_pfPasswordStore ${pfPasswordStoreTest_SOURCES})
+target_link_libraries(
+    test_pfPasswordStore
+    PRIVATE
+        gtest
+        gtest_main
+        pnUUID
+        pfPasswordStore
+)
+
+if(TARGET PkgConfig::LIBSECRET AND NOT TARGET Security::Security)
+    target_compile_definitions(test_pfPasswordStore PRIVATE HAVE_LIBSECRET)
+elseif(TARGET Security::Security)
+    target_compile_definitions(test_pfPasswordStore PRIVATE HAVE_SECURITY)
+endif()
+
+add_test(NAME test_pfPasswordStore COMMAND test_pfPasswordStore)
+add_dependencies(check test_pfPasswordStore)
--- a/Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
+++ b/Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
@@ -1,0 +1,130 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
+PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+licensors of this Program grant you additional
+permission to convey the resulting work. Corresponding Source for a
+non-source form of such a combination shall include the source code for
+the parts of OpenSSL and IJG JPEG Library used as well as that of the covered
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include <gtest/gtest.h>
+#include <string_theory/string>
+#include "plFileSystem.h"
+#include "pnUUID/pnUUID.h"
+#include "pfPasswordStore/pfPasswordStore.h"
+#include "pfPasswordStore/pfPasswordStore_impl.h"
+
+TEST(pfPasswordStore, service)
+{
+    pfPasswordStore* store = pfPasswordStore::Instance();
+    EXPECT_NE(store, nullptr);
+
+#if defined(HS_BUILD_FOR_WIN32)
+    EXPECT_EQ(typeid(*store), typeid(pfWin32PasswordStore));
+#elif defined(HAVE_SECURITY)
+    EXPECT_EQ(typeid(*store), typeid(pfApplePasswordStore));
+#elif defined(HAVE_LIBSECRET)
+    EXPECT_EQ(typeid(*store), typeid(pfUnixPasswordStore));
+#else
+    EXPECT_EQ(typeid(*store), typeid(pfFilePasswordStore));
+#endif
+}
+
+#if defined(HS_BUILD_FOR_WIN32)
+TEST(pfWin32PasswordStore, storing_and_retrieval)
+{
+    pfWin32PasswordStore store;
+    ST::string username = "test_pfPasswordStore";
+    ST::string password = plUUID::Generate().AsString();
+
+    bool success = store.SetPassword(username, password);
+    EXPECT_EQ(success, true);
+
+    ST::string pass = store.GetPassword(username);
+    EXPECT_EQ(pass.compare(password), 0);
+}
+#endif
+
+#if defined(HAVE_SECURITY)
+TEST(pfApplePasswordStore, storing_and_retrieval)
+{
+    pfApplePasswordStore store;
+    ST::string username = "test_pfPasswordStore";
+    ST::string password = plUUID::Generate().AsString();
+
+    bool success = store.SetPassword(username, password);
+    EXPECT_EQ(success, true);
+
+    ST::string pass = store.GetPassword(username);
+    EXPECT_EQ(pass.compare(password), 0);
+}
+#endif
+
+#if defined(HAVE_LIBSECRET)
+TEST(pfUnixPasswordStore, storing_and_retrieval)
+{
+    pfUnixPasswordStore store;
+    ST::string username = "test_pfPasswordStore";
+    ST::string password = plUUID::Generate().AsString();
+
+    bool success = store.SetPassword(username, password);
+    EXPECT_EQ(success, true);
+
+    ST::string pass = store.GetPassword(username);
+    EXPECT_EQ(pass.compare(password), 0);
+}
+#endif
+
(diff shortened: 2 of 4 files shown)


Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include_directories(${GTEST_INCLUDE_DIR})
+include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
+include_directories(../../../Plasma/CoreLib)
+include_directories(../../../Plasma/NucleusLib)
+include_directories(../../../Plasma/FeatureLib)
+
+SET(pfPasswordStoreTest_SOURCES
+    test_pfPasswordStore.cpp
+    )
```
Hoikas, line 9:
```suggestion
set(pfPasswordStoreTest_SOURCES
    test_pfPasswordStore.cpp
)
```

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
```diff
@@ -0,0 +1,81 @@
...
+*==LICENSE==*/
+
+#include <gtest/gtest.h>
+#include <string_theory/string>
+#include "plFileSystem.h"
+#include "pnUUID/pnUUID.h"
+#include "pfPasswordStore/pfPasswordStore.h"
+#include "pfPasswordStore/pfPasswordStore_impl.h"
+
+#if defined(HS_BUILD_FOR_WIN32) || defined(HS_BUILD_FOR_APPLE) || defined(HAVE_LIBSECRET)
```
Hoikas, line 50: IMO this should be a getter in `pfPasswordStore`, or we should have explicit platform tests as in `fallback_storing_and_retrieval`.

Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+include_directories(${GTEST_INCLUDE_DIR})
+include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
+include_directories(../../../Plasma/CoreLib)
+include_directories(../../../Plasma/NucleusLib)
+include_directories(../../../Plasma/FeatureLib)
```
zrax, line 5: I know we already have a lot of this pattern, but we should really be trying to clean it up (e.g. using `${CMAKE_SOURCE_DIR}` or `${<PROJECT-NAME>_SOURCE_DIR}` instead of relative paths)...

Hoikas: Don't forget [target_include_directories](https://cmake.org/cmake/help/latest/command/target_include_directories.html). Ideally, we'd move to a fully target-based system, but that's beyond the scope of this.

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
dpogue: I've split this up into separate test cases for each type of password store, as well as one ensuring that we get a valid store from the general API.

Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
dpogue: I went with `${CMAKE_SOURCE_DIR}` for now. I agree that `target_include_directories` would be great, but we don't _really_ have per-target include directories: they all use the same 4 top-level ones.

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
```diff
@@ -0,0 +1,127 @@
...
+    EXPECT_EQ(pass.compare(password), 0);
+}
+#endif
+
+TEST(pfFilePasswordStore, fallback_storing_and_retrieval)
+{
+    // Ensure we don't overwrite the real login.dat file
+    plFileSystem::CreateDir("init");
+    FILE* f = plFileSystem::Open(plFileName::Join("init", "login.dat"), "wb");
+    fclose(f);
```
Hoikas, line 116: I'm not really sure I understand what's going on here, even with the comment.

Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+include_directories(${GTEST_INCLUDE_DIR})
+include_directories(${gtest_SOURCE_DIR}/include ${gtest_SOURCE_DIR})
+include_directories(${CMAKE_SOURCE_DIR}/Sources/Plasma/CoreLib)
+include_directories(${CMAKE_SOURCE_DIR}/Sources/Plasma/NucleusLib)
+include_directories(${CMAKE_SOURCE_DIR}/Sources/Plasma/FeatureLib)
+
+if(HAVE_LIBSECRET)
+    add_definitions(-DHAVE_LIBSECRET)
+endif(HAVE_LIBSECRET)
```
Hoikas, line 9: With our, admittedly unstated, goal of modernizing our CMake, we should probably be using [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) for this.

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
dpogue: See https://github.com/H-uru/Plasma/blob/913276f4cd5bce8c5dbc78bed2d1f6ec6804d960/Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore.cpp#L117-L124

In External Release mode, it will read/write to the init/login.dat file in the user app data folder. In internal builds, it will use a local init/login.dat file if it exists.

For this test, I thought it would be best to ensure that local file exists, so there's a lower chance of people overwriting their login creds in user app data if they run the tests. It probably makes no difference for CI.

Hoikas: I think the comment should be expanded to include that information... and maybe this test disabled on external releases? I'm unsure about that last bit. Or maybe

Hoikas: I think the comment should be expanded to include that information. Or, perhaps the file password store should be modified such that we can override the hard coded login.dat file name in the case of tests.

.github/workflows/linux-ci.yml
```diff
@@ -52,4 +49,10 @@ jobs:
           -DPython_ADDITIONAL_VERSIONS=2.7 -DPLASMA_BUILD_TESTS=ON \
           -DPLASMA_BUILD_TOOLS=ON -DPLASMA_BUILD_RESOURCE_DAT=ON ..
         make -j2
+
+    - name: Test Plasma
+      run: |
+        cd ${GITHUB_WORKSPACE}/build
         make check
+      env:
+        CTEST_OUTPUT_ON_FAILURE: 1
```
Hoikas, line 58: I wonder if this could be done as an argument to the ctest command?

dpogue: Apparently yes, as `--output-on-failure`.
It can also be set as a default argument to ctest in the CMakeLists:
```
list(APPEND CMAKE_CTEST_ARGUMENTS "--output-on-failure")
```

Do we ever *not* want to see the output of the failing test on failure? Should this be turned on by default in our main CMakeLists?

Hoikas: I can't think of any reason why *not* seeing the test output would be desirable.

.github/workflows/linux-ci.yml
```diff
@@ -49,7 +46,17 @@ jobs:
...
         make -j2
+
+    - name: Test Plasma
+      run: |
+        sudo apt-get install xvfb gnome-keyring dbus-x11 libsecret-tools
+        export "$(dbus-launch)"
+        echo "" | gnome-keyring-daemon --unlock
+        gnome-keyring-daemon --start --daemonize --components=secrets
+        export "$(echo "" | gnome-keyring-daemon -r -d --unlock)"
+        echo -n "secret-tool-password-1" | secret-tool store --label=secret-tool-label-1 secret-tool-key-1 secret-tool-value-1
```
dpogue, line 59: I tried having all this junk as its own "Setup Test Environment" step, but it doesn't work because the exported env vars don't carry to other steps

.github/workflows/linux-ci.yml
```diff
@@ -49,7 +46,17 @@ jobs:
...
         cmake -DCMAKE_BUILD_TYPE=Debug \
           -DCMAKE_PREFIX_PATH="${GITHUB_WORKSPACE}/build_deps/prefix;${GITHUB_WORKSPACE}/build_deps/PhysX/physx/install" \
-          -DPython_ADDITIONAL_VERSIONS=2.7 -DPLASMA_BUILD_TESTS=ON \
-          -DPLASMA_BUILD_TOOLS=ON -DPLASMA_BUILD_RESOURCE_DAT=ON ..
+          -DPLASMA_BUILD_TESTS=ON -DPLASMA_BUILD_TOOLS=ON -DPLASMA_BUILD_RESOURCE_DAT=ON ..
         make -j2
+
+    - name: Test Plasma
+      run: |
+        sudo apt-get install xvfb gnome-keyring dbus-x11 libsecret-tools
```
Hoikas, line 54: Should probably happen in the dependencies step.

Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+
+target_link_libraries(test_pfPasswordStore gtest gtest_main)
+target_link_libraries(test_pfPasswordStore CoreLib)
+target_link_libraries(test_pfPasswordStore pnUUID)
+target_link_libraries(test_pfPasswordStore pfPasswordStore)
+target_link_libraries(test_pfPasswordStore ${STRING_THEORY_LIBRARIES})
+
+add_test(NAME test_pfPasswordStore COMMAND test_pfPasswordStore)
+add_dependencies(check test_pfPasswordStore)
+
```
zrax, line 21: This CMakeLists should be updated to use the new format we're starting to use...

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
```diff
@@ -0,0 +1,130 @@
...
+    pfWin32PasswordStore store;
+    ST::string username = "test_pfPasswordStore";
+    ST::string password = plUUID::Generate().AsString();
+
+    bool success = store.SetPassword(username, password);
+    EXPECT_EQ(success, true);
+
+    ST::string pass = store.GetPassword(username);
+    EXPECT_EQ(pass.compare(password), 0);
+}
```
zrax, line 78: Maybe I'm missing something, but why do we have three identical tests here?

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
```diff
@@ -0,0 +1,130 @@
...
+#endif
+
+TEST(pfFilePasswordStore, storing_and_retrieval)
+{
+    // In internal builds, if a local init/login.dat file exists, it will be
+    // used instead of the one in the User AppData folder. For testing, we'll
+    // ensure that file exists to avoid overwriting real login.dat credentials
+    // in the User AppData folder.
```
zrax, line 116: This smells like a hack...  If we really want to enable testing, we should allow a test override, instead of relying on implementation details that may change in the future.

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
dpogue: One for each password store implementation (Windows Credential Vault, macOS Keychain, Linux libsecret). I originally had a single test that tested whatever the native password store happened to be, but was asked to split them up as individual tests.

Hoikas: The body of the tests are all the same. Perhaps the `ifdef` could be moved closer to the point of difference?

Hoikas:

PR #744, opened 2020-12-15
Further work on `hsRef`

zrax:
This adds an `hsWeakRef` class which makes (lack of) ownership more clear in several APIs.  Furthermore, construction of an `hsRef` from a raw pointer is now disallowed.  You must either convert it to an `hsWeakRef` (which gets reffed when converted to an `hsRef`, matching the old conversion behavior), or steal it with the APIs added in #625.

In the process, this should plug a couple more leaks, but should be tested to ensure I didn't accidentally plug something that wasn't a leak.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -393,10 +393,10 @@ void DebugMsg(const char* fmt, ...);
     
     void    hsDebugMessage(const char* message, long refcon);
     #define hsDebugCode(code)                   code
-    #define hsIfDebugMessage(expr, msg, ref)    (void)( ((expr) != 0) || (hsDebugMessage(msg, ref), 0) )
-    #define hsAssert(expr, ...)                 (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
-    #define ASSERT(expr)                        (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, #expr), 0) )
-    #define ASSERTMSG(expr, ...)                (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
+    #define hsIfDebugMessage(expr, msg, ref)    (void)( (!!(expr)) || (hsDebugMessage(msg, ref), 0) )
+    #define hsAssert(expr, ...)                 (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
+    #define ASSERT(expr)                        (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, #expr), 0) )
+    #define ASSERTMSG(expr, ...)                (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
     #define FATAL(...)                          ErrorAssert(__LINE__, __FILE__, __VA_ARGS__)
     #define DEBUG_MSG                           DebugMsg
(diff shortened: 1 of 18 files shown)


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -393,10 +393,10 @@ void DebugMsg(const char* fmt, ...);
...
     void    hsDebugMessage(const char* message, long refcon);
     #define hsDebugCode(code)                   code
-    #define hsIfDebugMessage(expr, msg, ref)    (void)( ((expr) != 0) || (hsDebugMessage(msg, ref), 0) )
-    #define hsAssert(expr, ...)                 (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
-    #define ASSERT(expr)                        (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, #expr), 0) )
-    #define ASSERTMSG(expr, ...)                (void)( ((expr) != 0) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
+    #define hsIfDebugMessage(expr, msg, ref)    (void)( (!!(expr)) || (hsDebugMessage(msg, ref), 0) )
+    #define hsAssert(expr, ...)                 (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
+    #define ASSERT(expr)                        (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, #expr), 0) )
+    #define ASSERTMSG(expr, ...)                (void)( (!!(expr)) || (ErrorAssert(__LINE__, __FILE__, __VA_ARGS__), 0) )
```
Hoikas, line 399: !!! (this is me being silly because I didn't realize `!!` is valid syntax)

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -86,14 +86,14 @@ struct RelVaultNodeLink : THashKeyVal<unsigned> {
...
-        : THashKeyVal<unsigned>(nodeId), node(node), ownerId(ownerId), seen(seen)
+    RelVaultNodeLink (bool seen, unsigned ownerId, unsigned nodeId, hsRef<RelVaultNode> node)
+        : THashKeyVal<unsigned>(nodeId), node(std::move(node)), ownerId(ownerId), seen(seen)
     { }
 };
 
 
 struct IRelVaultNode {
-    RelVaultNode * node;        // MUST be a weak ref!
+    hsWeakRef<RelVaultNode> node;
```
Hoikas, line 96: :trollface:

Hoikas (commented):
This looks good - there are at least two or three (I lost count) leaks fixed in this PR. I'd like to run this with VLD on before merging though.

Hoikas (commented):
There appear to be no crashes or obvious runtime errors. I only see a few vault leaks... These may be old ones, but I'll let you be the judge of them: https://gist.github.com/Hoikas/1da14efc1ccbbed1f45c6ee0f346c122

The leak report is indicating lots of leaks from Python objects, so I didn't upload the whole thing. They all seem unrelated to these changes.

zrax:
All of those leaks seem to have a common pattern: They're all leaks from the `new RelVaultNodeLink` in `plVaultClientApi.cpp:360`.  It seems that the `s_nodes` cache isn't getting properly cleaned up, but that shouldn't be caused by my changes as far as I can tell.

Hoikas:
Yes, I ran VLD against master and saw the same pattern there. I'm :+1: on deferring fixing the cache until later and merging this as-is.

Hoikas (approved):

PR #745, opened 2020-12-15
INSTALL target

Hoikas:
This adds an INSTALL target that is supposed to install all of the files needed for the client and tools into an organized structure. It doesn't match the GNU standard of bin, lib, etc. because we are a game, not a system library or command line executable.

~~This is still considered a WIP because not all DLLs are copied into the installed tree on Windows. We currently run `windeployqt` on all GUI tools, and I would like to do something similar for the non-Qt DLLs. Vcpkg supports `VCPKG_APPLOCAL_DEPS`, which copies the DLLs into the binary directory. The installed variant of this is still experimental though and seems to error on my bare-bones `install` cmake commands. Do we want to try to work with vcpkg's experimental feature or roll that functionality ourselves? I realize, of course, that doing such on unix-like platforms is pointless... the goal I'm working toward here is to create good distributions for Windows.~~

The latest vcpkg master has a correctly functioning `X_VCPKG_APPLOCAL_DEPS_INSTALL`, so any DLLs needed by vcpkg are correctly copied into the install prefix. Further, our cmake now runs windeployqt on all GUI tools... because not everyone wants to build Qt5 from source using vcpkg .

This obsoletes the clunky "ScriptsSystem" target - now just `make install` or `cmake --build . --target INSTALL` to get that done.

Depends on #740.


CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ include(FeatureSummary)
 if(POLICY CMP0054)
     cmake_policy(SET CMP0054 NEW)
 endif()
```
zrax, line 12: We could remove CMP0054 (introduced in CMake 3.1) since we require CMake 3.12 now...

Scripts/CMakeLists.txt
```diff
@@ -6,60 +6,65 @@ source_group("Python" FILES ${GameScripts})
...
-                  COMMENT "Copying engine scripts..."
-
-                  # Copy core engine files
-                  COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/dat" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/dat"
-                  COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/Python" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/python"
-                  COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/SDL" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/SDL"
-
+add_custom_target(Scripts
                   SOURCES ${GameScripts} ${GameSDL} ${GameData}
 )
```
zrax, line 11: Is this target actually still needed?

cmake/DeployQt5.cmake
```diff
@@ -0,0 +1,49 @@
...
+    # Save some ruddy characters and install the target
+    install(TARGETS ${_pdqt_TARGET} DESTINATION ${_pdqt_DESTINATION})
+
+    if(TARGET Qt5::windeployqt)
+        # Copy Qt DLLs to the build directory so the stupid application can be run in Visual Studio
+        add_custom_command(
+            TARGET ${_pdqt_TARGET} POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E remove_directory "$<TARGET_FILE_DIR:${_pdqt_TARGET}>/qt"
+            COMMAND set PATH=%PATH%$<SEMICOLON>${qt5_install_prefix}/bin
+            COMMAND Qt5::windeployqt --dir "$<TARGET_FILE_DIR:${_pdqt_TARGET}>/qt" "$<TARGET_FILE:${_pdqt_TARGET}>"
```
zrax, line 40: We don't need ANGLE or OpenGL for our Qt tools, so you can add `--no-angle --no-opengl-sw` to the windeployqt command line.  Probably `--no-translations` too...

cmake/DeployQt5.cmake
```diff
@@ -0,0 +1,49 @@
...
+    endif()
+
+    # Save some ruddy characters and install the target
+    install(TARGETS ${_pdqt_TARGET} DESTINATION ${_pdqt_DESTINATION})
+
+    if(TARGET Qt5::windeployqt)
+        # Copy Qt DLLs to the build directory so the stupid application can be run in Visual Studio
+        add_custom_command(
+            TARGET ${_pdqt_TARGET} POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E remove_directory "$<TARGET_FILE_DIR:${_pdqt_TARGET}>/qt"
```
zrax, line 38: If this gets run more than once (which it does), each target will re-populate the deployment directory.  Our Qt apps all use the same Qt libs, so the end result is probably still fine, but this still does more work than necessary.

Perhaps instead we can make a list of executables that need to be considered for windeployqt, and then a single target that runs it on all of them at the same time (windeployqt can take multiple executables on the command line)...

zrax (changes requested):

Scripts/CMakeLists.txt
Hoikas: No, but I think @Deledrius liked being able to see the game scripts in VS.

cmake/DeployQt5.cmake
Hoikas: I was initially considering doing that by doing cmake's undocumented function overloading... But vcpkg already does that for `add_executable()`, which means we'd introduce an inifite loop. Maybe we could have a `plasma_add_qt_executable` that handles the book-keeping for us? Appending to a list in the CMakeLists.txt or having a hardcoded list seems kind of dirty to me.

CMakeLists.txt
Hoikas: "This policy was introduced in CMake version 3.1. CMake version 3.19.2 warns when the policy is not set and uses OLD behavior. Use the cmake_policy() command to set it to OLD or NEW explicitly." [(source)](https://cmake.org/cmake/help/latest/policy/CMP0054.html)

Hoikas: Nevermind, I see that the minimum_required thing overrides that,

Scripts/CMakeLists.txt
zrax: Ah, right, I hadn't considered the IDE integration...

cmake/DeployQt5.cmake
zrax: I don't know why you need to overload anything...  You could just switch to the normal `install` command for the Qt executables, and then run the deployment from the Tools directory's CMakeLists.txt, since that file already has knowledge of the full set of Qt executables to build...

Hoikas: I'm not certain how to obtain the list of those executables aside from hardocding a list of them (yuck) or injecting logic to gain it.

zrax: Another option, which still duplicates work but at least doesn't potentially delete needed files, would be to add a target that only cleans the `qt` staging directory once, and then still have each Qt executable run its own deployment...

Hoikas: One thing to note is that I've removed the default specification of `CMAKE_RUNTIME_DIRECTORY`, so, with Visual Studio, each executable is now building in its own directory and has its own qt staging directory. Previously, they were all going into `<builddir>/bin/<config>` So, if you're debugging one of these directly from the IDE, we might not want to clear its staging directory--you'd potentially be blasting the dependencies away before a debug session. I'm not sure what the behavior is with single config generators eg makefiles.

Scripts/CMakeLists.txt
Deledrius: If you feel it's extraneous, I won't argue the point.  The gamescripts straddle the line between source and content in annoying ways.  I just felt that if we have text source in the tree, and we're doing things with them, it would be nice to have them be accessible and organized in the editor, if nothing else but for transparency.

README.md
```diff
@@ -93,18 +94,17 @@ fully-patched installation of MOULa provided by Cyan Worlds.
...
    your *MOUL-OS* folder, and rename it as **server.ini**. If you are running
    your own dirtsand server or are connecting to one run by someone else, use
    the server.ini generated from that.
 4. Copy the **DLLs** from the development libraries bundle, as well as the DLLs
    PhysXLoader.dll, NxExtensions.dll, NxCooking.dll, and NxCharacter.dll from
    your PhysX SDK installation into the *MOUL-OS* folder.
-5. Copy the **resource.dat** file from `<build_dir>\bin` to the *MOUL-OS* folder,
-   or from http://www.guildofwriters.org/tools/resource.dat if you did not build
-   your own.
+5. Download and copy http://www.guildofwriters.org/tools/resource.dat into
```
Deledrius, line 106: Why remove the instructions for using the locally-built copy?

Hoikas: `resource.dat` will only be generated if you manually `pip install -r requirements.txt` or otherwise have the correct packages preinstalled. I figure that's a problem for another PR :smile:

zrax (approved):
Nice, I like it!

Hoikas:
:boom:

Issue #489, opened 2015-07-07
Lots of pipeline stuff includes from pfCamera


dpogue:
Maybe there's a camera base class that could be included in plPipeline (or plCamera) and let pfCamera just handle the brains?

Issue #488, opened 2015-07-07
plAgeLoader requires pfPatcher link


dpogue:
Move pfPatcher to plPatcher? Other suggestions?
/cc @Hoikas

Hoikas:
Sounds good to me.

Issue #487, opened 2015-07-07
Move pfKIMsg to PubUtilLib/plMessage


dpogue:
Lots of things in PubUtilLib that are referencing it.

Hoikas:
PubUtilLib references to pfKIMsg sound like a bug to me

dpogue:
There's a number of comments beside the includes indicating that the
intention was to promote it to PubUtilLib.

Issue #383, opened 2014-01-14
Validate target_link_libraries for all projects


dpogue:
Each project CMakeLists.txt should include `target_link_libraries` for all dependency projects.
In Visual Studio, this doesn't seem to be an issue, but it is helpful in *nix.

This is a checklist of all projects and whether their dependencies have been validate.
- [x] CoreLib
- [ ] NucleusLib/inc
- [x] NucleusLib/pnAsyncCore
- [x] NucleusLib/pnAsyncCoreExe
- [x] NucleusLib/pnDispatch
- [x] NucleusLib/pnEncryption
- [x] NucleusLib/pnFactory
- [x] NucleusLib/pnInputCore
- [x] NucleusLib/pnKeyedObject
- [x] NucleusLib/pnMessage
- [x] NucleusLib/pnModifier
- [x] NucleusLib/pnNetBase
- [x] NucleusLib/pnNetCli
- [x] NucleusLib/pnNetCommon
- [x] NucleusLib/pnNetProtocol
- [x] NucleusLib/pnSceneObject
- [x] NucleusLib/pnTimer
- [x] NucleusLib/pnUtils
- [x] NucleusLib/pnUUID
- [ ] PubUtilLib/plAgeDescription
- [ ] PubUtilLib/plAgeLoader
- [ ] PubUtilLib/plAudible
- [ ] PubUtilLib/plAudio
- [ ] PubUtilLib/plAudioCore
- [ ] PubUtilLib/plAvatar
- [ ] PubUtilLib/plClientResMgr
- [ ] PubUtilLib/plClipboard
- [ ] PubUtilLib/plCompression
- [ ] PubUtilLib/plContainer
- [ ] PubUtilLib/plDrawable
- [ ] PubUtilLib/plFile
- [ ] PubUtilLib/plGImage
- [ ] PubUtilLib/plGLight
- [ ] PubUtilLib/plInputCore
- [ ] PubUtilLib/plInterp
- [ ] PubUtilLib/plIntersect
- [ ] PubUtilLib/plMath
- [ ] PubUtilLib/plMessage
- [ ] PubUtilLib/plModifier
- [ ] PubUtilLib/plNetClient
- [ ] PubUtilLib/plNetClientComm
- [ ] PubUtilLib/plNetClientRecorder
- [ ] PubUtilLib/plNetCommon
- [ ] PubUtilLib/plNetCore
- [ ] PubUtilLib/plNetGameLib
[...]

PR #748, opened 2020-12-16
Clean up relative include paths in CMake

zrax:
... and some other minor cmake cleanup.


Sources/Plasma/FeatureLib/CMakeLists.txt
```diff
@@ -1,4 +1,9 @@
-add_definitions(-D_LIB)
+include_directories("${PLASMA_SOURCE_ROOT}/CoreLib")
+include_directories("${PLASMA_SOURCE_ROOT}/FeatureLib")
+include_directories("${PLASMA_SOURCE_ROOT}/NucleusLib")
+include_directories("${PLASMA_SOURCE_ROOT}/NucleusLib/inc")
+include_directories("${PLASMA_SOURCE_ROOT}/PubUtilLib")
+#include_directories("${PLASMA_SOURCE_ROOT}/PubUtilLib/inc")
```
Hoikas, line 6: Commented out?

Sources/Plasma/NucleusLib/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-add_definitions(-D_LIB)
+include_directories("${PLASMA_SOURCE_ROOT}/CoreLib")
+include_directories("${PLASMA_SOURCE_ROOT}/NucleusLib")
+include_directories("${PLASMA_SOURCE_ROOT}/NucleusLib/inc")
+include_directories("${PLASMA_SOURCE_ROOT}/PubUtilLib")
```
dpogue, line 4: TODO: Clean up anything from NucleusLib that depends on PubUtilLib classes

dpogue: Maybe it would be better to remove this at the top level and only include it in the subprojects that need it, with an eye towards removing it entirely?

Hoikas (commented):
Part of me wonders if we should be moving to a more target-based system with our includes and link libraries. That would require redoing pretty much every `#include` directive in the entire engine and making sure no header file names collided... But it certainly would clean up a lot of this sin. That's probably for another PR though...

Sources/Plasma/PubUtilLib/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-add_definitions(-D_LIB)
+include_directories("${PLASMA_SOURCE_ROOT}/CoreLib")
+include_directories("${PLASMA_SOURCE_ROOT}/FeatureLib")
```
dpogue, line 2: Likewise, PubUtilLib shouldn't depend on FeatureLib, but there is a **lot** of reaching across the boundary here for stuff like pfKIMsg and pfCamera stuff :(

dpogue (approved):

Hoikas:
@dpogue I think my comment about moving toward a proper target-based system would address a lot of your concerns. =P

Deledrius:
This makes things a lot simpler to read!

dpogue:
> @dpogue I think my comment about moving toward a proper target-based system would address a lot of your concerns. =P

:+1:
We should start a discussion around that, xref: https://github.com/H-uru/Plasma/issues/383, https://github.com/H-uru/Plasma/issues/487, https://github.com/H-uru/Plasma/issues/489, https://github.com/H-uru/Plasma/issues/488, etc.

Deledrius:
> 
> 
> > @dpogue I think my comment about moving toward a proper target-based system would address a lot of your concerns. =P
> 
> 👍
> We should start a discussion around that, xref: #383, #487, #489, #488, etc.

Excellent idea.

zrax:
> Part of me wonders if we should be moving to a more target-based system with our includes and link libraries. 

That was the original goal when I started, but it got out of control quickly, so I opted instead for something smaller that still had an immediate benefit...

Sources/Plasma/FeatureLib/CMakeLists.txt
zrax: Copypasta ;)

Hoikas (approved):

PR #753, opened 2020-12-19
Add some target_link_libraries

dpogue:
This is (very clearly) not all of the linking that needs to be done, but this is what I'd been carrying on my plGLClient branch to make it compile and link on Linux.

It seems better for these to exist on master, imperfect they may be, than to have them mixed in with GL pipeline work.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfAnimation/CMakeLists.txt
+++ b/Sources/Plasma/FeatureLib/pfAnimation/CMakeLists.txt
@@ -27,5 +27,12 @@ set(pfAnimation_HEADERS
 
 add_library(pfAnimation STATIC ${pfAnimation_SOURCES} ${pfAnimation_HEADERS})
 
+target_link_libraries(pfAnimation
+    PUBLIC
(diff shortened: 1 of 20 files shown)


Sources/Plasma/FeatureLib/pfAnimation/CMakeLists.txt
```diff
@@ -27,5 +27,9 @@ set(pfAnimation_HEADERS
 
 add_library(pfAnimation STATIC ${pfAnimation_SOURCES} ${pfAnimation_HEADERS})
 
+target_link_libraries(pfAnimation CoreLib)
+target_link_libraries(pfAnimation pnModifier)
+target_link_libraries(pfAnimation plModifier)
```
zrax, line 32: I would prefer we start moving to something like the following:

```cmake
target_link_libraries(pfAnimation
    PUBLIC
        CoreLib
        pnModifier
        plModifier
)
```

zrax (approved):

PR #755, opened 2020-12-19
Clean up some warning pragmas that are no longer needed

zrax:


Hoikas (commented):
I wonder if, at some point, all of these warning disable pragmas should be removed entirely in favor of fixing the warnings themselves.

zrax:
> 
> 
> I wonder if, at some point, all of these warning disable pragmas should be removed entirely in favor of fixing the warnings themselves.

Yep, they definitely should (or at least moved down to the point of use)...  But some of them (e.g. C4018) will still take a lot more work to clean up, as-is.

Hoikas:
Right, I'm just thinking we might as well turn on all the warning spam. Otherwise, it's a case of "out of sight, out of mind."

dpogue:
> Right, I'm just thinking we might as well turn on all the warning spam. Otherwise, it's a case of "out of sight, out of mind."

My approach for the x64 stuff is to turn on all the Clang warnings on Linux, and deal with whatever warnings MSVC prints out during build, and then fix them all. I'm happy to remove pragmas when I continue on that work and try to catch the rest of them too.

Hoikas (approved):

dpogue (approved):

PR #758, opened 2020-12-21
Fix some missing linked libs for plPhysX

dpogue:
Part of me feels like if PhysX requires dynamic library loading, it should be responsible for linking the appropriate lib, but it evidently doesn't work that way, and we need to `-ldl` ourselves.


Sources/Plasma/PubUtilLib/plPhysX/CMakeLists.txt
```diff
@@ -36,7 +36,17 @@ set(plPhysX_HEADERS
...
+target_link_libraries(plPhysX
+    PUBLIC
+        ${PHYSX_LIBRARIES}
+        CoreLib
+        plPhysical
+)
+
+if(UNIX)
+    target_link_libraries(plPhysX PUBLIC ${CMAKE_DL_LIBS})
+endif(UNIX)
```
Hoikas, line 49: If PhysX itself requires ldl, then it should be added to PHYSX_LIBRARIES in the find module.

Hoikas (changes requested):

Hoikas (approved):

PR #759, opened 2020-12-21
Resource.dat fixxors.

Hoikas:
This fixes an issue reported by @Deledrius in that since #745, the resource.dat file is always regenerated at build time. Also, it introduces the ability for our cmake to automatically install the dependency packages for this process. This is off by default because it seems inappropriate to `pip install` anything into someone's system Python. Future work when adding the vcpkg submodule will be to selectively enable the facility.

Also fixes an issue in which CI has not tested resource.dat in quite some time...


Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -7,18 +7,59 @@ if(MOVIE_AVAILABLE)
...
-        set(CAN_BUILD_RESOURCE_DAT ON)
-    endif((DEFINED PY_CAIROSVG) AND (DEFINED PY_PIL))
-endif(Python3_Interpreter_FOUND)
+find_program(
+    PNGCRUSH_EXECUTABLE pngcrush
+    DOC "Path to pngcrush"
+)
+
+# Can we generate the resource.dat file?
+python_install_modules(REQUIREMENTS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/external/requirements.txt")
```
Deledrius, line 30: If we're loading requirements here now, we should probably just remove [requirements.txt](https://github.com/H-uru/Plasma/blob/master/requirements.txt)

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -7,18 +7,59 @@ if(MOVIE_AVAILABLE)
...
+                --render --package
+                ${OPTIMIZE_ARGUMENT}
+                -i "${CMAKE_CURRENT_SOURCE_DIR}/external"
+                -w "${CMAKE_CURRENT_BINARY_DIR}"
+                -o "${CMAKE_CURRENT_BINARY_DIR}"
+        DEPENDS ${external_SCRIPTS} ${external_SOURCES}
+    )
+
+    install(
+        FILES "${CMAKE_CURRENT_BINARY_DIR}/resource.dat"
```
Deledrius, line 59: `${external_DAT}`?

Deledrius (changes requested):

Sources/Plasma/Apps/plClient/CMakeLists.txt
Hoikas: This is potentially a no-op. Right now, it only functions if the user `-DENABLE_PIP_INSTALL=ON`. Later, if we detect that we are using the vcpkg submodule,. then we'll actually download the requirements here.

Deledrius: Good point.  I suppose it retains its utility for now as a central requirement list.

cmake/PythonUtils.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+    if(NOT _pim_REQUIREMENTS_FILE AND NOT _pim_MODULES)
+        message(FATAL_ERROR "python_install_modules() requires either a REQUIREMENTS_FILE or a list of MODULES")
+    endif()
+    if(NOT Python3_EXECUTABLE)
+        message(WARNING "Python3 is not available, not installing requested modules.")
+    endif()
+
+    # Escape hatch for folks who don't want us mucking around with their system Python.
+    if(NOT ENABLE_PIP_INSTALL)
```
zrax, line 45: I wonder if this should be moved to a higher location (e.g. around the call to `python_install_modules`, rather than within it)...  Like Deledrius, I was surprised to see an unprotected global pip install, since this line wasn't obvious at first glance.

Hoikas: I was thinking that I wanted this to be future proof in case in that the next person who uses this function wouldn't need to guard it. Maybe, as a compromise, this could be merged with python_test_modules such that modules could be installed from a requirements file before being tested?

Deledrius (approved):

cmake/PythonUtils.cmake
```diff
@@ -0,0 +1,67 @@
...
+            message(STATUS "Bootstrapping pip...")
+            execute_process(
+                COMMAND "${Python3_EXECUTABLE}" -m ensurepip
+                RESULT_VARIABLE RETURNCODE
+                OUTPUT_QUIET ERROR_QUIET
+            )
+            if(RETURNCODE)
+                message(WARNING "Unable to bootstrap pip")
+            endif()
+            set(_PLASMA_ENSUREPIP_DONE ON CACHE INTERNAL "" FORCE)
```
zrax, line 24: For clarity, I would probably use `YES` or `TRUE` rather than `ON`...  (CMake treats all of these, and others, as equivalent boolean constants -- see https://cmake.org/cmake/help/latest/command/if.html#condition-syntax)

cmake/PythonUtils.cmake
```diff
@@ -0,0 +1,67 @@
...
+                RESULT_VARIABLE RETURNCODE
+                OUTPUT_QUIET ERROR_QUIET
+            )
+            if(NOT RETURNCODE)
+                set(${_module_NAME}_FOUND ON PARENT_SCOPE)
+            else()
+                set(${_module_NAME}_FOUND OFF PARENT_SCOPE)
+            endif()
+        else()
+            set(${_module_NAME}_FOUND OFF PARENT_SCOPE)
```
zrax, line 60: Same as above re: boolean constants.

zrax (approved):

PR #760, opened 2020-12-22
Fix #743

### LINKED ISSUE #743: vcpkg library detection issues
When using libraries built by vcpkg, the _debug_ builds for these libraries are automatically selected by cmake:
- ogg
- opus
- vorbis
- vorbisfile

This can cause errors when attempting to link Release builds due to the inclusion of symbols only present in the debug CRT.

If these are using our own built-in finders, we need to update those finders to properly select the library configuration. Alternatively, we should preferentially utilize any exported cmake targets and fall back on dumb library name detection.

Hoikas:
This fixes #743. Vcpkg builds both debug and release libraries, and its toolchain hooks up both the debug and release paths to the CMake prefix. Some libraries, such as ogg and vorbis have the same name in debug and release mode, causing cmake to potentially pick up the debug library for release builds. This can result in link-time errors due to debug-CRT symbols being used in the dependencies. We fix this by attempting to defer to any cmake config packages where possible and fall back on either our find modules or cmake's built in find modules.

Because cmake config packages are target based, this also does a significant amount of work toward making our cmake more target based. This means removing lots of overly promiscuous include directories and compile definitions. It does not completely address the requirements of #383, but it it another step in the right direction.

Note that only dependencies that can be acquired through vcpkg have been completely overhauled... with the exception of string-theory. As a side effect, this also fixes two reports of expat being broken in the current devlibs bundle.


cmake/FindOpus.cmake
```diff
@@ -15,6 +15,27 @@ find_library(Silk_LIBRARY NAMES silk_common
...
+    if(EXISTS ${Celt_LIBRARY})
+        set_property(
+            TARGET Opus::opus APPEND PROPERTY
+            INTERFACE_LINK_LIBRARIES ${Celt_LIBRARY}
+        )
+    endif()
+    if(EXISTS ${Silk_LIBRARY})
+        set_property(
+            TARGET Opus::opus APPEND PROPERTY
+            INTERFACE_LINK_LIBRARIES ${Cilk_LIBRARY}
```
Deledrius, line 34: Cilk, while pronounced the same, is not a good substitute for Silk.
:trollface:

Deledrius (changes requested):

cmake/FindOpus.cmake
Hoikas: Good thing all that cilk mess is very, very old opus :tongue:

Sources/Plasma/FeatureLib/pfPasswordStore/CMakeLists.txt
```diff
@@ -8,36 +8,32 @@ set(pfPasswordStore_SOURCES
...
-if(APPLE)
-    set(pfPasswordStore_SOURCES ${pfPasswordStore_SOURCES}
-        pfPasswordStore_Apple.cpp
-    )
-endif(APPLE)
+    list(APPEND pfPasswordStore_SOURCES pfPasswordStore_Win.cpp)
+elseif(TARGET PkgConfig::LIBSECRET)
+    list(APPEND pfPasswordStore_SOURCES pfPasswordStore_Unix.cpp)
+elseif(TARGET Security::Security)
+    list(APPEND pfPasswordStore_SOURCES pfPasswordStore_Apple.cpp)
```
dpogue, line 15: We should probably put the Security case above the libsecret case, since Security is guaranteed to exist on macOS and we want to use it, but someone hypothetically could also compile libsecret for macOS.

Hoikas: I realize the solution may look weird, but the cmake documentation implies that multiple `elseif()` blocks may execute.

Deledrius (approved):

dpogue:
@Hoikas I believe you said to hold off on merging this because you were going to make further changes... correct?

Hoikas:
This is ready for review (and merging, I suppose) @dpogue

Hoikas:
Reminder, this is the last known blocker for #736

cmake/FindVPX.cmake
```diff
@@ -1,16 +1,41 @@
...
+find_library(VPX_LIBRARY_DEBUG NAMES vpxmtd vpxmdd vpxd
              PATHS /usr/local/lib /usr/lib
 )
 
-if(VPX_INCLUDE_DIR AND VPX_LIBRARY)
-    set(VPX_FOUND TRUE)
+select_library_configurations(VPX)
+find_package_handle_standard_args(VPX REQUIRED_VARS VPX_INCLUDE_DIR VPX_LIBRARY)
+
+if(VPX_FOUNS AND NOT TARGET VPX::VPX)
```
zrax, line 19:
```suggestion
if(VPX_FOUND AND NOT TARGET VPX::VPX)
```

zrax:
Other than the typo, it appears to work on both my Windows and Linux environments...

dpogue:
On Linux I get a bunch of stuff like this when I run `cmake ..` (merged with my GL branch, so it's pulling in more of the client):

```
CMake Warning (dev) in CMakeLists.txt:
  Policy CMP0111 is not set: An imported target missing its location property
  fails during generation.  Run "cmake --help-policy CMP0111" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  IMPORTED_LOCATION not set for imported target "PhysX::Character".
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This seems to show up for all of the PhysX libraries.

Hoikas:
Are all of the PhysX libraries otherwise found and properly set in the cmake cache? Just so I know where to start looking... A missing location on an imported target is very bad(TM).

dpogue:
The OpenAL link doesn't seem to work on Linux

dpogue:
I copied all of the PhysX library paths from my existing CMakeCache and pasted them into the new one (to save time on copying paths), and they all show up properly in the CMake GUI.

Hoikas:
I see some potential issues with the PhysX find module, but OpenAL looks good. What is the exact error it gives?

dpogue:
It doesn't seem to find OpenAL, so then plGLClient tries to link against `OpenAL::OpenAL-NOTFOUND`.

```
Sources/Plasma/Apps/plGLClient/CMakeFiles/plGLClient.dir/build.make:300: *** target pattern contains no '%'.  Stop.
make[2]: *** [CMakeFiles/Makefile2:2425: Sources/Plasma/Apps/plGLClient/CMakeFiles/plGLClient.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2433: Sources/Plasma/Apps/plGLClient/CMakeFiles/plGLClient.dir/rule] Error 2
make: *** [Makefile:233: plGLClient] Error 2
```

In plGLClient.dir/build.make:

```
Sources/Plasma/Apps/plGLClient/plGLClient: /usr/lib/libz.so
Sources/Plasma/Apps/plGLClient/plGLClient: OpenAL::OpenAL-NOTFOUND
Sources/Plasma/Apps/plGLClient/plGLClient: /usr/lib/libspeex.so
```

dpogue (approved):
Tested and confirmed working on Linux with the latest commit

zrax (approved):

Issue #689, opened 2020-07-25
PlasmaMAX Camera Components


Hoikas:
Building the MaxComponent project yields the following linker warnings in Visual C++:

`1>plCameraComponent.obj : warning LNK4006: "public: __thiscall plCameraCmdComponent::plCameraCmdComponent(void)" (??0plCameraCmdComponent@@QAE@XZ) already defined in plCameraComponents.obj; second definition ignored
1>plCameraComponent.obj : warning LNK4006: "public: virtual bool __thiscall plCameraCmdComponent::SetupProperties(class plMaxNode *,class plErrorMsg *)" (?SetupProperties@plCameraCmdComponent@@UAE_NPAVplMaxNode@@PAVplErrorMsg@@@Z) already defined in plCameraComponents.obj; second definition ignored
1>plCameraComponent.obj : warning LNK4006: "public: virtual bool __thiscall plCameraCmdComponent::PreConvert(class plMaxNode *,class plErrorMsg *)" (?PreConvert@plCameraCmdComponent@@UAE_NPAVplMaxNode@@PAVplErrorMsg@@@Z) already defined in plCameraComponents.obj; second definition ignored
1>plCameraComponent.obj : warning LNK4006: "public: virtual bool __thiscall plCameraCmdComponent::Convert(class plMaxNode *,class plErrorMsg *)" (?Convert@plCameraCmdComponent@@UAE_NPAVplMaxNode@@PAVplErrorMsg@@@Z) already defined in plCameraComponents.obj; second definition ignored
1>plCameraComponent.obj : warning LNK4006: "class ParamBlockDesc2 gCameraCmdBlock" (?gCameraCmdBlock@@3VParamBlockDesc2@@A) already defined in plCameraComponents.obj; second definition ignored`

PR #768, opened 2021-01-11
CMake 3.16 Build Optimizations

Hoikas:
While working on the vcpkg buildsystem as described in #736, I discovered that our precompiled header trickery doesn't work with the Ninja generator as used by Visual Studio. Fortunately, CMake 3.16 added native support for precompiled headers, so `PLASMA_USE_PCH` has been updated to use the native CMake support if we are on CMake 3.16 or higher. This also means that we get PCH on gcc as well. Support for CMake 3.14 and 3.15 remains, but you no longer have precompiled headers on those old versions.

CMake 3.16 also added support for unity builds, which seems to greatly decrease the build time of several of our targets. While we could simply pass `CMAKE_UNITY_BUILD=TRUE` on the command line, it seems the codebase has several ODR violations. So, for now, I've gone with a per-target opt-in approach. Unity can be disabled with `-DPLASMA_UNITY_BUILD=OFF`.


Sources/Plasma/CoreLib/hsWindows.h
```diff
@@ -84,6 +84,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #       include <vld.h>
 #   endif // USE_VLD
 
+    // Conflicts with plDynamicTextMap
+#   ifdef DrawText
+#       undef DrawText
+#   endif
```
zrax, line 90: It looks like you explicitly changed it to use `DrawTextA` though -- is this still needed?

Sources/Tools/MaxComponent/CMakeLists.txt
```diff
@@ -107,7 +106,6 @@ set(MaxComponent_SOURCES
     plBehavioralComponents.cpp
     plBipedKiller.cpp
     plBlowComponent.cpp
-    plCameraComponent.cpp
```
zrax, line 110: Not sure how the deletion of plCameraComponent is related to PCH/Unity builds...  Perhaps this should be a separate PR?

Sources/Tools/MaxComponent/plDistribComponent.cpp
```diff
@@ -434,9 +433,9 @@ class plDistCompBitmapProc : public plRollupMgrProc
 
             cbox = GetDlgItem(hWnd, IDC_COMP_DISTRIB_PROBCOLORCHAN);
             int i;
-            for( i = 0; i < kNumColorChanOptions; i++ )
+            for( i = 0; i < std::size(kColorChanStrings); i++ )
```
zrax, line 436: You're now just one step off from making a range for loop -- why stop here?  ;)

Sources/Tools/MaxComponent/plDistribComponent_old.cpp
```diff
@@ -152,7 +150,7 @@ class plDistribComponentProc_old : public ParamMap2UserDlgProc
         case WM_INITDIALOG:
             cbox = GetDlgItem(hWnd, IDC_COMP_DISTRIB_PROBCOLORCHAN);
             int i;
-            for( i = 0; i < kNumColorChanOptions; i++ )
+            for( i = 0; i < std::size(kProbColorChanStrings); i++ )
```
zrax, line 153: Same as above

Sources/Tools/MaxComponent/plCameraComponent.cpp
```diff
@@ -1,180 +0,0 @@
...
-static plCameraCmdComponentProc gCameraCmdComponentProc;
-
-enum
-{
-    kCommandSetOffset,
-    kCommandSetFP,
-    kCommandSetFixedCam,
-};
-
-ParamBlockDesc2 gCameraCmdBlock
```
zrax, line 113: This file does have a (presumably self-registering) global...  Are you sure it's actually unneeded?

Sources/Plasma/CoreLib/hsWindows.h
Hoikas: Yes, without this, `pyDynamicText::DrawText()` is impacted... But I could rename that method if you prefer.

Sources/Tools/MaxComponent/CMakeLists.txt
Hoikas: This file's contents were a duplicated subset of plCameraComponents.cpp, causing the unity build to fail. This was previously a warning per #689.

Sources/Tools/MaxComponent/plCameraComponent.cpp
Hoikas: See plCameraComponent**s**.cpp :wink:

Sources/Plasma/CoreLib/hsWindows.h
zrax: Ah, I missed what was conflicting here...  Nevermind then

Sources/Tools/MaxComponent/CMakeLists.txt
zrax: *sigh*

zrax (approved):

PR #769, opened 2021-01-11
Fix "Dammit GCC"

Hoikas:
Doesn't work on VC++2019.


zrax:
Here's an alternative which makes both compilers happy (not sure what VC2019 is doing to the deduction rules for `decltype` to fail):

```c++
        typedef void (*ClearT_Type)(_ConfigT*);
        using ClearT_Null = std::integral_constant<ClearT_Type, nullptr>;
        using ClearT_Param = std::integral_constant<ClearT_Type, _ClearT>;
```

zrax (approved):

zrax (approved):

PR #770, opened 2021-01-11
Warning Cleanup

zrax:
Greatly reduces the number of warnings (when enabled) for both GCC and MSVC builds.


Hoikas:
I wonder if compilation times would be negatively affected on MSVC by simply dropping the `#pragma hdrstop`s...

zrax:
> 
> 
> I wonder if compilation times would be negatively affected on MSVC by simply dropping the `#pragma hdrstop`s...

On my machine (building with `msbuild /m:8` on an SSD), I see a clean build in ~3m50s with `#pragma hdrstop` and ~4m04s without it, so the difference is not significant for me.  However, I don't know if the IDE builds in parallel by default, or how much different the results would be on a slower machine (or slower drives).

Hoikas:
That seems acceptable to me - there are plenty of other low hanging build speedups available in header cleanups and unity builds.

Hoikas (approved):

PR #774, opened 2021-01-13
vcpkg-Powered Windows Builds

Hoikas:
This changes the world of building, on Windows anyway. After this is merged, we will be using GitHub Actions for building Windows CI. AppVeyor should be disabled. Edit: I should also point out that this bumps the recommended compiler to Visual C++ 2019.

We now use a pinned revision of the vcpkg toolchain on Windows by default. Instead of downloading the dependencies from a zip file and generating projects, you can open the repository in Visual Studio. Our cmake scripts will (ideally) then download the dependencies from the same cache that CI uses. **Note that this requires us to store an org token in repo using the _scope read:packages_.** Read the updated build instructions [here](https://github.com/Hoikas/Plasma/blob/vcpkg-manifest3/README.md#compiling-instructions).

If you are on Windows and want to continue to use custom dependency builds or the old devlibs bundle, then you need to set `-DUSE_VCPKG=OFF`. No change is needed for other operating systems, although you *may* opt-in to the vcpkg usage by setting `-DUSE_VCPKG=ON` -- note that non-Windows platforms are untested. You will still be able to generate Visual Studio solutions from CMake if you prefer that over Visual Studio's built-in CMake support.

As of this writing, this is a draft pull request because the dependency cache is currently pointing at my fork. The last blocker to merging is pointing everything to the H-uru organization. Once that is done, you will not be able to download the dependencies until the merged CI run is completed. This is due to the interesting mixture of vcpkg binary caching and GitHub secrets in use.


README.md
```diff
@@ -43,46 +43,34 @@ The following libraries are optional:
...
+3. Select **Open a local folder** and choose the folder where you cloned the
+   repository.
+4. Open the CMake Settings by chosing **Project > CMake Settings for Plasma**
+   from the Visual Studio menu bar.
+5. Add a new configuration in CMakeSettings.json by clicking the **green +**
+   button.
+6. Select **x86-Release** in the window that pops up.
+7. Compile the client by using **Build > Install Plasma** from the Visual
+   Studio menu bar.
+8. The client will be built and installed into the *out\install\x86-Release*
```
dpogue, line 71: Prefer to use backticks around the `out\install\x86-Release` path, because otherwise the backslashes look like vertical separators and it's hard to figure out what's intended.

(alternatively, use non-Windows slashes)

Hoikas: Can you submit that as a suggestion, please?

dpogue:
```suggestion
8. The client will be built and installed into the `out\install\x86-Release`
```

Hoikas:
I have switched over the NuGet credentials in the CMake to our H'uru organization. As mentioned in the body of the PR, if you want to test the ability to restore the development libraries from CI, you will need to strip off a612e96. The org CI cache will not be populated until the PR is merged. PRs cannot write to the packages store for security reasons. That also means that each Windows-CI run to this PR will take about an hour because all the dependencies are being rebuilt. With the package store properly populated, each run of Windows-CI takes approximately 15-17 minutes, and the builds are done in parallel.

dpogue (approved):

Hoikas:
This rebase was to fix conflicts from #778 and pick up the Briggs bug fix from vcpkg upstream.

Hoikas:
:boom: :boom: :boom: :boom: :boom:

PR #778, opened 2021-01-19
Update devlibs bundle to point to latest zip

dpogue:
This includes the fix for the OpenAL WINMM.lib path, as well as a bump to Python 3.9.0.

This is probably the final devlibs bundle, as we'll be moving to using vcpkg directly: #774 (and related discussion at #736)


Hoikas:
The `-DPython3_LIBRARY` on line 47 needs to be updated and the devlibs URL in the README.md... which apparently got missed last time 😞

Hoikas (approved):

PR #779, opened 2021-01-19
Remove pnUtArray and replace TArray with std::vector

zrax:


Sources/Plasma/PubUtilLib/plAvatar/plAvatarClothing.cpp
```diff
@@ -910,15 +910,13 @@ void plClothingOutfit::WriteToVault(const TArray<plStateDataRecord*> & SDRs)
...
         &SDRs,
         &morphs,
     };
-    for (unsigned arrIdx = 0; arrIdx < std::size(arrs); ++arrIdx) {
-        const TArray<plStateDataRecord*> * arr = arrs[arrIdx];
-        
+    for (const auto * arr : arrs) {
         // Write all SDL to to the outfit folder, reusing existing nodes and creating new ones as necessary
-        for (unsigned i = 0; i < arr->Count(); ++i) {
+        for (unsigned i = 0; i < arr->size(); ++i) {
```
Hoikas, line 919: Might as well go ahead and turn this into a range-based for. @dpogue will be happy...

Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.cpp
```diff
@@ -522,15 +522,15 @@ static void INetCliAuthChangePasswordCallback (
...
+    const std::vector<NetAgeInfo> & ages
 ) {
     NetCommParam * cp = (NetCommParam *) param;
     
     plNetCommPublicAgeListMsg * msg = new plNetCommPublicAgeListMsg;
     msg->result     = result;
     msg->param      = cp->param;
     msg->ptype      = cp->type;
-    msg->ages.Set(ages.Ptr(), ages.Count());
+    msg->ages       = ages;
```
Hoikas, line 533: Why not pass by value, then `std::move()` here?

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -414,16 +414,17 @@ static void FetchRefOwners (
...
     }
-    std::sort(ownerIds.Ptr(), ownerIds.Term());
+    std::sort(ownerIds.begin(), ownerIds.end());
     NetVaultNode templateNode;
     templateNode.SetNodeType(plVault::kNodeType_PlayerInfo);
-    {   unsigned prevId = 0;
-        for (unsigned i = 0; i < ownerIds.Count(); ++i) {
+    {
+        unsigned prevId = 0;
+        for (unsigned i = 0; i < ownerIds.size(); ++i) {
```
Hoikas, line 427: Might as well go ahead with this being a range loop.

zrax: In this loop, `i` is used to index both `ownerIds` and `refs`

Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.cpp
zrax: `INetCliAuthGetPublicAgeListCallback` is used as a parameter of the function pointer type `FNetCliAuthGetPublicAgeListCallback`, which I didn't feel like committing to removing the reference on...  It's possible(probable) this is the only use though, in which case it would be safe.

Sources/Plasma/PubUtilLib/plAvatar/plAvatarClothing.cpp
zrax: Hmm, I thought I converted all the simple cases, but I guess I missed this one.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
dpogue: `size_t` in that case?

Hoikas (approved):

PR #782, opened 2021-01-21
Add a Resize method to hsTArray which performs Expand() + SetCount().

zrax:
This also fixes a bug introduced years ago where a call to ExpandAndZero() was replaced with Expand().  The former automatically adjusts the use count, whereas the latter does not :(.

### DIFF
--- a/Sources/Plasma/CoreLib/hsTemplates.h
+++ b/Sources/Plasma/CoreLib/hsTemplates.h
@@ -148,6 +148,13 @@ public:
     int     Count() const { return fUseCount; }
     int     GetCount() const { return fUseCount; }
     inline void SetCount(int count);
+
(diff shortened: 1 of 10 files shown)


Sources/Plasma/CoreLib/hsTemplates.h
```diff
@@ -148,6 +148,13 @@ template <class T> class hsTArray : public hsTArrayBase
     int     Count() const { return fUseCount; }
     int     GetCount() const { return fUseCount; }
     inline void SetCount(int count);
+
+    void Resize(int count)
+    {
+        // WARNING: By design (sigh), the new elements are not (re)initialized...
```
Hoikas, line 154: I think this should be moved up above the method as a doxygen style comment so it will pop up as a tip in the IDE.

Hoikas (approved):

PR #783, opened 2021-01-21
I see you like stack allocations...

Hoikas:
Visual C++ 2019's static analysis doesn't like it when you allocate 64 KiB blocks on the stack. I don't think these functions are actually *used* anywhere, but I tend to agree that doing such a thing is probably not wise.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plCompression/plZlibCompress.cpp
+++ b/Sources/Plasma/PubUtilLib/plCompression/plZlibCompress.cpp
@@ -44,6 +44,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsMemory.h"
 #include "hsStream.h"
 
+#include <memory>
+
 bool plZlibCompress::Uncompress(uint8_t* bufOut, uint32_t* bufLenOut, const uint8_t* bufIn, uint32_t bufLenIn)
 {
     unsigned long buflen_out = *bufLenOut;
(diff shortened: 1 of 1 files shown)


Sources/Plasma/PubUtilLib/plCompression/plZlibCompress.cpp
```diff
@@ -138,7 +141,7 @@ bool  plZlibCompress::UncompressFile( const char *compressedPath, const char *de
         {
             for( ;; )
             {
-                length = gzread( inFile, buffer, sizeof( buffer ) );
+                uint32_t length = gzread( inFile, buffer.get(), kGzBufferSize );
```
zrax, line 144: `gzread` returns an `int` (which makes the next line, `if( length < 0 )` make more sense)

Sources/Plasma/PubUtilLib/plCompression/plZlibCompress.cpp
```diff
@@ -246,34 +248,33 @@ bool  plZlibCompress::CompressToFile( hsStream * s, const char * filename )
...
+    auto buffer = std::make_unique<uint8_t[]>(kGzBufferSize);
 
     outFile = gzopen( filename, "wb" );
     if( outFile != nil )
     {
         for( ;; )
         {
-            int avail = s->GetEOF()-s->GetPosition();
-            int n = ( avail>sizeof( buffer ) ) ? sizeof( buffer ) : avail;
+            uint32_t avail = s->GetEOF()-s->GetPosition();
```
zrax, line 260:
```suggestion
            uint32_t avail = s->GetEOF() - s->GetPosition();
```
(The subtraction wasn't obvious to me)

zrax (changes requested):

zrax (approved):

PR #784, opened 2021-01-21
Prevent "coning" by using PhysX dominance groups.

Hoikas:
This fixes an observed issue by which other players can be moved around by pushing cones (or fire marbles...) into their character controllers.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPhysX/plPXSimulation.cpp
+++ b/Sources/Plasma/PubUtilLib/plPhysX/plPXSimulation.cpp
@@ -598,6 +598,8 @@ physx::PxScene* plPXSimulation::InitSubworld(const plKey& world)
     }
 #endif
 
+    scene->setDominanceGroupPair(plSimDefs::kGroupAvatar, plSimDefs::kGroupDynamic, { 0, 1 });
+
     auto it = fWorlds.try_emplace(world, scene);
     return it.first->second;
 }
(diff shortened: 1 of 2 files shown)


Sources/Plasma/PubUtilLib/plPhysX/plPXSimulation.cpp
```diff
@@ -598,6 +598,8 @@ physx::PxScene* plPXSimulation::InitSubworld(const plKey& world)
     }
 #endif
 
+    scene->setDominanceGroupPair(plSimDefs::kGroupAvatar, plSimDefs::kGroupDynamic, { 0, 1 });
```
Deledrius, line 601: What's the significance here of `{0, 1}`?

Hoikas: It's a `physx::PxDominanceGroupPair`. The usage of it is a little strange, though... You specify either 0 or 1 to indicate the dominance, but they act sort of like a multplier. So in a collision between a cone and an avatar, a cone imparts no force onto an avatar, but the avatar imparts its force unchanged onto a cone.

Deledrius: Ah, that seems pretty straightforward.

dpogue (approved):

Hoikas:
We're going to need to wait on merging this one. With this change, attempting to walk on top of the fish baskets in Eder Gira causes them to launch into the stratosphere.

Hoikas:
Maybe one day, I will come back to this.

PR #785, opened 2021-01-22
More pfLocalizationMgr optimizations.

Hoikas:
This is a minor performance increase to one of the client init hot paths. The goal was to remove lots of data copying. As a bonus, there is now a benchmarking tool for this in plLocalizationBenchmark. I'm OK with removing this tool, but I would like for it to at least live in the repo history somewhere.

I tested before and after on both the MOULa and GULP (aka "hell") localization data sets -- 100 iterations (not 1000), times in seconds:
```csv
,Before Total,After Total
MOULa,12.4428,11.0254
GULP,34.0057,29.9308
,,
,Before Avg,After Avg
MOULa,0.1244,0.1103
GULP,0.3401,0.2993
```

I did try using `std::unordered_map`, but that seemed to cause a performance regression... likely due to the constant iteration over the maps. The only ways to squeeze more out of this code might be to move to a DOM parser (eg pugixml) to avoid all the function calls and selectively reduce the validations, which only print warnings to the log.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
+++ b/Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
@@ -180,20 +180,16 @@ XML_Memory_Handling_Suite gHeapAllocator = {
 
 void XMLCALL LocalizationXMLFile::StartTag(void *userData, const XML_Char *element, const XML_Char **attributes)
 {
-    ST::string wElement = element;
     LocalizationXMLFile *file = (LocalizationXMLFile*)userData;
-    std::map<ST::string, ST::string> wAttributes;
 
-    for (int i = 0; attributes[i]; i += 2)
-        wAttributes[attributes[i]] = attributes[i+1];
+    LocalizationXMLFile::tagInfo newTag;
+    newTag.fTag = element;
+    for (size_t i = 0; attributes[i]; i += 2)
+        newTag.fAttributes[attributes[i]] = attributes[i+1];
 
     LocalizationXMLFile::tagInfo parentTag;
     if (!file->fTagStack.empty())
         parentTag = file->fTagStack.top();
-
-    LocalizationXMLFile::tagInfo newTag;
-    newTag.fTag = wElement;
-    newTag.fAttributes = wAttributes;
 
     file->fTagStack.push(newTag);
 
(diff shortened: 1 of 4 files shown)


Hoikas:
Link errors on Linux will be fixed by #794.

Hoikas:
Rebased on top of #794 to demonstrated the fixage.

Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
```diff
@@ -100,7 +102,7 @@ class LocalizationXMLFile
         ST::string fTag;
         std::map<ST::string, ST::string> fAttributes;
     };
-    std::stack<tagInfo> fTagStack;
+    std::deque<tagInfo> fTagStack;
```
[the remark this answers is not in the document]
Hoikas: I preferred `std::deque`'s emplacement methods, which can operate on either the back or front, unlike `std::stack's` which could only operate on the top. But, I guess I could use `std::stack::emplace()`.

Hoikas:
Rebased after all the chaos 😃

Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
```diff
@@ -542,16 +545,15 @@ void LocalizationDatabase::IMergeData()
 
 void LocalizationDatabase::IVerifyElement(const ST::string &ageName, const ST::string &setName, LocalizationXMLFile::set::iterator& curElement)
 {
-    std::vector<ST::string> languageNames;
-    ST::string defaultLanguage;
+    std::unordered_set<ST::string, ST::hash> languageNames;
```
zrax, line 548:
```suggestion
    std::unordered_set<ST::string> languageNames;
```

`std::hash` is already specialized for `ST::string`

Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
```diff
@@ -169,39 +182,34 @@ XML_Memory_Handling_Suite gHeapAllocator = {
...
-    LocalizationXMLFile::tagInfo parentTag;
-    if (!file->fTagStack.empty())
-        parentTag = file->fTagStack.top();
 
-    LocalizationXMLFile::tagInfo newTag;
-    newTag.fTag = wElement;
-    newTag.fAttributes = wAttributes;
+    // Need a dummy parent tag on the stack at minimum
+    if (file->fTagStack.empty())
+        file->fTagStack.emplace();
```
zrax, line 189: This initial tag emplacement could be moved to `LocalizationXMLFile::Parse` instead of checking it on every `StartTag`

Sources/Tools/CMakeLists.txt
```diff
@@ -41,10 +41,10 @@ function(_add_tools)
 
     add_dependencies(tools ${_at_TOOLS})
     if(_at_GUI)
-        install(TARGETS ${_at_TOOLS} DESTINATION tools_gui)
+        install(TARGETS ${_at_TOOLS} OPTIONAL DESTINATION tools_gui)
         set(_TOOLS_GUI ${_at_TOOLS} ${_TOOLS_GUI} PARENT_SCOPE)
     else()
-        install(TARGETS ${_at_TOOLS} DESTINATION tools_cli)
+        install(TARGETS ${_at_TOOLS} OPTIONAL DESTINATION tools_cli)
```
zrax, line 47: I can't imagine we ever want to actually install the benchmark program...  Perhaps it should just be a simple `add_subdirectory` instead of part of the tools target...

Hoikas: The downside is that it then becomes more difficult to get to the benchmark program if you want it, especially on multi-config generators. Navigating the build tree to the generated artifacts is unpleasant IME. :/

zrax (approved):

PR #789, opened 2021-01-24
Fix some minor Windows-CI Issues.

Hoikas:
Fixes two observed issues with the Windows CI:
- Runs the install target as a build step so that `python.39.dll` is properly available in the artifact bundle. Vcpkg is internally using `dumpbin.exe` for this but cannot find that executable if cmake is in install mode.
- Avoid caching the Qt install -- the cache can be poisoned if the Qt server goes down. This happened immediately on the merge of #744, causing the GUI tools to be omitted from the CI build.

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -19,13 +19,6 @@ jobs:
         with:
           submodules: true
 
-      - name: Cache Qt
-        id: cache-qt
-        uses: actions/cache@v2
-        with:
-          path: ${{ github.workspace }}/qt
-          key: qt-${{ matrix.platform.qt-version }} ${{ matrix.platform.qt-arch }}
-
       - name: Install Qt
         continue-on-error: true
         uses: jurplel/install-qt-action@v2
@@ -33,7 +26,6 @@ jobs:
           version: ${{ matrix.platform.qt-version }}
           arch: ${{ matrix.platform.qt-arch }}
           dir: ${{ github.workspace }}/qt
-          cached: ${{ steps.cache-qt.outputs.cache-hit }}
 
       - name: Build
         run: |
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -49,8 +49,7 @@ jobs:
             -DPLASMA_VCPKG_NUGET_TOKEN="${{ secrets.GITHUB_TOKEN }}" `
             -DPLASMA_VCPKG_NUGET_RW=TRUE `
             ..
-          cmake --build . --config "${{ matrix.cfg.type }}" -j 2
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-ci.yml
```diff
@@ -49,8 +41,7 @@ jobs:
             -DPLASMA_VCPKG_NUGET_TOKEN="${{ secrets.GITHUB_TOKEN }}" `
             -DPLASMA_VCPKG_NUGET_RW=TRUE `
             ..
-          cmake --build . --config "${{ matrix.cfg.type }}" -j 2
-          cmake --install . --config "${{ matrix.cfg.type }}"
+          cmake --build . --target INSTALL --config "${{ matrix.cfg.type }}" -j 2
```
dpogue, line 44: I was curious if we need to pass `--config` in to each step (although I guess there's only one now), or if we could pass it in to the first cmake command and have it set for the remainder of them

zrax: For multi-config generators (e.g. MSVC projects), I believe you do have to specify the config on each `--build` command...

Personal prefernce:  I would prefer to see the build and install targets kept separate, so they're easier to see in the build logs.

Also, we should probably run the `check` target here too; we may not have many tests, but we should try to keep them healthy.  Note that you'll need the Python DLL in PATH (or in the target's build directory) for the pfPythonTest to work.

Hoikas: > Personal prefernce: I would prefer to see the build and install targets kept separate, so they're easier to see in the build logs.

Would you like them in a completely separate step in the YAML, or would you want separate calls `cmake --build .` for the `ALL_BUILD` and `INSTALL` targets?

> Also, we should probably run the check target here too; we may not have many tests, but we should try to keep them healthy. Note that you'll need the Python DLL in PATH (or in the target's build directory) for the pfPythonTest to work.

The check target is run as the last step in the YAML already. `VCPKG_APPLOCAL_DEPS` takes care of copying over the python DLL for us :smile:. I did it this way so we could get the installed artifacts, even if the tests fail.

zrax: > 
> 
> Would you like them in a completely separate step in the YAML, or would you want separate calls `cmake --build .` for the `ALL_BUILD` and `INSTALL` targets in the current step?

Having configure, build, and install all as separate steps in the YAML makes the navigation nice in the build results, but it also adds a bit of extra complexity to the file...  I'd be fine with just separate cmake calls.

> 
> 
> The check target is run as the last step in the YAML already. `VCPKG_APPLOCAL_DEPS` takes care of copying over the python DLL for us 😄. I did it this way so we could get the installed artifacts, even if the tests fail.

Ah, I missed that, sorry (didn't expect it to be after "upload artifacts" ^_^)

zrax (approved):

PR #791, opened 2021-01-25
Fix some issues with vcpkg on Linux.

Hoikas:
This completely fixes `-DUSE_VCPKG=ON` on Linux. One thing to be aware of is that many vcpkg ports have dependencies on system packages. If any of these are missing, the configure may error out with a mysterious message like "Build program Ninja not found" or something similar. In this case, you'd need to check the `vcpkg-manifest-install.log` to see what is missing. I find these packages to be the most common stumbling blocks: `autoconf libtool nasm ninja-build`, but there may be others.

I expect this will also completely fix macOS as well, but I can't afford the Apple Tax.


dpogue (approved):
Would slightly prefer to wait until https://github.com/microsoft/vcpkg/pull/15855 is resolved so that we don't lose speex support

Hoikas:
The current vcpkg revision kills the entire build with an error on speex, sadly. I think when that PR gets in we can look into updating our submodule :smile:

PR #793, opened 2021-01-27
Fix Shroomie Gate state consistency

### LINKED ISSUE #792: Shroomie Gate, Inconsistent State
via Ehren on the Guild of Writers forum: https://forum.guildofwriters.org/viewtopic.php?f=117&t=7230

> This is a glitch that exists on every MOULa server, but it just occurred to me that someone might actually care to fix it in a future version here. In Teledahn if you pull the lever to open/close the dam with the power cut off to it (either by having the main power off or the lever for the spinning shaft disengaged), the gate will appear to not move, but relinking will show its state actually did toggle despite the lack of power.
>
> In other words, having the power off merely prevents the gate from animating, and relinking or starting up the power and pulling the lever again afterwards shows that the state did in fact change. For the sake of logical consistency I think it would be worth a small fix, I can't imagine it would be too hard to do.

dpogue:
The state was being modified regardless of whether the gate had power or not, but should only be changed when power is on (and when the relevant responders are run).

Closes #792.

### DIFF
--- a/Scripts/Python/tldnShroomieGate.py
+++ b/Scripts/Python/tldnShroomieGate.py
@@ -76,7 +76,7 @@ class tldnShroomieGate(ptResponder):
 
         elif id == respLeverPull.id:
             ageSDL = PtGetAgeSDL()
-            PtDebugPrint("tldnShroomieGate:\t---Shroomie Gate Up SDL: %d" % (ageSDL[kGateVariable][0]))
+            PtDebugPrint(f"tldnShroomieGate:\t---Shroomie Gate Up SDL: {ageSDL[kGateVariable][0]}")
             if ageSDL["tldnShroomieGatePowerOn"][0] and self.sceneobject.isLocallyOwned():
                 ageSDL[kGateVariable] = (not ageSDL[kGateVariable][0],)
 
@@ -87,15 +87,9 @@ class tldnShroomieGate(ptResponder):
         ageSDL = PtGetAgeSDL()
         value = ageSDL[kGateVariable][0]
 
-        if playerID:
-            try:
-                avatar = PtGetAvatarKeyFromClientID(playerID).getSceneObject()
-            except:
-                avatar = None
-            ff = False
-        else:
-            avatar = None
-            ff = True
+        avatarKey = PtGetAvatarKeyFromClientID(playerID)
+        avatar = avatarKey.getSceneObject() if avatarKey else None
+        ff = avatar is None
 
         if value:
             respGateUp.run(self.key, avatar=avatar, fastforward=ff)
@@ -116,7 +110,7 @@ class tldnShroomieGate(ptResponder):
 
         if ageSDL[kGateVariable][0]:
             PtDebugPrint("tldnShroomieGate:\tInit---Shroomie Gate Up")
-            respGateUp.run(self.key,fastforward=1)
+            respGateUp.run(self.key, fastforward=True)
         else:
             PtDebugPrint("tldnShroomieGate:\tInit---Shroomie Gate Down")
-            respGateDown.run(self.key,fastforward=1)
+            respGateDown.run(self.key, fastforward=True)
--- a/Scripts/Python/tldnShroomieGate.py
+++ b/Scripts/Python/tldnShroomieGate.py
@@ -56,17 +56,19 @@ respLeverPull = ptAttribResponder(2, "resp: Lever Pull", netForce=1)
 respGateDown = ptAttribResponder(3, "resp: Gate Down", netForce=1)
 respGateUp = ptAttribResponder(4, "resp: Gate Up", netForce=1)
 
+kGateVariable = "tldnShroomieGateUp"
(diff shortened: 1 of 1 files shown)


Deledrius (approved):

Hoikas (changes requested):
This only fixes half of the issue in that it prevents the state from changing when the power is off. The shroomie gate needs to respond (read: animate) when its state is changed.

dpogue:
My understand from the bug report is that it **does** animate properly when the power is on.

When the power is off, it (correctly) does not animate, but it did (incorrectly) change the state.

Hoikas:
My complaint is with something you will not observe in normal gameplay. That is, if you manually toggle the SDL variable `tldnShroomieGateUp`, then the visual state and the logical state will become desynchronized because the animation is triggered by the lever pull responder, not the actual state change. This will become a visual desynchronization if anyone links in after this operation.

EhrenCG:
Hey, I remembered my password!  I find GitHub a little complex and intimidating so thank you for submitting this!

Hoikas, I think I understand what you're talking about since you mentioned the scenario of what if someone changed the SDL in the background without physically pulling the lever?  Isn't this the sort of thing that needs an "OnSDLNotify" to make sure people would see it animate?

I guess this counts as a case of *ahem*... "appreciating" Cyan's programming.

Hoikas:
That's correct, @EhrenCG. Ideally, we want all of the game scripts to respond to SDL updates a) in case someone changes the SDL state and b) to ensure proper synchronization. Those kinds of changes might come from someone who is a shard admin using admin tools, someone playing around with an internal client, or a bot. As the fix currently stands, those actors would not be able to raise/lower the gate without requiring people link out and back into the Age, as you described in your bug report.

EhrenCG:
@Hoikas How widespread is this problem?  I just happened to look into GiraDoor.py.  It also has the same synch issue.  It even has OnSDLNotify commented out.  (Also, the info under the license refers to GiraDoor.py as "Module: Garden.py", heh.) I'm far from an expert in this stuff but it seems like a mess to me.  So maybe a lot of files have this issue, would take awhile to go through them all...

dpogue:
@Hoikas updated to use `OnSDLNotify`

Scripts/Python/tldnShroomieGate.py
```diff
@@ -56,48 +56,67 @@
...
+        ageSDL.setNotify(self.key, kGateVariable, 0.0)
+
+        if ageSDL[kGateVariable][0]:
             PtDebugPrint("tldnShroomieGate:\tInit---Shroomie Gate Up")
             respGateUp.run(self.key,fastforward=1)
         else:
             PtDebugPrint("tldnShroomieGate:\tInit---Shroomie Gate Down")
-            respGateDown.run(self.key,fastforward=1)
\ No newline at end of file
+            respGateDown.run(self.key,fastforward=1)
```
Hoikas, line 122:
```suggestion
            respGateDown.run(self.key, fastforward=True)
```

Scripts/Python/tldnShroomieGate.py
```diff
@@ -56,48 +56,67 @@
...
-                if ageSDL["tldnShroomieGateUp"][0]:
-                    respGateDown.run(self.key)
-                    PtDebugPrint("tldnShroomieGate:\t---Shroomie Gate Going Down")
-                else:
-                    respGateUp.run(self.key)
-                    PtDebugPrint("tldnShroomieGate:\t---Shroomie Gate Going Up")
-            ageSDL["tldnShroomieGateUp"] = (not ageSDL["tldnShroomieGateUp"][0],)
-                            
-        
+            PtDebugPrint("tldnShroomieGate:\t---Shroomie Gate Up SDL: %d" % (ageSDL[kGateVariable][0]))
```
Hoikas, line 79:
```suggestion
            PtDebugPrint(f"tldnShroomieGate:\t---Shroomie Gate Up SDL: {ageSDL[kGateVariable][0]}")
```

Scripts/Python/tldnShroomieGate.py
```diff
@@ -56,48 +56,67 @@
...
+
+        if playerID:
+            try:
+                avatar = PtGetAvatarKeyFromClientID(playerID).getSceneObject()
+            except:
+                avatar = None
+            ff = False
+        else:
+            avatar = None
+            ff = True
```
Hoikas, line 98:
```suggestion
        avatarKey = PtGetAvatarKeyFromClientID(playerID)
        avatar = avatarKey.getSceneObject() if avatarKey else None
        ff = avatar is None
```

Hoikas (commented):
This could be cleaned up even more with `@property` accessors and mutators for the AgeSDL, but I won't insist on it.

Hoikas:
@EhrenCG Yep, looks like that one has the same issue. It's less noticeable, I expect, because it works fine under normal gameplay.

Hoikas (approved):

PR #796, opened 2021-01-28
Fix cursor resolution (#795)

### LINKED ISSUE #795: Cursor resolution
Hey,
I don't know if it's actually an issue but when I'm compiling the client by my self the ingame cursor is way to small. :D

Screenshot 2021-01-27 231158

Hoikas:
Fixes #795.

### DIFF
--- a/Sources/Plasma/Apps/plClient/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plClient/CMakeLists.txt
@@ -5,13 +5,12 @@ set(external_SCRIPTS
     external/scalergba.py
 )
 
-set(external_SCRIPTS
-    external/create_resource_dat.py
-    external/makeres.py
-    external/render_svg.py
-    external/scalergba.py
+set(external_SOURCES
+    external/Cursor_Base.svg
+    external/Linking_Book.svg
+    external/Loading_Text_rasterfont.svg
+    external/Voice_Chat.svg
 )
-
 find_program(
     PNGCRUSH_EXECUTABLE pngcrush
     DOC "Path to pngcrush"
--- a/Sources/Plasma/Apps/plClient/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plClient/CMakeLists.txt
@@ -1,3 +1,17 @@
+set(external_SCRIPTS
+    external/create_resource_dat.py
+    external/makeres.py
+    external/render_svg.py
+    external/scalergba.py
+)
+
+set(external_SCRIPTS
+    external/create_resource_dat.py
+    external/makeres.py
+    external/render_svg.py
+    external/scalergba.py
+)
+
 find_program(
     PNGCRUSH_EXECUTABLE pngcrush
     DOC "Path to pngcrush"
--- a/Sources/Plasma/Apps/plClient/external/render_svg.py
+++ b/Sources/Plasma/Apps/plClient/external/render_svg.py
@@ -146,8 +146,11 @@ def render_cursors(inpath, outpath):
 			shift_all_layers(layers, *cursorOffsetList.get(cursor, [0, 0]))
 
 			outfile = os.path.join(outpath, cursor + ".png")
+			cursorSVG.documentElement.setAttribute("width", str(scalefactor*svgwidth))
+			cursorSVG.documentElement.setAttribute("height", str(scalefactor*svgheight))
+			cursorSVG.documentElement.setAttribute("viewBox", "0 0 {} {}".format(svgwidth, svgheight))
 			cairosvg.svg2png(bytestring=cursorSVG.toxml().encode('utf-8'), write_to=outfile, 
-				parent_width=svgwidth, parent_height=svgheight, scale=scalefactor)
+				parent_width=scalefactor*svgwidth, parent_height=scalefactor*svgheight)
 			scalergba.scale(outfile, outfile, scalefactor)
 
 def render_loading_books(inpath, outpath):


cwalther:
> The cursor is already scaled by the SVG processor. Don't scale it again.

But does the SVG processor (whatever that is, I didn’t check) do it right? I spent a lot of effort on the algorithm in `scalergba.py`.

If you do throw it out, you could remove that file too.

Hoikas:
I was not aware there was a wrong way to scale an image. How can I determine that?

cwalther:
Pretty much everyone does it wrong. The comments at the top of `scalergba.py` explain how.

The purpose of `scalergba.py` is not so much scaling in and of itself, but achieving a file with RGBA values that will look right when composited by DirectX. For that purpose, it needs additional information that is obtained by getting an oversized image from the previous stage of the pipeline.

Hoikas:
Unfortunately, I do not understand the comments at the top of `scalergba.py` . I don't have enough experience with graphics to understand what the practical implications. What I can tell is that my old resource.dat from years ago looks correct, resource.dat generated by master have a small cursor, and resource.dat generated with this change look correct. If this is unacceptable, someone else will have to take over.

Cursor with this PR:
Screenshot 2021-01-28 06 48 00

cwalther:
Could you send me the new resource.dat generated with this change, or compare it to the old one yourself? I suspect you’ll find a systematic difference in the alpha values of partially covered pixels. If not, then great.

For a visual check, the book sharing cursor with its thin diagonal lines is best, on dark and light backgrounds.

Hoikas:
I didn't see anything terribly different with the book sharing cursor, but I do think I remember there being issues with that in the past. Here is the resulting resource.dat:

[resource.zip](https://github.com/H-uru/Plasma/files/5886924/resource.zip)

cwalther:
Thanks, I’ll take a look tonight.

Regarding the graphics background: It all comes down to http://poynton.ca/GammaFAQ.html.
Basically, when trying to present a pixel whose area is half covered with white, half with black, most software (at least pixel-based ones, I don’t know about SVG renderers) will give you an 8-bit value of 128. That is wrong. The correct solution is “emit half the light”. 128 does not mean “emit half the light”, it means “send half the voltage to the CRT”. (Which, conveniently, happens to result in something that looks like “medium brightness” to human vision, which is the reason why we still do it that way even though we no longer use CRTs.) The correct value for “emit half the light” is approximately 0.5^(1/2.2)*255 = 186.

Hoikas:
Thanks for the background - it's interesting and explains how what you're saying would be hard to observe. I expect that some folks would be more able to discern that difference than others.

I can't help but wonder if the update to Python 3 and some subtle behavior difference is the cause of scalergba resulting in a too-small cursor. The code looks Python 3 aware to me, but I'm just struggling to understand what broke.

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -1,3 +1,17 @@
...
+    external/scalergba.py
+)
+
+set(external_SCRIPTS
+    external/create_resource_dat.py
+    external/makeres.py
+    external/render_svg.py
+    external/scalergba.py
+)
+
```
Deledrius, line 14:
```suggestion
set(external_SOURCES
    external/Cursor_Base.svg
    external/Linking_Book.svg
    external/Loading_Text_rasterfont.svg
    external/Voice_Chat.svg
)
```
I don't think we need it twice to make up for not having it at all.

Deledrius (changes requested):

Sources/Plasma/Apps/plClient/CMakeLists.txt
Hoikas: :trollface:

cwalther:
I haven’t checked the details yet (it may be more than just the scaling and gamma treatment), but there is clearly something wrong here:

image

Left is what it’s supposed to look like (one I had lying around from 2014), right is your new one.

cwalther:
Alright:

- What broke the scaling appears to be 50cc61c52a0b01675b824b4f746d149b1efc9bb3. I get the correct scale with the current CairoSVG 2.5.1 and the tiny one with 1.0.22 as specified in requirements.txt. It looks like 1.0.22 doesn’t support the `scale` argument, so I wonder, has this ever worked for you, @Deledrius, since you already specified it that way in [178e44d](https://github.com/H-uru/Plasma/commit/178e44dc8c9b65ee49bff5b6d0c97cdabf09604b#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dc)?
- The lack of shadows is explained by
  https://cairosvg.org/documentation/#colors-gradients-patterns-clipping-masking-filtering-compositing:
  > Only `feOffset`, `feBlend` and `feFlood` filters are supported.
  
  https://github.com/H-uru/Plasma/blob/c8ba46cabfae08e046b5577567fc1b79d9145805/Sources/Plasma/Apps/plClient/external/Cursor_Base.svg#L32 uses `feGaussianBlur`.

So – find a better SVG renderer?

Hoikas:
CairoSVG 2.x doesn't function properly on Windows. 😞

Deledrius:
> I get the correct scale with the current CairoSVG 2.5.1 and the tiny one with 1.0.22 as specified in requirements.txt. It looks like 1.0.22 doesn’t support the `scale` argument, so I wonder, has this ever worked for you, @Deledrius, since you already specified it that way in [178e44d](https://github.com/H-uru/Plasma/commit/178e44dc8c9b65ee49bff5b6d0c97cdabf09604b#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dc)?

It did, in my testing, however as I recall some of that version locking was done after the majority of testing in order to satisfy build requirements that cropped up after the work was done. :/  I'm guessing I never noticed it after we fought to get it working on other build systems.

Hoikas:
Do we need to rewrite these tools in C++?

Deledrius:
> 
> 
> Do we need to rewrite these tools in C++?

We may have to at this point.  Finding a decent library is hard enough these days; finding one also with working Python bindings seems to be impossible.

cwalther:
Quick fix for the scaling with CairoSVG 1.0.22:

```diff
diff --git a/Sources/Plasma/Apps/plClient/external/render_svg.py b/Sources/Plasma/Apps/plClient/external/render_svg.py
index 72f24db17..989b22c67 100644
--- a/Sources/Plasma/Apps/plClient/external/render_svg.py
+++ b/Sources/Plasma/Apps/plClient/external/render_svg.py
@@ -146,8 +146,11 @@ def render_cursors(inpath, outpath):
                        shift_all_layers(layers, *cursorOffsetList.get(cursor, [0, 0]))
 
                        outfile = os.path.join(outpath, cursor + ".png")
+                       cursorSVG.documentElement.setAttribute("width", str(scalefactor*svgwidth))
+                       cursorSVG.documentElement.setAttribute("height", str(scalefactor*svgheight))
+                       cursorSVG.documentElement.setAttribute("viewBox", "0 0 {} {}".format(svgwidth, svgheight))
                        cairosvg.svg2png(bytestring=cursorSVG.toxml().encode('utf-8'), write_to=outfile, 
-                               parent_width=svgwidth, parent_height=svgheight, scale=scalefactor)
+                               parent_width=scalefactor*svgwidth, parent_height=scalefactor*svgheight)
                        scalergba.scale(outfile, outfile, scalefactor)
 
 def render_loading_books(inpath, outpath):
```

cwalther:
Which is actually all that was asked for in #795, so I can make a PR for that after I give it a bit more testing, and we can treat the missing shadows as a separate issue. But first, dinner.

Deledrius:
Thanks for looking into the problem, @cwalther.

cwalther:
Okay, tested, the cursor PNGs are now identical between CairoSVG 1.0.22 and 2.5.1, and look good apart from the missing shadows: https://github.com/cwalther/Plasma/commit/6aec4b670d7a332a4f9fbb1010b7dbab59950fd7. @Hoikas, do you want to cherry-pick/force-push that onto this PR (replacing 0b0c2bf but keeping the other two commits) so we don’t split the discussion?

I noticed another issue with CairoSVG 1.0.22 though: _ui_microphone.png_ and _ui_speaker.png_ are not rendered correctly either, some parts are missing. They look okay on 2.5.1.

Regarding possible replacements: Inkscape can be called on the command line to export PNGs – may be a too large dependency though.

And another side note regarding the usefulness of scaling, independently of the gamma treatment done in _scalergba.py_: In the right picture in my “clearly something wrong” screenshot above, notice the gray pixels along the edges? These shouldn’t be there. The black shadow (that erroneously isn’t blurred) behind the white figure has the exact same shape as the white figure, it should disappear completely behind it. There should only be white pixels, of varying opacity. But apparently the renderer rasterizes each shape individually, so for a partially covered pixel you first get semi-transparent black and then semi-transparent white on top of that, which makes gray. Rendering at a higher resolution and then scaling down in pixel-land, effectively supersampling, reduces that.

Hoikas:
OK, I've stripped off 0b0c2bf and picked cwalther@6aec4b6 onto the end of this branch.

What we were aiming for with dependencies was to allow for the complete client to be built on Windows by installing Visual Studio, then executing Visual Studio. No fumbling with other libraries or dependencies involved. So, that makes Inkscape incompatible with that goal. 

I'm hopeful that we can figure out a complete fix for these problems, given the sorry state of SVG rendering libraries.

cwalther (approved):

Hoikas:
I have a more complete fix for the svg rendering in progress, but it looks like I am going to upstream a change to the vcpkg cairo port in the process. The end result is that we will be able to use cairosvg 2.5.1 on all platforms with the current resource scripts. That being said, upstreaming to vcpkg may or may not be very fast, so I think we should merge @cwalther's quick fix for now.

cwalther:
Great! Using CairoSVG 2.5.1 does not fix the missing shadows though, so it’s still not a complete fix.

PR #797, opened 2021-01-28
Override all the things

zrax:
This refactors most of Plasma to use "override" on overridden virtual non-destructor methods.

For code that can be built on Linux, this was done with the assistance of GCC's `-Werror=suggest-override` and should be complete (*).  For the rest, including the massive number of virtual functions in the 3ds Max plugin code, this was done manually (and therefore, some overrides may have been missed).

(*) NOTE:  This does not address other possible issues with virtual functions, including but not limited to:
* Failure to override (functions which should override but don't because of a parameter mismatch, typo, etc).
* Failure to virtualize (functions which hide a parent method which is not virtual)
* Unnecessary virtual (functions which are never overridden, or whose implementation is identical to the parent's)
* Missing virtual destructors
* Dead code elimination, or any other non-whitespace cleanup in the affected code.


Hoikas:
Have you tried using Visual Studio's copy of clang-tidy to run `modernize-use-override`? Since the codebase now plays nicely with VS's built-in CMake support, it shouldn't be too hard to do. Not that I think this PR should be held up on that, just curious.

zrax:
I did try, actually, but I had some other troubles with clang not parsing the Max headers correctly, which I haven't tried to look into solving yet :/

Hoikas:
Ugh. It doesn't help that they are older than dirt.

Sources/Plasma/PubUtilLib/plAvatar/plNPCSpawnMod.h
```diff
@@ -57,14 +57,14 @@ class plNPCSpawnMod : public plSingleModifier
     CLASSNAME_REGISTER( plNPCSpawnMod );
     GETINTERFACE_ANY( plNPCSpawnMod, plSingleModifier );
     
-    virtual void AddTarget(plSceneObject* so);
+    void AddTarget(plSceneObject* so) override;
 //  bool MsgReceive(plMessage* msg);
```
Hoikas, line 61:
```suggestion
//  bool MsgReceive(plMessage* msg) override;
```

Or delete it...

Hoikas (commented):
"270 / 581 files viewed"

Hoikas (approved):

Issue #800, opened 2021-02-01
Client Resource.dat Rendering Roundup


Hoikas:
This is a meta-issue for tracking the outstanding issues with cursor rendering. See the discussion in #796 for more detail.

* [x] Cursor appears too small in game.
    * Reported in #795 
    * Fixed by #796
* [x] Microphone icon appears garbled.
    * Would be fixed by a bump to cairosvg 2.5.1 in #801
* [x] Cursor lacks shadows.
    * Cairosvg does not support the filter `feGaussianBlur`, a regression from rsvg
    * Fixed by #814

cwalther:
_Cursor lacks shadows_ is addressed by #814.

Possible additional issue: `xLoading_*.png` are generated in higher resolution than they used to, years ago – e.g. 1175×1177 instead of 256×256 for `xLoading_Linking.*.png`. Check what today’s client actually expects.

Hoikas:
This has been fixed for awhile. Closing.

PR #801, opened 2021-02-01
Bump to cairosvg 2.5.1

Hoikas:
This bumps the resource.dat generation scripts to cairosvg 2.5.1 to fix the garbled microphone icon as mentioned in #796 and #800. Unfortunately, modern cairosvg rely on cairocffi but do not include libcairo-2.dll on Windows. To work around that, I have written a vcpkg overlay port that builds cairo and copies it into the vcpkg python3 tools directory on Windows. Other platforms should simply install cairo from their package manager.

Currently, this will fail CI due to the vcpkg cairo port depending on fontconfig, which has dependencies that fail to build under a static triplet. I could list them all as dynamic in our triplet, but these dependencies collectively take over 30 minutes to build. So, we are blocking on upstream:
* [x] microsoft/vcpkg#15965
* [x] ~~microsoft/vcpkg#15944~~
* [x] microsoft/vcpkg#15474
* [x] #895 

Note that because the triplet files have changed, the hash of each dependency package changes as well... In case you're wondering why CI is slow here :smile:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,6 +139,7 @@ jobs:
         run: |
           brew install \
               autoconf \
+              automake \
               cairo \
               libtool \
               nasm \
--- a/Scripts/Ports/python3-cairosvg/portfile.cmake
+++ b/Scripts/Ports/python3-cairosvg/portfile.cmake
@@ -6,24 +6,85 @@ if(NOT VCPKG_TARGET_IS_WINDOWS)
 endif()
 
 set(CAIROSVG_VERSION "2.5.1")
-set(CURRENT_PYTHON3_PREFIX "${CURRENT_PACKAGES_DIR}/tools/python3")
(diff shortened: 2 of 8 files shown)


Hoikas:
The vcpkg PRs we needed have now landed. I'll update this PR soon-ish (RL work beckons...) - it will include a small change to our vcpkg.json manifest due to upstream vcpkg learning how to build speex on linux and macOS :smile:

Hoikas:
OK, it looks like everything is working. From the vcpkg upgrade you get:
* speex on Linux and macOS, for whatever value that is
* PhysX debugging symbols are now included
* vcpkg.exe is downloaded on Windows instead of being built each CI run

Bonus:
* If the CI build fails, we now upload any log files we can find. Nice when the Tim Cook error strikes.

Now, let the evisceration of the overlay port begin 😆

Hoikas:
Rebased on top of #895, so merge that first.

Deledrius (approved):

PR #803, opened 2021-02-01
Fix an incorrect loop variable...

zrax:
I have no idea what it might affect, but I noticed this oddity while working on signed/unsigned comparison fixes...


Hoikas (approved):
Practically speaking, this would generally only be triggered in fan Ages exported by PyPRP. PlasmaMax and Korman exclude all Responder SDL states by default, so there's never any Responder SDL to restore. Still, better fix it.

PR #805, opened 2021-02-03
Make plKey a movable type.

zrax:


Hoikas (commented):
Could we move these into the header to encourage inlining? AFAIK don't have the proper CMake instrumentation set up (yet) to enable LTCG.

Hoikas (approved):

PR #806, opened 2021-02-04
macOS CI

dpogue:
I hate how it formats the diff here 🙃

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
             -DPLASMA_VCPKG_NUGET_RW=TRUE \
             -S . -B build
         env:
-          PKG_CONFIG_PATH: "/usr/local/opt/qt/lib/pkgconfig"
+          CMAKE_PREFIX_PATH: "/usr/local/opt/qt"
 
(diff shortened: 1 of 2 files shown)


zrax:
```
-- The following OPTIONAL packages have not been found:

 * Speex, Free and open audio codec specifically designed for speech, <http://www.speex.org/>
   Required for legacy in-game voice chat in the client
 * Qt5, Cross-platform GUI framework for C++, <http://www.qt.io/>
   Required for building GUI tools for Plasma
```

It seems despite installing qt5, the build doesn't find it...

dpogue:
Looks like trying to use the `PKG_CONFIG_PATH` envvar didn't work either. One option is probably to `brew link --force qt5`, but it specifically calls out that it doesn't play nice with CMake 😕

zrax:
Where does homebrew install it?  You could just add that to the CMAKE_PREFIX_PATH...

dpogue:
Thanks Zrax, `CMAKE_PREFIX_PATH` seems to work :)

zrax (approved):

Hoikas (approved):

PR #808, opened 2021-02-04
Fix several issues with libsecret find module

zrax:
Fix several issues with libsecret find module:
* Incorrect test for REQUIRED flag
* Incorrect parameter name for REQUIRED flag
* Typo in PkgConfig_FOUND test
* Incorrect capitalization of pkg-config module name

This was clearly never tested, and I unfortunately didn't notice it until recently because I already had a cached copy of the required variables from before it was reworked (and broken) in 9d48c967


Hoikas:
:trollface:

dpogue (approved):
This looks like it's working in the Linux CI now :+1:

Hoikas (approved):

PR #809, opened 2021-02-05
Fix some HRESULT -> bool conversions

dpogue:
It's not safe to treat an HRESULT as a boolean, and Microsoft provides the `FAILED()` macro to specifically check if the result is an error.

In the case of the DirectX Enumerate, it was actually returning bool but the function declaration said it returned an HRESULT.


Hoikas (approved):
Is `hsGDirect3DTnLEnumerate::SelectFromDevMode()` even used? We know that `plLayerAVI` junk isn't. Still, better fix it...

dpogue:
> Is `hsGDirect3DTnLEnumerate::SelectFromDevMode()` even used? We know that `plLayerAVI` junk isn't. Still, better fix it...

https://github.com/dpogue/Plasma/blob/4fec893a69ba72db0392846fc5c9ee8dc3cca467/Sources/Plasma/PubUtilLib/plPipeline/DX/plDXPipeline.cpp#L583

PR #810, opened 2021-02-05
Enable the /LTCG switch where applicable.

Hoikas:
CMake does not by default enable link-time code generation (read: optimizations). This adds the instrumentation to opt-in to such a thing without manual user intervention. This can be turned off by setting `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF`. While I am skeptical that LTCG will offer any significant performance benefit, it does seem like the produced plClient.exe is smaller (down from ~22,000 KB to ~16,000 KB).

### DIFF
--- a/cmake/CompilerChecks.cmake
+++ b/cmake/CompilerChecks.cmake
@@ -10,6 +10,17 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANGXX)
     set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
 endif()
 
+# Use LTCG if available.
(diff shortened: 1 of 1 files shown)


zrax (commented):
That much of a size reduction is probably worth it alone, regardless of performance...  However, did you verify this doesn't remove those "unreferenced" objects MSVC sometimes warns about?

Hoikas:
I noticed today that there seems to be an observable benefit to using this in debug mode. I didn't bother measuring anything though because meh, debug performance.

> did you verify this doesn't remove those "unreferenced" objects MSVC sometimes warns about?

Good question. Off the top of my head, the only ones I could think of are in the console commands. Those still seem to be working. Are there any others I should be looking for in particular?

zrax:
If the ones you found in the console are working, it's probably safe to say it works...

zrax (approved):

PR #814, opened 2021-02-07
Fix cursor shadow rendering with CairoSVG

cwalther:
See #800: Because CairoSVG does not support `feGaussianBlur`, do the blurring and compositing externally in _render_svg.py_.

It turns out CairoSVG also doesn’t fully support masks (it only applies the alpha of the mask, rather than the product of luminance and alpha as [specified](https://www.w3.org/TR/SVG11/masking.html#Masking)), but that doesn’t matter because the masking needs to be applied after the blurring, so we need to do it externally too anyway.

The result is not a pixel-perfect match for the old rsvg-rendered cursors, but the differences are hardly noticeable. (Mainly, in “poised” cursors, the shadow is a bit lighter in the center and a bit darker between the two rings.)

before:
cursorsheet-2 5 1

rsvg:
cursorsheet-b4206e5

new:
cursorsheet-out

Probably requires #801 – I only tested it with a locally self-installed CairoSVG 2.5.1 on macOS.

### DIFF
--- a/Sources/Plasma/Apps/plClient/external/render_svg.py
+++ b/Sources/Plasma/Apps/plClient/external/render_svg.py
@@ -56,6 +56,11 @@ try:
 except ImportError as e:
 	print("Rendering SVG resources requires CairoSVG.  Exiting...")
 	exit(1)
+try:
+	from PIL import Image, ImageFilter
+except ImportError:
+	print("Rendering SVG resources requires Pillow.  Exiting...")
+	exit(1)
 
 cursorList = {
 	"cursor_up": ["circleOuter"],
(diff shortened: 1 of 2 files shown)


cwalther:
_resource.dat_ in the CI artifacts is identical to what I got except for the broken _ui_microphone.png_ and _ui_speaker.png_.

Sources/Plasma/Apps/plClient/external/render_svg.py
```diff
@@ -56,6 +57,11 @@
 except ImportError as e:
 	print("Rendering SVG resources requires CairoSVG.  Exiting...")
 	exit(1)
+try:
+	from PIL import Image, ImageFilter
+except ImportError:
+	print("Rendering SVG resources requires Pillow.  Exiting...")
+	exit(1)
```
Hoikas, line 64: FWIW, if you execute `pip install cairosvg`, PIL(low) is installed as a dependency automatically.

Hoikas (approved):
Thanks for working on this, @cwalther! I had to look very closely indeed to see the difference you referenced. I'll let @Deledrius have the final word here, but this gets a big :+1: from me.

Sources/Plasma/Apps/plClient/external/render_svg.py
cwalther: Ah, good point. That was just mindless copy-paste. But in the spirit of “include what you use”, I’ll leave it here if you don’t mind (a future CairoSVG might stop requiring PIL).

Deledrius (approved):
This looks great.  We've been doing a lot of shake-ups on the entire pipeline and this regression was missed.  Thanks for not only noticing but providing a great working fix.

PR #815, opened 2021-02-08
De-duplicate lots of `NetTransRecv` code with modern C++.

Hoikas:
I've been digging deep into the networking code, planning its refactor into something a bit more modern and cross-platform. This is one of the low-hanging fruits from my notes. I am disappointed that the deletions didn't make it to 1k LOC, but oh well.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetGameLib/Intern.h
+++ b/Sources/Plasma/PubUtilLib/plNetGameLib/Intern.h
@@ -385,7 +385,7 @@ struct HasTransId<T, std::void_t<decltype(&T::transId)>> : std::true_type { };
 
 template<typename T>
 inline std::enable_if_t<HasTransId<T>::value, bool>
-NetTransRecvFromMsg(const uint8_t msg[], unsigned bytes, void*)
+NetTransRecvFromMsgGeneric(const uint8_t msg[], unsigned bytes, void*)
 {
     const T* reply = reinterpret_cast<const T*>(msg);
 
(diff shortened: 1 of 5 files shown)


Sources/Plasma/PubUtilLib/plNetGameLib/Intern.h
```diff
@@ -377,6 +377,26 @@ void NetTransCancelByConnId (unsigned connId, ENetError error);
 void NetTransCancelAll (ENetError error);
 void NetTransUpdate ();
 
+template<typename T, typename = void>
+struct HasTransId : std::false_type { };
+
+template<typename T>
+struct HasTransId<T, decltype((void)T::transId, void())> : std::true_type { };
```
zrax, line 384: This is what C++17's `void_t` is for :)

```suggestion
struct HasTransId<T, std::void_t<decltype(&T::transId)>> : std::true_type { };
```

Sources/Plasma/PubUtilLib/plNetGameLib/Intern.h
```diff
@@ -377,6 +377,26 @@ void NetTransCancelByConnId (unsigned connId, ENetError error);
...
+
+template<typename T>
+inline std::enable_if_t<HasTransId<T>::value, bool>
+NetTransRecvFromMsg(const uint8_t msg[], unsigned bytes, void*)
```
zrax, line 388: Maybe this should be named something to indicates it's a generic catch-all?  Something like `NetTransRecvFromMsgGeneric`?

Sources/Plasma/PubUtilLib/plNetGameLib/Intern.h
Hoikas: Cool! I'm still trying to get the hang of SFINAE and all these little helpers.

zrax (approved):

PR #816, opened 2021-02-09
Refactor all NucleusLib `hsTArray`s to `std::vector`

zrax:


Sources/Plasma/PubUtilLib/plInputCore/plSceneInputInterface.cpp
```diff
@@ -545,9 +544,9 @@ bool    plSceneInputInterface::MsgReceive( plMessage *msg )
     plNotifyMsg* pNMsg = plNotifyMsg::ConvertNoRef(msg);
     if (pNMsg)
     {
-        for(int x=0; x < pNMsg->GetEventCount();x++)
+        for(size_t x = 0; x < pNMsg->GetEventCount(); x++)
         {
-            proEventData* pED = pNMsg->GetEventRecord(0);
+            proEventData* pED = pNMsg->GetEventRecord(x);
```
zrax, line 549: :trollface:

Hoikas: lolwut

zrax: It's another case of attempting to iterate over a range but not actually using the iterator variable...

Sources/Plasma/NucleusLib/pnTimer/plTimerCallbackManager.cpp
```diff
@@ -122,16 +112,16 @@ bool plTimerCallbackManager::CancelCallbacksToKey(const plKey& key)
...
         {
-            const plKey rKey = fCallbacks[i]->fMsg->GetReceiver(j);
+            const plKey rKey = (*iter)->fMsg->GetReceiver(j);
             
             if (rKey == key)
             {
-                delete fCallbacks[i];
-                fCallbacks.Remove(i);
+                delete *iter;
+                fCallbacks.erase((iter + 1).base());    // Yuck
```
Hoikas, line 124: Ok, I have to admit that I'm bamboozled here. I've been staring at this line for five minutes and still don't understand 😞

Sources/Tools/MaxComponent/plGUIComponents.cpp
```diff
@@ -4442,8 +4441,8 @@ bool plGUIProgressCtrlComponent::Convert(plMaxNode *node, plErrorMsg *pErrMsg)
         // so to avoid breaking old formats, if we can't grab an animObjInterface, we just grab the key
         // of the master mod of our node, like we would've before
         plAGMasterMod   *master = node->GetAGMasterMod();
-        hsTArray<plKey> keys;
-        keys.Append( master->GetKey() );
+        std::vector<plKey> keys;
+        keys.emplace_back(master->GetKey());
         ctrl->SetAnimationKeys( keys, ENTIRE_ANIMATION_NAME );
```
Hoikas, line 4446: Could we use an initializer list here, eg

```suggestion
        ctrl->SetAnimationKeys( { master->GetKey() }, ENTIRE_ANIMATION_NAME );
```

Sources/Tools/MaxComponent/plGUIComponents.cpp
```diff
@@ -2724,8 +2723,8 @@ bool plGUIKnobCtrlComponent::Convert(plMaxNode *node, plErrorMsg *pErrMsg)
...
```
Hoikas, line 2728: Same question as below re initializer

```suggestion
        ctrl->SetAnimationKeys( { master->GetKey() }, ENTIRE_ANIMATION_NAME );
```

Sources/Plasma/NucleusLib/pnTimer/plTimerCallbackManager.cpp
zrax: Yes, this is an unfortunate ugly wart on STL...  You can't erase a `reverse_iterator`, so you have to convert it to a forward iterator...  But reverse iterators actually point to one after their "referenced" element (adjusted when dereferencing), so it needs to be moved forward one to get the correct element...   See the discussion [here](https://stackoverflow.com/questions/1830158/how-to-call-erase-with-a-reverse-iterator).

An alternate syntax would be `std::next(iter).base()`, but I'm not sure that's any more clear either.

Hoikas: Ugh. Could the comment be amended to say something to the effect of `// Ugh - can't erase a reverse iterator`? I didn't realize that myself, and it would have (hopefully) been enough of a hint to go read up.

Maybe one day I will actually read the standard... 😛

Hoikas (approved):

PR #818, opened 2021-02-09
Fix Python 3.x compatibility with a couple of inline Python strings.

zrax:
Note: I have not tested the MaxMain one, and I'm not even sure it's still relevant...


Hoikas (approved):
Whoops

PR #822, opened 2021-02-11
Don't use the QFileDialog native dialog backend in the Qt tools.

zrax:
Qt apparently has a long-standing bug that causes the native file dialog on Windows to hang indefinitely if the application initializes COM (CoInitialize[Ex]) before the dialog is shown.  Since Plasma does this for us, all file dialogs are broken with the native backend on Windows. The non-native dialog may not look as well integrated, but it works.


Hoikas:
I wonder if it would make more sense to replace the current unconditional COM init in hsCOM with a more explicit init gated behind `std::once_flag`? IIRC the only place that actually *needs* COM is `plAudioEndpointVolume` for Windows Core Audio.

zrax:
Dropped in favor of #826

PR #824, opened 2021-02-12
plProduct Improvements

Hoikas:
Now that we are uploading functioning clients to CI, we need to be able to differentiate these clients to our users. Currently, the client always identifies itself as `1.918` -- this cannot be changed because it's part of the server handshake. That means, currently, the only way a user can identify which client they are using is to do something like `sha1sum plClient.exe`. This pull request revamps `plProduct` to embed the current git tag into the product identifier and display it on the launcher. If no tag is available, a pretty version of the current SHA is used instead.

This also throws away the old Win32-specific `pnBuildDate` junk with a proper pre-build step. Care was taken to make sure that pressing build with no changes will not cause the entire engine to need to re-link :smile:.

If you do not want the git revision embedded in your client, set `-DPRODUCT_EMBED_BUILD_INFO=OFF`. This will also turn off the build time. To turn off only the build time, set `-DPRODUCT_EMBED_BUILD_TIME=OFF`.


Hoikas:
Rebased to resolve merge conflicts.

cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,112 @@
...
+                COMMAND ${GIT_EXECUTABLE} log --pretty=format:%h -n 1
+                OUTPUT_VARIABLE GIT_TAG
+                OUTPUT_STRIP_TRAILING_WHITESPACE
+                ERROR_QUIET
+            )
+        endif()
+        if(GIT_DIFF EQUAL 1)
+            string(APPEND GIT_REV "_dirty")
+            string(APPEND GIT_TAG "_dirty")
+        endif()
```
zrax, line 42: Why not just use `git describe --tags --dirty`, which automatically appends `-dirty` when the repo is dirty?

Obviously we'd need to put some more reasonable tags in the repo...

cmake/hsBuildInfo.inl.cmake
```diff
@@ -0,0 +1,20 @@
...
+
+namespace plProduct
+{
+    constexpr uint32_t PRODUCT_BRANCH_ID = @PRODUCT_BRANCH_ID@;
+    constexpr uint32_t PRODUCT_BUILD_ID = @PRODUCT_BUILD_ID@;
+    constexpr uint32_t PRODUCT_BUILD_TYPE = @PRODUCT_BUILD_TYPE@;
+
+    static ST::string PRODUCT_CORE_NAME = ST_LITERAL("@PRODUCT_CORE_NAME@");
+    static ST::string PRODUCT_SHORT_NAME = ST_LITERAL("@PRODUCT_SHORT_NAME@");
+    static ST::string PRODUCT_LONG_NAME = ST_LITERAL("@PRODUCT_LONG_NAME@");
```
zrax, line 12: Even though we can't have `constexpr` yet (😞), the strings should still be `const`

cmake/BuildInfo.cmake
Hoikas: I remember bypassing `git describe` because it was using reaaaaaly old tags. We need to do a better job of tagging, heh. I'm not sure how I feel about seeing `UruLive.1.918 - gamemgr-1437-g87f077d1c - Internal.Release` over `UruLive.1.918 - g87f077d1c - Internal.Release`. In particular, I chose the `_dirty` prefix over `-dirty` to avoid confusion with the other dash separators in the product string.

Hoikas: OK, I think I've got a workable solution such that we don't do this whole manual dirty thing and obsolete tags aren't picked up.

Sources/Plasma/CoreLib/CMakeLists.txt
```diff
@@ -83,6 +75,29 @@ set(CoreLib_HEADERS
...
+)
+add_custom_target(pcBuildInfo
+    COMMAND ${BUILD_INFO_COMMAND}
+    BYPRODUCTS "${BUILD_INFO_FILE}"
+)
+list(APPEND CoreLib_HEADERS "${BUILD_INFO_FILE}")
+
+# Always generate the hsBuildInfo.inl at configure time to ensure the file is present-and-accounted
+# for. Otherwise, people like me will be staring at the include going, "WTF mate?"
+execute_process(COMMAND ${BUILD_INFO_COMMAND} OUTPUT_QUIET WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
zrax, line 99: Eh, I'm not sure this is really necessary...  We already have cases of generated files that don't show up until you build (e.g. the Qt automoc/autouic/autorcc stuff), and this will increase the time to re-run cmake each time.

cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,104 @@
...
+if(PRODUCT_EMBED_BUILD_TIME)
+    # Include the hash of the diff to ensure that non-commit changes trigger a build-time update.
+    string(SHA256 CURRENT_DIFF_HASH "${GIT_DIFF_CONTENTS}")
+
+    if(WIN32)
+        execute_process(
+            COMMAND powershell -Command date +%Y-%m-%d
+            OUTPUT_VARIABLE BUILD_DATE
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
```
zrax, line 54:
```
Get-Date : Cannot bind parameter 'Date'. Cannot convert value "+%Y-%m-%d" to
type "System.DateTime". Error: "String was not recognized as a valid DateTime."
At line:1 char:6
+ date +%Y-%m-%d
+      ~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-Date], ParameterBindin
   gException
    + FullyQualifiedErrorId : CannotConvertArgumentNoMessage,Microsoft.PowerSh
   ell.Commands.GetDateCommand
```

cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,104 @@
...
+        execute_process(
+            COMMAND powershell -Command date +%H:%M%z
+            OUTPUT_VARIABLE BUILD_TIME
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
```
zrax, line 59:
```
Get-Date : Cannot bind parameter 'Date'. Cannot convert value "+%H:%M%z" to
type "System.DateTime". Error: "String was not recognized as a valid DateTime."
At line:1 char:6
+ date +%H:%M%z
+      ~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-Date], ParameterBindin
   gException
    + FullyQualifiedErrorId : CannotConvertArgumentNoMessage,Microsoft.PowerSh
   ell.Commands.GetDateCommand
```

cmake/BuildInfo.cmake
zrax: Perhaps you meant `Get-Date -UFormat +%Y-%m-%d`?

cmake/BuildInfo.cmake
zrax: Also, `Get-Date` apparently doesn't support `%z`, only `%Z` (which is two digits instead of four)

cmake/BuildInfo.cmake
```diff
@@ -0,0 +1,106 @@
...
+            OUTPUT_VARIABLE BUILD_DATE
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
+        execute_process(
+            COMMAND powershell -Command "Get-Date -UFormat +%H:%M%Z"
+            OUTPUT_VARIABLE BUILD_TIME
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
+        # Force UTC offset to four digits instead of two
+        string(REGEX REPLACE [[(-|\+)([0-9][0-9])$]] [[\1\200]] BUILD_TIME "${BUILD_TIME}")
```
zrax, line 61: Heh, I hope that doesn't look for substitution #200 ;)

zrax (approved):

PR #828, opened 2021-02-12
Use use-after-free issue in Minkata.

Hoikas:
When you touch the linking stone in Minkata, it pages the ground out from under you. Because we were not using the ground's smart pointer, we happily dereferenced freed memory (boom). Amusingly, the ground seems to be freed between `IFindGroundCandidate()` and actually using the ground information. At one point, I tried to properly ref and deref the physical object, but that added a lot of bp considering our use of `std::optional<plControllerHitRecord>`.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plPhysicalControllerCore.h
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plPhysicalControllerCore.h
@@ -69,7 +69,7 @@ struct plControllerHitRecord
     float Displacement;
 
     plControllerHitRecord()
-        : PhysHit(), Point(), Normal(), Displacement()
+        : Point(), Normal(), Displacement()
     { }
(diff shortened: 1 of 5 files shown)


Sources/Plasma/PubUtilLib/plAvatar/plPhysicalControllerCore.h
```diff
@@ -60,17 +63,25 @@ class plSceneObject;
...
     hsPoint3 Point;
     hsVector3 Normal;
     float Displacement;
 
     plControllerHitRecord()
-        : ObjHit(), Point(), Normal(), Displacement()
-    { }
-    plControllerHitRecord(plPhysical* phys, hsPoint3 p, hsVector3 n, float d=0.f)
-        : ObjHit(phys), Point(std::move(p)), Normal(std::move(n)), Displacement(d)
+        : PhysHit(), Point(), Normal(), Displacement()
```
zrax, line 72: Don't plKeys have a default constructor?

zrax (approved):

PR #829, opened 2021-02-13
Throw away more CMake Cruft.

Hoikas:
This finishes the work of making our find modules target based and lays the groundwork for building 64-bit 3ds Max plugins. Note that the option to build the max plugin has changed to `-DPLASMA_BUILD_MAX_PLUGIN=ON` to be consistent with the rest of the already available options.


dpogue:
I have a slight preference to keep the 3DSM and VLD `find_package` calls guarded by `if(WIN32)`, otherwise running a build on Linux will always print out that various packages are missing.
(Also, put the Security `find_package` in a `if(APPLE)` call for the same reason)

CMakeLists.txt
```diff
@@ -107,7 +104,8 @@ if(PLASMA_PIPELINE STREQUAL "OpenGL")
     add_definitions(-DPLASMA_PIPELINE_GL)
 endif(PLASMA_PIPELINE STREQUAL "OpenGL")
 
-#Allow us to disable certain parts of the build
+# Allow us to disable certain parts of the build
+cmake_dependent_option(PLASMA_BUILD_MAX "Do we want to build the 3ds Max plugin?" OFF "TARGET 3dsm" OFF)
```
zrax, line 108: `PLASMA_BUILD_MAX_PLUGIN`?  Longer, but more clear, IMO

cmake/Find3dsm.cmake
```diff
@@ -1,50 +1,37 @@
...
+        list(APPEND _TEST_FILES "x64/lib/gup.lib")
+    else()
+        set(_LIBDIR "lib")
+    endif()
 
-find_path(3dsm_INCLUDE_DIR maxapi.h
-    ${3dsm_PATH}/include
-)
+    find_path(3dsm_PATH NAMES ${_TEST_FILES}
+        PATHS "C:/3dsmax7" ENV ADSK_3DSMAX_SDK_2012
```
zrax, line 14: Is this PATHS line really needed?  What are the odds that someone will be using Max 7 AND have it installed at `C:\3dsmax7`?  Also, for the `ENV` line, do we need additional entries for other (marginally) supported versions?  I'd probably just remove this whole line, personally.

Hoikas: `C:\3dsmax7` is the default install location for 3dsmax7 (yikes) - and yeah, it looks like each version would have its own `ENV` entry 😞

zrax (approved):

PR #834, opened 2021-02-23
Cleanup some unused / unneeded code in pnAsyncCoreExe

zrax:
* Port AsyncThread's backend to `std::thread`.
* Remove several Async APIs that aren't used by any of the client code.

This greatly reduces the surface of what needs to be ported or replaced for cross-platform support.


Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.cpp
```diff
@@ -47,6 +47,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plNetClientComm.h"
 
+#include <thread>
+#include <chrono>
```
Hoikas, line 51:
```suggestion
#include <chrono>
#include <thread>
```

dpogue:
```diff
-        AsyncSleep(10);
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
```

but does it still sleep asynchonously? :trollface:

zrax:
> but does it still sleep asynchonously? :trollface:

If you read the commit message for 391e10d57eebb3a6f0fb13c583fa4681347b84d2, you would see the answer to that question :trollface:

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
```diff
@@ -50,6 +50,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #endif
 #define PLASMA20_SOURCES_PLASMA_NUCLEUSLIB_PNASYNCCORE_PRIVATE_PNACTHREAD_H
 
+#include <thread>
```
Hoikas, line 53: This should also be included in the `Pch.h`, I think.

Hoikas (approved):

PR #835, opened 2021-02-24
Port Async DNS resolution to ASIO

zrax:
This is the first part of a goal to make the async netcode cross-platform using Standalone ASIO.

* ~Depends on #834.  (#834 should be merged first, since this is build on top of it to avoid conflicts)~
* Supersedes #376.


Hoikas:
Closes #376.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+static void AddressResolved(const asio::error_code& err,
+                            const tcp::resolver::results_type& results,
+                            const DnsResolveData& data)
+{
+    if (err || results.empty()) {
+        if (err)
+            LogMsg(kLogFatal, "DNS: Failed to resolve {}: {}", data.m_name, err.message());
```
dpogue, line 95: Can we move towards using `plNetApp::StaticErrorMessage` instead of pnAsyncCore's `LogMsg`? I'd like to be able to get rid of `LogMsg` entirely to break the pnAsyncCore dependency on plStatusLog

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,155 @@
...
+    PerfSubCounter(kAsyncPerfNameLookupAttemptsCurr, 1);
+}
+
+void AsyncAddressLookupName(FAsyncLookupProc lookupProc,
+                            const ST::string& name, unsigned port, void* param)
+{
+    ASSERT(lookupProc);
+
+    PerfAddCounter(kAsyncPerfNameLookupAttemptsCurr, 1);
+    PerfAddCounter(kAsyncPerfNameLookupAttemptsTotal, 1);
```
dpogue, line 125: Is it possible to use plProfile for these counters? Maybe not without making plProfile thread-safe and atomic?

zrax: I used these because they were already in place from the old DNS code...  As you mentioned, plProfile is not atomic, which would be a problem for these counters with the current design, so I feel like that sort of change belongs in a separate PR.

Hoikas: I'd prefer for us to avoid making `plProfile` atomic if at all possible.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
Hoikas: One thing to consider with these log messages is that any `plStatusLog`-backed logger (including `plNetApp::Static*Message` is that they block. Originally, `LogMsg` did not block, so we'll want to be mindful of that.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+*==LICENSE==*/
+
+#include "Pch.h"
+
+using tcp = asio::ip::tcp;
+
+struct DnsResolver
+{
+    asio::io_context m_context;
```
Hoikas, line 49: Does this mean we are abandoning the `f` (field) prefix for `m_` (member)?

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,155 @@
...
+    asio::executor_work_guard<asio::io_context::executor_type> m_workGuard;
+    tcp::resolver m_resolver;
+    std::thread m_lookupThread;
+
+    DnsResolver()
+        : m_context(), m_resolver(m_context), m_workGuard(m_context.get_executor())
+    {
+        // Start the resolver thread
+        m_lookupThread = std::thread([this] { m_context.run(); });
```
Hoikas, line 58: Needs to enable VLD.

cmake/FindASIO.cmake
```diff
@@ -0,0 +1,33 @@
...
+if(ASIO_INCLUDE_DIR AND EXISTS "${ASIO_INCLUDE_DIR}/asio/version.hpp")
+    file(STRINGS "${ASIO_INCLUDE_DIR}/asio/version.hpp" _asio_version_line
+         REGEX "^#[ \t]*define[ \t]+ASIO_VERSION[ \t]+([0-9]+).*")
+    string(REGEX REPLACE "^.*ASIO_VERSION[ \t]+([0-9]+).*$" "\\1"
+           _asio_version "${_asio_version_line}")
+    math(EXPR _asio_major "${_asio_version} / 100000")
+    math(EXPR _asio_minor "(${_asio_version} / 100) % 1000")
+    math(EXPR _asio_patch "${_asio_version} % 100")
+    set(ASIO_VERSION "${_asio_major}.${_asio_minor}.${_asio_patch}")
+endif()
```
Hoikas, line 16: Wow that's complicated... And unused except for probably a log message.

cmake/FindASIO.cmake
```diff
@@ -0,0 +1,33 @@
...
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(ASIO
+    REQUIRED_VARS ASIO_INCLUDE_DIR
+    VERSION_VAR ASIO_VERSION
+)
+
+if(ASIO_FOUND AND NOT TARGET ASIO::Standalone)
+    find_package(Threads REQUIRED)
+
+    add_library(ASIO::Standalone INTERFACE IMPORTED)
```
Hoikas, line 27: I have a slight preference for `ASIO::ASIO` because this implies the library's name is `Standalone` and it better matches what namespaced cmake exports look like in general.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,155 @@
...
+        m_resolver.cancel();
+        AsyncThreadTimedJoin(m_lookupThread, exitThreadWaitMs);
+        m_lookupThread = {};
+    }
+};
+
+struct DnsResolveData
+{
+    ST::string          m_name;
+    FAsyncLookupProc    m_lookupProc;
```
Hoikas, line 73: I would really like to see these C function pointers go away in favor of `std::function` so that we can use stateful lambdas in the higher level code. I'd be ok with deferring that to another PR, but I think that could be done fairly transparently.

Hoikas (commented):
Another general observation - IMO, it would be nice to see pnAsyncCore and pnAsyncCoreExe be merged into a single project. Especially without the rather goofy PCH configuration that causes the whole project to rebuild when a single project header changes. That's probably outside the scope of this PR, but it's on my wishlist 😄.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
zrax: Alas, linking in plNetApp also requires some creatable registration which is unavailable in plUruLauncher...  So this cannot be done yet.

cmake/FindASIO.cmake
zrax: Thanks, that reminds me I forgot to add a minimum version check ;)

Hoikas: This explains much 😃

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
zrax: Blargh, habits...

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
Hoikas: IMO, it would be better to demote plStatusLog to pnStatusLog if we're going to be using it down here in NucleusLand.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+struct DnsResolveData
+{
+    ST::string          fName;
+    FAsyncLookupProc    fLookupProc;
+    void*               fParam;
+};
+
+static std::recursive_mutex     s_critsect;
+static DnsResolver             *s_resolver = nullptr;
```
Hoikas, line 83:
```suggestion
static std::recursive_mutex     s_critsect;
static DnsResolver*             s_resolver = nullptr;
```

Hoikas (approved):

PR #837, opened 2021-02-24
Port pnAceTimer to ASIO

zrax:
Also removes the unimplemented Unix backend stubs for AsyncCoreExe, since that will eventually become unnecessary.


zrax:
Note:  It's more useful to look at pnAceTimer.cpp as if it were a new file rather than a diff against the old version, since that's how it was developed.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceTimer.cpp
```diff
@@ -39,321 +39,144 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    FAsyncTimerProc     fTimerProc;
+    FAsyncTimerProc     fDestroyProc;
+    void*               fParam;
 
-/****************************************************************************
-*
-*   Private
-*
-***/
+    AsyncTimer(asio::io_context& context, FAsyncTimerProc &&timerProc, void* param)
```
Hoikas, line 52:
```suggestion
    AsyncTimer(asio::io_context& context, FAsyncTimerProc&& timerProc, void* param)
```

Hoikas (commented):
I would leave this comment in line, but it's basically impossible to comment on the diff, heh. It looks like there is a potential crash in `TimerDestroy()` in that we start looping over the timers in a range-based loop without locking the timer mutex.

zrax:
Not locking the mutex was intentional, since we already stopped the thread...  However, it is indeed possible the thread didn't stop before the timeout, so I added a commit to terminate the event loop, which should ensure no more tasks are run on the thread (making the termination loop safe).  This was handled by the old code by closing the event object that wakes the thread, but in both cases, it is expected that the callbacks are executed with the mutex unlocked.

Hoikas (approved):

PR #838, opened 2021-02-25
Clean up hsBitVector and add some unit tests

dpogue:
* Fixes a bug where `ToggleBit` didn't actually work if the bitvector was empty
* Fixes some signed/unsigned integer comparison issues (although some still exist in hsBitIterator)
* Adds some basic unit tests because tests are good and they uncover bugs (see point 1)

### DIFF
--- a/Sources/Plasma/CoreLib/hsBitVector.cpp
+++ b/Sources/Plasma/CoreLib/hsBitVector.cpp
@@ -50,31 +50,33 @@ void hsBitVector::IGrow(uint32_t newNumBitVectors)
     hsAssert(newNumBitVectors > fNumBitVectors, "Growing smaller");
     uint32_t *old = fBitVectors;
     fBitVectors = new uint32_t[newNumBitVectors];
-    int i;
-    for( i = 0; i < fNumBitVectors; i++ )
+
+    uint32_t i;
+    for (i = 0; i < fNumBitVectors; i++)
         fBitVectors[i] = old[i];
-    for( ; i < newNumBitVectors; i++ )
+    for (; i < newNumBitVectors; i++)
         fBitVectors[i] = 0;
-    delete [] old;
+    delete[] old;
     fNumBitVectors = newNumBitVectors;
 }
 
 hsBitVector& hsBitVector::Compact()
 {
-    if( !fBitVectors )
+    if (!fBitVectors)
         return *this;
 
-    if( fBitVectors[fNumBitVectors-1] )
+    if (fBitVectors[fNumBitVectors-1])
         return *this;
 
-    int hiVec = 0;
-    for( hiVec = fNumBitVectors-1; (hiVec >= 0)&& !fBitVectors[hiVec]; --hiVec );
-    if( hiVec >= 0 )
+    uint32_t hiVec = 0;
+    for (hiVec = fNumBitVectors-1; (int32_t(hiVec) >= 0) && !fBitVectors[hiVec]; --hiVec)
+        ;
+
+    if (hiVec >= 0)
     {
         uint32_t *old = fBitVectors;
         fBitVectors = new uint32_t[++hiVec];
-        int i;
-        for( i = 0; i < hiVec; i++ )
+        for (uint32_t i = 0; i < hiVec; i++)
             fBitVectors[i] = old[i];
         fNumBitVectors = hiVec;
         delete [] old;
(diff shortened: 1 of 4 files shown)


Sources/Tests/CoreTests/test_hsBitVector.cpp
```diff
@@ -0,0 +1,258 @@
...
+TEST(hsBitVector, bit_setting)
+{
+    hsBitVector bv;
+
+    bv.SetBit(1, true);
+
+    EXPECT_EQ(bv.Empty(), false);
+    EXPECT_EQ(bv.GetSize(), 32);
+    EXPECT_EQ(bv.GetNumBitVectors(), 1);
+    EXPECT_NE(bv.GetBitVector(0), 0);
```
zrax, line 67: Why not `EXPECT_EQ(bg.GetBitVector(0), 2);` ?

Sources/Tests/CoreTests/test_hsBitVector.cpp
```diff
@@ -0,0 +1,258 @@
...
+{
+    hsBitVector bv;
+
+    bv.SetBit(48, true);
+
+    EXPECT_EQ(bv.Empty(), false);
+    EXPECT_EQ(bv.GetSize(), 64);
+    EXPECT_EQ(bv.GetNumBitVectors(), 2);
+    EXPECT_EQ(bv.GetBitVector(0), 0);
+    EXPECT_NE(bv.GetBitVector(1), 0);
```
zrax, line 85: Why not `EXPECT_EQ(bv.GetBitVector(1), 0x10000);` ?

Sources/Plasma/CoreLib/hsBitVector.cpp
```diff
@@ -50,31 +50,33 @@ void hsBitVector::IGrow(uint32_t newNumBitVectors)
...
     fBitVectors = new uint32_t[newNumBitVectors];
-    int i;
-    for( i = 0; i < fNumBitVectors; i++ )
+
+    uint32_t i;
+    for (i = 0; i < fNumBitVectors; i++)
         fBitVectors[i] = old[i];
-    for( ; i < newNumBitVectors; i++ )
+    for (; i < newNumBitVectors; i++)
         fBitVectors[i] = 0;
```
zrax, line 58: These two loops could be replaced by a `memcpy` and `memset`...

Sources/Plasma/CoreLib/hsBitVector.cpp
```diff
@@ -50,31 +50,33 @@ void hsBitVector::IGrow(uint32_t newNumBitVectors)
...
         return *this;
 
-    int hiVec = 0;
-    for( hiVec = fNumBitVectors-1; (hiVec >= 0)&& !fBitVectors[hiVec]; --hiVec );
-    if( hiVec >= 0 )
+    uint32_t hiVec = 0;
+    for (hiVec = fNumBitVectors-1; (int32_t(hiVec) >= 0) && !fBitVectors[hiVec]; --hiVec)
+        ;
+
+    if (hiVec >= 0)
```
zrax, line 75: `hiVec` should be `int32_t` -- despite initializing it with a `uint32_t - 1`, it is compared with 0 in two places (the latter without a cast, which will always be `true`)

zrax (changes requested):
Additional thoughts:
* It seems like `hsBitVector` and `hsBitIterator` could be simplified (and the latter possibly removed) with a port to `std::vector` for storage instead of manually bookkeeping the array...  That's a bigger refactor though, so I'd be fine with deferring that to another PR.
* The unit test should probably test the edge cases (bit 31 and bit 32) directly.

Sources/Plasma/CoreLib/hsBitVector.cpp
dpogue: Thinking about this more, the core issue seems to be that limits defined/enforced for what the maximum value in a bit vector is. At one point in `Enumerate`, it tries to store all the contained bits in a `vector<uint16_t>`. It's clearly possible to store bits that exceed the capacity of a 16-bit unsigned integer, and even that exceed a 32-bit signed integer.

Should we start enforcing that we only support storing up to `uint16_t` max value, which would make it entirely safe to use a `int32_t` for the number of bit vectors and allow us to use -1 with no jumping between signedness?

zrax: Worse, `Enumerate` populates a vector of `int16_t`, even though it will never use negative indices...  I'm fine with enforcing a maximum size of 2^15 bits, but I don't necessarily feel that `fNumBitVectors` needs to become signed.  For reference, I already use `hsSsize_t` for backwards iteration in several places with `std::vector`s, with the implicit assumption that those vectors will never be greater than 2^31 (or 2^63 on 64-bit) elements.  If we have that many of anything in Plasma, Something Terribly has Gone Wrong ;)

Hoikas:
Seeing as how this has been maturing for over a year, what's our current status on this?

dpogue:
> what's our current status on this?

I've not followed up on this, and I see there are conflicts, so let's close and it can be rebased and opened as a new PR if we want to revive it in the future. Maybe flag it as a good Hacktoberfest project, because hsBitVector should definitely be testable.

PR #841, opened 2021-02-26
Clean up NT socket code

zrax:
This does NOT port AsyncCoreExe sockets to ASIO yet...  But it does remove large swaths of unused code, as well as flattening the Async API dispatch table into direct calls.  More importantly, this means we can now build pnAsyncCoreExe on Linux and macOS (even though they're useless without the implementation of most of the socket code, which will come with the ASIO port later).


Sources/Plasma/NucleusLib/pnAsyncCoreExe/CMakeLists.txt
```diff
@@ -28,12 +23,16 @@ set(pnAsyncCoreExe_PRIVATE_WIN32
...
-            ${pnAsyncCoreExe_PRIVATE} ${pnAsyncCoreExe_PRIVATE_NT}
-            ${pnAsyncCoreExe_PRIVATE_WIN32})
+            ${pnAsyncCoreExe_PRIVATE})
+
+if(WIN32)
+    target_sources(pnAsyncCoreExe PRIVATE
+        ${pnAsyncCoreExe_PRIVATE_NT}
+        ${pnAsyncCoreExe_PRIVATE_WIN32}
+    )
+endif()
```
Hoikas, line 35: This could also be a generator expression in the main `add_library()` call:
```cmake
add_library(foobar STATIC $<$<PLATFORM_ID:Windows>:${WINDOWS_SOURCES}>)
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
```diff
@@ -56,28 +56,21 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 #include "pnUtils/pnUtils.h"
 #include "pnNetBase/pnNetBase.h"
 #include "pnAsyncCore/pnAsyncCore.h"
 
 #include "Private/pnAceInt.h"
-#include "Private/Nt/pnAceNt.h"
 
+// Target Windows 7 or later for windows socket compatibility
+// ASIO needs to know this to warn about some API incompatibilities
+#define _WIN32_WINNT 0x0601
```
Hoikas, line 68: Redefining `_WIN32_WINNT` outside of hsWindows.h, eh? I see you like living dangerously.-- This macro is defined as 0x501 in hsWindows.h. I would just bump it there and move along with life.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
```diff
@@ -56,28 +56,21 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plProduct.h"
 #include "hsThread.h"
-#include "hsWindows.h"
```
Hoikas, line 59: I would leave this in, considering it's still being included in pnAceW32Thread

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNtSocket.cpp
```diff
@@ -255,46 +220,10 @@ static bool SocketDispatchRead (NtSock * sock) {
...
-            return false;
-    }
-
-    // add bytes used by IOsFindListenProc and kNotifySocketListenSuccess
-    sock->opRead.read.bytesProcessed += bytesProcessed;
-    return true;
+    // Probably unneeded since we just tested for nullptr above, but it
+    // maintains compatibility with what we had before...
+    sock->notifyProc = nullptr;
+    return false;
```
Hoikas, line 226: I don't see where the deleted code nulls out the notifyProc? This looks a bit iffy IMO.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/CMakeLists.txt
zrax: Eh, it could...  But I don't really see the benefit of a generator expression in this case, especially since this whole block will eventually go away (including the Win32Thread bit).

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
zrax: Is there any actual need to keep it consistent?  Also, `hsWindows.h` does NOT set `_WIN32_WINNT` except on MinGW; otherwise it gets the default `windows.h` value of `0x0A00` (Windows 10) with the current VC2017 and 2019 SDKs

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
zrax: Which is why I moved the include down to `pnAceW32Thread` instead of here globally...

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNtSocket.cpp
zrax: Yeah, this one definitely needs some more clarification, since there were several steps involved in factoring it out...  Essentially:

* `s_notifyProcs` in pnAceIo.cpp was only ever read, not written to, meaning...
* `AsyncSocketFindNotifyProc` would never find an entry in `s_notifyProcs`, meaning...
* Whether `GetConnHash` fails or we look up an entry in `s_notifyProcs`, the return value of `AsyncSocketFindNotifyProc` is always `nullptr`, meaning...
* The call to `AsyncSocketFindNotifyProc` can be simplified to `nullptr`.

So you say, "But the call also had side effects!"...  Well, the side effects (setting values in `notify` and `bytesProcessed`) are only used if `sock->notifyProc` is non-null, which we've already proven is not the case.  Therefore, we always take the shortcut (`return false`) and the rest of the function is not used, meaning that even the call to `AsyncSocketFindNotifyProc` can be eliminated.

zrax:
Added some additional cleanup around AsyncThreads

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNtSocket.cpp
Hoikas: OK, I think I understand. From my understanding, this was probably used on the server to register notify procedures for each connection type.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNt.cpp
```diff
@@ -312,28 +267,15 @@ void NtInitialize () {
...
-    if (!s_pageSizeMask) {
-        SYSTEM_INFO si;
-        GetSystemInfo(&si);
-        s_pageSizeMask = si.dwPageSize - 1;
-
+    if (!s_ioThreadCount) {
         // Set worker thread count
-        s_ioThreadCount = si.dwNumberOfProcessors * 2;
+        s_ioThreadCount = std::thread::hardware_concurrency() * 2;
+        ASSERT(s_ioThreadCount);  // The standard indicates the above can return zero...
```
Hoikas, line 278: O.o - I'd be tempted to `s_ioThreadCount = std::max(1, std::thread::hardware_concurrency() * 2);`

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNt.cpp
```diff
@@ -162,23 +127,13 @@ static void INtOpDispatch (
...
                 Operation * next = ntObj->opList.Next(op);
                 ntObj->opList.Delete(op);
                 op = next;
 
-                // set event *after* operation is complete
-                if (signalComplete) {
-                    signalComplete->SignalObject();
-                    signalComplete->UnRef();
-                }
-
```
Hoikas, line 181: I'm having a trouble understanding how we can remove this wait, mostly because I'm not 100% on why it's here in the first place. That being said, it looks like only `kOpSocketWrite` will get here (a synchronous-ish write) but any `kOpQueuedSocketWrite` posts a `kOpSocketWrite` in `INtSocketOpCompleteQueuedSocketWrite()`. The event seems to only be waited on when the operation is destroyed, though.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
Hoikas: This project is goofy in that this is both the precompiled header and the global private header. What a mess. 😞

Sources/Plasma/NucleusLib/pnAsyncCoreExe/Private/Nt/pnAceNt.cpp
zrax: We're not actually removing the wait, we're removing the `signalComplete` member from `op` entirely, since it was never set to anything other than `nullptr`.  Since it was never set, the local variable would always be `nullptr` too, meaning this if statement was always false.

Hoikas (approved):

PR #842, opened 2021-02-27
Add Serene to MOULa

DoobesURU:
AGE, FNI, PY and SDL files for the Age of Serene for MOULa.

### DIFF
--- a/Scripts/Python/Serene.py
+++ b/Scripts/Python/Serene.py
@@ -1,0 +1,79 @@
+# /*==LICENSE==*
+#
+# CyanWorlds.com Engine - MMOG client, server and tools
+# Copyright (C) 2011  Cyan Worlds, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


Scripts/Python/Serene.py
```diff
@@ -0,0 +1,79 @@
...
+            return None
+        
+        dnitime = PtGetDniTime()
+        month = time.strftime('%m', time.gmtime(dnitime))
+        day = time.strftime('%d', time.gmtime(dnitime))
+        ageSDL = PtGetAgeSDL()
+        if int(month) == 12 and int(day) > 13:
+            ageSDL['sereneChristmasVis'] = (1,)
+        else:
+            ageSDL['sereneChristmasVis'] = (0,)
```
Hoikas, line 71: `sereneChristmasVis` is marked has coming from the VAULT in the SDL file, but this modifies the AgeSDLHook blob. I wonder if it would be better to just remove this and let the shard admin handle it for now.

PR #843, opened 2021-02-27
Kill some C-string spin-washing.

Hoikas:
This eradicates some spin-washing of strings from `ST::string` to `char*` back to `ST::string` by removing some uses of `hsStrcpy` in favor of `ST::string` itself. Note this does not remove all uses of `hsStrcpy`.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -310,7 +310,7 @@ bool    pfConsole::MsgReceive( plMessage *msg )
     }
 
     plConsoleMsg *cmd = plConsoleMsg::ConvertNoRef( msg );
-    if (cmd != nullptr && cmd->GetString() != nullptr)
+    if (cmd != nullptr && !cmd->GetString().empty())
     {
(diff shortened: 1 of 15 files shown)


Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -332,10 +332,10 @@ bool    pfConsole::MsgReceive( plMessage *msg )
             }
         }
         else if( cmd->GetCmd() == plConsoleMsg::kAddLine )
-            IAddParagraph( cmd->GetString() );
+            IAddParagraph( cmd->GetString().c_str() );
         else if( cmd->GetCmd() == plConsoleMsg::kExecuteLine )
         {
-            if( !fEngine->RunCommand( (char *)cmd->GetString(), IAddLineCallback ) )
+            if( !fEngine->RunCommand( const_cast<char*>(cmd->GetString().c_str()), IAddLineCallback ) )
```
zrax, line 338: The console will (destructively) `strtok` this, which violate's `ST::string`'s immutability...  It would be better to use an `ST::char_buffer`

zrax (approved):

PR #844, opened 2021-02-27
More Build Time Improvements

Hoikas:
This is another batch of build improvements by enabling unity builds and adding precompiled headers. With this round, I used `vcperf` to profile what exactly MSVC was doing. According to profiling, the parsing of `st_string.h` and `st_formatter.h` are what we are spending the most time on, so the focus was primarily on adding precompiled headers including those files.

Further profiling indicates the following future work is available: 
- `windows.h` is leaking into `plDrawable` and `pfPython` via `pnInputCore/plKeyDef.h`
- Most of `PubUtilLib` is building serially due to dependency spaghetti. The average build concurrency is 1.5 parallel targets, according to my profiling.

As a bonus, enabling unity builds on `plSurface` exposed that the creatable `plLayerMultiply` was never registered nor did it ever have a class index. This has been corrected to fix linker errors.

On my machine, the following improvements have been measured:
- Total: 8m 14s -> 5m 28s
- plSurface: 9.981s -> 1.878s
- plSDL: 7.168s -> 2.422s
- plPhysX: 10.683s -> 2.577s
- plGImage: 11.651s -> 2.237s
- plDrawable: 38.684s -> 4.226s

The above list is not intended to be comprehensive of the proposed changes. The full list of [build timings before and after is attached](https://github.com/H-uru/Plasma/files/6055872/times.txt). They were generated using [this extension](https://github.com/VioletGiraffe/BuildMonitor).


Sources/Plasma/PubUtilLib/plDrawable/plDynaFootMgr.cpp
```diff
@@ -41,30 +41,28 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 #include "plAvatar/plAvBrainHuman.h"
+#include "plMessage/plAvatarFootMsg.h"
+#include "plMessage/plDynaDecalEnableMsg.h"
 
-#include "pnEncryption/plRandom.h"
 static plRandom sRand;
 
-static const uint32_t kNumPrintIDs = 2;
-static const uint32_t kPrintIDs[kNumPrintIDs] =
+constexpr std::array<uint32_t, 2> kFootPrintIDs
```
zrax, line 65:
```suggestion
static constexpr uint32_t kFootPrintIDs[] =
```

(Instead of hard-coding the size)

Sources/Plasma/PubUtilLib/plDrawable/plDynaFootMgr.cpp
```diff
@@ -81,10 +79,9 @@ int plDynaFootMgr::INewDecal()
 
 plDynaFootMgr::plDynaFootMgr()
 {
-    fPartIDs.SetCount(kNumPrintIDs);
-    int i;
-    for( i = 0; i < kNumPrintIDs; i++ )
-        fPartIDs[i] = kPrintIDs[i];
+    fPartIDs.SetCount(kFootPrintIDs.size());
+    for (size_t i = 0; i < kFootPrintIDs.size(); i++)
```
zrax, line 83:
```suggestion
    fPartIDs.SetCount(std::size(kFootPrintIDs));
    for (size_t i = 0; i < std::size(kFootPrintIDs); i++)
```
(So the hard-coded size isn't needed in `std::array`)

And yes, this will eventually become simpler when everything is converted to `std::vector` >.>

Sources/Plasma/PubUtilLib/plDrawable/plDynaPuddleMgr.cpp
```diff
@@ -41,22 +41,20 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#include "plDynaPuddleMgr.h"
+#include "plPrintShape.h"
 
 #include "plAvatar/plAvBrainHuman.h"
 #include "plAvatar/plArmatureMod.h"
+#include "plMessage/plAvatarFootMsg.h"
 
-static const uint32_t kNumPrintIDs = 2;
-static const uint32_t kPrintIDs[kNumPrintIDs] =
+constexpr std::array<uint32_t, 2> kPuddlePrintIDs
```
zrax, line 57:
```suggestion
static constexpr uint32_t kPuddlePrintIDs[] =
```

Sources/Plasma/PubUtilLib/plDrawable/plDynaPuddleMgr.cpp
```diff
@@ -65,10 +63,9 @@ static const uint32_t kPrintIDs[kNumPrintIDs] =
 
 plDynaPuddleMgr::plDynaPuddleMgr()
 {
-    fPartIDs.SetCount(kNumPrintIDs);
-    int i;
-    for( i = 0; i < kNumPrintIDs; i++ )
-        fPartIDs[i] = kPrintIDs[i];
+    fPartIDs.SetCount(kPuddlePrintIDs.size());
+    for (size_t i = 0; i < kPuddlePrintIDs.size(); i++)
```
zrax, line 67:
```suggestion
    fPartIDs.SetCount(std::size(kPuddlePrintIDs));
    for (size_t i = 0; i < std::size(kPuddlePrintIDs); i++)
```

Sources/Plasma/PubUtilLib/plDrawable/plDynaRippleMgr.cpp
```diff
@@ -41,33 +41,28 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
-#include "plTweak.h"
+#include "plAvatar/plArmatureMod.h"
+#include "plAvatar/plAvBrainHuman.h"
+#include "plMessage/plAvatarMsg.h"
+#include "plMessage/plDynaDecalEnableMsg.h"
+#include "plMessage/plRippleShapeMsg.h"
 
-static const uint32_t kNumPrintIDs = 5;
-static const uint32_t kPrintIDs[kNumPrintIDs] =
+constexpr std::array<uint32_t, 5> kRipplePrintIDs
```
zrax, line 65:
```suggestion
static constexpr uint32_t kRipplePrintIDs[] =
```

Sources/Plasma/PubUtilLib/plDrawable/plDynaRippleMgr.cpp
```diff
@@ -105,10 +100,9 @@ plDynaRippleMgr::plDynaRippleMgr()
     fInitUVW(1.f,1.f,1.f),
     fFinalUVW(1.f,1.f,1.f)
 {
-    fPartIDs.SetCount(kNumPrintIDs);
-    int i;
-    for( i = 0; i < kNumPrintIDs; i++ )
-        fPartIDs[i] = kPrintIDs[i];
+    fPartIDs.SetCount(kRipplePrintIDs.size());
+    for (size_t i = 0; i < kRipplePrintIDs.size(); i++)
```
zrax, line 104:
```suggestion
    fPartIDs.SetCount(std::size(kRipplePrintIDs));
    for (size_t i = 0; i < std::size(kRipplePrintIDs); i++)
```

Sources/Plasma/PubUtilLib/plDrawable/plDynaTorpedoMgr.cpp
```diff
@@ -56,13 +56,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 #include "pnEncryption/plRandom.h"
 
-static const uint32_t kNumPrintIDs = 0;
-
-static plRandom sRand;
-
 plDynaTorpedoMgr::plDynaTorpedoMgr()
 {
-    fPartIDs.SetCount(kNumPrintIDs);
+    fPartIDs.SetCount(0);
```
zrax, line 61: What's the advantage of using a literal zero here instead of a constant?  More specifically, if this should be empty, why not remove this line entirely since the default state of an `hsTArray` should be empty.

Sources/Plasma/NucleusLib/inc/plCreatableIndex.h
```diff
@@ -143,7 +143,7 @@ CLASS_INDEX_LIST_START
     CLASS_INDEX(plNetClientApp),
     CLASS_INDEX(plNetServerApp),
     CLASS_INDEX(plClient),
-    CLASS_INDEX(UNUSED_plCompoundTMModifier),
+    CLASS_INDEX(plLayerMultiply),
```
zrax, line 146: If this class was never registered, do we actually need it?

Sources/Plasma/PubUtilLib/plDrawable/plDynaFootMgr.cpp
Hoikas: I figured you'd be able to make use of a constexpr array when vector-izing all of this ^_^

Sources/Plasma/NucleusLib/inc/plCreatableIndex.h
Hoikas: Good question. I'm willing to remove this commit in favor of deleting the class, if you prefer.

Sources/Plasma/PubUtilLib/plSurface/CMakeLists.txt
```diff
@@ -19,7 +18,6 @@ set(plSurface_HEADERS
     plLayerAnimation.h
     plLayerDepth.h
     plLayerInterface.h
-    plLayerMultiply.h
```
zrax, line 22: This file wasn't deleted?

Sources/Plasma/NucleusLib/inc/plCreatableIndex.h
zrax: I'm fine with keeping it if you think it's useful...  We'd have to adjust other tools to make use of it though, if it's never been used before.

Hoikas: I'm not sure it is useful... In a modern, shader based system these layers are just a legacy data source. :/

Sources/Plasma/PubUtilLib/plSurface/CMakeLists.txt
Hoikas: It doesn't [seem to exist](https://github.com/Hoikas/Plasma/tree/vcperf/Sources/Plasma/PubUtilLib/plSurface/plLayerMultiply.h)?

Hoikas: Looks like the deletion is shown in the commit view for 52f736d61e9a4c714d8820b49a650a5eb3a7985f

zrax: I see, git seems to think you renamed it pfConditional/Pch.h  :facepalm:

zrax (approved):

Sources/Plasma/PubUtilLib/plSurface/CMakeLists.txt
Deledrius: I am suspicious of those being the same file... :trollface:

Hoikas: 🚬

PR #846, opened 2021-02-28
De-vendor libwebm.

Hoikas:
#831 spent some time refactoring code that we vendored from libwebm. IMO, our time is better spent focusing on Plasma itself, so this de-vendors that dependency.


Deledrius:
Hallelujah!

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -72,7 +72,9 @@ class plMoviePlayer
     class plMipmap* fTexture;
 
     mkvparser::MkvReader* fReader;
+#ifdef USE_WEBM
```
zrax, line 75: Why not put fReader in here too?

Hoikas: It's only ever dereferenced if `USE_WEBM`, but it is unconditionally initialized. `std::unique_ptr`, OTOH, doesn't like deleting an forward-declared type whose value is null.

zrax: Yeah, I get why `fSegment` is protected...  But my point is that every other use of `fReader` other than the initializer is already protected, so why not fix the initializer too?

Hoikas: True. Another compelling solution would be to just get rid of all of this `USE_WEBM` stuff -- back when this was written, building libvpx was nontrivial, but vcpkg makes it trivial, if time consuming.

Deledrius (commented):
Missing a description via `set_package_properties` in [cmake/Dependencies.cmake](https://github.com/H-uru/Plasma/blob/master/cmake/Dependencies.cmake).

PR #847, opened 2021-03-01
Add some missing includes required when PCH/unity builds are disabled.

zrax:
Coverity builds seem to work again with the latest cov-build, but PCH and unity builds cause the build output to be about 10x larger (and slower to build)...  These changes fix the build with those features disabled.


Hoikas:
Looks like git wasn't configured correctly 😉

zrax:
Dammit...  Fixed

Hoikas (approved):

PR #849, opened 2021-03-02
Don't leak the TnLEnum.

Hoikas:
Who wrote this junk? *Checks git blame...* Moving right along...


Sources/Plasma/PubUtilLib/plPipeline/DX/hsGDirect3D.cpp
```diff
@@ -60,7 +60,7 @@ hsGDirect3DTnLEnumerate& hsGDirect3D::EnumerateTnL(bool reenum)
 
 void hsGDirect3D::ReleaseTnLEnum()
 {
-    s_tnlEnum.release();
+    s_tnlEnum.reset(nullptr);
```
zrax, line 63:
```suggestion
    s_tnlEnum.reset();
```

zrax (approved):

PR #850, opened 2021-03-03
Improve pyEnum.

Hoikas:
The following issues were corrected:
- Fixed a ~hypothetical~ memory leak when adding the enum object to the module fails. (Apparently this leak is not so hypothetical. VLD says this fixes many leaks.)
- Fixed the failure return for EnumValue_hash to actually indicate failure.
- Avoided copying an std::map.
- Replaced an std::map used only for iteration with a vector of tuples.
- Fixed the code style such that I no longer want to stab my eyes out.

Sorry about the noise from all the semicolon removals 😞

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyEnum.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyEnum.cpp
@@ -82,20 +82,21 @@ static PyObject* EnumValue_new(PyTypeObject* type, PyObject* args, PyObject*)
     return (PyObject*)self;
 }
 
-static int EnumValue_clear(EnumValue* self)
+static int EnumValue_clear(PyObject* self)
 {
-    Py_CLEAR(self->name);
+    Py_CLEAR(((EnumValue*)self)->name);
     return 0;
 }
 
-static void EnumValue_dealloc(EnumValue* self)
+static void EnumValue_dealloc(PyObject* self)
 {
     EnumValue_clear(self);
-    Py_TYPE(self)->tp_free((PyObject*)self);
+    Py_TYPE(self)->tp_free(self);
 }
 
-static PyObject* EnumValue_repr(EnumValue* self)
+static PyObject* EnumValue_repr(PyObject* obj)
 {
+    EnumValue* self = (EnumValue*)obj;
     if (self->name == nullptr) {
         // no name, so just output our value
         return PyUnicode_FromFormat("%s(%ld)", Py_TYPE(self)->tp_name, self->value);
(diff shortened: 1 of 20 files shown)


Sources/Plasma/FeatureLib/pfPython/pyEnum.cpp
```diff
@@ -50,212 +50,151 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
+#include "pyEnum.h"
+#include "pyGlueHelpers.h"
+#include "pyObjectRef.h"
+
 struct EnumValue {
     PyObject_HEAD
-    long value;
-    char* name;
+    hsSsize_t value{};
```
zrax, line 62: `hsSsize_t` and `Py_ssize_t` are not guaranteed to be the same type...  You should use `Py_ssize_t` (or another type documented/supported by Python) to better guarantee compatibility with the Python API.

Sources/Plasma/FeatureLib/pfPython/pyEnum.cpp
```diff
@@ -360,21 +299,21 @@ PYTHON_TYPE_START(EnumValue)
     nullptr,                            /* tp_getattr */
     nullptr,                            /* tp_setattr */
     nullptr,                            /* tp_as_async */
-    EnumValue_repr,                     /* tp_repr */
+    (reprfunc)EnumValue_repr,           /* tp_repr */
```
zrax, line 302: IIRC even with a cast, this will trigger a warning on newer versions of GCC...

Hoikas: Thanks GCC. So much for avoiding dumb casts 😞

Sources/Plasma/FeatureLib/pfPython/pyGlueHelpers.h
```diff
@@ -611,12 +611,12 @@ static PyObject *methodName(PyObject *self) /* and now for the actual function *
 /////////////////////////////////////////////////////////////////////
 
 // the start of an enum block
-#define PYTHON_ENUM_START(enumName) std::map<std::string, int> enumName##_enumValues
+#define PYTHON_ENUM_START(enumName) std::vector<std::tuple<ST::string, hsSsize_t>> enumName##_enumValues{
```
zrax, line 614: Missed one...

Sources/Plasma/FeatureLib/pfPython/pyEnum.cpp
```diff
@@ -366,15 +304,15 @@ PYTHON_TYPE_START(EnumValue)
...
+    nullptr,                            /* tp_str */
     nullptr,                            /* tp_getattro */
     nullptr,                            /* tp_setattro */
     nullptr,                            /* tp_as_buffer */
     Py_TPFLAGS_DEFAULT
     | Py_TPFLAGS_BASETYPE,              /* tp_flags */
     "A basic enumeration value",        /* tp_doc */
     nullptr,                            /* tp_traverse */
-    nullptr,                            /* tp_clear */
+    (inquiry)EnumValue_clear,           /* tp_clear */
```
zrax, line 315: This cast should probably be fixed too

zrax (approved):

PR #851, opened 2021-03-03
Fix a crash and a silly identified by Coverity.

Hoikas:


Sources/Plasma/FeatureLib/pfPython/cyPhysics.cpp
```diff
@@ -195,7 +195,6 @@ void cyPhysics::Warp(pyPoint3& pos)
     // create message
     PyObject* matObj = pyMatrix44::New();
     pyMatrix44* mat = pyMatrix44::ConvertFrom(matObj);
-    mat->fMatrix.IdentityMatrix();
```
zrax, line 198: I expect the actual intent of this line was to set `mat->fMatrix = hsMatrix44::IdentityMatrix()`, since `SetTranslate` does not affect the entire matrix.

Hoikas: That was my thought too. Fortunately, the `pyMatrix44` constructor calls `hsMatrix44::Resset()` for us, so this is sufficient.

PR #854, opened 2021-03-04
First class clang-tidy support.

Hoikas:
This adds first class support for clang-tidy for Makefile, Ninja, and Visual Studio generators. It is loosely based on the implementation found in DIRTSAND. ~~Currently, only the modernize-use-nullptr check is enabled; modernize-use-override is still complaining about some destructors, which is outside the scope of this PR.~~

To use clang-tidy:
- If using a Visual Studio **solution**, click `Analyze` > `Run Code Analsys` > `On Solution` - this will run clang-tidy now instead of the Microsoft code analysis.
- In other generators, including Visual Studio CMake, build the `tidy` target. Any checks with fixers can be applied by building the `fix` target.

I had to clean up the SIMD/CPUID stuff a bit for clang-tidy+MSVC to not blow up on the CoreLib and plPipeline targets. As a side effect, SIMD is now used on GCC and Clang, where it evidently was not before.  Due to the way GCC does SIMD intrinsics, each SIMD instruction set must be limited to its own file for cross-platform code; otherwise, GCC may emit instructions that the end user's CPU does not implement. (NOTE: SIMD stuff was split off to #862)

Potentially outside the scope of this PR, but relevant: which checks should we be using? Many of them look useful (eg bugprone-unused-raii) but others are actually legitimate (anti-)patterns we use (eg modernize-avoid-c-arrays).

### DIFF
--- a/cmake/PlasmaTargets.cmake
+++ b/cmake/PlasmaTargets.cmake
@@ -1,11 +1,11 @@
 include(QtDeployment)
-#TODO (#854): include(Sanitizers)
+include(Sanitizers)
 
 cmake_dependent_option(
     PLASMA_USE_PCH
     "Enable precompiled headers?"
     ON
-    [[${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16"]]
+    "CMAKE_VERSION VERSION_GREATER_EQUAL 3.16;ALLOW_BUILD_OPTIMIZATIONS"
     OFF
 )
 
@@ -13,13 +13,13 @@ cmake_dependent_option(
     PLASMA_UNITY_BUILD
     "Enable unity build?"
     ON
-    [[${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.16"]]
+    "CMAKE_VERSION VERSION_GREATER_EQUAL 3.16;ALLOW_BUILD_OPTIMIZATIONS"
     OFF
 )
 
 function(plasma_executable TARGET)
     cmake_parse_arguments(PARSE_ARGV 1 _pex
-        "WIN32;CLIENT;TOOL;QT_GUI;EXCLUDE_FROM_ALL"
+        "WIN32;CLIENT;TOOL;QT_GUI;EXCLUDE_FROM_ALL;NO_SANITIZE"
         ""
         "SOURCES"
     )
@@ -60,12 +60,14 @@ function(plasma_executable TARGET)
         set_property(GLOBAL APPEND PROPERTY _PLASMA_GUI_TOOLS ${TARGET})
     endif()
 
-    #TODO (#854): plasma_sanitize_target(${TARGET})
+    if(NOT _pex_NO_SANITIZE)
+        plasma_sanitize_target(${TARGET})
+    endif()
 endfunction()
 
 function(plasma_library TARGET)
     cmake_parse_arguments(PARSE_ARGV 1 _plib
-        "UNITY_BUILD;SHARED"
+        "UNITY_BUILD;SHARED;NO_SANITIZE"
         ""
         "PRECOMPILED_HEADERS;SOURCES"
     )
@@ -85,7 +87,9 @@ function(plasma_library TARGET)
         set_target_properties(${TARGET} PROPERTIES UNITY_BUILD TRUE)
     endif()
 
-    #TODO (#854): plasma_sanitize_target(${TARGET})
+    if(NOT _plib_NO_SANITIZE)
+        plasma_sanitize_target(${TARGET})
+    endif()
 endfunction()
 
 function(plasma_test TARGET)
--- a/cmake/PlasmaTargets.cmake
+++ b/cmake/PlasmaTargets.cmake
@@ -93,7 +93,7 @@ function(plasma_library TARGET)
 endfunction()
 
 function(plasma_test TARGET)
-    plasma_executable(${TARGET} ${ARGN})
+    plasma_executable(${TARGET} NO_SANTIZE ${ARGN})
     add_test(NAME ${TARGET} COMMAND ${TARGET})
     add_dependencies(check ${TARGET})
 endfunction()
--- a/cmake/Sanitizers.cmake
+++ b/cmake/Sanitizers.cmake
@@ -1,0 +1,97 @@
+find_program(CLANG_APPLY_EXE NAMES "clang-apply-replacements" DOC "Path to clang-apply-replacements executable")
+find_program(CLANG_TIDY_EXE NAMES "clang-tidy" DOC "Path to clang-tidy executable")
+
+if(CMAKE_GENERATOR MATCHES "Makefile|Ninja")
+    set(USE_SANITIZE_TARGETS TRUE)
+    # Default such that we don't clobber PlasmaTargets.
+    set(SANITIZE_DEFAULT_VALUE OFF)
+else()
+    set(SANITIZE_DEFAULT_VALUE ON)
+endif()
+
+# Conditional that evaluates to whether or not build optimizations are allowed.
+# This is due to the sanitizers needing the compile commands of each file, not
+# the unity files. CMake will generate files like cmake_pch.cxx, and unity_0.cxx.
+# Both of these files are listed in the compile commands JSON, however, with
+# the current tidy target implementation, the sanitizer cannot know about them.
+# This results in bad/corrupted PCH errors and missing compile commands (errors)
+# for source files included in unity builds.
+if(USE_SANITIZE_TARGETS AND USE_CLANG_TIDY)
+    set(ALLOW_BUILD_OPTIMIZATIONS FALSE)
+else()
+    set(ALLOW_BUILD_OPTIMIZATIONS TRUE)
+endif()
+
+cmake_dependent_option(
+    USE_CLANG_TIDY
+    "Allow the codebase to use the clang-tidy sanitizer"
+    ${SANITIZE_DEFAULT_VALUE}
+    "CLANG_TIDY_EXE"
+    OFF
+)
+
+if(USE_CLANG_TIDY)
+    # Don't set CMAKE_CXX_CLANG_TIDY. CMake+Ninja+MSVC+clang-tidy are broken in this mode, see:
+    # https://gitlab.kitware.com/cmake/cmake/-/issues/20512
+    # https://bugs.llvm.org/show_bug.cgi?id=45356
+    # Use the tidy target instead.
+    if(USE_SANITIZE_TARGETS)
+        set(_TIDY_REPLACEMENTS_DIR "${CMAKE_BINARY_DIR}/clang-tidy-replacements")
+        add_custom_target(tidy_init
+            COMMENT "Preparing to tidy..."
+            COMMAND ${CMAKE_COMMAND} -E remove_directory "${_TIDY_REPLACEMENTS_DIR}"
+            COMMAND ${CMAKE_COMMAND} -E make_directory "${_TIDY_REPLACEMENTS_DIR}"
+        )
+        add_custom_target(tidy DEPENDS tidy_init)
+        if(CLANG_APPLY_EXE)
+            add_custom_target(fix
+                DEPENDS tidy
+                COMMENT "Applying fixes..."
+                COMMAND ${CLANG_APPLY_EXE} "${_TIDY_REPLACEMENTS_DIR}"
+                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+            )
+        endif()
+
+        # If we were on CMake 3.20, we could do this per-target...
+        set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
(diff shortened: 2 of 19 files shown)


zrax:
> modernize-use-override is still complaining about some destructors, which is outside the scope of this PR.

You can set `IgnoreDestructors` to, well, ignore the destructors:  https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html

Hoikas:
I've got the following .clang-tidy file locally, but it still seems to be complaining about destructors.
```
Checks: '-*,modernize-use-override,modernize-use-nullptr'
CheckOptions: [
    {
        key: modernize-use-override.IgnoreDestructors,
        value: 1
    }
]
HeaderFilterRegex: '.*'
```

zrax:
Hmm, I don't know how it's working for you, but for me (using Ninja) the compiler seems to think exceptions aren't enabled, so everything barfs...

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,67 @@
+find_program(CLANG_APPLY_EXE NAMES "clang-apply-replacements" DOC "Path to clang-apply-replacements executable")
+find_program(CLANG_TIDY_EXE NAMES "clang-tidy" DOC "Path to clang-tidy executable")
+
+cmake_dependent_option(USE_CLANG_TIDY "Allow the codebase to use the clang-tidy sanitizer" ON "CLANG_TIDY_EXE" OFF)
+
+if(CMAKE_GENERATOR MATCHES "[Mm][Aa][Kk][Ee][Ff][Ii][Ll][Ee]|[Nn][Ii][Nn][Jj][Aa]")
```
zrax, line 6: I think this is set to a constant string, so the silliness around case sensitivity is unnecessary...

Hoikas:
Yikes. Is this on Windows? Ninja is working well on Linux for me.

cmake/CompilerChecks.cmake
```diff
@@ -23,11 +23,59 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
+    cmake_parse_arguments(_passf "" "SOURCE_GROUP" "${_INSTRUCTION_SETS}" ${ARGN})
+
+    # Hack: if we ever bump to CMake 3.17, use ZIP_LISTS.
+    list(LENGTH _INSTRUCTION_SETS _LENGTH)
+    math(EXPR _LENGTH "${_LENGTH} - 1")
+
+    foreach(i RANGE ${_LENGTH})
+        # foreach(i IN ZIP_LISTS _INSTRUCTION_SETS _GCC_ARGS)
+        list(GET _INSTRUCTION_SETS ${i} i_0)
+        list(GET _GCC_ARGS ${i} i_1)
```
zrax, line 58: Even if you are using ZIP_LISTS, I would prefer to see reasonable variable names, rather than `${i_0}` and `${i_1}`

zrax:
Windows (Ninja + MSVC)

Hoikas:
Hmm... It was "working" with MSVC+CMake last night, but I haven't tested it lately. I'll investigate tomorrow 😄

cmake/CompilerChecks.cmake
Hoikas: Sadly, we aren't using `ZIP_LISTS`, so reasonable variable names are definitely an option.

cmake/Sanitizers.cmake
Hoikas: I aM rEmInDeD oF wHeN iT wAs KeWl To CaPiTaLiZe LiKe tHiS. :trollface:

Hoikas:
OK, Ninja+MSVC should be good now.

zrax:
I'll try it later when I get the chance...  But I like the option of having `tidy` be a separate build target rather than a normal part of compilation anyway :)

cmake/PlasmaTargets.cmake
```diff
@@ -85,11 +85,12 @@ function(plasma_library TARGET)
...
 
-    #TODO (#854): plasma_sanitize_target(${TARGET})
+    plasma_sanitize_target(${TARGET})
 endfunction()
 
 function(plasma_test TARGET)
     plasma_executable(${TARGET} ${ARGN})
     add_test(NAME ${TARGET} COMMAND ${TARGET})
     add_dependencies(check ${TARGET})
+    plasma_sanitize_target(${TARGET})
```
zrax, line 95: `plasma_executable()` already calls `plasma_sanitize_target()`

Hoikas: Whoops!

Hoikas:
I fixed some issues where clang-tidy was trying to tidy bitmaps and Windows resource files, hosing the whole process.

Deledrius:
> 
> 
> I fixed some issues where clang-tidy was trying to tidy bitmaps and Windows resource files, hosing the whole process.

Silly clang, you can't tidy those.

Hoikas:
`USE_CLANG_TIDY` now defaults to `OFF` when using the Unix Makefiles or Ninja generator. Setting `USE_CLANG_TIDY=ON` will disable build optimizations (PCH, Unity) on these generators. This is due to clang-tidy only knowing about, for example, plNetMsgScreener.cpp but not unity_{n}.cxx while the inverse is true for the compile commands JSON file. This results in fascinating but irrelevant errors.

cmake/hsConfig.h.cmake
```diff
@@ -45,6 +45,14 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 /* Compiler settings */
 #cmakedefine HAVE_CPUID
+#cmakedefine HS_AVX2
+#cmakedefine HS_AVX
+#cmakedefine HS_SSE42
+#cmakedefine HS_SSE41
+#cmakedefine HS_SSE4
+#cmakedefine HS_SSE3
+#cmakedefine HS_SSE2
+#cmakedefine HS_SSE1
```
zrax, line 55: I realize these are the names we used before, but perhaps `HAVE_<instruction set>` would be simpler / more readable?

cmake/CompilerChecks.cmake
```diff
@@ -23,11 +23,67 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     endif()
 endif()
 
-# Check for CPUID headers
-try_compile(HAVE_CPUID ${PROJECT_BINARY_DIR}
-            ${PROJECT_SOURCE_DIR}/cmake/check_cpuid.cpp)
```
zrax, line 28: The big advantage of the check file is that it checked in exactly the same way as the use in hsCpuID.h...  By pulling this into CMake, the check is no longer the same, which may result in odd errors when someone uses different compilers or tools :/

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,94 @@
...
+            )
+        endif()
+
+        if(USE_SANITIZE_TARGETS)
+            get_target_property(_TARGET_SOURCES ${TARGET} SOURCES)
+            get_target_property(_TARGET_SOURCE_DIR ${TARGET} SOURCE_DIR)
+            foreach(_SOURCE_FILE IN LISTS _TARGET_SOURCES)
+                # Prevent CMake from stoopidly feeding Python files and bitmaps to clang-tidy.
+                # Yes, CMake actually does try to do that...
+                if(_SOURCE_FILE MATCHES [[\.c$|\.cpp$|\.cxx$|\.cc$]])
```
zrax, line 77: No headers?

cmake/Sanitizers.cmake
```diff
@@ -0,0 +1,94 @@
...
+    set(ALLOW_BUILD_OPTIMIZATIONS TRUE)
+endif()
+
+cmake_dependent_option(
+    USE_CLANG_TIDY
+    "Allow the codebase to use the clang-tidy sanitizer"
+    ${ALLOW_BUILD_OPTIMIZATIONS} # Default such that we don't clobber PlasmaTargets.
+    "CLANG_TIDY_EXE"
+    OFF
+)
```
zrax, line 27: Circular dependency:  `USE_CLANG_TIDY` depends on `ALLOW_BUILD_OPTIMIZATIONS`, which is set based on `[...] AND USE_CLANG_TIDY`

cmake/Sanitizers.cmake
Hoikas: We can only tidy files listed in the compile commands JSON. I think clang-tidy smart enough to discard headers, knowing that they'll be checked on include. It isn't smart enough to filter out python files, resource files, or bitmaps though.

cmake/CompilerChecks.cmake
Hoikas: I could pull the cpuid test back out into its own file, but I think the various simd header checks are good like this.

zrax:
Clang-tidy can't currently handle the Max headers, which breaks the global 'tidy' target -- perhaps we should add an option to skip creating a tidy target for certain libs?

Hoikas:
Or we could add a clang-tidy file that skips the max headers? Max includes should be jailed to MaxAPI.h, so skipping that file should be sufficient... But, now that I'm thinking about it, I thought the global clang-tidy file should prevent any header outside Sources/ from being processed.

cmake/CompilerChecks.cmake
zrax: Yeah, the SIMD tests should definitely be kept in CMake; I was just looking specifically at the cpuid check, which is more convoluted.

zrax:
The problem isn't that clang-tidy needs to *tidy* the max headers, just that it needs to *parse* them in order to process the rest of the source file...  It's the parsing that clang-tidy has trouble with

cmake/Sanitizers.cmake
zrax: Very odd, but it raises an interesting question:  Why are those files listed as target sources to begin with?  Can't they be added to the source group without being a target source?

Hoikas: Unfortunately not. Omitting them from the target prevents them from being listed in the IDE at all. Also, you want *.rc files to be compiled into the target by MSVC.

Hoikas:
Oh, right. Yeah, at bare minimum we could have a `NO_TIDY` (or should that be `NO_SANITIZE`?) option on all of our target functions. Maybe with some further investigation, we could find which, if any, Max headers clang can parse.

cmake/CompilerChecks.cmake
Hoikas: Note for posterity, the SIMD stuff was split off to #862.

zrax:
I'd vote for `NO_SANITIZE` for consistency with `plasma_sanitize_target`

Hoikas:
Ok, I turned off clang-tidy for the Max targets. I saw the checks were firing on some Plasma code, but the Max headers are just too broken for clang to parse... At least with Max 2012.

zrax (commented):
It seems gtest headers are caught, since they are under **Sources**/Tests/gtest/, but I'm fine with leaving them if they're tricky to exclude (they're not that noisy).

I also had to make an additional change to make clang happy:
```diff
--- a/Sources/Plasma/PubUtilLib/plPipeline/DX/plDXPipeline.cpp
+++ b/Sources/Plasma/PubUtilLib/plPipeline/DX/plDXPipeline.cpp
@@ -11576,7 +11576,7 @@ void plDXPipeline::IRenderShadowsOntoSpan(const plRenderPrimFunc& render, const
                 if( selfShadowNow )
                 {
                     plConst(float) kMaxSelfPower = 0.3f;
-                    float power = fShadows[i]->fPower > kMaxSelfPower ? kMaxSelfPower : fShadows[i]->fPower;
+                    float power = fShadows[i]->fPower > kMaxSelfPower ? (float)kMaxSelfPower : fShadows[i]->fPower;
                     lRef->fD3DInfo.Diffuse.r
                         = lRef->fD3DInfo.Diffuse.g
                         = lRef->fD3DInfo.Diffuse.b
```

Other than that, it looks great 👍

Sources/Tests/.clang-tidy
```diff
@@ -0,0 +1,2 @@
+InheritParentConfig: true
```
zrax, line 1:
```
[89/100] Running clang-tidy on test_CoreLib...
YAML:1:22: error: unknown key 'InheritParentConfig'
InheritParentConfig: true
                     ^~~~
```

Hoikas: It looks like this option is [relatively new](https://reviews.llvm.org/D75184).

zrax: That would explain it...  I'm using whatever version ships with MSVC

Hoikas: The best I can do is disable sanitizing the tests, unless you can write a regex to filter out `Sources/Tests/gtests` but not `Sources/Plasma` and `Sources/Tools`...

zrax (approved):

PR #856, opened 2021-03-05
Allow case-insensitive matching of player names in KI commands

Alderem:
Commands such as /p, /addbuddy, etc. are case-sensitive to names when they shouldn't be.  This change corrects that problem.  Supplements boq's tab to autocomplete names, which is already case-insensitive as it should be.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -265,8 +265,8 @@ class xKIChat(object):
                 for player in self.BKPlayerList:
                     # Is the player in this Age?
                     if isinstance(player, ptPlayer):
-                        plyrName = player.getPlayerName()
-                        if pWords[1].startswith(plyrName + " "):
+                        plyrName = player.getPlayerName().lower()
+                        if pWords[1].lower().startswith(plyrName + " "):
                             selPlyrList.append(player)
                             cFlags.private = True
                             foundBuddy = True
@@ -279,8 +279,8 @@ class xKIChat(object):
                         ePlyr = player.getChild()
                         ePlyr = ePlyr.upcastToPlayerInfoNode()
                         if ePlyr is not None:
-                            plyrName = ePlyr.playerGetName()
-                            if pWords[1].startswith(plyrName + " "):
+                            plyrName = ePlyr.playerGetName().lower()
+                            if pWords[1].lower().startswith(plyrName + " "):
                                 selPlyrList.append(ptPlayer(ePlyr.playerGetName(), ePlyr.playerGetID()))
                                 cFlags.private = True
                                 cFlags.interAge = True
@@ -940,8 +940,8 @@ class CommandsProcessor:
         except ValueError:
             for player in self.chatMgr.BKPlayerList:
                 if isinstance(player, ptPlayer):
-                    plyrName = player.getPlayerName()
-                    if params == plyrName:
+                    plyrName = player.getPlayerName().lower()
+                    if params.lower() == plyrName:
                         return player.getPlayerID()
             return 0
 
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -205,7 +205,7 @@ class xKIChat(object):
         message = self.commandsProcessor(message)
         if not message:
             return
-        msg = message.lower()
+        msg = message.casefold()
 
         # Get any selected players.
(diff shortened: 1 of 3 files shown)


Hoikas (approved):

Scripts/Python/ki/xKIChat.py
```diff
@@ -265,8 +265,8 @@ def SendMessage(self, message):
                 for player in self.BKPlayerList:
                     # Is the player in this Age?
                     if isinstance(player, ptPlayer):
-                        plyrName = player.getPlayerName()
-                        if pWords[1].startswith(plyrName + " "):
+                        plyrName = player.getPlayerName().lower()
+                        if pWords[1].lower().startswith(plyrName + " "):
```
zrax, line 269: Probably better to use [casefold](https://docs.python.org/3/library/stdtypes.html#str.casefold) to avoid potential issues with non-Latin-1 characters in player names

Alderem: I could do that.  Should I also apply that change in all the other places where we're (still) using lower() on player names?  There are a few.  I'd think if there were an issue using lower() with non-Latin-1, these other places would also be problematic:

- detecting if your name was mentioned in chat in order to flash the window
- detecting a player name in coop animations
- comparing player names in CMPplayerOnline()
- name autocomplete's pickFirst()

(That's just a quick search, there could be more)

Hoikas: I think those would be a great addition to this PR as a batch of case-sensitivity roundup fixes. :smile:

Alderem: I've replaced lower() with casefold() accordingly when working with data that is user-inputted (messages, typed player names), i.e. data which could be Unicode.

Some points:
- I have not replaced all instances of lower(), only those in dealing with user-inputted data.  E.g. Age name comparisons still use lower().
- Casefold is just a Unicode-safe lower(), it **does not normalize** characters (does not strip accents).

Clarifying on that last point with some examples:
- "der Fluß" becomes "der Fluss" (still a capital F, to my English-accustomed eyes - German seems weird) 
- François should become françois, lowercasing the F but retaining the cedilla on the C

For now, I expect using casefold() in lieu of lower() will have no perceivable impact on the user experience.  Player names, which are the primary target of string comparisons using lower()/casefold(), seem to (still) be limited to Latin-1.  On a DirtSand server, the H'uru client ignores my clicks on the "Explore Uru" (create player) button when I attempt to create a character named François.

To allow player names beyond Latin-1, I think we'd need to change both servers and clients:
- Servers (DirtSand, MOSS, and MOULa) would need to allow Unicode in player names, and would need to treat François and Francois as equal names so only one may exist at a time.
- Clients would need extended python name matching code, so "/p francois" would send a message to François.

Some python unicode decoding pointers:
* https://stackoverflow.com/questions/517923/what-is-the-best-way-to-remove-accents-normalize-in-a-python-unicode-string
* https://stackoverflow.com/questions/44431730/how-to-replace-accented-characters-in-python

Extending player names beyond Latin-1 is outside the scope of this PR, so I'm including the above links for anyone who might want to pick up from here.

Hoikas: I'll give the code and behavior a more detailed look later today...

> "der Fluß" becomes "der Fluss" (still a capital F, to my English-accustomed eyes - German seems weird)
```
>>> "ASS".casefold()
'ass'
>>> "der Fluß".casefold()
'der fluss'
```

> Servers (DirtSand, MOSS, and MOULa) would need to allow Unicode in player names,

The servers and networking protocol already handle strings as either UTF-8 or UTF-16, depending on the context. I suspect the most pressing concern is finding places in the Python glue that hands UTF-8 off as ASCII. With the update to Python 3, the difference between the getString, getStringW, setString, and setStringW methods has been erased, so ignore them :wink:. Also, per @dpogue, we need to remove our dependence on or extend the charset of the P2Fs (font files). Currently, they only contain latin-1 glyphs.

But yeah, there's still work to be done there. :wink:

zrax (approved):

Scripts/Python/ki/xKIChat.py
Alderem: Thanks!  And yes, I didn't test that specific case, I trusted information I found on the internet (https://www.programiz.com/python-programming/methods/string/casefold) - shame on me.

Hoikas:
Thanks for the PR!

PR #857, opened 2021-03-05
Marten/enhanced chat history browsing

Alderem:
Extends Filtik's chat history browsing feature:
- Allows the user to return to text they'd entered before they accessed chat history (more like Bourne shell than CMD shell)
- Adds navigation to the microKI (was originally only in miniKI)

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1661,6 +1661,8 @@ class xKI(ptModifier):
 
         if (set == "up"):
             if (self.chatMgr.MessageHistoryIs < len(self.chatMgr.MessageHistoryList)-1):
+                if (self.chatMgr.MessageHistoryIs == -1):
+                    self.chatMgr.MessageCurrentLine = control.getString()
                 self.chatMgr.MessageHistoryIs = self.chatMgr.MessageHistoryIs +1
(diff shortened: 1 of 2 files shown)


Scripts/Python/ki/__init__.py
```diff
@@ -5520,6 +5527,15 @@ def ProcessNotifyMicro(self, control, event):
...
                 self.Autocomplete(control)
+        # Up or Down key to scroll in the chat history
+        elif event == kMessageHistoryUp:
+            ctrlID = control.getTagID()
+            if ctrlID == kGUI.ChatEditboxID:
+                self.MessageHistory(control, "up")
+        elif event == kMessageHistoryDown:
+            ctrlID = control.getTagID()
+            if ctrlID == kGUI.ChatEditboxID:
+                self.MessageHistory(control, "down")
```
Hoikas, line 5538: I thought the KI already did this?

Alderem: Only in ProcessNotifyMini(), not in ProcessNotifyMicro().  This won't change the world, but it does fix a needless inconsistency.

Hoikas (approved):

Hoikas:
Thanks for the PR!

PR #858, opened 2021-03-05
Add cmake functions to simplify common cmake target patterns

zrax:
This is built on #854, since that change made the need for something like this more obvious (and I wanted to see what the end delta looked like)...  However, it could easily be rebased to not include `plasma_sanitize_target()` and let that merge happen separately depending on which gets merged first.


Hoikas:
Looks like PlasmaTargets.cmake is missing 😢

zrax:
Rebased to exclude #854

Hoikas:
This may be outside the scope of this PR, but it would be nice if `plasma_executable(TARGET QT_GUI` would take care of `windeployqt` and remove the need for all the hacks in the tools CMakeLists.

zrax:
Good idea; I think I can make it work...

cmake/PlasmaTargets.cmake
```diff
@@ -1,26 +1,32 @@
 include(BuildOptimizations)
+include(QtDeployment)
 #TODO (#854): include(Sanitizers)
 
 function(plasma_executable TARGET)
     cmake_parse_arguments(_pex
-        "WIN32;QT_GUI;EXCLUDE_FROM_ALL"     # Flags
-        ""                                  # Single values
-        "SOURCES"                           # Multiple values
+        "WIN32;CLI_TOOL;QT_TOOL;CLIENT;EXCLUDE_FROM_ALL"
```
Hoikas, line 7: FWIW, I am planning to convert plCrashHandler, a client app, into a Qt GUI to allow crash submissions. I think this function would be better both with that in mind and in general by retaining `QT_GUI`, and having just `TOOL` and `CLIENT` as your install switches. Whether to install into tools_cli or tools_gui could be a switch on `if(_pex_WIN32 OR _pex_QT_GUI)`

cmake/QtDeployment.cmake
```diff
@@ -0,0 +1,58 @@
+
```
Hoikas, line 1: Mmm blank line

cmake/PlasmaTargets.cmake
```diff
@@ -0,0 +1,84 @@
+include(BuildOptimizations)
```
Hoikas, line 1: Since BuildOptimizations.cmake is now only 15 lines long, it might be preferable to just move all of its contents to this file.

cmake/PlasmaTargets.cmake
```diff
@@ -0,0 +1,84 @@
...
+include(QtDeployment)
+#TODO (#854): include(Sanitizers)
+
+function(plasma_executable TARGET)
+    cmake_parse_arguments(_pex
+        "WIN32;CLIENT;TOOL;QT_GUI;EXCLUDE_FROM_ALL"
+        ""
+        "SOURCES"
+        ${ARGN}
+    )
```
Hoikas, line 11: Since plasma_test forwards to this, we may want to use the [PARSE_ARGV ](https://cmake.org/cmake/help/v3.14/command/cmake_parse_arguments.html?highlight=parse_argv) form of this function to be safe. The documentation calls out `;`, but I have read about issues with `ARGN` stomping on other websites.

cmake/PlasmaTargets.cmake
```diff
@@ -0,0 +1,84 @@
...
+        target_include_directories(${TARGET}
+            PRIVATE
+                # Needed for custom widget includes
+                ${CMAKE_CURRENT_SOURCE_DIR}
+        )
+
+        # Add to the list of tools which need windeployqt
+        get_property(gui_tools GLOBAL PROPERTY _PLASMA_GUI_TOOLS)
+        list(APPEND gui_tools ${TARGET})
+        set_property(GLOBAL PROPERTY _PLASMA_GUI_TOOLS ${gui_tools})
```
Hoikas, line 48: Why not:

```suggestion
        set_property(GLOBAL PROPERTY _PLASMA_GUI_TOOLS APPEND ${TARGET})
```

zrax: Cool, I didn't know about `APPEND`...

Hoikas (approved):
❤️

PR #862, opened 2021-03-09
Fix SIMD usage on GCC/clang.

Hoikas:
This is needed for clang-tidy to be able to parse files built with MSVC.

If the -m(instruction set) flag is not set when compiling a source file, then intrinsics from that instruction set are not available. This means that when parsing a source file that uses, eg, SSE3, then the file will fail to compile. Unfortunately, we can't just turn on every instruction set we want to use because that will allow the compiler to emit the given instruction set anywhere. This will result in fun-to-debug illegal instruction crashes for our end users.

Anyway, if you have a file that needs to compile with gcc, be sure to jail any SIMD code to its own file and add it to the target using `plasma_target_simd_sources(TARGET [instruction set] [files...])`.

### DIFF
--- a/cmake/CompilerChecks.cmake
+++ b/cmake/CompilerChecks.cmake
@@ -23,11 +23,65 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     endif()
 endif()
 
+
 # Check for CPUID headers
 try_compile(HAVE_CPUID ${PROJECT_BINARY_DIR}
             ${PROJECT_SOURCE_DIR}/cmake/check_cpuid.cpp)
-if(HAVE_CPUID)
-    message(STATUS "CPUID header found -- using hardware math acceleration when available")
-else()
-    message(STATUS "CPUID header not found -- using software math")
-endif()
+
+# Check for SIMD headers
+CHECK_INCLUDE_FILE("immintrin.h" HAVE_AVX2)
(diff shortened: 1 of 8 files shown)


cmake/CompilerChecks.cmake
```diff
@@ -23,11 +23,65 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
...
+        set(_CURRENT_SOURCE_FILES ${_passf_${_CURRENT_INSTRUCTION_SET}})
+        if(_CURRENT_SOURCE_FILES)
+            target_sources(${TARGET} PRIVATE ${_CURRENT_SOURCE_FILES})
+            if(_passf_SOURCE_GROUP)
+                source_group(${_passf_SOURCE_GROUP} FILES ${_CURRENT_SOURCE_FILES})
+            endif()
+            if(HAVE_${_CURRENT_INSTRUCTION_SET} AND CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
+                set_source_files_properties(
+                    ${_CURRENT_SOURCE_FILES} PROPERTIES
+                    COMPILE_OPTIONS ${_CURRENT_COMPILE_OPTION}
```
zrax, line 68: Does this replace the property or add to it?

Hoikas: Replace, which is a bad thing, I suppose...

zrax (approved):

PR #863, opened 2021-03-09
Introduce hsPoolVector for plShadowMaster's object pools.

zrax:
hsPoolVector caters to the specific (ab)use of hsTArray's peculiar allocation quirks to reuse items in a pool separately from tracking its
size.  Unlike hsTArray, this class makes the intent clear to the reader, rather than relying on hidden and non-obvious implementation details.

This is currently only used by plShadowMaster and its children.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plGLight/plDirectShadowMaster.cpp
+++ b/Sources/Plasma/PubUtilLib/plGLight/plDirectShadowMaster.cpp
@@ -66,16 +66,14 @@ plDirectShadowMaster::plDirectShadowMaster()
 
 plDirectShadowMaster::~plDirectShadowMaster()
 {
-    for (plBoundsIsect* isect : fIsectPool.pool())
-        delete isect;
 }
 
-plShadowSlave* plDirectShadowMaster::INewSlave(const plShadowCaster* caster)
+std::unique_ptr<plShadowSlave> plDirectShadowMaster::INewSlave(const plShadowCaster* caster)
 {
     if( caster->GetPerspective() )
-        return new plPerspDirSlave;
+        return std::make_unique<plPerspDirSlave>();
     
-    return new plDirectShadowSlave;
+    return std::make_unique<plDirectShadowSlave>();
 }
 
 plShadowSlave* plDirectShadowMaster::INextSlave(const plShadowCaster* caster)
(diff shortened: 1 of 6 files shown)


Sources/Plasma/PubUtilLib/plGLight/plShadowMaster.h
```diff
@@ -43,7 +43,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #ifndef plShadowMaster_inc
 #define plShadowMaster_inc
 
-#include "hsTemplates.h"
+#include <type_traits>
```
Hoikas, line 46: Do we need this massive header?

Sources/Plasma/PubUtilLib/plGLight/plShadowMaster.h
```diff
@@ -58,6 +59,59 @@ class plMessage;
...
+// ugly hack altogether.
+template <class T>
+class hsPoolVector
+{
+public:
+    hsPoolVector() : fUsed() { }
+
+    // Access the underlying pool
+    std::vector<T> &pool() { return fPool; }
+    const std::vector<T> &pool() const { return fPool; }
```
Hoikas, line 75:
```suggestion
    std::vector<T>& pool() { return fPool; }
    const std::vector<T>& pool() const { return fPool; }
```

Sources/Plasma/PubUtilLib/plGLight/plDirectShadowMaster.cpp
```diff
@@ -66,10 +66,8 @@ plDirectShadowMaster::plDirectShadowMaster()
 
 plDirectShadowMaster::~plDirectShadowMaster()
 {
-    fIsectPool.SetCount(fIsectPool.GetNumAlloc());
-    int i;
-    for( i = 0; i < fIsectPool.GetCount(); i++ )
-        delete fIsectPool[i];
+    for (plBoundsIsect* isect : fIsectPool.pool())
+        delete isect;
```
Hoikas, line 70: This seems a little ugly to me. Since this is a small PR, maybe it would be better to go ahead and convert these naked pointers to either an `hsRef` or `std::unique_ptr`?

Hoikas (commented):
This is a lot easier to understand 👍

Sources/Plasma/PubUtilLib/plGLight/plShadowMaster.h
zrax: Whoops, I did originally but forgot to remove the include later

Hoikas (approved):

PR #865, opened 2021-03-10
Clean up a handful of size_t warnings in common headers

dpogue:
These amount to a lot of the warnings printed out by the x64 builders, because the warning is issued every time the header is included.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetMessage/plNetMsgHelpers.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetMessage/plNetMsgHelpers.cpp
@@ -400,29 +400,28 @@ void plNetMsgObjectHelper::WriteVersion(hsStream* s, hsResMgr* mgr)
 ////////////////////////////////////////////////////////
 plNetMsgObjectListHelper::~plNetMsgObjectListHelper()
 {
-    Reset();         
+    Reset();
 }
 
 void plNetMsgObjectListHelper::Reset()
 {
-    int i;
-    for( i=0 ; i<GetNumObjects() ; i++  )
-    {
+    for (size_t i = 0; i < GetNumObjects(); i++) {
         delete GetObject(i);
         fObjects[i] = nullptr;
-    } // for    
+    }
+
     fObjects.clear();
 }
 
 int plNetMsgObjectListHelper::Poke(hsStream* stream, uint32_t peekOptions)
 {
-    int16_t num = GetNumObjects();
-    stream->WriteLE(num);
-    int i;
-    for( i=0 ;i<num  ;i++  )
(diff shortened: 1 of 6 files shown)


Sources/Plasma/PubUtilLib/plNetMessage/plNetMsgHelpers.cpp
```diff
@@ -400,29 +400,26 @@ void plNetMsgObjectHelper::WriteVersion(hsStream* s, hsResMgr* mgr)
...
 
 int plNetMsgObjectListHelper::Poke(hsStream* stream, uint32_t peekOptions)
 {
-    int16_t num = GetNumObjects();
-    stream->WriteLE(num);
-    int i;
-    for( i=0 ;i<num  ;i++  )
-    {
+    size_t num = GetNumObjects();
+    stream->WriteLE(int16_t(num));
```
zrax, line 419: Whenever I truncate a `size_t` to 16-bit or smaller, I like to put an `hsAssert` to check that it isn't too large.  Also, probably better to use `uint16_t` rather than `int16_t`, since that's what hsStream actually takes.

Sources/Plasma/PubUtilLib/plNetMessage/plNetMsgHelpers.cpp
```diff
@@ -502,11 +498,10 @@ int plNetMsgMemberListHelper::Peek(hsStream* stream, const uint32_t peekOptions)
 
 int plNetMsgMemberListHelper::Poke(hsStream* stream, const uint32_t peekOptions)
 {
-    int16_t numMembers = (int16_t)GetNumMembers();
-    stream->WriteLE(numMembers);
+    size_t numMembers = GetNumMembers();
+    stream->WriteLE(int16_t(numMembers));
```
zrax, line 502: Same comments

Sources/Plasma/PubUtilLib/plNetMessage/plNetMsgHelpers.cpp
```diff
@@ -542,11 +537,10 @@ int plNetMsgReceiversListHelper::Peek(hsStream* stream, const uint32_t peekOptio
 
 int plNetMsgReceiversListHelper::Poke(hsStream* stream, const uint32_t peekOptions)
 {
-    uint8_t numIDs = (uint8_t)GetNumReceivers();
-    stream->WriteLE(numIDs);
+    size_t numIDs = GetNumReceivers();
+    stream->WriteLE(uint8_t(numIDs));
```
zrax, line 541: Same size comment

zrax (approved):

PR #866, opened 2021-03-10
Fix a crash introduced by c19ae73a

zrax:
The pool vector only calls the createItem code when adding a new element to the pool, but plDXPipeline will explicitly null out items in the pool prior to resetting its use count. This change adds a reset method which also clears the pool (note: per the C++17 standard, clear() will not cause a reallocation of the underlying storage, so this is not a performance regression).

### DIFF
--- a/Sources/Plasma/CoreLib/hsPoolVector.h
+++ b/Sources/Plasma/CoreLib/hsPoolVector.h
@@ -94,6 +94,12 @@ public:
         fUsed = 0;
     }
 
+    void reset() noexcept
+    {
+        fPool.clear();
+        fUsed = 0;
+    }
+
 private:
     std::vector<T>  fPool;
     size_t          fUsed;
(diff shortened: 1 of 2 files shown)


Hoikas (commented):
I have to admit, it's a little confusing to me that the `reset()` method `clear()`s the vector, but the `clear()` method does not.

zrax:
The idea is that `clear()` was originally chosen to match STL, and it sets the in-use count back to zero (so our pool is "empty"), whereas `reset()` does both a `clear()` and releases the underlying re-use pool back to empty as well (thus "reset" is stronger than "clear").  I agree the names are confusing though, so let me know if you have better suggestions...

Deledrius:
It's a bit verbose, but perhaps `clear()` and `release_and_clear()` would be more obvious?

Hoikas:
I think I would be OK if both methods had doxy comments to explain the difference.

Hoikas (approved):

PR #690, opened 2020-07-25
PhysX 4.1

Hoikas:
This updates PhysX from version 2.6 to 4.1. I decided to go with PhysX 4.1 instead of other options such as Bullet Physics or Newton Game Dynamics due to our previous usage of PhysX. While the API has changed significantly in places, much of the simulation behavior is similar, which makes life easier. PhysX 4.1 is open source, cross platform, and supports 64-bit.

Note that this is intended to be tested with a set of minor asset fixes in H-uru/moul-assets#10. Please ensure that you are using these files before submitting commentary.

On a high level, this changeset:
- Ports the libHSPlasma uncooking code for PhysX 2.6 data, to minimize data file churn.
- Updates PhysX 2.6 APIs to PhysX 4.1 APIs.
- Rewrites the CMake FindPhysX script for PhysX 4.1 with special considerations to the idiosyncrasies of vcpkg.
- As much as possible, separates PhysX logic from Plasma logic by pulling the code into a separate file.
- Removes or simplifies many of the bespoke aspects of the character controller in favor of functionality built in to PhysX 4.1.
- Spreads some nice `override` all over the joint.

Aside from dealing with API changes, this changeset overhauls the character controller significantly. PhysX's built in character controller can at best be described as "functional" in the old PhysX 2.6 case. Indeed, in the case of Jalak, it is insufficient and a bespoke mode had to be engaged. In PhysX 4.1, I saw issues where the avatar would be able to get on top of kickables and "ride" them such that it was trivial to go out of bounds while running in place on a kickable. So, the character controller is a true dynamic actor now, controlled by directly setting its linear velocity each simulation frame. This results in more realistic world interactions for the most part and greatly simplified code.

Some changes to note from the 2.6 movement handling:
- When attempting to walk on slopes, when the avatar reaches a slope of greater than 57 degrees, the falling animation begins. The falling animation discontinues when the slope is under 40 degrees.
- Downward velocity (due to gravity) is calculated based on whether or not we are on the ground--if the ground is a kickable, we inherit its downward velocity. Otherwise, we compute the velocity based on the ground normal. This greatly improves walking on top of the fish baskets in Eder Gira.

Closes #4.
Depends on Microsoft/vcpkg#12413 or NVIDIAGameWorks/PhysX#322.
Fixes #609.
Obsoletes and closes #667.
Depends on H-uru/moul-assets#10.

PR #867, opened 2021-03-10
Clean up hsStreams a bit

zrax:
... And remove the broken and unused stream logging functionality.


Sources/Plasma/NucleusLib/pnMessage/plMessage.cpp
```diff
@@ -344,13 +333,10 @@ int plMsgCStringHelper::Peek(char *& str, hsStream* stream, const uint32_t peekO
         {
             str = new char[strlen+1];
             str[strlen] = '\0';
-            if (strlen) {
-                stream->LogRead(strlen,str,"CString");
-                stream->LogStringString(ST::format("Value: {}", str).c_str());
-            }
+            if (strlen)
+                stream->Read(strlen, str);
```
Hoikas, line 337: While we're here, I think we should move the null-terimation to be done post-read instead of pre-read. Just in case someone is sending us very gnawty strings....

Sources/Plasma/PubUtilLib/plAnimation/plAGAnim.cpp
```diff
@@ -406,14 +406,14 @@ void plATCAnim::Read(hsStream *stream, hsResMgr *mgr)
...
     {
         ST::string name = stream->ReadSafeString();
-        float begin = stream->ReadLEScalar();
-        float end = stream->ReadLEScalar();
+        float begin = stream->ReadLEFloat();
+        float end = stream->ReadLEFloat();
         fLoops[name] = std::pair<float,float>(begin,end);
     }
 
     int numStops = stream->ReadLE32();
```
Hoikas, line 414:
```suggestion
    uint32_t numStops = stream->ReadLE32();
```

Sources/Plasma/PubUtilLib/plNetCommon/plServerGuid.cpp
```diff
@@ -42,7 +42,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsWindows.h"
 #include "plServerGuid.h"
 #include "pnMessage/plMessage.h"
-#include "PubUtilLib/plStreamLogger/plStreamLogger.h"
```
Hoikas, line 45: 😓

Sources/Plasma/PubUtilLib/plNetMessage/plNetMessage.h
```diff
@@ -44,6 +44,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "HeadSpin.h"
 #include "hsBitVector.h"
+#include "hsStream.h"
```
Hoikas, line 47: The include-graph mindful part of me would like to see this moved into the source file if it would be a trivial change.

Sources/Plasma/PubUtilLib/plPhysX/plPXStream.h
```diff
@@ -55,14 +55,14 @@ class plPXStream : public NxStream
     NxU8        readByte() const override { return fStream->ReadByte(); }
     NxU16       readWord() const override { return fStream->ReadLE16(); }
     NxU32       readDword() const override { return fStream->ReadLE32(); }
-    float       readFloat() const override { return fStream->ReadLEScalar(); }
+    float       readFloat() const override { return fStream->ReadLEFloat(); }
```
Hoikas, line 58: ... This file should have been deleted in #690... 😞 (sigh)

Sources/Plasma/NucleusLib/pnMessage/plMessage.cpp
zrax: It actually doesn't matter, since we're not reading into that final element...  But yeah, it could be moved for clarity.

Hoikas (approved):

PR #869, opened 2021-03-11
More hsStream cleanup

zrax:
* Always make bool I/O explicit, since we can stream both `bool`s and `BOOL`s (😢)
* Convert overloaded type readers and writers to a template
* Fix several cases of mismatched I/O type vs. its underlying storage.

### DIFF
--- a/Sources/Plasma/CoreLib/hsStream.h
+++ b/Sources/Plasma/CoreLib/hsStream.h
@@ -119,11 +119,9 @@ public:
 
     /* Overloaded  Begin (8 & 16 & 32 int)*/
     /* yes, swapping an 8 bit value does nothing, just useful*/
-    void            ReadLE(bool* value) { *value = this->ReadByte() ? true : false; }
     void            ReadLE(uint8_t* value) { *value = this->ReadByte(); }
     void            ReadLE(uint16_t* value) { *value = this->ReadLE16(); }
     void            ReadLE(uint32_t* value) { *value = this->ReadLE32(); }
-    void            WriteLE(bool value) { this->Write(1,&value); }
     void            WriteLE(uint8_t value) { this->Write(1,&value); }
     void            WriteLE(uint16_t value) { this->WriteLE16(value); }
     void            WriteLE(uint32_t value) { this->WriteLE32(value); }
--- a/Sources/Plasma/CoreLib/hsStream.h
+++ b/Sources/Plasma/CoreLib/hsStream.h
@@ -116,42 +116,29 @@ public:
     void            WriteLE32(uint32_t value);
     void            WriteLE32(size_t count, const uint32_t values[]);
 
-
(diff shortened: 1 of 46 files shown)


Sources/Plasma/PubUtilLib/plDrawable/plDynaDecalMgr.cpp
```diff
@@ -210,8 +210,8 @@ void plDynaDecalMgr::Read(hsStream* stream, hsResMgr* mgr)
         mgr->ReadKeyNotifyMe(stream, new plGenRefMsg(GetKey(), plRefMsg::kOnCreate, 0, kRefPartyObject), plRefFlags::kPassiveRef);
     }
 
-    fMaxNumVerts = (uint16_t)(stream->ReadLE32());
-    fMaxNumIdx = (uint16_t)(stream->ReadLE32());
+    fMaxNumVerts = stream->ReadLE32();
+    fMaxNumIdx = stream->ReadLE32();
```
Hoikas, line 214: FWIW, vertex and index buffers are limited to 16-bits. Not that I'm suggesting changing anything here, just making a note.

Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.h
```diff
@@ -187,8 +187,8 @@ class plDynamicTextMap : public plMipmap
 
         bool    MsgReceive(plMessage *msg) override;
 
-        uint16_t  GetVisibleWidth() { return fVisWidth; }
-        uint16_t  GetVisibleHeight() { return fVisHeight; }
+        uint32_t  GetVisibleWidth() { return fVisWidth; }
+        uint32_t  GetVisibleHeight() { return fVisHeight; }
```
Hoikas, line 191: While we're here...
```suggestion
        uint32_t  GetVisibleWidth() const { return fVisWidth; }
        uint32_t  GetVisibleHeight() const { return fVisHeight; }
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -103,57 +103,42 @@ enum {
...
     void            WriteLEDouble(double value);
-    void            WriteLEDouble(int count, const double values[]);
+    void            WriteLEDouble(size_t count, const double values[]);
 
+    template <typename T>
+    inline void ReadLE(T* value)
+    {
+        Read(sizeof(T), value);
+        *value = hsToLE(*value);
+    }
```
Hoikas, line 134: Could we delete potentially bugprone instances of this, eg `ReadLE<size_t>` and `ReadLE<hsSsize_t>`?

zrax: Not really, since `size_t` may alias `uint32_t` on some environments (for example) :/

Hoikas (approved):

PR #871, opened 2021-03-12
Make hsStream typed readers and writers type-safe

zrax:
This should prevent future surprises when trying to change the type of a member variable which is also streamed, and also ensures compatibility with 64-bit builds.

### DIFF
--- a/Sources/Plasma/CoreLib/hsStream.h
+++ b/Sources/Plasma/CoreLib/hsStream.h
@@ -126,19 +126,30 @@ public:
     void            WriteLEDouble(double value);
     void            WriteLEDouble(size_t count, const double values[]);
 
-    template <typename T>
-    inline void ReadLE(T* value)
-    {
-        Read(sizeof(T), value);
-        *value = hsToLE(*value);
-    }
+    // Type-safe placement readers
+    template <typename T> inline void ReadByte(T*) = delete;
+    void ReadByte(uint8_t* v) { *v = ReadByte(); }
+    void ReadByte(int8_t* v) { *v = (int8_t)ReadByte(); }
+    template <typename T> inline void ReadLE16(T*) = delete;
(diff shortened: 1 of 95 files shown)


Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -126,19 +126,30 @@ enum {
...
+    template <typename T> inline void ReadLE16(T*) = delete;
+    void ReadLE16(uint16_t* v) { *v = ReadLE16(); }
+    void ReadLE16(int16_t* v) { *v = (int16_t)ReadLE16(); }
+    template <typename T> inline void ReadLE32(T*) = delete;
+    void ReadLE32(uint32_t* v) { *v = ReadLE32(); }
+    void ReadLE32(int32_t* v) { *v = (int32_t)ReadLE32(); }
+    template <typename T> inline void ReadLEFloat(T*) = delete;
+    void ReadLEFloat(float* v) { *v = ReadLEFloat(); }
+    template <typename T> inline void ReadLEDouble(T*) = delete;
+    void ReadLEDouble(double* v) { *v = ReadLEDouble(); }
```
Hoikas, line 142: I wonder if these should be marked `[[nodiscard]]`? I'm assuming the reason for the `(void)` when reading and throwing away with `(void)s->ReadLE32();` is to silence warnings about return values?

Hoikas (approved):

Sources/Plasma/CoreLib/hsStream.h
zrax: Yeah, I actually did have `[[nodiscard]]` at one point (hence the `(void)` casts), but I later changed my mind...  I could add them back in if preferred.

Hoikas: https://www.youtube.com/watch?v=CxK_nA2iVXw

PR #875, opened 2021-03-15
Marten/more casefold

Alderem:
Two casefold()-related fixes.

(1) Under the comment `Make sure it's still just a "/p"`, a comparison is performed between the ChatPrivate command and a derived form of the original message (rather than the lowercased `msg`), causing the ChatPrivate command to be unintentionally case-sensitive (unlike all other commands).  You may now /P to your heart's content.

(2) If we introduce non-Latin1 characters in player names, a casefolded player name may not be the same length as what a player originally typed in their message (e.g. "ß" could become "ss").  This would cause the split between the player name and the rest of the message to be incorrectly determined.  The split is now figured out by counting the number of spaces in the player name.


Hoikas:
I am really having trouble conceptualizing what's being done here. If the issue is case sensitivity, why not explicitly chop off the casefolded command prefix? Also, I don't think 150518e should be included in the PR.

Alderem:
> I am really having trouble conceptualizing what's being done here. If the issue is case sensitivity, why not explicitly chop off the casefolded command prefix? Also, I don't think [150518e](https://github.com/H-uru/Plasma/commit/150518efa263609ab511e6173a5ca4ec43dd7ea2) should be included in the PR.

The issue isn't case sensitivity, it's that casefold() can _change the length of a string_.  This won't manifest as an issue until we have names with non-Latin1 characters.  If there's no desire to support non-Latin1 character names, this change isn't needed, but then much of the casefold() stuff I committed in my previous PR would also be moot.

In German, you can have a name like _Reiß_.  Not everyone has a ß (eszett) on their keyboard.  In that case, you adapt the ß to ss, so the name becomes _Reiss_.  (If the name is normally spelled Reiss, you would never go the other direction and spell it Reiß.)  See https://en.wikipedia.org/wiki/%C3%9F. 

So here are two scenarios if you've created a character in Uru named Reiß.  Both of these scenarios should (_and will, with the casefold() changes_) trigger a match to your character name, but note the outcome is different.

Scenario 1:
- I message you with "/p Reiß hello"
- You see a message from me:  "ello"

Scenario 2:
- I message you with "/p Reiss hello"
- You see a message from me:  "Hello"

Notice how the 'h' is missing in Scenario 1.  Here's why:
- The original code to strip the player name is `message = pWords[1][len(plyrName) + 1:]`.  That uses the _length of plyrName_ (which has been casefolded into "reiss") as an offset.  But what I typed in my message is "Reiß".  So the casefolded text is one character longer than the name typed, and the offset into the message of len(plyrName) + 1 will be 6 instead of 5.

Suppose we _didn't casefold() the variable plyrName_?  That would work for Scenario 1, but would do the wrong thing on Scenario 2.  The solution I've put forth will work in both cases.

Re: 150518e, thank you for catching that!  Clearly I need more guidance on how to use 'git'.  I did not expect my commit to include the PR I used to refresh my fork from master. 
* How can I fix this?
* When drawing up future PR, how do I prevent refreshes from the main repo from being echoed like this?

Scripts/Python/ki/xKIChat.py
```diff
@@ -272,7 +272,11 @@ def SendMessage(self, message):
                             cFlags.private = True
                             foundBuddy = True
                             # Remove the "/p buddyname" from the message.
-                            message = pWords[1][len(plyrName) + 1:]
+                            # due to casefold, len(plyrName) may not equal the length of what the user typed
+                            # count number of spaces in player name and split message based on that
+                            pNameWords = plyrName.split()
+                            pWords2 = pWords[1].split(" ",len(pNameWords))
+                            message = pWords2[len(pNameWords)]
```
zrax, line 279: A simpler option here might be to use the original string instead of the casefolded one, e.g.:

```python
message = pWords[1][len(player.getPlayerName()) + 1:]
```

Alderem:
Hmm, did those things, but the PR still shows 3 commits.  Any ideas?

**D:\HuruPlasma-jj>git fetch origin**

**D:\HuruPlasma-jj>git rebase origin/master**
Current branch Marten/more-casefold is up to date.

**D:\HuruPlasma-jj>git log origin/master..HEAD**
commit 36f0635e58454803a8144b5e24b6c683b9559b9c (HEAD -> Marten/more-casefold, origin/Marten/more-casefold)
Author: John Johns <johnsjrii+github@gmail.com>
Date:   Sun Mar 14 18:10:52 2021 -0700

    Fix missing () and add better comments

commit 632748dd2d873124b878cc4b7f3ff750910e2cde
Author: John Johns <johnsjrii+github@gmail.com>
Date:   Sun Mar 14 18:04:42 2021 -0700

    Fix an oversight where /p was still case-sensitive
    Split messages from player names in a casefold-robust manner

**D:\HuruPlasma-jj>git push**
Everything up-to-date

cwalther:
> If we introduce non-Latin1 characters […] (e.g. "ß" could become "ss").

Not that it matters for the topic, but I’d like to point out that ß _is_ a Latin-1 character (0xDF).

> * I message you with "/p Reiss hello"

This seems odd to me (as a German speaker, although in Switzerland we don’t use ß). I wouldn’t expect a message addressed to “Reiss” to go to “Reiß”. These are different names. What if there is both a Reiss and a Reiß in the age, will it find the right one? (I assume that is possible, or does avatar name registration already use casefolding?) For those who can’t type an ß on their keyboard, wouldn’t tab completion already solve the problem in this case?

Alderem:
> Not that it matters for the topic, but I’d like to point out that ß is a Latin-1 character (0xDF).

True.  Technically, ß is part of the Latin-1 Supplement.  What's a better way to refer to the character set allowed in names today? 
 So far as I can tell, characters like ß and ç aren't permitted in names currently.  I tried creating "François" on Adam's test shard, and I didn't even get an error message, the client ignored clicks on the "Explore Uru" button until I changed the name to "Francois".

> This seems odd to me (as a German speaker, although in Switzerland we don’t use ß). I wouldn’t expect a message addressed to “Reiss” to go to “Reiß”

I wasn't sure about this myself, so before I submitted my last response, I double-checked it with a friend who is German, and he confirmed it to be true.   I also used [Is the eszett (ß) used in last names?](https://german.stackexchange.com/questions/4994/is-the-eszett-%C3%9F-used-in-last-names) as a reference.

> What if there is both a Reiss and a Reiß in the age, will it find the right one?

Today, I don't believe there can be a Reiß, but I think we would need to implement casefolding into avatar name registration to disallow both Reiß and Reiss from coexisting.  I believe we would also want to disallow Francois and François from existing, and casefold doesn't help with that, so we'd need to do further work to de-accent characters for comparing against existing avatars.

All of this is hypothetical on my part, and presumes that we might want to expand beyond the current allowed character set.  If I receive guidance through this PR that we don't want to do that, then I'll close/cancel this PR.

> wouldn’t tab completion already solve the problem in this case?

Tab auto-complete is certainly helpful in the example cases above, but what if a special or accented character is the first or second character in a name?

Postscript:  While writing this up, I checked https://en.wikipedia.org/wiki/Latin-1_Supplement_%28Unicode_block%29 and searched the page for ß.  The finder matched every instance of "SS" and "ss" on the page as well as "ß".

Alderem:
Cancelling this pull request.  Will resubmit the fix for (1) and leave out the work for (2) which isn't needed until such time as people care about extending the character set for player names.

PR #876, opened 2021-03-15
Break the pnMessage->plSDL dependency with new msg

dpogue:
plSDLModifierStateMsg is a subclass of plSDLModifierMsg that contains the state variable handling. It lives in plMessage, which can safely link against plSDL, and is used only by PubUtilLib classes.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plMessage/CMakeLists.txt
+++ b/Sources/Plasma/PubUtilLib/plMessage/CMakeLists.txt
@@ -40,6 +40,7 @@ set(plMessage_SOURCES
     plRideAnimatedPhysMsg.cpp
     plRippleShapeMsg.cpp
     plRoomLoadNotifyMsg.cpp
+    plSDLModifierStateMsg.cpp
     plSimStateMsg.cpp
     plSwimMsg.cpp
     plSynchEnableMsg.cpp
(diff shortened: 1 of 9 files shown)


Sources/Plasma/PubUtilLib/plMessage/CMakeLists.txt
```diff
@@ -150,6 +152,7 @@ target_link_libraries(plMessage
         plGImage
         plInputCore # plInputIfaceMgrMsg
         plNetGameLib
+        plSDL # plSDLModifierStateMsg
```
Hoikas, line 155: Should be `PUBLIC` since `plSDL/plSDL.h` is included in a public header.

Hoikas (changes requested):

Sources/Plasma/PubUtilLib/plMessage/CMakeLists.txt
dpogue: It is not included in the header, only in the implementation

Hoikas: Whoops, looking at wrong filename.

Hoikas (approved):

Hoikas:
Anything left to be done here?

Hoikas:
Works great in game. Merging.

PR #877, opened 2021-03-17
Fix syntax errors hosing Python.

Hoikas:
If a single unpacked Python file has a syntax error, Python will be hosed for an indeterminate time due to the loader not properly processing errors. This fixes that issue. Related: `DisplayPythonOutput()` does nothing - all calls to this method need to be audited (in another PR) for intent and corrections applied where needed.

This issue was discovered when VS Code "helpfully" inserted an invalid `import` statement at the top of a file I was working on.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/plPythonFileMod.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/plPythonFileMod.cpp
@@ -490,18 +490,13 @@ bool plPythonFileMod::ILoadPythonCode()
         if (PythonInterface::RunFile(pyfile, fModule)) {
             // we've loaded the code into our module
             // now attach the glue python code to the end
-            if (!PythonInterface::RunFile(gluefile, fModule)) {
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfPython/plPythonFileMod.cpp
```diff
@@ -490,18 +490,15 @@ bool plPythonFileMod::ILoadPythonCode()
         if (PythonInterface::RunFile(pyfile, fModule)) {
             // we've loaded the code into our module
             // now attach the glue python code to the end
-            if (!PythonInterface::RunFile(gluefile, fModule)) {
+            if (PythonInterface::RunFile(gluefile, fModule)) {
                 // display any output (NOTE: this would be disabled in production)
```
zrax, line 494: This comment doesn't apply here any more ;)

zrax (approved):

PR #880, opened 2021-03-19
GitHub Landing Page updates

dpogue:
Preview: https://dpogue.github.io/Plasma

Expands the GitHub landing page for the project to include information about downloading builds from CI runs, how to compile Plasma with the newer vcpkg build system, how to run Plasma builds locally, and how to get involved.

This is intended to be the canonical instruction guide for building and running Plasma, rather than having poorly maintained information scattered across GitHub readme files, GoW and OpenUru wikis, and IRC logs.


building.html
```diff
@@ -0,0 +1,193 @@
...
+
+            <dt>32-bit Release:</dt>
+            <dd><tt>out\install\x86-Release\client</tt></dd>
+
+            <dt>64-bit Debug:</dt>
+            <dd><tt>out\install\x64-Debug\client</tt></dd>
+
+            <dt>64-bit Release:</dt>
+            <dd><tt>out\install\x64-Release\client</tt></dd>
+          </dl>
```
Hoikas, line 145: I think this section needs more detail. When I initially saw it, my reaction was: "wait, what are we talking about?" There's no mention of arch or build types before now by default.

building.html
```diff
@@ -0,0 +1,193 @@
...
+        <li>
+          <p>Launch Visual Studio.</p>
+        </li>
+
+        <li>
+          <p>Select <strong>Open a local folder</strong> and choose the folder where you cloned the repository.</p>
+        </li>
+
+        <li>
+          <p>Compile the project by using <strong>Build &gt; Install Plasma</strong>. All required dependencies will automatically be downloaded from vcpkg as part of the build process.</p>
```
Hoikas, line 93: This will build a debug client.

building.html
```diff
@@ -0,0 +1,193 @@
...
+          <li><a href="https://www.opus-codec.org/">Opus</a> (for video and voice chat)</li>
+          <li><a href="https://www.speex.org/downloads/">Speex</a> (for legacy voice chat)</li>
+          <li>Autodesk 3ds Max 7 (non-free SDK; for 3ds Max plugin)</li>
+        </ul>
+      </details>
+
+
+      <h3 id="building-windows">Building on Windows</h3>
+
+      <p>Compiling on Windows requires Visual Studio and the Windows SDK. We recommend <a href="https://visualstudio.microsoft.com/downloads/">Visual Studio 2019 Community Edition</a>, which is available as a free download. During the installation of Visual Studio, the Windows SDK can be installed by selecting the “Desktop development with C++” workload.</p>
```
Hoikas, line 72: This is a little wordy and overly specific IMO. The goal is to get a working Windows C++ compiler, so I think it should just say 'install the "desktop development with C++" workload". I'm not certain the wording in the suggestion is perfect, though.

```suggestion
      <p>Compiling on Windows requires Visual Studio. We recommend <a href="https://visualstudio.microsoft.com/downloads/">Visual Studio 2019 Community Edition</a>, which is available as a free download. Install the “Desktop development with C++” workload.</p>
```

building.html
dpogue: Is that what we want the default to be? Can we change the default by attempting to set a default `CMAKE_BUILD_TYPE`?

building.html
dpogue: I don't know how to tell people where to find the files when the output folder varies, without also mentioning the fact that it varies 😛

style.css
```diff
@@ -26,23 +46,38 @@ body {
 
 a {
     color: inherit;
-    text-shadow: 0 1px 2px white;
+    font-weight: 400;
+}
+
+b, strong {
+    /* Firefox uses "bolder" be default, which is... not bold enough */
```
Deledrius, line 53: Typo. ;)

style.css
```diff
@@ -26,23 +46,38 @@ body {
...
-    text-shadow: 0 1px 2px #2f2f2e;
     text-align: center;
     margin: 0;
     margin-bottom: 0.25rem;
     padding-top: 1.25rem;
 }
 
 h2 {
-    font-family: 'Cinzel', monospace;
+    font-family: 'Charter', 'XCharter', 'Bitstream Charter', serif;
```
Deledrius, line 69: Charter isn't being imported.

style.css
```diff
@@ -4,18 +4,38 @@
...
+
+
 body {
-    background: #37372f; /*#5c5c5c;*/
+    background: #37372f;
+    background: var(--page-background);
     padding: 0;
     margin: 0 auto;
-    font-family: 'Roboto', sans-serif;
+    font-family: 'Roboto', system-ui, sans-serif;
```
Deledrius, line 38: Roboto is no longer being imported.

Deledrius (approved):

style.css
dpogue: Charter is considered a web-safe font because it ships with Windows, macOS, and iOS. XCharter seems to be another common name for it when installed on Linux systems.

Source: https://iainbean.com/posts/2021/system-fonts-dont-have-to-be-ugly/#Charter

style.css
dpogue: This will try to use Roboto if it's installed, falling back to the standard system sans-serif UI font (Segoe UI on Windows, San Francisco on macOS and iOS, Roboto on Android), and finally falling back to the default sans-serif browser font on other browsers (typically Arial).

This might result in minor font differences between browsers and operating systems, but it should be better for performance because it doesn't require loading additional fonts from another server.

style.css
Deledrius: Weird.  I suspected that was the intent so I checked my freshly-installed system and found no font by this name...

style.css
Deledrius: I'm aware of the fallback system, but I don't see why we can't retain the import for a freely-available font and ensure it's available if we're explicitly intending to use it.  What does system-ui and sans-serif default to on Android?

dpogue: `system-ui` and `sans-serif` should both be Roboto on Android. `sans-serif` will be Droid Sans on older versions of Android (pre 6.0 I think?)

style.css
zrax: I tend to agree with Deledrius; I see no font matching "(X?)Charter" on either Windows 10 or Arch Linux by default.  Do web browsers typically ship with additional fonts not found in the system?  I didn't see any changes trying to poke Charter into Firefox's web inspector (whereas I did see changes poking other fonts into it), so I assume Firefox doesn't provide it either; but then I'm not really that knowledgeable about web dev...

dpogue: I'll play around a bit more with this and maybe look at using something like Georgia that should be more widely available. I'm also really not picky about what font actually gets used as long as it's readable and looks half-decent 😛

Deledrius: Same, it looks good on my system as it is.  I just wanted to point out what seemed to be a misplaced assumption in case there is a better solution.

dpogue: Updated to use Georgia, which actually seems a bit more readable (on macOS anyways).

PR #883, opened 2021-03-19
Port remaining 3ds Max plugin components to std::vector

zrax:
This is built on #882 (merge that one first), as this finalizes the removal of `hsTArray` from Plasma.


Hoikas:
Will review when this is rebased against master.

Hoikas (approved):
`hsTArray` beggone 💥

PR #884, opened 2021-03-20
Confirmation (Yes/No/OK/Quit/Logout) Overhaul

Hoikas:
While working on the [clickable links in KI Chat](https://www.youtube.com/watch?v=qvsb0mhxUig) concept, it became apparent that I needed to issue a confirmation or Yes/No dialog from a Python script that did not have its own `plKey`. Currently, Yes/No dialogs are handled by the xKI python scripts over many hundreds of lines in an especially ad-hoc way and can only be responded to using a `plNotifyMsg`. This implements a singleton whose job is to receive confirmation requests from anyone who links against plMessage and respond to them via either a (potentially stateful) lambda or using a legacy `plNotifyMsg`.

This new functionality will allow us to be very flexible with where we issue and how we respond to confirmation requests. Some examples floating around in my mind:
- `PtYesNoDialog()` can accept a Python callable to call back in addition to a key to notify.
- If linking to an Age fails, we can ask if we want to quit or return to the previous Age... instead of popping up an OS dialog that says "Link to Age failed"
- Localized dialogs can be issued from code in C++ that does not link against pfLocalizationMgr.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
@@ -44,6 +44,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include <Python.h>
 
+#include <string_view>
+
 #include "pyEnum.h"
 #include "pyColor.h"
 #include "pyGlueHelpers.h"
@@ -58,7 +60,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 namespace plPythonCallable
 {
-    static inline void IBuildTupleArg(PyObject* tuple, size_t idx, plConfirmationMsg::Result value)
+    template<>
+    inline void IBuildTupleArg(PyObject* tuple, size_t idx, plConfirmationMsg::Result value)
     {
         PyTuple_SET_ITEM(tuple, idx, PyLong_FromSsize_t((Py_ssize_t)value));
     }
--- a/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
@@ -40,34 +40,151 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
-#include <Python.h>
-#include "pyKey.h"
-
 #include "cyMisc.h"
-#include "pyGlueHelpers.h"
-#include "pyColor.h"
-#include "pyPlayer.h"
-#include "pyEnum.h"
 
-// for enums
+#include <Python.h>
+
+#include "pyEnum.h"
+#include "pyColor.h"
+#include "pyGlueHelpers.h"
+#include "pyKey.h"
+#include "pyPlayer.h"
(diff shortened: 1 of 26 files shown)


dpogue:
The implementation and overall design looks good to me.

Do we want this as a new fixed key singleton, or could we use pfGameGUIMgr as the receiver of plConfirmationMsg, and have it own a plConfirmationMgr that it passes them to?

Putting this in pfCharacter seems a bit weird, but pfCharacter is already a bit weird... 🤷🏼‍♂️

Hoikas:
I considered putting this into the pfGameGUIMgr *target*, but I felt like pfCharacter was more appropriate, since that used to house player interface elements many moons ago. I didn't really consider adding it directly to pfGameGUIMgr the class because everything is terrible over there. 😞

Hoikas:
I've adding the Python bindings for `plConfirmationMgr`. To ease porting, I kept the name `PtYesNoDialog` and its argument signature largely unchanged to ease porting.

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
```diff
@@ -40,34 +40,143 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
     }
-    pyKey* key = pyKey::ConvertFrom(keyObj);
-    cyMisc::YesNoDialog(*key, text);
-    PYTHON_RETURN_NONE
+
+    plConfirmationMsg::Callback cb;
+    if (pyKey::Check(cbObj)) {
+        cb = pyKey::ConvertFrom(cbObj)->getKey();
+    } else if (PyCallable_Check(cbObj)) {
+        cb = plPythonCallable::BuildCallback<plConfirmationMsg::Result>("PtYesNoDialog", cbObj);
```
Hoikas, line 93: @zrax I'm a little unhappy that this function requires manual specification of the callback arguments. I'm sure there's a way to erase those, but I'm drawing a blank. Any ideas?

zrax: I don't think you can in this case, at least not without moving the construction of the tuple outside the callback...

Hoikas: I got it down to specifying the variant index, which is good enough... I guess...

Hoikas:
We're ready for more detailed review :rocket:

Alderem:
One of my grievances with the old Logout dialog is that it is not keyboard-accessible.  A good UI should let a person tab between the buttons, with some indication of which is selected, and ENTER or SPACE or something like that should trigger the button.  

Should I get my hopes up that this might be addressed with this overhaul? :)

dpogue:
> Should I get my hopes up that this might be addressed with this overhaul?

That's probably a task for a separate round of changes, since it probably involves adding a concept of "focus" to the GUIDialog control classes and potentially changes to the art asset files.

Accessibility is important though, and thanks for raising this. Reporting a new issue is probably the best way to keep it from getting lost.

Hoikas:
Rebased to merge without conflicts.

dpogue (approved):

Deledrius (approved):

PR #885, opened 2021-03-21
Convert runtime factory checks to compile-time checks.

Hoikas:
When working on #884, I accidentally triggered `plFactory::IValidate()`. This was annoying because it should have been an early compile failure. This changeset makes that happen. As a bonus, the new checks reveal `plFactory::IValidate()` has missed for many years that `pfConsole`, an `hsKeyedObject` derivative, was not in the keyed object class index range... 😞


zrax (approved):
Nice!

PR #888, opened 2021-03-25
plNetGameLib depends on pnAsyncCoreExe

dpogue:


Hoikas:
This doesn't seem right... pnAsyncCoreExe is the *private* implementation of pnAsyncCore, so only pnAsyncCore should link against pnAsyncCoreExe. What error are you seeing that makes this change needed?

Hoikas:
Pinging @dpogue for response to earlier question.

dpogue:
I'm getting errors when trying to compile a command-line patcher on Linux, but it's hard to tell which ones are expected because of the missing AsyncSocket stuff and which ones are unexpected because of linking issues. I definitely get fewer errors with this change, but I'm going to mark this as a draft until the socket stuff builds on Linux and I can do more testing.

Hoikas:
Closing per discussion in IRC.

dpogue:
For the sake of documentation, I did figure out what was happening here: pnAsyncCore currently only links with pnAsyncCoreExe on Windows, which results in the undefined references when compiling on Linux.
https://github.com/H-uru/Plasma/blob/88ba573fede78a0c486cea641f7632519c88b724/Sources/Plasma/NucleusLib/pnAsyncCore/CMakeLists.txt#L26

PR #892, opened 2021-03-28
Marten/fix case sensitive page

Alderem:
All commands are case sensitive EXCEPT /p, and this fixes the reason for that.


Alderem:
Ugh.  A 'pull' commit is showing up in this again.  What the heck is wrong with Git that it does that by default and I have to go through extra steps to make it NOT do that.

PR #893, opened 2021-03-29
Fix /p to not be case sensitive

Alderem:
Despite all other commands being case-insensitive, /P didn't work.
This fixes the cause.


Hoikas (approved):

Hoikas:
Thanks for the PR!

PR #890, opened 2021-03-27
Clickable URLs in KI Chat and KI Mail.

Hoikas:
This implements clickable URLs in KI chat and KI mail. This is done by adding a new control code to the `pfGUIMultiLineEditCtrl` that indicates a "link". The actual detection and handling of these links is done in Python where the rest of the text handling is done. Note that due to technical limitations, there are some behavior changes in the KI. Specifically, KI mail is no longer censored in the vault. Rather, it is censored at display time to prevent the corruption of URLs containing censored words. Further, KI mail can only be edited by the sender, preventing a trivial opportunity to escape the naughty word censor... When edit mode is activated, all censored words become unmasked.

[Demonstration video here](https://www.youtube.com/watch?v=ewpgJrTKv1w)

Depends on:
- [ ] H-uru/moul-assets#26
- [x] #884 
- [ ] ~~#889~~

So merge those first.

PR #894, opened 2021-03-29
Change or fix bindings that conflict with chatting.

Hoikas:
When working on the [clickable URLs in KI chat](https://www.youtube.com/watch?v=wgRlXdGxluA) functionality (see #890), it became apparent that the internal client could not start chatting by pressing the H key. This is problematic because, at this time, all URLs start with H. This fixes that problem and others.

Changed bindings:
- (Internal Client) Open Logs: Ctrl+L (was L)
- (Internal Client) Camera Drive: Ctrl+D (was Shift+C)
- (Internal Client) Flymode: Ctrl+P (was Shift+P)
- (All Clients) Strafe Modifier: Alt (was Z)

Unchanged:
- Flymode Up: U
- Flymode Down: H

The unchanged keys now bind only when they are needed.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plInputCore/plAvatarInputInterface.cpp
+++ b/Sources/Plasma/PubUtilLib/plInputCore/plAvatarInputInterface.cpp
@@ -157,6 +157,11 @@ plAvatarInputInterface::plAvatarInputInterface()
     fControlMap->AddCode( B_CONTROL_MOVE_UP,                kControlFlagNormal | kControlFlagNoRepeat );
     fControlMap->AddCode( B_CONTROL_MOVE_DOWN,              kControlFlagNormal | kControlFlagNoRepeat );
     fControlMap->AddCode( B_TOGGLE_DRIVE_MODE,              kControlFlagDownEvent | kControlFlagNoRepeat );
+
+    // These by default conflict with text characters, so we deactivate the bindings until they are needed.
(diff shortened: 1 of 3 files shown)


Sources/Plasma/PubUtilLib/plInputCore/plAvatarInputInterface.cpp
```diff
@@ -308,7 +313,7 @@ void    plAvatarInputInterface::RestoreDefaultKeyMappings()
     fControlMap->BindKey( KEY_NUMPAD_SUBTRACT,          B_CAMERA_ZOOM_OUT );
 
     fControlMap->BindKey( KEY_SHIFT,                    B_CONTROL_MODIFIER_FAST );
-    fControlMap->BindKey( KEY_Z,                        B_CONTROL_MODIFIER_STRAFE );
+    fControlMap->BindKey( KEY_ALT,                      B_CONTROL_MODIFIER_STRAFE );
```
zrax, line 316: Ah, the infamous Z bug...  I guess people can stop calling me "rax" now ;)

zrax (approved):

Sources/Plasma/PubUtilLib/plInputCore/plAvatarInputInterface.cpp
Hoikas: :trollface:

PR #896, opened 2021-03-30
Cope with more goofy colliders.

Hoikas:
This fixes the issue by which attempting to use any of the Teledahn power tower machinery would result in the avatar getting stuck while seeking. Also nukes the old PhysX 2.6-era skin width off the character capsule and tries to use tighter bounds for convex hulls.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plAvTaskSeek.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plAvTaskSeek.cpp
@@ -109,6 +109,7 @@ void plAvTaskSeek::IInitDefaults()
     fAnimName = "";
     fPosGoalHit = false;
     fRotGoalHit = false;
+    fSweepTestHit = false;
     fStillPositioning = true;
     fStillRotating = true;
     fShuffleRange = kDefaultShuffleRange;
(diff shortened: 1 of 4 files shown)


Sources/Plasma/PubUtilLib/plAvatar/plAvTaskSeek.cpp
```diff
@@ -379,6 +380,27 @@ bool plAvTaskSeek::IMoveTowardsGoal(plArmatureMod *avatar, plAvBrainHuman *brain
         // We just set the pos/rot, so we know these are hit.
         fPosGoalHit = true;
         fRotGoalHit = true;
+    } else if (fRotGoalHit && fDistance <= .5f) {
```
branan, line 383: I've totally forgotten what units plasma uses. Feet? So this is ~6 inches of maximum warp?

Hoikas: That's correct. I don't remember how much experimenting I did to come up with this magic number, but it works in my tests. For reference, [this](https://www.youtube.com/watch?v=Hj6SoWpejfA) is the issue I'm trying to fix.

Hoikas:
Pulling for inclusion in Gehn.28 Beta 3.

Deledrius (approved):

PR #897, opened 2021-04-01
Fix static initialization order issue in plClientLauncher.

zrax:
This fixes getting junk data in plProduct in plUruLauncher.exe


Hoikas (approved):
I think it would be better to mitigate this in `plProduct` itself because its public interface doesn't reveal that we're accessing static objects. But, I'm not sure how feasible that is, and it doesn't seem worth holding up this potential crash fix for.

PR #900, opened 2021-04-02
Enable (optional) compatibility with Qt 6.0

zrax:
Due to various issues with Qt6's CMake modules, I recommend leaving Qt5 the default for now; However, those adventurous enough to try should get working tools with Qt6 as well with these changes.

This was tested with Qt 5.9.9, 5.15.2 and 6.0.3


Sources/Tools/plFontConverter/plFontConverter.cpp
```diff
@@ -74,6 +74,13 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "res/ui_FreeType.h"
 #include "res/ui_FreeTypeBatch.h"
 
+// Dammit Qt
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+#   define QT_SKIP_EMPTY_PARTS Qt::SkipEmptyParts
+#else
+#   define QT_SKIP_EMPTY_PARTS QString::SkipEmptyParts
+#endif
+
```
Hoikas, line 83: I was wondering how we were going to deal with this. Deprecated in the next to last 5.x and deleted in 6.x... Give me a break, Qt.

Hoikas (approved):

PR #901, opened 2021-04-03
Fix some resource.dat bugs.

Hoikas:
Fixes:
- resource.dat being junked if the number of frames in the spinning book changes or if the number of zeroes changes. Happens when switching to and from the Gehn Shard branch
- removes the listing for all resource files from plClient's target if `PLASMA_BUILD_RESOURCE_DAT` is `OFF`.

### DIFF
--- a/Sources/Plasma/Apps/plClient/external/render_svg.py
+++ b/Sources/Plasma/Apps/plClient/external/render_svg.py
@@ -50,6 +50,7 @@ import math
 import io
 from xml.dom.minidom import parse
 from optparse import OptionParser
+from glob import glob
 import scalergba
 
 try:
@@ -247,7 +248,12 @@ if __name__ == '__main__':
 	outpath = os.path.expanduser(options.outpath)
 	inpath = os.path.expanduser(options.inpath)
 
-	if not os.path.exists(outpath):
+	# Ensure that the render directory is empty, preventing old renders from junking the output
+	if os.path.exists(outpath):
+		print("Cleaning render directory...")
+		for i in glob(os.path.join(outpath, "*.png")):
+			os.unlink(os.path.join(outpath, i))
+	else:
 		os.makedirs(outpath)
 
 	## Do the work!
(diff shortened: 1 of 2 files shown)


Sources/Plasma/Apps/plClient/external/render_svg.py
```diff
@@ -247,7 +248,12 @@ def render_voice_icons(inpath, outpath):
 	outpath = os.path.expanduser(options.outpath)
 	inpath = os.path.expanduser(options.inpath)
 
-	if not os.path.exists(outpath):
+	# Ensure that the render directory is empty, preventing old renders from junking the output
+	if os.path.exists(outpath):
+		print("Cleaning render directory...")
+		for i in glob(os.path.join(outpath, "*.png")):
```
branan, line 254: I forget the actual file structure - Is there any reason to preserve any file in the outpath? Can we just purge it entirely? That way if the renders are changed to a different format, nothing breaks here.

Hoikas: I initially considered deleting the entire directory, but I decided that since the script is invocable with an arbitrary output path outside of the build process, simply deleting the entire directory would a nasty GOTCHA.

branan: Fair enough. On that case it'd be good to DRY up that "*.png" with the rendering code. But that's outside the scope of this PR

branan (approved):

branan:
Gonna let someone who's had more hands on experience with this code over the past year review as well, bit it all LGTM given the conversation on my comment

Sources/Plasma/Apps/plClient/external/render_svg.py
```diff
@@ -247,7 +248,12 @@ def render_voice_icons(inpath, outpath):
...
+			os.unlink(os.path.join(outpath, i))
```
Deledrius, line 255: I'm slightly uncomfortable with an automated tool doing wildcard unlinks, but this looks relatively safe.

Deledrius (approved):
This has been a long-standing low-key annoyance of mine when I swap back and forth between our builds.  I've been thinking for some time that a proper solution would be to output a list of assets created, and pass that to the `resource.dat` generation as an input which would solve this problem "more correctly", but it's held up by the notion that these tools themselves need a rewrite/replacement.

Due to that, this seems like an appropriate fix for the medium-term, with sufficient safeguards for its current use case.

PR #902, opened 2021-04-04
Fix improper ground filter logic.

Hoikas:
The previous logic could result in a 90-degree contact being returned as the ground. At best, this caused the player to be "flung" backwards after jumping into a wall. At worst, this caused the player to start falling in place against sheer walls.


branan (approved):
Programming is hard. New logic seems obviously more correct.

PR #905, opened 2021-04-19
Fix missing cluster spans regression

zrax:
Fixes a regression from 44f3f64e2478384e91d1230633c033daa963c284, which caused cluster spans to be missing (e.g. Minkata craters)


dpogue (approved):

Hoikas (approved):
TY for figuring this out!

PR #907, opened 2021-04-24
Add chronicle helper scripts.

### LINKED ISSUE #906: Python Script for show/hide based on Chronicles
Rather than showing/hiding objects based on Age SDL values (which are global for everyone in the Age), I'd like to be able to show and hide objects based on a player's Chronicle values.

Example use cases:
* something that is only visible when a player has completed a Marker Mission
* enabling/disabling things based on player-specific accessibility preferences

It seems like it's easy enough to read the value of a chronicle and set the correct state, but there's no obvious/easy way to be notified when the chronicle value changes.

Hoikas:
Closes #906.

xPlayerChronicleRespond: runs a responder based off the boolean conversion of a chronicle's value. **WARNING**: NEVER USE THIS TO CONTROL AN AVATAR. Use SDL for that.
xPlayerChronicleSet: sets a chronicle's value when an action happens
xPlayerChronicleShowHide: shows or hides objects based off the boolean conversion of a chronicle's value.

Reminder... Chronicles are stored in the player's vault. They should be used only when you explicitly do not want something to be synchronized among players, rather, specific to the local player.

Danger: completely untested

### DIFF
--- a/Scripts/Python/xPlayerChronicleRespond.py
+++ b/Scripts/Python/xPlayerChronicleRespond.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+""" *==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


Hoikas:
Rebased to fix spurious CI failure.

Hoikas:

PR #915, opened 2021-05-22
Adds /Dance2 to list of emotes

Hazado:
Adds Dance to to list of emotes and loopable emotes.


Hazado:
so I havent added an updated globalanimations.age, because the one on this repo has entries for the female wall climbing animations already. It also includeds other animations that dont exist in standard moula.

Should I use the page entries for the female wall climbing animations or change the numbers to what is used in https://github.com/H-uru/moul-assets/pull/41?

Hazado:
Okay, both branches are updated to reflect the current page numbers

Hoikas:
I think this would be a good opportunity to generalize the `/dance` command so that multiple dances can be more easily added in the future. I'm thinking that the `/dance` command should be modified to take an argument... I suppose that can be numeric, eg, `/dance1` for the current one and `/dance2` for the retro dance. If just `/dance` is specified, we pick a random one in the range. That should also remove the need to change anything in the localization files.

On a related topic, I think we should use descriptive names for our PRPs. Those are binary assets, and if two people try to add a new dance at the same time (unlikely, I know), then we will wind up with a conflict where two people are adding different Dance3.prps at the same time. As you know, these are not trivial to resolve. So, I think these should be eg DanceRetro.prp

PR #917, opened 2021-05-27
Automating Holidays

Hazado:
I believe that this code will work for setting the SDL states for each holiday automatically.

One thing we might want to remove is the one for Eder Delin as historically its set at first snow at the Cyan Headquarters. Since I have not found a good api yet for pulling snow data for that area...

Hanukkah is programmed out to 2050. If Moula makes it past 2050, ill add more dates then.


Scripts/Python/BaronCityOffice.py
```diff
@@ -84,4 +84,18 @@ def Load(self):
...
+        agevault = ptAgeVault()
+        if agevault:
+            ageSDL = agevault.getAgeSDL()
+            if ageSDL:
+                if int(month) == 12 and int(day) > 13 and ageSDL.findVar('bcoChristmasVis').getBool() == 0:
+                    ageSDL.findVar('bcoChristmasVis').setBool(1)
+                    agevault.updateAgeSDL(ageSDL)
+                elif ageSDL.findVar('bcoChristmasVis').getBool() == 1:
+                    ageSDL.findVar('bcoChristmasVis').setBool(0)
+                    agevault.updateAgeSDL(ageSDL)
```
Hoikas, line 101: This makes the code more Pythonic and fixes a logic error by which the Christmas deco will be toggled off on any subsequent link-in. Prefer `""` for strings and `''` for single characters.

```suggestion
        st = time.gmtime(PtGetDniTime())
        if agevault := ptAgeVault():
            if ageSDL := agevault.getAgeSDL():
                if st.tm_mon == 12 and st.tm_mday > 13 and not ageSDL.findVar("bcoChristmasVis").getBool():
                    ageSDL.findVar("bcoChristmasVis").setBool(True)
                    agevault.updateAgeSDL(ageSDL)
                elif not (st.tm_mon == 12 and st.tm_mday > 13) and ageSDL.findVar("bcoChristmasVis").getBool():
                    ageSDL.findVar('bcoChristmasVis').setBool(False)
                    agevault.updateAgeSDL(ageSDL)
```

You'll probably want to propagate these kinds of changes on down...

Scripts/Python/Neighborhood.py
```diff
@@ -81,3 +91,42 @@ def OnFirstUpdate(self):
...
+
+    def OnServerInitComplete(self):
+        dnitime = PtGetDniTime()
+        month = time.strftime('%m', time.gmtime(dnitime))
+        day = time.strftime('%d', time.gmtime(dnitime))
+        agevault = ptAgeVault()
+        if agevault:
+            ageSDL = agevault.getAgeSDL()
+            if ageSDL:
+                if int(month) == 1 and int(day) <= 3: #New Years
```
Hoikas, line 103: [PEP8](https://www.python.org/dev/peps/pep-0008/#inline-comments) says inline comments should be avoided.

Scripts/Python/hanukkah.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
```
Hoikas, line 1:
```suggestion
```

Hoikas (commented):
A few concerns:
These scripts, as exist, will do changes while other players are in the Age. To avoid that, we'll only want to operate when the Age is empty, eg by bailing:
```python
if PtGetPlayerList():
    return
```


I'm also not sure this should be accepted into the main line. These kinds of decisions, IMO, should rest with the shard operator. And ideally we would adjust the server's global state instead of the per-Age state, as requested by Chogon. That being said, this may be useful for some shard operators, so it's still useful.

Scripts/Python/hanukkah.py
Hazado: Removed.

Scripts/Python/Neighborhood.py
Hazado: Removed.

Scripts/Python/BaronCityOffice.py
Hazado: Changes made, please check the code again

Hoikas:
Per previous observations and discussion in IRC, I am closing this. This kind of a decision should probably be made at the shard/fork level.

PR #928, opened 2021-06-09
WaveSet Buoys

dpogue:
These were mostly already implemented, but the physics messages had gone missing, and someone added a wildly incorrect multiplication factor.

* Fixes some 32-bit assumptions in hsFastMath, but I can drop this commit if it seems misplaced
* Adds back plImpulseMsg and plDampMsg and hooks them up in the physics simulation
  * I suspect the damping handling could be improved, but it gives a reasonably decent effect right now
* Adds console commands for testing
* Adds Python API for fun runtime shenanigans
* Adds a new flag to the existing plWaveSet7 bitvector, which causes it to read an extra array of buoys to attach automatically
  * I have code ready to go for libHSPlasma and korman
  * This should be compatible with all existing data, but not compatible with old clients (although maybe they just neglect to read the extra data and it's fine?)

[Demo video](https://www.youtube.com/watch?v=m_74DyDb0pY)

Wishlist: A way to specify that some dynamic physical objects should not interact with the avatar

### DIFF
--- a/Scripts/Python/plasma/Plasma.py
+++ b/Scripts/Python/plasma/Plasma.py
@@ -9667,6 +9667,10 @@ class ptWaveSet:
         """None"""
         pass
 
+    def addBuoy(self, soKey: ptKey) -> None:
+        """Adds the specified object as a buoy"""
+        pass
+
     def getDepthFalloff(self):
         """Returns the attribute's value"""
         pass
@@ -9793,6 +9797,10 @@ class ptWaveSet:
 
     def getWindDir(self):
         """Returns the attribute's value"""
+        pass
+
+    def removeBuoy(self, soKey: ptKey) -> None:
+        """Removes the specified object as a buoy"""
         pass
 
     def setDepthFalloff(self,s, secs = 0):
@@ -9922,4 +9930,3 @@ class ptWaveSet:
     def setWindDir(self,s, secs = 0):
         """Sets the attribute to s over secs time"""
         pass
-
--- a/Sources/Plasma/FeatureLib/pfPython/pyWaveSet.h
+++ b/Sources/Plasma/FeatureLib/pfPython/pyWaveSet.h
@@ -185,6 +185,13 @@ public:
 
     PyObject* GetEnvCenter() const; // returns pyPoint3
     float GetEnvRadius() const;
+
+    // ==============================================================================
+    // Buoy functions
+    // ==============================================================================
+
+    void AddBuoy(const pyKey& soKey) const;
+    void RemoveBuoy(const pyKey& soKey) const;
 };
 
 
--- a/Sources/Plasma/PubUtilLib/plMessage/plDampMsg.h
+++ b/Sources/Plasma/PubUtilLib/plMessage/plDampMsg.h
@@ -1,0 +1,66 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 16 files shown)


Sources/Plasma/PubUtilLib/plMessage/plDampMsg.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include "pnMessage/plSimulationMsg.h"
+#include "hsGeometry3.h"
+
+class plDampMsg :  public plSimulationMsg
+{
+public:
+    plDampMsg() = default;
+    plDampMsg(const plKey& sender, const plKey& receiver, float damp)
+        : plSimulationMsg(sender, receiver, nullptr), fDampFactor(damp) {};
```
Hoikas, line 54:
```suggestion
        : plSimulationMsg(sender, receiver, nullptr), fDampFactor(damp) { }
```

Sources/Plasma/FeatureLib/pfPython/pyWaveSet.h
```diff
@@ -185,6 +185,13 @@ class pyWaveSet
 
     PyObject* GetEnvCenter() const; // returns pyPoint3
     float GetEnvRadius() const;
+
+    // ==============================================================================
+    // Buoy functions
+    // ==============================================================================
+
+    void AddBuoy(pyKey& soKey);
+    void RemoveBuoy(pyKey& soKey);
```
Hoikas, line 194: I feel like these can be const references (and the methods themselves could also be const).

Scripts/Python/plasma/Plasma.py
```diff
@@ -9667,6 +9667,10 @@ def __init__(self,ey):
         """None"""
         pass
 
+    def addBuoy(self, soKey):
```
Hoikas, line 9670:
```suggestion
    def addBuoy(self, soKey: ptKey) -> None:
```

Scripts/Python/plasma/Plasma.py
```diff
@@ -9795,6 +9799,10 @@ def getWindDir(self):
         """Returns the attribute's value"""
         pass
 
+    def removeBuoy(self, soKey):
```
Hoikas, line 9802:
```suggestion
    def removeBuoy(self, soKey: ptKey) -> None:
```

Sources/Plasma/PubUtilLib/plMessage/plDampMsg.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+    plDampMsg() = default;
+    plDampMsg(const plKey& sender, const plKey& receiver, float damp)
+        : plSimulationMsg(sender, receiver, nullptr), fDampFactor(damp) { }
+
+    CLASSNAME_REGISTER(plDampMsg);
+    GETINTERFACE_ANY(plDampMsg, plSimulationMsg);
+
+    void SetDamp(float d) { fDampFactor = d; }
+    float Damp() { return fDampFactor; }
```
Hoikas, line 60:
```suggestion
    float Damp() const { return fDampFactor; }
```

Sources/Plasma/PubUtilLib/plMessage/plImpulseMsg.h
```diff
@@ -0,0 +1,66 @@
...
+public:
+    plImpulseMsg() = default;
+    plImpulseMsg(const plKey& sender, const plKey& receiver, const hsVector3& impulse)
+        : plSimulationMsg(sender, receiver, nullptr), fImpulse(impulse) {};
+
+    CLASSNAME_REGISTER(plImpulseMsg);
+    GETINTERFACE_ANY(plImpulseMsg, plSimulationMsg);
+
+    void SetImpulse(const hsVector3& i) { fImpulse = i; }
+    const hsVector3& Impulse() { return fImpulse; }
```
Hoikas, line 60:
```suggestion
    const hsVector3& Impulse() const { return fImpulse; }
```

Hoikas (commented):
Aside from the two const nits, this LGTM. I think the non-avatar dynamic but should be deferred to another PR.

Hoikas (approved):

Issue #930, opened 2021-06-12
Support Max 2022


Hoikas:
The 3ds Max plugin fails to build against the 3ds Max 2022 SDK.

PR #931, opened 2021-06-12
Adds building the Max Plugin to CI

Hoikas:
As discussed in IRC, this adds 3ds Max plugin building to our CI. The main build job currently builds for Max 7 in Win32 and Max 2020 in x64. Additional jobs are spawned for 2012/Win32, 2013/x64, and 2021/x64. I feel like the Maxes that are in the main pass are the two we should most officially "support" and we should explicitly tie H-uru/moul-assets to those versions. I picked 2020 because EAA recently shared a screenshot of viewing Kahlo in Max 2020, however this is open to discussion.

Because we are not allowed to redistribute the 3ds Max SDK, to get max plugin CI in your fork, you will need a token allowing access to the *private* repository H-uruMachineUser/3dsMaxSDK.

This does not attempt to address #930.


dpogue (commented):
If building the multiple Max versions is only happening on push, does it make sense to split that (back) out into its own workflow file?

Beyond that, :+1: on this

Hoikas:
That's a good question... Additionally, since the big max matrix can't run on PRs, should we take that as an opening to add more SDKs to the matrix?

dpogue:
Sure, why not?

.github/workflows/ci.yml
```diff
@@ -60,6 +70,76 @@ jobs:
...
+    # Can only run if we have a token for our super seekrit Max SDK repo. Sad.
+    needs: ["max-secrets"]
+    if: needs.max-secrets.outputs.HAS_MACHINE_USER_TOKEN == 'true'
+
+    runs-on: windows-latest
+    name: maxplugin-${{ matrix.cfg.str }}
+    strategy:
+      matrix:
+        cfg:
+          - { sdk-version: 2008, generator: Visual Studio 16 2019, arch: Win32, str: 2008-windows-x86 }
```
dpogue, line 99: A comment here indicating that Max 7 and 2020 are tested as part of the CI workflows would be helpful for some future person looking here and being confused about missing versions 😛

dpogue (approved):

PR #935, opened 2021-06-15
Great Tree Pub Puzzle Area

Hazado:
Files needed for GTP puzzle to work correctly.
Has not been tested for Multiplayer yet

Tied to https://github.com/H-uru/moul-assets/pull/51


Hazado:
Closing until we have a working blend file for these

PR #937, opened 2021-06-19
The Great `std::move()`-ing.

Hoikas:
Adds the clang-tidy fixers [performance-move-const-arg](https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html), [performance-unnecessary-copy-initialization](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-copy-initialization.html), [performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html) to our .clang-tidy file and clean up all the resulting chaos. This involves making sure all keys, maps, functions, uoids, and strings are *not* copied when passing by reference or moving is more appropriate. Further, trivial types that were being moved have been reverted to copying. This *does not* attempt to clean up the goofy constructors that don't initialize properly or the mess of duplicate `std::string`, `std::wstring`, `const char*`, and `const wchar_t*` overloads in certain python bindings. The max plugin remains unaudited due to clang diagnostic errors in the plugin.

Good future work would be to turn on [modernize-pass-by-value](https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html).


dpogue (approved):
_reh gahro kroahl_, as they might say

Sources/Plasma/NucleusLib/pnKeyedObject/plUoid.h
```diff
@@ -158,6 +163,15 @@ class plUoid
     plUoid(const plLocation& location, uint16_t classType, const ST::string& objectName, const plLoadMask& m=plLoadMask::kAlways);
     plUoid(plFixedKeyId fixedKey);
     plUoid(const plUoid& src);
+    plUoid(plUoid&& move)
+        : fObjectID(move.fObjectID), fClonePlayerID(move.fClonePlayerID),
+          fCloneID(move.fCloneID), fClassType(move.fClassType),
+          fObjectName(std::move(move.fObjectName)),
+          fLocation(move.fLocation), fLoadMask(move.fLoadMask)
+    {
+        move.Invalidate();
```
zrax, line 172: Do we actually need to invalidate the other object?  The only member that matters (`fObjectName`) is already taken care of by the move above, and the destructor calls `Invalidate()` anyway (probably also not necessary, TBH).

In fact, this whole move constructor could probably be `= default`.

Sources/Plasma/NucleusLib/pnKeyedObject/plUoid.h
```diff
@@ -168,10 +182,33 @@ class plUoid
...
+    plUoid& operator=(plUoid&& move)
+    {
+        fObjectID = move.fObjectID;
+        fClonePlayerID = move.fClonePlayerID;
+        fCloneID = move.fCloneID;
+        fClassType = move.fClassType;
+        fObjectName = std::move(move.fObjectName);
+        fLocation = move.fLocation;
+        fLoadMask = move.fLoadMask;
+        move.Invalidate();
```
zrax, line 208: As above, this move assignment operator could probably just be `= default`

Sources/Plasma/NucleusLib/pnNetCommon/plNetGroup.h
```diff
@@ -66,7 +66,13 @@ class plNetGroupId
    plNetGroupId() : fFlags(0) {}
    plNetGroupId(const plLocation& id, const uint8_t flags) : fId(id), fFlags(flags) {  }
    plNetGroupId(const plLocation& id) : fId(id), fFlags(0) {  }
-   
+   plNetGroupId(const plNetGroupId& copy) = default;
+   plNetGroupId(plNetGroupId&& move)
+       : fId(move.fId), fFlags(move.fFlags), fDesc(std::move(move.fDesc))
+   {
+       move.fId.Invalidate();
+   }
```
zrax, line 74: These can be `= default` as above...  Furthermore, if all copy and move constructors are `= default` and there is no destructor, than all of them can be omitted to automatically get the default implementations.

Sources/Plasma/PubUtilLib/plNetMessage/plNetMessage.h
```diff
@@ -635,7 +635,11 @@ class plNetMsgGroupOwner: public plNetMsgServerToClient
         void Write(hsStream* s) const;
 
         GroupInfo() : fGroupID(plNetGroup::kNetGroupUnknown), fOwnIt(false) {}
-        GroupInfo(plNetGroupId gID, bool o) : fGroupID(gID),fOwnIt(o) {}
+        GroupInfo(plNetGroupId gID, bool o) : fGroupID(std::move(gID)),fOwnIt(o) {}
+        GroupInfo(const GroupInfo&) = default;
+        GroupInfo(GroupInfo&& move)
+            : fGroupID(std::move(move.fGroupID)), fOwnIt(move.fOwnIt)
+        { }
```
zrax, line 642: Can be defaulted / removed.

Sources/Plasma/PubUtilLib/plSDL/plSDL.h
```diff
@@ -189,11 +188,16 @@ class plStateChangeNotifier
...
+    int IRemoveKey(const plKey& k);
 public:
-    plStateChangeNotifier();
+    plStateChangeNotifier() : fDelta(0.f) { }
     plStateChangeNotifier(float i, plKey k);
-    virtual ~plStateChangeNotifier();
+    plStateChangeNotifier(const plStateChangeNotifier& copy) = default;
+    plStateChangeNotifier(plStateChangeNotifier&& move)
+        : fDelta(move.fDelta), fKeys(std::move(move.fKeys))
+    { }
```
zrax, line 198: Can be defaulted / removed

Sources/Plasma/PubUtilLib/plSDL/plSDL.h
```diff
@@ -209,6 +213,15 @@ class plStateChangeNotifier
...
     bool operator==(const plStateChangeNotifier &) const;
+
+    plStateChangeNotifier& operator=(const plStateChangeNotifier& copy) = default;
+
+    plStateChangeNotifier& operator=(plStateChangeNotifier&& move)
+    {
+        fDelta = move.fDelta;
+        fKeys = std::move(move.fKeys);
+        return *this;
+    }
```
zrax, line 224: Can be defaulted / removed

Sources/Plasma/NucleusLib/pnNetCommon/plNetGroup.h
Hoikas: clang-tidy indicated that without the inclusion of the move constructor and operator that a copy would be performed. 😕 I'll try defaulting them all.

zrax: If there's a copy constructor or copy operator declared at all (even if it's `= default`), it will suppress the creation of an automatic move constructor and move assignment operator.  You'd need to delete all 4 declarations (or set them all `= default`) to get all 4 automatically created.

Sources/Plasma/NucleusLib/pnKeyedObject/plUoid.h
Hoikas: I was on the fence about this myself - I was thinking that by invalidating the other object, it would be clearer that a move had taken place. Later, I added the bugprone-use-after-move check, so the invalidation is probably redundant.

Sources/Plasma/NucleusLib/pnNetCommon/plNetGroup.h
Hoikas: clang-tidy is pretty aggressively complaining about copying if I don't explicitly `default` the move constructor, even with all four declarations removed. So I guess that's what we'll have to do (sigh).

zrax: Hmm, that may mean that something else is suppressing the move (maybe a destructor?)...  Making them explicit is certainly fine though

zrax (approved):

Sources/Plasma/NucleusLib/pnNetCommon/plNetGroup.h
Hoikas: Ah! It must be the defaulted virtual constructor on `plStateChangeNotifier` 😄

PR #943, opened 2021-06-27
Fix more char*/TCHAR* confusion in Max.

Hoikas:
This fixes a number of property pages in Max being junked due to foolishness such as casting `char*` to `LPARAM` when Windows is expecting a `wchar_t*`. I audited every cast matching `(LPARAM)foo`, so hopefully this should be all of them.

### DIFF
--- a/Sources/Tools/MaxComponent/plAvatarComponent.cpp
+++ b/Sources/Tools/MaxComponent/plAvatarComponent.cpp
@@ -607,7 +607,6 @@ public:
 
         IParamBlock2 *pb = map->GetParamBlock();
         HWND cbox = nullptr;
-        char* buffer = nullptr;
 
         int selection;
         switch (msg)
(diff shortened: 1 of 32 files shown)


Sources/Tools/MaxComponent/plAvatarComponent.cpp
```diff
@@ -1045,9 +1043,9 @@ plLODAvatarComponent::plLODAvatarComponent() : fMaterial()
     fClassDesc = &gLODAvatarCompDesc;
     fClassDesc->MakeAutoParamBlocks(this);
 
-    fLODLevels.push_back("High");
-    fLODLevels.push_back("Medium");
-    fLODLevels.push_back("Low");
+    fLODLevels.emplace_back(_M("High"));
+    fLODLevels.emplace_back(_M("Medium"));
+    fLODLevels.emplace_back(_M("Low"));
```
zrax, line 1048:
```suggestion
    fLODLevels = {
        _M("High"),
        _M("Medium"),
        _M("Low),
    };
```

zrax (approved):

PR #948, opened 2021-07-01
Safety checking for avatars without clothing outfits

dpogue:
Most of the code properly handles `fClothingOutfit` being null, but the LayerLinkAnimation check does not.

Currently all the avatars have clothing outfits (PlasmaMax always exports them) but it doesn't necessarily make sense for things like Quabs.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
@@ -1317,7 +1317,8 @@ bool plArmatureMod::MsgReceive(plMessage* msg)
     {
         // We got a clothing file and are supposed to load our avatar from it.
         // Let's tell our outfit to do so!
-        fClothingOutfit->SetClothingFile(clothingMsg->GetClothingFile());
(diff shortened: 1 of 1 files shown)


Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
```diff
@@ -1313,7 +1313,7 @@ bool plArmatureMod::MsgReceive(plMessage* msg)
     }
 
     plLoadClothingMsg *clothingMsg = plLoadClothingMsg::ConvertNoRef(msg);
-    if (clothingMsg)
+    if (clothingMsg && fClothingOutfit)
```
Hoikas, line 1316: This change will cause the `plLoadClothingMsg` to be propagated down the chain of `MsgReceive()` overloads. Is this intended?

dpogue: I think that's _probably_ fine, but I'll change it to eat the message just to be safe

Hoikas (approved):

PR #951, opened 2021-07-03
Removes Kadish Moon Elevator subworld from interacting with player

Hazado:
This fixes the issue reported by CalumTraveler in OU#Dev

Calum entered the Kadish Elevator at the bottom and tried to leave before it started to rise.
He was prevented from leaving.

### DIFF
--- a/Scripts/Python/kdshGlowInTheDark.py
+++ b/Scripts/Python/kdshGlowInTheDark.py
@@ -284,7 +284,7 @@ class kdshGlowInTheDark(ptResponder):
 
             elif id == rgnEnterSubBtm.id:
                 PtDebugPrint("You stepped on the elevator at the bottom. Joining subworld.")
-                rgnExitSubBtm.disable()
+                #rgnExitSubBtm.disable()
                 me.avatar.enterSubWorld(elevatorsubworld.value)
                 return
 
--- a/Scripts/Python/kdshGlowInTheDark.py
+++ b/Scripts/Python/kdshGlowInTheDark.py
@@ -164,7 +164,7 @@ class kdshGlowInTheDark(ptResponder):
             if len(PtGetPlayerList()) == 0:
                 PtDebugPrint("\t I'm alone, so I'm starting the elevator.")
                 respElevDown.run(self.key)
-                PtAtTimeCallback(self.key,3,2) # clear xRgn at top
+                PtAtTimeCallback(self.key,4,2) # clear xRgn at top
             else:
                 PtDebugPrint("\t Someone else was already in this Kadish instance. Depending on Overriding HighSDL component to synch elevator.")
         else:
@@ -225,7 +225,6 @@ class kdshGlowInTheDark(ptResponder):
         elif id == respElevUp.id:
             PtDebugPrint("kdshGlowInTheDark: The elevator has reached the top.")
             PtAtTimeCallback(self.key,ElevatorDelay,3) # wait 10 seconds, then lower elevator again
-            xRgnTop.releaseNow(self.key)
             rgnEnterSubTop.enable()
             #~ rgnExitSubTop.enable()
             for region in objExitTop.value:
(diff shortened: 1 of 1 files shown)


Scripts/Python/kdshGlowInTheDark.py
```diff
@@ -284,7 +284,7 @@ def OnNotify(self,state,id,events):
 
             elif id == rgnEnterSubBtm.id:
                 PtDebugPrint("You stepped on the elevator at the bottom. Joining subworld.")
-                rgnExitSubBtm.disable()
+                #rgnExitSubBtm.disable()
```
Deledrius, line 287:
```suggestion
```
No need to comment out something we want to remove, as version control tracks that for us nicely. :)  It's best to simply remove the line.

Deledrius (changes requested):

Hoikas (commented):
This doesn't seem right to me. The problem is that the blocker objects prevent the avatar from entering/exiting the elevator due to mis-timing. This, OTOH, disables a subworld transition region.

Hazado:
my change removes the exit region disable when you enter the subworld. The script also disables it when the elevator timer is up amd it starts rising. This change allows you to exit the elevator before the timer is up.

Hoikas:
This won't fix the fact that the main world has a blocker physically blocking you from exiting the elevator.

Hazado:
But it does fix this issue that Calum reported here - https://discord.com/channels/282045216221822978/282046743338811392/860696685909639208

Which is what I was fixing

Hoikas:
Ok, I understand now. One thing to be aware of is that this disable was there for a reason - with this change, there is now a way to eject yourself from the Age by running into the corner of the elevator.

Hazado:
Okay, so would a better fix to convert this to a jalak style elevator?
No more subworlds to fuzt with?

Hoikas:
I think that would be the best fix. Hopefully the real floor in Kadish will not be in the way.

Scripts/Python/kdshGlowInTheDark.py
```diff
@@ -538,3 +513,11 @@ def OnTimer(self,id):
...
             PtDebugPrint("kdshGlowInTheDark: It's been one second. FF'ing floor to glow state.")
             respFloorGlow.run(self.key,fastforward=1)
+
+        if id == 6:
+            note = ptNotify(self.key)
+            note.clearReceivers()
+            note.addReceiver(self.key)
+            note.setActivate(1)
+            note.addVarNumber("foo",id)
+            note.send()
```
Hoikas, line 523: What's this for?

Hazado: It triggers the early release of the top exclude region.

Scripts/Python/kdshGlowInTheDark.py
```diff
@@ -295,14 +271,14 @@ def OnNotify(self,state,id,events):
...
                     PtDebugPrint("kdshGlowInTheDark: Timer 1 Callback. Raising elevator again.")
-                    xRgnBottom.clearNow(self.key)
                     rgnExitSubBtm.disable()
 
                     if not self.sceneobject.isLocallyOwned():
                         PtDebugPrint("\tI'm not the owner, so I'll let another client netforce raise the elevator.")
                         return
                     else:
                         respElevUp.run(self.key)
+                        PtAtTimeCallback(self.key,14,6)
```
Hoikas, line 281: Last thing, I think. Since this timer sends a local notification that toggles an exclude region, it needs to be sent without regard to the ownership of the PFM object. In other words, move this up to where `xRgnButton.clearNow(self.key)` was. Otherwise, there will be some sneaky bugs around players quitting the game or linking out while the elevator is moving.

Hazado: Change made.

Hoikas (approved):

Hoikas:
Riding physics will be improved in #953.

PR #952, opened 2021-07-03
Allow the max plugin to auto-generate PlasmaMAX2.ini

Hoikas:
Previously, if you launched Max without a PlasmaMAX2.ini file, you were SOL. The functionality to specify a client directory was already present, so this simply enables it. As a bonus, we now use the folder picker dialog from Vista (and above) instead of the old Windows 9x dialog, and windows.h becomes much less viral.

### DIFF
--- a/Sources/Tools/MaxMain/plMaxCFGFile.cpp
+++ b/Sources/Tools/MaxMain/plMaxCFGFile.cpp
@@ -69,7 +69,7 @@ plFileName plMaxConfig::GetClientPath(bool getNew, bool quiet)
     if ((len == 0 || getNew) && !quiet)
     {
         // If the user selects one, save it
-        plasmaPath = plBrowseFolder::GetFolder(plasmaPath, "Specify your client folder");
+        plasmaPath = plBrowseFolder::GetFolder(plasmaPath, "Select your Plasma 2.0 client folder");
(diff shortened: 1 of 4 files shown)


Sources/Tools/MaxMain/plMaxCFGFile.cpp
```diff
@@ -69,7 +69,7 @@ plFileName plMaxConfig::GetClientPath(bool getNew, bool quiet)
     if ((len == 0 || getNew) && !quiet)
     {
         // If the user selects one, save it
-        plasmaPath = plBrowseFolder::GetFolder(plasmaPath, "Specify your client folder");
+        plasmaPath = plBrowseFolder::GetFolder(plasmaPath, "Select your Plasma 2.0 client folder");
         if (plasmaPath.IsValid())
             WritePrivateProfileStringW(L"SceneViewer", L"Directory", plasmaPath.WideString().data(),
```
dpogue, line 74: Do we want to keep using `SceneViewer.Directory` or would something like `Export.Directory` be better?

Hoikas: I would rather not break old PlasmaMAX2.ini files 😄

Hoikas:
Merging per discussion in IRC.

PR #953, opened 2021-07-05
PhysX Behavior Rollup

Hoikas:
Some corrections to physical behaviors resulting from the PhysX 4.1 upgrade. See each commit description for details. This unfortunately does not include #784, which needs more investigation.


Sources/Plasma/PubUtilLib/plPhysX/plPXSimulation.cpp
```diff
@@ -538,8 +538,12 @@ physx::PxRigidActor* plPXSimulation::CreateRigidActor(const physx::PxGeometry& g
         actor = fPxPhysics->createRigidStatic(globalPose);
         break;
     case plPXActorType::kDynamicActor:
-        actor = fPxPhysics->createRigidDynamic(globalPose);
-        break;
+        {
+            physx::PxRigidDynamic* dynamic = fPxPhysics->createRigidDynamic(globalPose);
+            dynamic->setMaxDepenetrationVelocity(10.f);
```
Deledrius, line 543: What are the units used here?  The commit message says 20ft/s, but the value is 10.0.

Hoikas: The commit message needs to say 10ft/s. 😞

Hoikas: @Deledrius, I have corrected the commit message.

Deledrius (approved):

PR #954, opened 2021-07-06
Correct harvester/car ramp logic.

Hoikas:
This corrects the Er'cana Harvester to only put the ramps down when the car and harvester are stationary in the same location. Previously, it was possible to cleverly desynchronize this and wind up with the ramps down while the machine was moving, potentially without the car attached.

See companion m-a PR for more details.

Since this PR does not attempt to rewrite the script to not be a pile of garbage, these are my notes about the harvester states:
```
car pos
0: not ready/eligible to go up into the mill
1: at bottom
2: going up
3: at top
4: going down

harvester pos
0: far shell cloth
2: moving
4: at mill
```


Hoikas:

PR #956, opened 2021-07-07
Fix improper warp when exiting the Garrison gear.

Hoikas:
In response to reports on OpenUru#dev...

The regions are goofed up somehow such that we get ambiguous notifications. So, we keep track of what's open and enter/exit as appropriate. As a bonus, because the stupid animation is not properly synchronized between players (needs a PRP fix), don't net force the regions being enabled/disabled.


Hoikas:
This looked good over in OU land, but I forgot to turn off draft mode.

Hoikas:

PR #958, opened 2021-07-15
Use XML CDATA for journals in LOC files.

Hoikas:
This adds some very basic heuristics for detecting journals contained in the localization files and exports their data enclosed in XML's `<![CDATA[thingy]]>` to prevent having to review localization changes with XML-encoded Plasma esHTML.

Bonus: a spurious copy was eliminated.


zrax (approved):

Hoikas:
A critical bug with parameterized strings has been fixed. This issue only appeared when serializing the localization database.

zrax (approved):

PR #970, opened 2021-07-23
Suppress vault callbacks when loading or unloading VNodeMgrs.

Hoikas:
On MOULa, there have been complaints about linking to some Neighborhoods and activating some players taking a very long time. One of these Neighborhoods in particular is the "DRC(67) Bevin". When I attempted to link to this Age, I found the link took approximately 85 seconds each time. On profiling, I discovered that for every node downloaded during the initialization phase, we were calling into Python at least once. Suppressing vault callbacks during times when they are obviously going to storm and be useless decreases the link time to 7 seconds.

Initially, I suppressed just the Python side of the callbacks when an Age is being initialized, but it seems to me that a more reasonable solution is to not run callbacks during vault loads and unloads. Clearly, things are happening at this stage, so we don't need that level of granulatrity about what is happening to the vault(s) that we are subscribing to or unsubscribing from.


Hoikas:

PR #972, opened 2021-07-25
Add ptAvatar.findBone method

dpogue:
This allows getting an avatar's bones using Python, which enables hackery with `PtAttachObject` to them

### DIFF
--- a/Scripts/Python/plasma/Plasma.py
+++ b/Scripts/Python/plasma/Plasma.py
@@ -1476,6 +1476,10 @@ class ptAvatar:
         """Exits the avatar from the subWorld where it was"""
         pass
 
+    def findBone(self, bone_name):
+        """Find the ptSceneObject for the requested bone"""
+        pass
+
     def getAllWithSameMesh(self,clothing_name):
         """Returns a lilst of all clothing items that use the same mesh as the specified one"""
         pass
--- a/Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp
@@ -1620,6 +1620,27 @@ bool cyAvatar::LoadClothingFromFile(plFileName filename)
     return false;
 }
 
+PyObject* cyAvatar::FindBone(const ST::string& boneName)
+{
+    // make sure that there is atleast one avatar scene object attached (should be)
+    if (!fRecvr.empty() && fRecvr[0] != nullptr)
+    {
+        // find the armature modifier
+        plArmatureMod* avatar = (plArmatureMod*)IFindArmatureMod((plKey)fRecvr[0]);
+        if(avatar)
+        {
+            const plSceneObject* bone = avatar->FindBone(boneName);
+            if (bone) {
+                return pySceneObject::New(bone->GetKey());
+            }
+        }
+    }
+
+    ST::string errmsg = ST::format("Bone {} not found", boneName);
+    PyErr_SetString(PyExc_NameError, errmsg.c_str());
+    return nullptr; // return nullptr cause we errored
+}
+
 /////////////////////////////////////////////////////////////////////////////
 //
 //  Function   : ChangeAvatar
--- a/Sources/Plasma/FeatureLib/pfPython/cyAvatarGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyAvatarGlue.cpp
@@ -639,6 +639,18 @@ PYTHON_METHOD_DEFINITION(ptAvatar, loadClothingFromFile, args)
     PYTHON_RETURN_BOOL(self->fThis->LoadClothingFromFile(filename));
 }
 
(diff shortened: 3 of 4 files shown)


Hoikas:
I think we need to spend some time designing a better system. `PtAttachObject` has no persistence and doesn't seem like the best method IMO.

dpogue:
"No persistence" meaning there's no state to be synched to newly joined players? hmm...

A visual issue like that wouldn't be a barrier for my particular use case, but does seem like something worth addressing.

Hoikas:
Correct. To my mind, I think it would be better to have a more formalized inventory system for attaching objects to the avatar (this system probably involves avatar.sdl) with the ability to anchor objects to the left hip or off-hand built-in, for example. That way, we discourage bespoke solutions in each Age that results in an increased maintenance burden.

dpogue:
I pulled these comments into a discussion https://github.com/H-uru/Plasma/discussions/974 so we can come up with a plan before doing any further implementation.

Scripts/Python/plasma/Plasma.py
```diff
@@ -1476,6 +1476,10 @@ def exitSubWorld(self):
         """Exits the avatar from the subWorld where it was"""
         pass
 
+    def findBone(self, bone_name):
+        """Find the ptSceneObject for the requested bone"""
+        pass
+
```
dgelessus: I think this stub is already formatted correctly, but it would be good to rebase this PR and use generate_stubs.py to do this part automatically.

Sources/Plasma/FeatureLib/pfPython/cyAvatarGlue.cpp
```diff
@@ -639,6 +639,18 @@ PYTHON_METHOD_DEFINITION(ptAvatar, loadClothingFromFile, args)
...
+PYTHON_METHOD_DEFINITION(ptAvatar, findBone, args)
+{
+    char* boneName = nullptr;
+    if (!PyArg_ParseTuple(args, "s", &boneName))
+    {
+        PyErr_SetString(PyExc_TypeError, "findBone expects a string bone name");
+        PYTHON_RETURN_ERROR;
+    }
+
+    return self->fThis->FindBone(ST::string::from_utf8(boneName));
```
dgelessus: Can parse directly to a `ST::string`:

```suggestion
    ST::string boneName;
    if (!PyArg_ParseTuple(args, "O&|", PyUnicode_STStringConverter, &boneName))
    {
        PyErr_SetString(PyExc_TypeError, "findBone expects a string bone name");
        PYTHON_RETURN_ERROR;
    }

    return self->fThis->FindBone(boneName);
```

Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp
```diff
@@ -1620,6 +1620,27 @@ bool cyAvatar::LoadClothingFromFile(plFileName filename)
     return false;
 }
 
+PyObject* cyAvatar::FindBone(const ST::string& boneName)
+{
+    // make sure that there is atleast one avatar scene object attached (should be)
+    if (!fRecvr.empty() && fRecvr[0] != nullptr)
+    {
+        // find the armature modifier
+        plArmatureMod* avatar = (plArmatureMod*)IFindArmatureMod((plKey)fRecvr[0]);
```
dgelessus: Is either of these casts necessary? A `const plArmatureMod*` should be enough (`plArmatureMod::FindBone` is `const`) and `fRecvr` is a vector of `plKeys`.

```suggestion
        const plArmatureMod* avatar = IFindArmatureMod(fRecvr[0]);
```

dpogue: This seems like it should work (and compiles fine), but the original pattern is used in pretty much every function in this file, so maybe we'll want to go through and to a cleanup task at some point.

dpogue:
I've rebased and updated this based on feedback (but not tested it in the client after rebasing)

Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp
```diff
@@ -1614,6 +1614,33 @@ bool cyAvatar::LoadClothingFromFile(plFileName filename)
...
+        if (avatar) {
+            const plSceneObject* bone = avatar->FindBone(boneName);
+            if (bone) {
+                return pySceneObject::New(bone->GetKey());
+            }
+        }
+    }
+
+    ST::string errmsg = ST::format("Bone {} not found", boneName);
+    PyErr_SetString(PyExc_NameError, errmsg.c_str());
```
Hoikas, line 1640:
```suggestion
    PyErr_SetString(PyExc_ValueError, errmsg.c_str());
```

This is a little pedantic, but [`NameError`](https://docs.python.org/3/library/exceptions.html#NameError) is raised for global or local variable names that don't exist.

Hoikas (commented):
Other than one nitpick. LGTM

PR #973, opened 2021-07-25
K'veer Update August 2021

DoobesURU:
Adds new SDL states for Atrus clothing item.

### DIFF
--- a/Scripts/SDL/Kveer.sdl
+++ b/Scripts/SDL/Kveer.sdl
@@ -59,3 +59,16 @@ VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
 ## Reward Clothing
 
 }
+
+STATEDESC Kveer
+{
+	VERSION 2
+
+# Boolean variables
+VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
+
+## Reward Clothing
+VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
+VAR BOOL    kverAtrusGet[1]              DEFAULT=1
+
+}
--- a/Scripts/SDL/Kveer.sdl
+++ b/Scripts/SDL/Kveer.sdl
@@ -69,5 +69,6 @@ STATEDESC Kveer
 
     ## Reward Clothing
     VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
+    VAR BOOL    kverAtrusGet[1]              DEFAULT=1
 
 }
--- a/Scripts/SDL/Kveer.sdl
+++ b/Scripts/SDL/Kveer.sdl
@@ -64,11 +64,11 @@ STATEDESC Kveer
 {
 	VERSION 2
 
-# Boolean variables
-VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
+    # Boolean variables
+    VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
 
-## Reward Clothing
-VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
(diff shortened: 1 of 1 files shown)


DoobesURU:
Since K'veer is instanced as hood and public, the clothing item has two states: one for the clickable to get the clothing and a separate one for the visibility. This is so the clothing item can be introduced later after the update and it won't disappear once the first avatar clicks on it.

Scripts/SDL/Kveer.sdl
```diff
@@ -59,3 +59,16 @@ VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
...
+STATEDESC Kveer
+{
+	VERSION 2
+
+# Boolean variables
+VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
+
+## Reward Clothing
+VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
+VAR BOOL    kverAtrusGet[1]              DEFAULT=1
```
Hoikas, line 72:
```suggestion
    # Boolean variables
    VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT

    ## Reward Clothing
    VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
    VAR BOOL    kverAtrusGet[1]              DEFAULT=1
```

Hoikas, line 72: I don't think we want this variable (kverAtrusGet); all possible variations of K'veer may be shared with multiple players, so we don't want the clothing item to disappear for future players when a single player collects it.

DoobesURU: I mentioned this in my supplemental comment above actually. That's actually why there's two states.

In normal cases, it appears the get variable and the show/hide variable use the same SDL bool, so when you click the item to get it, it also toggles the show/hide, causing the item to disappear...which is annoying if someone grabs your clothing item before you do. An exception to this is Catherine's tunic on Myst Island...it only has the get script and no show/hide script, so it doesn't disappear when you click it.

Having the two separate bool variables will allow us to appear/disappear the clothing item separately while still allowing the xTakableClothing script to remain active.

Hoikas: Does xTakableClothing not allow for the SDL variable to be empty?

DoobesURU: > 
> 
> Does xTakableClothing not allow for the SDL variable to be empty?

I...haven't tried that. I'll check it out.

At the same time, we'd probably still want the show/hide bool so we can have it "magically" appear sometime after the big update itself, yes? Or would we rather it just be constantly on? Honestly, I'm fine with either way. 😄

Hoikas: Yeah, I think keeping the global show/hide is definitely a good thing... If we really want to change the SDL 👅

Scripts/SDL/Kveer.sdl
```diff
@@ -59,3 +59,16 @@ VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
...
+STATEDESC Kveer
+{
+	VERSION 2
+
+    # Boolean variables
+    VAR BOOL    kverMystLinkingBookVis[1]    DEFAULT=0 DEFAULTOPTION=VAULT
+
+    ## Reward Clothing
+    VAR BOOL    kverAtrusVis[1]              DEFAULT=0 DEFAULTOPTION=VAULT
+    VAR BOOL    kverAtrusGet[1]              DEFAULT=1
```
Hoikas, line 72: I seem to remember that this variable is no longer needed...? Should this PR be closed?

DoobesURU: Sorry I'm only just now getting back to this (finally got around to cleaning up git branches.

This is exactly how it now appears now in MOULa. In hindsight, probably should have adjusted the takable clothing script node tree, but I didn't know that was a thing until after this deed was done. 😊

Hoikas:
Merging to match MOULa. Note the spurious SDL that needs to be deleted on next revision.

PR #980, opened 2021-07-27
Fix #979.

### LINKED ISSUE #979: plVisLOSMgr does not detect avatars.
If using `plVisLOSMgr` to check the visibility of an object, the visibility check will not fire on avatars. This can be trivially observed in that Korman's lamp flares are not hidden by players; they draw on top of players instead.

Hoikas:
As it turns out, avatars are on the render level 0x0FFFFFFE. They are still opaque (Hint: `0x0FFFFFFE >> 28 == 0`) but the entire level is not zero. The bad news about this is that the yeesha hologram now counts as visible, but oh well, what can you do?


Deledrius:
The actual hologram or the retcon hologram? ;)

Hoikas:
:trollface:

Hoikas:
Merging per discussion in IRC.

PR #986, opened 2021-08-06
Fixes issue with using Great Zero Mission Scopes at the end

Hazado:
If you have someone open the Great Zero Mission room for you and you have not done both red and green calibration, you will get stuck in the mission GUI.

### DIFF
--- a/Scripts/Python/grtzMarkerScopeGUI.py
+++ b/Scripts/Python/grtzMarkerScopeGUI.py
@@ -229,8 +229,7 @@ class grtzMarkerScopeGUI(ptModifier):
         if id in self._scopes:
             if PtDetermineKIMarkerLevel() < kKIMarkerNormalLevel:
                 MarkerGameDlg.dialog.hide()
-            else:
-                MGMachineOffResp.run(self.key, netPropagate=False)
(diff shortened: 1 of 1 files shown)


Hoikas (commented):
This file is a bizarre mess of implicit state change callback soup. I'm having trouble understanding just how the GUI gets turned ON (it looks like it just gets turned OFF). So, with that in mind, if this does what it says in testing, I'm inclined to merge.

Hoikas:
I just tested master (without this change), and could not find any problems with entering and exiting the scopes without having the red and green missions. Perhaps this issue is endemic to the OU codebase?

Hazado:
Since i tested this with the huru client, im not sure why your not seeing the issue. If your character has not completed both red and green missions it should prevent you from leaving the gui when you go to the mission scopes at the end.

Hoikas:
It's working fine for me 🤷‍♂️

Hoikas:
I "completed" the green missions using `Python.Cheat GiveMeMarkerTag`, and used the machine to receive the link into the big room. After linking normally and flying through the doors, I was able to enter and exit the scopes without any issue.

Hazado:
This issue does not happen with Huru code base.

PR #988, opened 2021-08-13
Kadish Elevator Fix for bottom exclude region

Hazado:
Re-adds the bottom exclude region clear


Hoikas:

PR #989, opened 2021-08-16
Korman 0.12 Localization Stuff

Hoikas:
Adds xDynTextLoc.py, a reimplementation of xDynTextObj,py that does not use a hardcoded lookup table. Also fixes an attribute type problem in xDynaTextBoy.py.

### DIFF
--- a/Scripts/Python/xDynTextLoc.py
+++ b/Scripts/Python/xDynTextLoc.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+""" *==LICENSE==*
+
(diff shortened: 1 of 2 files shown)


dpogue:
For what it's worth, I don't recall any instances of xDynaTextBoy being used in Uru. It doesn't support localization or much in the way of text formatting (with newlines or stuff)

Hoikas:
It's not, but I used it in the GUI test Age to verify changes to Korman's dynamic text map handling. The incorrect attribute types completely hose the script.

Deledrius (approved):

PR #992, opened 2021-09-20
Fix PlasmaMax sound component crashes

dpogue:
I'm happy to drop the third and fourth commits if you prefer them not to be included. The fourth one in particular probably needs more investigation to figure out what to do (otoh, it happens on exit, so leaking memory maybe isn't a concern?)

### DIFF
--- a/Sources/Tools/MaxComponent/plMaxWaveUtils.cpp
+++ b/Sources/Tools/MaxComponent/plMaxWaveUtils.cpp
@@ -47,7 +47,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 SegmentMap *GetWaveSegmentMap(const MCHAR* file, plErrorMsg *pErrMsg)
 {
     CWaveFile waveFile;
-    waveFile.Open(M2ST(file), nullptr, WAVEFILE_READ);
+    const plFileName fileName = M2ST(file);
+    plFileInfo fileInfo(fileName);
+    if (!fileInfo.Exists())
+        return nullptr;
+
+    waveFile.Open(fileName, nullptr, WAVEFILE_READ);
     int numMarkers = waveFile.GetNumMarkers();
     if (numMarkers == 0)
         return nullptr;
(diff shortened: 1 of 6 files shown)


Hoikas:
The second commit looks fine to me. I'll look at the first and third in more detail when I'm next at a PC.

Sources/Tools/MaxComponent/plMaxWaveUtils.cpp
```diff
@@ -62,3 +68,23 @@ SegmentMap *GetWaveSegmentMap(const MCHAR* file, plErrorMsg *pErrMsg)
...
+        waveFile.Open(M2ST(file), nullptr, WAVEFILE_READ);
+
+        return waveFile.GetHeader().fNumChannels;
+    } else {
+        plSoundBuffer* buf = new plSoundBuffer(fileName);
+        if (buf != nullptr && buf->IsValid())
+            return buf->GetHeader().fNumChannels;
+        return 0;
+    }
+}
```
Hoikas, line 90: Newline at EOF.

Sources/Tools/MaxComponent/plMaxWaveUtils.cpp
```diff
@@ -62,3 +68,23 @@ SegmentMap *GetWaveSegmentMap(const MCHAR* file, plErrorMsg *pErrMsg)
...
+uint16_t GetWaveNumChannels(const MCHAR* file)
+{
+    const plFileName fileName = M2ST(file);
+    plFileInfo fileInfo(fileName);
+    if (!fileInfo.Exists())
+        return 0;
+
+    if (fileName.GetFileExt().compare_ni("wav", 3) == 0) {
+        CWaveFile waveFile;
+        waveFile.Open(M2ST(file), nullptr, WAVEFILE_READ);
```
Hoikas, line 81: Why not:

```suggestion
        waveFile.Open(fileName, nullptr, WAVEFILE_READ);
```

Sources/Tools/MaxExport/plExportDlg.cpp
```diff
@@ -89,7 +89,7 @@ class plExportDlgImp : public plExportDlg
 
 public:
     plExportDlgImp();
-    ~plExportDlgImp();
+    ~plExportDlgImp() { }
```
Hoikas, line 92:
```suggestion
```

Sources/Tools/MaxComponent/plAutoUIBase.cpp
```diff
@@ -59,6 +59,7 @@ plAutoUIBase::~plAutoUIBase()
 {
     if (fDesc)
     {
+#if 0
```
Hoikas, line 62: We should probably strip this commit off and figure out the real problem.

dpogue:
updated

Hoikas:
What do you see? An unmerged-but-approved PR. Fixing...

PR #993, opened 2021-09-25
Fix custom avatar footsteps

dpogue:
The age and page names for footstep sounds are stored in the plArmatureMod data, but the effects were getting initialized before those values were read, so they were always using the defaults of "GlobalAvatars" and "Audio". This results in the custom footstep sounds for Quabs not taking effect in-game.

However, this will break footstep sounds for other custom avatars who *do* want to use the global sounds, because the age name is exported as "CustomAvatars" instead of "GlobalAvatars" in their plLODAvatarMod data.  
The custom avatar data can easily be fixed in the compiled data files in the moul-assets repo.

This change include an update to PlasmaMax so that if the avatar footstep page is set to "Audio" it will export with a hardcoded "GlobalAvatars" age, otherwise it will use the current age name. If the footstep page field is left blank, it will export an avatar with no plArmatureEffectsMgr (which is valid and supported by the engine).

This also includes a change to PlasmaMax to stop generating unneeded plClothingOutfit and plClothingBase objects for avatars with "(No Clothing Options)" selected. The engine is perfectly happy with avatars that have a null clothing outfit.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
@@ -1746,47 +1746,11 @@ void plArmatureMod::Read(hsStream * stream, hsResMgr *mgr)
 
     fBodyType = stream->ReadLE32();
 
-    if( stream->ReadBool() )
-    {
-        plKey effectMgrKey = mgr->ReadKey(stream);
-        mgr->AddViaNotify(effectMgrKey, new plGenRefMsg(GetKey(), plRefMsg::kOnCreate, -1, -1), plRefFlags::kActiveRef); // plArmatureEffects       
-
-        // Attach the Footstep emitter scene object
-        hsResMgr *mgr = hsgResMgr::ResMgr();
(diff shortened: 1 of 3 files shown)


Hoikas:
> However, this will break footstep sounds for other custom avatars who do want to use the global sounds, because the age name is exported as "CustomAvatars" instead of "GlobalAvatars" in their plLODAvatarMod data.
> The custom avatar data can easily be fixed in the compiled data files in the moul-assets repo.

Could we add a fallback to `GlobalAvatars` if the custom avatar age's audio page doesn't exist?

dpogue:
> Could we add a fallback to `GlobalAvatars` if the custom avatar age's audio page doesn't exist?

yes, fairly easily. Do we want to fall back to `GlobalAvatars_<customname>` or `GlobalAvatars_Audio`?

Hoikas:
Hmm, that's a good question. Do you have a preference? I can see arguments for both.

dpogue:
`GlobalAvatars_Audio` probably makes the most sense, that's the only option in GlobalAvatars currently

dpogue:
This should be ready to go

PR #994, opened 2021-10-02
vcpkg upgrade

Hoikas:
Routine vcpkg upgrade with some not-so-routine fixes. This pulls down a patch I submitted to fix the python3 build on GHA windows-2019 runners due to what appears to be a bug in the Windows SDK. Also, we now use the newfangled `VCPKG_DISABLE_COMPILER_TRACKING` to avoid including the hash of cl.exe in the binarycache. So, hopefully, folks will actually download the libraries from GitHub Packages now, instead of rebuilding them from scratch because they have a slightly newer or older VS.

Before merging:
- [x] Check that python39.dll is present in the artifacts.
- [x] Ensure all 3ds Max builds succeed.

Updated Libraries:
- asio 1.19.2
- curl 7.79.1
- libjpeg-tubo 2.0.6#1
- libvpx 1.10.0#2
- openssl 1.1.1l#1
- Python 3.9.7#1
- sqlite3 3.36.0
- zlib 1.2.11#12

### DIFF
--- a/Scripts/Triplets/x64-windows-plasma.cmake
+++ b/Scripts/Triplets/x64-windows-plasma.cmake
@@ -1,6 +1,6 @@
 set(VCPKG_TARGET_ARCHITECTURE x64)
 set(VCPKG_CRT_LINKAGE dynamic)
-
+set(VCPKG_DISABLE_COMPILER_TRACKING TRUE)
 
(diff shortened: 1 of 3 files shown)


Hoikas:
The binary cache hashes are still showing a mismatch from my local build and CI, but I think that may be because my CMake version does not match what's on CI. I don't think that should block this PR, though.

dpogue (approved):

PR #997, opened 2021-10-05
More PhysX Character Handling

Hoikas:
More work to battle trouble points with the PhysX Character Handling. Specifically addressed:
- It should no longer be possible to get stuck falling wedged between nearly parallel 90 degree colliders or against collisions of angles > 90 degrees (Relto Hut Lamps)
- There should no longer be a visible toss-back effect when jumping into a wall.
- Remaining toe-stubs should be less grabby than before (resolved problems on the Relto stairs, Kahlo Pub, Kadish X2 Vault stairs).
- Increased the fall threshold back to 60 degrees to better match the previous behavior.
- The avatar should no longer fall down the threshold of stairs.
- The stair slow down effect should be imperceptible now (we do **not** go ***faster up the stairs***, though, unlike MOULa/PX26)
- Grazing walls should no longer kill the avatar's momentum, especially when ascending stairs.


Hoikas:
Closing in favor of #1127.

PR #1001, opened 2021-10-08
SDL Console Command Improvements

### LINKED ISSUE #987: Add Console command to reset all Age SDL to default values
I couldn't see anything that obviously achieved this result, but it seems like a useful thing for Age developers to be able to reset an Age to its initial state.

There's probably a more convoluted way to achieve this by removing the Age instance and creating a new one, but a simple `Age.ResetSDLStates` that happened in real-time would be ideal

Hoikas:
This addresses some problems with the Age SDL console commands:

There are three different forms of Age SDL, two of which the client knows about: vault SDL and python/gameserver SDL. On DirtSand, these are always equivalent, but not so on other server types. Previously, the ShowSDL command showed the vault SDL but the SetSDL commands modified the python SDL. Therefore, `Age.ShowSDL` has been split into two commands: `Age.ShowPythonSDL` and `Age.ShowVaultSDL`. @Deledrius can now be happy.

`Age.ResetPythonSDL` has been added. Closes #987.

Fixed bugs where scripts seem to assume that SDL bools will str to numbers. Fixed memory leaks in the SetSDL commands.

### DIFF
--- a/Scripts/Python/xPodBattery.py
+++ b/Scripts/Python/xPodBattery.py
@@ -83,16 +83,16 @@ behSpeaker04            = ptAttribResponder(25, "Beh: Speaker04")
 behSpotlight01          = ptAttribResponder(26, "Beh: Spotlight01")
 behSpotlight02          = ptAttribResponder(27, "Beh: Spotlight02")
 behSpotlight03          = ptAttribResponder(28, "Beh: Spotlight03")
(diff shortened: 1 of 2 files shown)


Hoikas:
I can certainly do that, but I'm uncertain as to how prevalent this hack is. So, I'll take some time to look through all of the scripts for it.

Hoikas:
It looks like it was just the one script that was being naughty. 😄

PR #1004, opened 2021-10-09
Fix memory leaks in pyVaultNodeOperationCallback.

Hoikas:
A long standing issue has been mysterious "key only" memory leaks when the `plResManager` destructs, as seen in the leak report in plasmadbg.log:
```
(10/09 07:00:12) 	Leaks in page Personal>psnlMYSTII[000d0022]:
(10/09 07:00:12) 		plSceneObject: (S:0xd0022 F:0 I:526 N:ImgrPhotoPlane02 C:[0,0]) (key only, 10 refs left)
(10/09 07:00:12) 		plLogicModifier: (S:0xd0022 F:0 I:134 N:cRgnImagerActivate_Enter C:[0,0]) (key only, 1 refs left)
(10/09 07:00:12) 		plLogicModifier: (S:0xd0022 F:0 I:135 N:cRgnImagerActivate_Exit C:[0,0]) (key only, 1 refs left)
(10/09 07:00:12) 		plResponderModifier: (S:0xd0022 F:0 I:336 N:RespImager2Button C:[0,0]) (key only, 1 refs left)
(10/09 07:00:12) 		plPythonFileMod: (S:0xd0022 F:0 I:86 N:PythonFileImager02 C:[0,0]) (key only, 13 refs left)
(10/09 07:00:12) 		plDynamicTextMap: (S:0xd0022 F:0 I:1 N:ImagerDynamic02_dynText C:[0,0]) (key only, 1 refs left)
```

These leaks correspond to the xSimpleImager.py script, which uses vaultOperation callbacks. The vault operation callback C++ code was incorrectly assuming that `PyObject_GetAttrString()` would return a borrowed reference and not decrementing the reference count. This is fixed by using `pyObjectRef` to correct and clarify the reference counting.

Unfortunately, Python still reports errors on shutdown, but this is a good step toward correcting them.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyObjectRef.h
+++ b/Sources/Plasma/FeatureLib/pfPython/pyObjectRef.h
@@ -64,7 +64,7 @@ public:
     pyObjectRef(PyObject* object, pyObjectNewRef_Type)
         : fPyObject(object)
     {
-        Py_INCREF(object);
+        Py_XINCREF(object);
     }
 
     pyObjectRef(std::nullptr_t, pyObjectNewRef_Type) = delete;


Hoikas:
I need to change `pyObjectNewRef` to safely accept `nullptr`, don't merge this yet.

PR #1009, opened 2021-11-01
Add remote python console based on named pipes

roblabla:
This allows easily debugging python by connecting with PuTTY to a named pipe.

How to use:

- Start Uru, wait for intro movie to finish
- Start PuTTY, and try to connect to the following Serial port: `\\.\pipe\URU-PYTHON-<PID>` (For instance, if URU is running with pid 5151, connect to `\\.\pipe\URU-PYTHON-5151`).
image
- Upon connecting, you will be prompted with a python prompt. You can type python in there!

How it works:

The `pfRemoteConsole` class is responsible for setting everything up. It essentially creates a thread that handles the named pipe (connecting it, reading from it to get a script to run, and writing the output of the script back). When it gets a python script to run, it sends a message back to the main thread by dispatching a `pfRemoteConsoleMsg`. The `pfRemoteConsole::MsgReceive` will then take care of executing this message, and signaling the thread back when the output is acquired. This ensures that the python logic executes on the main thread, which should avoid data races.

Note that I'm not 100% sure this design is sound. Mainly, is it OK to call `plMessage::Send` from a new thread?

TODO:

- [ ] Properly handle backspace and left/right arrow keys to edit lines
- [ ] Handle line history with the up/down arrows
- [ ] Clean up unhandled control characters so they don't end up in the python script.

### DIFF
--- a/Sources/Plasma/Apps/plClient/plClient.cpp
+++ b/Sources/Plasma/Apps/plClient/plClient.cpp
@@ -138,6 +138,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pfCharacter/pfConfirmationMgr.h"
 #include "pfCharacter/pfMarkerMgr.h"
 #include "pfConsole/pfConsole.h"
+#include "pfConsole/pfRemoteConsole.h"
 #include "pfConsole/pfConsoleDirSrc.h"
 #include "pfConsoleCore/pfConsoleEngine.h"
 #if defined(PLASMA_PIPELINE_DX)
@@ -1388,6 +1389,12 @@ bool plClient::StartInit()
     pfConsole::SetPipeline( fPipeline );
     fConsole->RegisterAs( kConsoleObject_KEY );     // fixedKey from plFixedKey.h
     fConsole->Init( fConsoleEngine );
+
+#ifndef PLASMA_EXTERNAL_RELEASE
+    fRemoteConsole = new pfRemoteConsole();
+    fRemoteConsole->RegisterAs( kRemoteConsoleObject_KEY );
+    fRemoteConsole->Init();
+#endif
 
     /// Init the font cache
     fFontCache = new plFontCache();
--- a/Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
@@ -1,0 +1,206 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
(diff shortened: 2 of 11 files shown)


Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -1389,6 +1390,12 @@ bool plClient::StartInit()
     fConsole->RegisterAs( kConsoleObject_KEY );     // fixedKey from plFixedKey.h
     fConsole->Init( fConsoleEngine );
 
+#ifndef PLASMA_EXTERNAL_RELEASE
+    fRemoteConsole = new pfRemoteConsole();
+    fRemoteConsole->RegisterAs( kRemoteConsoleObject_KEY );
```
dpogue, line 1395: It doesn't look like we ever need to address messages to the RemoteConsole by its key, so we probably don't need a fixed key entry for it.

Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
```diff
@@ -0,0 +1,206 @@
...
+        return true;
+    }
+    return hsKeyedObject::MsgReceive(msg);
+}
+
+bool canRun(ST::string& script) {
+    return script.contains("\n");
+}
+
+void handlePipeServer() {
```
dpogue, line 94: As the person trying to get this engine running on macOS & Linux, I'd really love a cross platform way of doing this

Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
```diff
@@ -0,0 +1,206 @@
...
+        if (isMultiline && !curScript.ends_with("\n\n")) {
+            SendFullBuf(pipe, "... ", 4);
+            continue;
+        }
+
+        // Send message.
+        pfRemoteConsoleMsg *cMsg = new pfRemoteConsoleMsg(curScript);
+        std::shared_ptr<pfRemoteConsoleMsgOutput> output(cMsg->GetOutput());
+
+        cMsg->Send(nullptr, true);
```
dpogue, line 181: Based on the implementation of [`plDispatch::IMsgEnqueue`](https://github.com/H-uru/Plasma/blob/37d26cae22728af72a6abcb9676f102421654488/Sources/Plasma/NucleusLib/pnDispatch/plDispatch.cpp#L231-L250) this should actually be safe from to call from a thread, as long as the second parameter is `true`.

It will lock a mutex before adding it to the queue of messages to be dispatched.

If the second parameter is false, it will also immediately try to dispatch it and probably call the receiver's `MsgReceive` handler on the thread which is probably not safe.

dpogue (commented):
This is a cool idea, and I'm glad someone else is brave enough to play with messages on threads 😛

Sources/Plasma/Apps/plClient/plClient.cpp
Hoikas: You still need a key to receive messages though, so this is not really incorrect.

Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
Hoikas: Part of me wonders if pnAsyncCore should learn about named pipes, then you get callbacks on the main thread for free. Then this silly worker thread and the synchronization questions go away. Maybe @zrax will have some input on this, considering the asio work.

Sources/Plasma/Apps/plClient/plClient.cpp
roblabla: This is mostly copy pasted from the fConsole code above. I used to not have the `RegisterAs` call, but the messages wouldn't ever arrive to the console. Is there another way to register the `plRemoteConsole` without a fixed key?

Sources/Plasma/FeatureLib/pfConsole/pfRemoteConsole.cpp
roblabla: So I actually gave some thought to this while writing the code. It should actually be fairly trivial to make the code cross-platform, by making a very lightweight abstraction layer between named pipes and unix sockets on linux (and you'd connect to the unix socket with socat or something when on macOS/Linux).

Regarding pnAsyncCore, I actually started by trying to add pipe support there, but was a bit daunted by how complex it looked, and since I wanted to have a quick proof of concept runnig, I quickly changed to the worker thread approach. I do agree that it'd probably be cleaner that way though. I'll try to look into it when I've fixed the remaining TODOs.

Hoikas:
This PR is maturing, so I recommend closing it. It can be reopened or recreated if more progress is made.

PR #1011, opened 2021-11-11
Make AsyncThreadTimedJoin available on other platforms.

zrax:
On Linux, this uses `pthread_timedjoin_np()`.  For other OSes (not Windows or Linux), we just fall back to a blocking join :/


dpogue (approved):

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
```diff
@@ -61,3 +64,32 @@ void ThreadDestroy (unsigned exitThreadWaitMs) {
...
+//============================================================================
+void AsyncThreadTimedJoin(std::thread& thread, unsigned timeoutMs)
+{
+    // HACK: No cross-platform way to perform a timed join :(
+#if defined(HS_BUILD_FOR_WIN32)
+    DWORD rc = WaitForSingleObject(thread.native_handle(), timeoutMs);
+    if (rc == WAIT_TIMEOUT)
+        LogMsg(kLogDebug, "Thread did not terminate after {} ms", timeoutMs);
+    thread.detach();
+#elif defined(HS_BUILD_FOR_LINUX)
```
Hoikas, line 77: I wonder if this should be a cmake check based on the provider of `Threads::Threads`. Doesn't pthread exist and offer this functionality on BSD?

zrax: pthreads exist on BSD and Darwin, but the `pthread_timedwait_np` is a non-portable extension, which AFAIK is Linux-only.

Hoikas: Some BSDs seem to advertise `pthread_timedjoin_np` in their manpages, but Darwin support is probably more important :/. Example: https://www.freebsd.org/cgi/man.cgi?pthread_join

zrax: Ah, good to know...  Perhaps a better way to do this would be to test for the presence of the function.  If I don't forget, I'll make a PR in the future to do that

PR #1016, opened 2021-11-12
Add RelWithPDB configuration for MSVC.

Hoikas:
This is for the upcoming redesigned crash dialog.

The RelWithDebInfo configuration is great for trivially getting a PDB file, but it doesn't do the full level of optimizations that Release
does. So, this adds a new build type that sets the optimization levels up to Release but ensures a PDB file is generated. This configuration should only be used for actual releases. For general development level hacking, RelWithDebInfo is more appropriate.


CMakeLists.txt
```diff
@@ -48,6 +48,32 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
...
+    endif()
+
+    # Yeah, this looks strange. The semicolon is to establish an empty list item
+    # for the configuration-less import.
+    set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHPDB "Release;")
+
+    # Need /Zi for the PDB
+    set(CMAKE_C_FLAGS_RELWITHPDB   "/MD /Zi /O2 /Ob2 /DNDEBUG" CACHE STRING "")
+    set(CMAKE_CXX_FLAGS_RELWITHPDB "/MD /Zi /O2 /Ob2 /DNDEBUG" CACHE STRING "")
+    set(CMAKE_EXE_LINKER_FLAGS_RELWITHPDB "/INCREMENTAL:NO /DEBUG:FULL /OPT:REF,NOICF" CACHE STRING "")
```
Hoikas, line 64: The `/DEBUG` flag doesn't mean "debug build", rather, it means "generate debug information". The PDB will not be generated without it.

Deledrius (approved):
Nice work! :)

CMakeLists.txt
```diff
@@ -48,6 +48,32 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
...
+        set(CMAKE_CONFIGURATION_TYPES "${CMAKE_CONFIGURATION_TYPES};RelWithPDB" CACHE STRING "" FORCE)
+    endif()
+
+    # Yeah, this looks strange. The semicolon is to establish an empty list item
+    # for the configuration-less import.
+    set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHPDB "Release;")
+
+    # Need /Zi for the PDB
+    set(CMAKE_C_FLAGS_RELWITHPDB   "/MD /Zi /O2 /Ob2 /DNDEBUG" CACHE STRING "")
+    set(CMAKE_CXX_FLAGS_RELWITHPDB "/MD /Zi /O2 /Ob2 /DNDEBUG" CACHE STRING "")
```
zrax, line 63: Maybe it would be more future-proof to say
```suggestion
    set(CMAKE_C_FLAGS_RELWITHPDB   "${CMAKE_C_FLAGS_RELEASE} /Zi" CACHE STRING "")
    set(CMAKE_CXX_FLAGS_RELWITHPDB "${CMAKE_CXX_FLAGS_RELEASE} /Zi" CACHE STRING "")
```

zrax (approved):

PR #1018, opened 2021-11-12
Improve detection of pthread_timedjoin_np

zrax:
This will now use `pthread_timedjoin_np()` on any platform which provides it with a compatible API, rather than restricting its use to Linux.

### DIFF
--- a/cmake/CompilerChecks.cmake
+++ b/cmake/CompilerChecks.cmake
@@ -23,6 +23,10 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     endif()
 endif()
 
+try_compile(HAVE_PTHREAD_TIMEDJOIN_NP ${PROJECT_BINARY_DIR}
+            ${PROJECT_SOURCE_DIR}/cmake/check_pthread_timedjoin_np.cpp
+            LINK_LIBRARIES Threads::Threads)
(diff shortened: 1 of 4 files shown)


dpogue (approved):

cmake/CompilerChecks.cmake
```diff
@@ -23,6 +23,10 @@ if(NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
     endif()
 endif()
 
+try_compile(HAVE_PTHREAD_TIMEDJOIN_NP ${PROJECT_BINARY_DIR}
+            ${PROJECT_SOURCE_DIR}/cmake/check_pthread_timedjoin_np.cpp
+            LINK_LIBRARIES Threads::Threads)
+
```
Hoikas, line 29: Why not
```suggestion
include(CheckCXXSymbolExists)
check_cxx_symbol_exists(pthread_timedjoin_np pthread.h HAVE_PTHREAD_TIMEDJOIN_NP)
```

zrax: That was my first attempt, but getting it to correctly use the pthread library was proving both difficult and ugly...  `try_compile` makes it easy to add the appropriate library flags.

Hoikas (approved):

PR #1024, opened 2021-11-17
Better Crash Dialog (Part 1)

Hoikas:
This moves the crash dialog out of the plClient.exe process and enhances it to display a stack trace again by using the Windows DbgHelp library. The reason for this change is that with many age creators now using our clients to iterate their Ages, we need a more trivially communicable way to diagnose problems. While the crash.dmp functionality is nice (and has been retained), there are a number of hoops to jump through before any worthwhile debugging can commence. This gives us options in that if a user takes a screenshot of the dialog (a common anti-pattern), we at least get some information. We get even more information if they press the copy button.

Future work will be to add a third command link button to enable uploading the crash dumps to a remote collection. This will likely require some design discussion, so I am punting on that for now. I would appreciate better help text suggestions for the two links that I have currently because they kind of suck 😛.

Screenshot 2021-11-10 022253

Note that this screenshot was made when plClient.pdb is a stripped pdb. If a full-fat pdb is present, then file and line numbers will be present. If no pdb is present, then only addresses (no function names) will be shown.


Hoikas:
1. Done in #1106.
2. I think that's fair. IIRC Microsoft's guidelines allow for command links to *not* have the subline text, they just recommend it.
3. The default size of the columns is hardcoded. I remember that it was really easy to get the columns into odd states, and the whole thing was just very frustrating in general. I can try to make the symbol size column a touch wider, but no promises.
4. Right now, the trace starts at the top of the stack and goes until DbgHelp is no longer able to unwind it or find matching symbols. It wouldn't be hard to add a limit, but why do you think that would be needed?

ZarothYe:
> 1. Done in [Fix crash when `plClipboard::SetClipboardText()` is called twice. #1106](https://github.com/H-uru/Plasma/pull/1106).
> 2. I think that's fair. IIRC Microsoft's guidelines allow for command links to _not_ have the subline text, they just recommend it.
> 3. The default size of the columns is hardcoded. I remember that it was really easy to get the columns into odd states, and the whole thing was just very frustrating in general. I can try to make the symbol size column a touch wider, but no promises.
> 4. Right now, the trace starts at the top of the stack and goes until DbgHelp is no longer able to unwind it or find matching symbols. It wouldn't be hard to add a limit, but why do you think that would be needed?

I think I followed up on this in DM, but forgot to record here. I am happy with the state of this crash dialog, for whatever that's worth. It was working well with your updates to items 1 and 2 on the current TL build from what I could tell. Regarding your answer to 4, I didn't realize it would stop automatically rather than at a particular depth limit, so that seems fine to me.

Hoikas:
I've rebased this and fixed plCrashHandler.exe to use a manifest. The crash dialog looks like this, assuming no crash information:

image

ZarothYe (approved):

ZarothYe:
@Deledrius @dpogue @zrax Would it be possible to get a review on the latest version so we could get this goodness merged, possibly? Consider this a bat signal... I think we could use y'all's help. :)

dpogue (approved):
This seems fine to me (with a non-blocking "please think about how to make this cross-platform" plea 😛)

Hoikas:
Once we have a baseline "working" macOS or Linux client in mainline, we'll have a lot of those kinds of questions to answer 😄

PR #1027, opened 2021-11-25
Prevent sending blank/space KI messages in variety of contexts; Fix /invite so it doesn't silently crash

ZarothYe:
- Update KI chat command processor to discard blank/whitespace messages, including messages used with /me and /buddies, among other commands
- Remove usage of nonexistent constant so /invite command doesn't silently crash and instead actually creates an invite with passkey

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -866,27 +866,38 @@ class CommandsProcessor:
                 else:
                     statusMsg = PtGetLocalizedString(emote[1], [PtGetLocalPlayer().getPlayerName()])
                 self.chatMgr.DisplayStatusMessage(statusMsg, 1)
+
+                # Get remaining message string after emote command
                 message = message[len(words[0]):]
-                if message == "":
+                if message == "" or message.isspace():
                     return None
                 return message[1:]
             except LookupError:
                 try:
                     command = xKIExtChatCommands.xChatExtendedChat[str(words[0][1:].casefold())]
                     if isinstance(command, str):
+                        # Retrieved command is just a plain string
                         args = message[len(words[0]):]
                         PtConsole(command + args)
                     else:
+                        # Retrieved command is not a string (it's a function)
                         try:
+                            # Get remaining message string after chat command
                             args = message[len(words[0]) + 1:]
                             if args:
                                 try:
                                     retDisp = command(args)
                                 except TypeError:
                                     retDisp = command()
+
+                                    # Command took no args, so return args as new message
+                                    if args == "" or args.isspace():
+                                        return None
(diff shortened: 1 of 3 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -866,27 +866,38 @@ def __call__(self, message):
                 else:
                     statusMsg = PtGetLocalizedString(emote[1], [PtGetLocalPlayer().getPlayerName()])
                 self.chatMgr.DisplayStatusMessage(statusMsg, 1)
+
+                # Get remaining message string after emote command
                 message = message[len(words[0]):]
-                if message == "":
+                if message == "" or message.isspace():
```
Hoikas, line 872: Seeing as how this is repeated a few times, I wonder if it would make more sense to compile the regex `\s*` somewhere and use that as our magic "is this a blank or empty string" test.

Scripts/Python/xCheat.py
```diff
@@ -678,6 +678,7 @@ def ImportMarkers(args):
                     if jnode.folderGetName() == mg[1]:
                         # yes, add the markers to this game
                         for marker in mg[0][4]:
+                            # TODO: PlasmaVaultConstants.PtVaultNodePermissionFlags does not exist
```
Hoikas, line 681: This code is actually dead because markers don't work this way anymore. I'd remove this given there are bigger problems afoot 😉

Scripts/Python/xInvite.py
```diff
@@ -172,7 +172,7 @@ def DeleteInvitation(params=None):
     return PtGetLocalizedString("KI.Invitation.DeletedInvitation") + str(passkey)
 
 def MeChat(params=None):
-    if (params == None):
+    if (params == None or params == "" or params.isspace()):
```
Hoikas, line 175:
```suggestion
    if (not params or params.isspace()):
```

Empty strings are Falsey in Python.

Scripts/Python/xCheat.py
ZarothYe: @Hoikas How much code did you want me to remove there? Or did you just mean remove my comment?

Scripts/Python/xInvite.py
ZarothYe: Updated according to your suggestions

Scripts/Python/xCheat.py
Hoikas: Just the comment - fixing the actual function is definitely a "later" task.

ZarothYe: TODO Comment has been removed from this code

Hoikas (approved):

PR #1028, opened 2021-11-25
Fix Python deprecation warnings because of \s and \d escape characters in regex patterns

ZarothYe:
- Change \s and \d escape characters in some python regex pattern strings to use \\s and \\d so they no longer throw deprecation warnings
- Update missing player name check on Create Player dialog so it doesn't repeat (playerName == "") twice

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -1452,7 +1452,7 @@ class CommandsProcessor:
             return
 
         # Parse common dice notation
-        dice_opt = re.match("^(\d+)d(\d+)$", dice_str)
+        dice_opt = re.match("^(\\d+)d(\\d+)$", dice_str)
         if not dice_opt:
             self.chatMgr.AddChatLine(None, "I'm sorry, I don't know how to roll {}.".format(dice_str), kChat.SystemMessage)
             return
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -1452,7 +1452,7 @@ class CommandsProcessor:
             return
 
         # Parse common dice notation
-        dice_opt = re.match("^(\\d+)d(\\d+)$", dice_str)
+        dice_opt = re.match(r"^(\d+)d(\d+)$", dice_str)
         if not dice_opt:
             self.chatMgr.AddChatLine(None, "I'm sorry, I don't know how to roll {}.".format(dice_str), kChat.SystemMessage)
             return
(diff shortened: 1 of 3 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -1452,7 +1452,7 @@ def RollDice(self, dice_str):
             return
 
         # Parse common dice notation
-        dice_opt = re.match("^(\d+)d(\d+)$", dice_str)
+        dice_opt = re.match("^(\\d+)d(\\d+)$", dice_str)
```
zrax, line 1455: Perhaps an even better fix would be to use raw string literals:
```suggestion
        dice_opt = re.match(r"^(\d+)d(\d+)$", dice_str)
```

Scripts/Python/xDialogStartUp.py
```diff
@@ -353,7 +353,7 @@ def OnGUINotify(self,id,control,event):
                         self.ToggleColor(GUIDiag4b, k4bPlayer03)
                     else:
                         fixedPlayerName = playerName.strip()
-                        (fixedPlayerName, whitespacefixedcount) = re.subn("\s{2,}|[\t\n\r\f\v]", " ", fixedPlayerName)
+                        (fixedPlayerName, whitespacefixedcount) = re.subn("\\s{2,}|[\t\n\r\f\v]", " ", fixedPlayerName)
```
zrax, line 356:
```suggestion
                        (fixedPlayerName, whitespacefixedcount) = re.subn(r"\s{2,}|[\t\n\r\f\v]", " ", fixedPlayerName)
```

Scripts/Python/ki/xKIChat.py
ZarothYe: I saw raw strings as an alternate option in my researching, but I was familiar with using \\ in other languages so I favored this option. It looked like \\s was being used in a couple places in xKIHelpers.py as well. Is there a reason why a raw string would be better than using \\ escapes?

Hoikas: We generally prefer raw strings instead of manually escaping backslashes due to the clarity it offers. This way, we're looking at the exact regex instead of having to go through the cognitive load of decoding the regex any time we want to tweak it. It's nice to be able to copy and paste directly from regex101.com, for example. The only time we'd probably want to *not* use raw strings is if there is a need to use specific unicode codepoints.

ZarothYe: Ok, I will update those places. Did you want me to update the uses in xKIHelpers.py as well to use raw strings?

Hoikas: If those are also raising DeprecationWarnings, that would be great!

ZarothYe: They are actually using the \\\\s version already, so they don't throw any warnings. But if we wanted to keep things more consistent?

Hoikas: It would be a welcome change 😄

ZarothYe: Adjustments complete to use raw strings in the 2 deprecation warning locations and the 2 locations that were using \\\\s

Hoikas (approved):

PR #1029, opened 2021-11-26
Change chat mentions to only highlight player's name instead of entire message

ZarothYe:
- Chat mentions will now only highlight the player's name in the message instead of the entire message. Intended to avoid confusion for players who mistake mentions for private/whisper messages
- When player's name is contained within a URL, the URL handling will take precedence and the name will not be highlighted
- Add DniBlueDk (medium dusky blue) color that matches KI lights for potential use later

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -102,7 +102,6 @@ class xKIChat(object):
 
         # Set a fake player name to avoid errors.
         self.playerName = None
-        self._chatMentionRegex = None
 
         # Message History
         self.MessageHistoryIs = -1 # Current position in message history (up/down key)
@@ -144,11 +143,16 @@ class xKIChat(object):
 
     # Update the current player name.
     def _setPlayerName(self, value):
-        if value:
-            value = re.escape(value)
-            regex = rf"(?:^|[\s\W](?<!\/\/|\w\.))(?P<mention>({value}\s?)+)(?=$|[\s\W])"
-            PtDebugPrint(f"xKIChat: The chat mention regex is now `{regex}`", level=kWarningLevel)
-            self._chatMentionRegex = re.compile(regex, re.IGNORECASE)
+        value = re.escape(value if value else "Anonymous Coward")
+
+        # (?:^|[\s\W](?<!\/\/|\w\.)) - non-capturing group: line start or whitespace or non-word character
+        #                              with lookbehind that excludes matches preceded by // or word character and .
+        #                              trying to prevent mentions that are part of a URL
+        # (?P<mention>({value}\s?)+) - named capture group: one or more occurrence of name, optionally split by a space
+        # (?=$|[\s\W])               - lookahead ensures match is followed by line end or whitespace or non-word character
+        regex = rf"(?:^|[\s\W](?<!\/\/|\w\.))(?P<mention>({value}\s?)+)(?=$|[\s\W])"
+        PtDebugPrint(f"xKIChat: The chat mention regex is now `{regex}`", level = kWarningLevel)
+        self._chatMentionRegex = re.compile(regex, re.IGNORECASE)
     playerName = property(None, _setPlayerName)
 
     ## Make the player enter or exit chat mode.
@@ -488,7 +492,7 @@ class xKIChat(object):
                         self.lastPrivatePlayerID = (player.getPlayerName(), player.getPlayerID(), 1)
                         PtFlashWindow()
                     # Are we mentioned in the message?
-                    elif self._chatMentionRegex is not None and self._chatMentionRegex.search(message) is not None:
+                    elif self._chatMentionRegex.search(message) is not None:
                         hasMention = True
                         PtFlashWindow()
 
@@ -525,19 +529,13 @@ class xKIChat(object):
                 if cFlags.toSelf:
                     headerColor = kColors.ChatHeaderBroadcast
                     pretext = PtGetLocalizedString("KI.Chat.BroadcastSendTo")
-
-                    # Are we mentioned in the message?
-                    if self._chatMentionRegex is not None and self._chatMentionRegex.search(message) is not None:
-                        hasMention = True
-                        forceKI = True
-                        PtFlashWindow()
                 else:
                     headerColor = kColors.ChatHeaderBroadcast
                     pretext = PtGetLocalizedString("KI.Chat.BroadcastMsgRecvd")
                     self.AddPlayerToRecents(player.getPlayerID())
 
                     # Are we mentioned in the message?
-                    if self._chatMentionRegex is not None and self._chatMentionRegex.search(message) is not None:
+                    if self._chatMentionRegex.search(message) is not None:
                         hasMention = True
                         forceKI = True
                         PtFlashWindow()
@@ -583,9 +581,10 @@ class xKIChat(object):
             else:
                 chatMessageFormatted = " {}".format(message)
 
-        chatMentions = []
         if hasMention:
-            chatMentions = list(((i.start("mention"), i.end("mention"), i.group("mention")) for i in self._chatMentionRegex.finditer(chatMessageFormatted)))
+            chatMentions = [(i.start("mention"), i.end("mention"), i.group("mention")) for i in self._chatMentionRegex.finditer(chatMessageFormatted)]
+        else:
+            chatMentions = []
 
         for chatArea in (self.miniChatArea, self.microChatArea):
             with PtBeginGUIUpdate(chatArea):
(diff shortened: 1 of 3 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -401,6 +404,21 @@ def SendMessage(self, message):
     ## Adds a line to the RT chat.
     def AddChatLine(self, player, message, cFlags, forceKI=True):
 
+        playerName = PtGetClientName()
+        playerNameMentionRegex = None
+
+        # Compose the regex differently depending on if player name starts and ends with a word or with special characters
+        if _STARTS_WITH_WORD_REGEX.search(playerName) != None:
```
Hoikas, line 411:
```suggestion
        if _STARTS_WITH_WORD_REGEX.search(playerName) is not None:
```

`is` is preferred for checking singletons like `None`.

Scripts/Python/ki/xKIChat.py
```diff
@@ -401,6 +404,21 @@ def SendMessage(self, message):
...
+            if _ENDS_WITH_WORD_REGEX.search(playerName) != None:
+                playerNameMentionRegex = re.compile(fr"\b({re.escape(playerName)})\b", re.IGNORECASE)
+            else:
+                playerNameMentionRegex = re.compile(fr"\b({re.escape(playerName)})(?=$|\W)", re.IGNORECASE)
+        else:
+            if _ENDS_WITH_WORD_REGEX.search(playerName) != None:
+                playerNameMentionRegex = re.compile(fr"(?=^|\W)({re.escape(playerName)})\b", re.IGNORECASE)
+            else:
+                playerNameMentionRegex = re.compile(fr"(?=^|\W)({re.escape(playerName)})(?=$|\W)", re.IGNORECASE)
```
Hoikas, line 420: @Deledrius You're a bit better with regexes than I am. Any thoughts about this?

Scripts/Python/ki/xKIChat.py
ZarothYe: Updated all the occurrences I made with "!= None" to be "is not None" instead.

Scripts/Python/ki/xKIChat.py
Hoikas: I've posted an alternative at ZarothYe/Plasma#1 for review.

Scripts/Python/ki/xKIChat.py
```diff
@@ -103,6 +100,10 @@ def __init__(self, StartFadeTimer, ResetFadeState, FadeCompletely, GetCensorLeve
         # Add the commands processor.
         self.commandsProcessor = CommandsProcessor(self)
 
+        # Set a fake player name to avoid errors.
+        self.playerName = None
+        self._chatMentionRegex = None
```
Hoikas, line 105:
```suggestion
```

Overrides the "default" (read: error case) regex, complicating the code below. The player name is guaranteed to be set by `OnAccountUpdate()` before any chat can come in because there is no connection to any game (read: chat) server when the player is unset.

Scripts/Python/ki/xKIChat.py
```diff
@@ -529,15 +525,20 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
                 if cFlags.toSelf:
                     headerColor = kColors.ChatHeaderBroadcast
                     pretext = PtGetLocalizedString("KI.Chat.BroadcastSendTo")
+
+                    # Are we mentioned in the message?
+                    if self._chatMentionRegex is not None and self._chatMentionRegex.search(message) is not None:
```
Hoikas, line 530:
```suggestion
                    if self._chatMentionRegex.search(message) is not None:
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -529,15 +525,20 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
...
                 else:
                     headerColor = kColors.ChatHeaderBroadcast
                     pretext = PtGetLocalizedString("KI.Chat.BroadcastMsgRecvd")
                     self.AddPlayerToRecents(player.getPlayerID())
 
                     # Are we mentioned in the message?
-                    if playerNameMentionRegex.search(message) is not None:
-                        specialColor = kColors.ChatMessageMention
-                        specialMatchPattern = playerNameMentionRegex
+                    if self._chatMentionRegex is not None and self._chatMentionRegex.search(message) is not None:
```
Hoikas, line 540:
```suggestion
                    if self._chatMentionRegex.search(message) is not None:
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -582,6 +583,10 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
             else:
                 chatMessageFormatted = " {}".format(message)
 
+        chatMentions = []
+        if hasMention:
+            chatMentions = list(((i.start("mention"), i.end("mention"), i.group("mention")) for i in self._chatMentionRegex.finditer(chatMessageFormatted)))
```
Hoikas, line 588:
```suggestion
        if hasMention:
            chatMentions = ((i.start("mention"), i.end("mention"), i.group("mention")) for i in self._chatMentionRegex.finditer(chatMessageFormatted)))
        else:
            charMentions = ()
```

Generator comprehensions avoid allocating the entire list

ZarothYe: I needed it **not** to be a generator because it will be requested once and iterated twice (which I found out a couple days ago is not possible to do with an iterator/generator) -- for chatArea in (miniKI, microKI). I figured doing the .finditer once prior to that for-in loop might improve performance, thus the need for it to be iterable more than one time.

Hoikas: That's right. This has to be iterated through for two chat areas. In that case, one small adjustment. Instead of creating a list from a generator comprehension `list((...))`, use a list comprehension instead `[...]`.

Scripts/Python/ki/xKIChat.py
ZarothYe: I changed the playerName property setter not to set the _chatMentionRegex if the value was None (it seemed a little silly to always have to compile a "fake" regex the first time when we will never want it or use it), and I added the `self._chatMentionRegex = None` because I like variable declarations... even when not strictly necessary. I can get rid of the declaration, though, and simplify the mention check cases you noted if we are certain AddChatLine will never be called without a real playerName set. Is it important to do the default/error/fake playerName set process, though?

Scripts/Python/ki/xKIChat.py
ZarothYe: Ok, I will do that. :) I will need to google python comprehensions because I don't understand it yet, though. Expect a push later today after work hours.

Scripts/Python/ki/xKIChat.py
Hoikas: I wanted it to be that way for consistency... when you return to the character selection screen, `OnAccountUpdate()` will be called again, and `PtGetClientName()` will return an empty string. So, it's important in my mind to handle "crap, there is an empty string here" and do it the same way each time. Minimal surprises and all. Once you login to another avatar, the same thing happens, except `PtGetClientName()` will return the avatar name. Related: the engine itself will specify that KI messages from unknown players are from "Anonymous Coward", which is where I got that.

Scripts/Python/ki/xKIChat.py
ZarothYe: Oh, I get it. So `chatMentions = [(i.start("mention"), i.end("mention"), i.group("mention")) for i in self._chatMentionRegex.finditer(chatMessageFormatted)]` will create a list from the iterable without going through the intermediate generator comprehension step that I had before, right?

Hoikas: You've got it :+1:

Scripts/Python/ki/xKIChat.py
ZarothYe: Hm, alright. I'll remove the if check in the setter. I guess the extra computation is minimal.

Scripts/Python/ki/xKIChat.py
ZarothYe: Committed adjustments for this and for your previous comments as well. I've finished kicking the tires on the regex (as well as added piece-by-piece comments for those who have to read that crap later), so I'm good with this all if you're good.

Hoikas (approved):
Yeah, this all looks good to me. Since I was a little heavy-handed involved here, I want to give other maintainers an opportunity to chime in as well.

Scripts/Python/ki/xKIChat.py
Deledrius: The new method seems more robust to me. :)

Deledrius (approved):

PR #1032, opened 2021-12-10
[Do Not Merge] Temp stub pnAceSocket impl for Linux

dpogue:
I'm guessing this is not useful, except as a local patch for people wanting to _compile_ the full engine on Linux without wanting to _run_ it.

This allows pnAsyncCoreExe and all the net code to compile on Linux/macOS. However, trying to run a client/tool with this will almost definitely result in weird and broken behaviour as soon as it tries to do anything useful with a socket.


Hoikas:
CI failures are due to a defective release of actions/upload-artifact. See actions/upload-artifact#270.

Hoikas:
Probably conflicts with #1066.

dpogue:
yeah, this should never be merged under any circumstances 😛  #1066 is/will be the correct implementation.

PR #1036, opened 2021-12-16
Remove TIam from Nexus

DoobesURU:
Temporarily removes Tiam listings from the Nexus until a Linking Panel can be added.

### DIFF
--- a/Scripts/Python/nxusBookMachine.py
+++ b/Scripts/Python/nxusBookMachine.py
@@ -203,9 +203,9 @@ kLinkPanels = {
 }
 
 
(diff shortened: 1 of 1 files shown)


Hoikas:
There is currently no (nor any PR adding) an Age called Tiam.

DoobesURU:
> There is currently no (nor any PR adding an) Age called Tiam.

Was waiting to see if ametist would add it, but sounds like she's not familiar with git, so I just did it for her. Two new PRs in Plasma and moul-assets have now been added for Tiam.

PR #1039, opened 2021-12-18
Runtime renderer selection

dpogue:
Refactors the pipeline projects such that both DirectX and OpenGL can be compiled at the same time, and implements a `--Renderer=opengl` option to pick the preferred renderer.

Metal can be handled the same way when it is in a usable state.

### Renderer arguments

| Renderer | Valid argument values (case-insensitive) |
| --- | --- |
| **pfDXPipeline** | `directx`, `dx`, `direct3d`, `d3d` |
| **pfGLPipeline** | `opengl`, `gl` |

### Notes:
* Building pfGLPipeline is disabled by default
* hsG3DDeviceSelector will currently never return an OpenGL device
* Specifying a renderer argument that is not enabled/found will exit with an error about no devices found
* hsG3DDevice stuff is a mess and needs some thought about how to restructure it


Sources/Plasma/Apps/plClient/winmain.cpp
```diff
@@ -972,6 +975,25 @@ LONG WINAPI plCustomUnhandledExceptionFilter( struct _EXCEPTION_POINTERS *Except
...
 }
 
+uint32_t ParseRendererArgument(const ST::string& requested)
+{
+    using namespace ST::literals;
+
+    if (requested.compare_i("directx"_st) == 0
+            || requested.compare_i("direct3d"_st) == 0
+            || requested.compare_i("dx"_st) == 0
+            || requested.compare_i("d3d"_st) == 0) {
```
Hoikas, line 985: Part of me wonders if it would be better to check against a set of values (or use a regex) for cleaner code... eg
```C++
static std::unordered_map<ST::string, ST::hash_i, ST::equal_i> dx{
    "directx"_st, "direct3d"_st, "dx"_st, "d3d"_st
};
if (dx.find(requested) != dx.end())
    return hsG3DDeviceSelector::kDevTypeDirect3D;
```

dpogue: Sounds like a good plan to me

Deledrius (approved):

Hoikas (approved):

PR #1040, opened 2021-12-18
Allow mouse-free and message-free KI commands to select KI folders: /age, /neighbors, and /buddies

ZarothYe:
This allows users to change chat context for future messages without clicking on the KI folders. For example, if you are PMing/whispering to someone, then want to stop whispering and start sending future messages to all Age Players, you can switch context by using "/age\<Enter\>" and it behaves like you have clicked the AGE PLAYERS folder from your KI list.

Should be merged alongside [this moul-assets PR](https://github.com/H-uru/moul-assets/pull/95).

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -952,11 +952,45 @@ class CommandsProcessor:
                         except:
                             PtDebugPrint("xKIChat.commandsProcessor(): Chat command function did not run.", command, level=kErrorLevel)
                 except LookupError:
-                    if str(words[0].casefold()) in xKIExtChatCommands.xChatSpecialHandledCommands:
+                    firstWordLower = str(words[0].casefold())
+                    if firstWordLower in xKIExtChatCommands.xChatSpecialHandledCommands:
                         # Get remaining message string after special chat command
                         remainingMsg = message[len(words[0]):]
+
+                        # If remaining message is empty, try to do mousefree KI folder selection instead
                         if remainingMsg == "" or remainingMsg.isspace():
+                            userListBox = ptGUIControlListBox(KIMini.dialog.getControlFromTag(kGUI.PlayerList))
+                            caret = ptGUIControlTextBox(KIMini.dialog.getControlFromTag(kGUI.ChatCaretID))
+                            privateChbox = ptGUIControlCheckBox(KIMini.dialog.getControlFromTag(kGUI.miniPrivateToggle))
+
+                            # Handling for selecting Age Players, Buddies, or Neighbors
+                            if firstWordLower == PtGetLocalizedString("KI.Commands.ChatAge"):
+                                folderName = xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kAgeMembersFolder)
+                                caretValue = ">"
+                            elif firstWordLower == PtGetLocalizedString("KI.Commands.ChatBuddies"):
+                                folderName = xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kBuddyListFolder)
+                                caretValue = PtGetLocalizedString("KI.Chat.TOPrompt") + folderName + " >"
+                            elif firstWordLower == PtGetLocalizedString("KI.Commands.ChatNeighbors"):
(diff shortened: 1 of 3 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -952,11 +952,45 @@ def __call__(self, message):
...
+                                folderIdx = 0
+                                while folderIdx < numElements:
+                                    if userListBox.getElement(folderIdx).casefold() == folderName.casefold():
+                                        break;
+                                    folderIdx += 1
+
+                                if folderIdx < numElements:
+                                    userListBox.setSelection(folderIdx)
+                                    caret.setStringW(caretValue)
+                                    privateChbox.setChecked(0)
```
Hoikas, line 989:
```suggestion
                                    privateChbox.setChecked(False)
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -952,11 +952,45 @@ def __call__(self, message):
                         except:
                             PtDebugPrint("xKIChat.commandsProcessor(): Chat command function did not run.", command, level=kErrorLevel)
                 except LookupError:
-                    if str(words[0].casefold()) in xKIExtChatCommands.xChatSpecialHandledCommands:
+                    firstWordLower = str(words[0].casefold())
```
Hoikas, line 955:
```suggestion
                    firstWordLower = words[0].casefold()
```

Yeah, it was goofy before, but let's go ahead and fix it.

Deledrius:
Awesome!  Looks like this PR has a few extraneous commits in it, though.

Scripts/Python/ki/xKIChat.py
```diff
@@ -952,11 +952,45 @@ def __call__(self, message):
...
+                            # Only try to find in list if it was one of the 3 expected folders, not a reply command
+                            if folderName is not None:
+                                numElements = userListBox.getNumElements()
+                                folderIdx = 0
+                                while folderIdx < numElements:
+                                    if userListBox.getElement(folderIdx).casefold() == folderName.casefold():
+                                        break;
+                                    folderIdx += 1
+
+                                if folderIdx < numElements:
```
Hoikas, line 986: Let's use a for loop instead. We could also flatten this out to be more Pythonic, eg:

```suggestion
                                try:
                                    folderIdx = next((i for i in range(userListBox.getNumElements()) if userListBox.getElement(folderIdx).casefold() == folderName.casefold()), None)
                                except StopIteration:
                                    # Should probably indicate an error to the user here because the folder was not found.
                                    pass # fixme, see above
                                else:
```

Scripts/Python/ki/xKIChat.py
ZarothYe: Oh, setChecked(0) was what __init__.py used in some similar handling, but it should handle both the same, right?

Hoikas: Yes, in Python boolean is a subclass of int:

```
>>> isinstance(False, int)
True
```

Scripts/Python/ki/xKIChat.py
ZarothYe: I trust you, but I need to read up on what this is doing for my own edification. I don't fully understand it.

ZarothYe: Suggestion has been implemented, and I added a line that will show a KI Error message to the user if StopIteration is hit.

Deledrius (approved):

Hoikas:
This looks good in testing. Question, though, when you sent buddy chat via `/buddies<enter>` then message, the resulting error is different from if you send the message via `/buddies hello!<enter>`. Is that related to this PR or just some old issue?

image

ZarothYe:
> This looks good in testing. Question, though, when you sent buddy chat via `/buddies<enter>` then message, the resulting error is different from if you send the message via `/buddies hello!<enter>`. Is that related to this PR or just some old issue?
> 
> image

This appears to be just because it's going through different handling in IAddChatLine. See xKIChat line #333 vs line #384. I think it would end up doing the "Everyone in list" message if you clicked the KI "BUDDIES" folder and sent a message rather than using /buddies \<message\> as well.

Hoikas (approved):

PR #1045, opened 2021-12-25
Add a Brewfile for macOS Homebrew

dpogue:
This allows [Homebrew](https://brew.sh) users to run `brew bundle` to automatically install all the needed dependencies to build Plasma on macOS. The PhysX and string-theory dependencies are not in the core Homebrew repos, and come from https://github.com/dpogue/homebrew-plasma-deps instead.

To test this, I've also updated the GitHub Action CI workflow to use brew for installing dependencies rather than vcpkg. This seems to result in faster builds (although that's probably because the NuGet package caching still fails to upload sometimes, resulting in always recompiling some libs).

In theory, Homebrew can also be used on Linux, but I've never tried that...


CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 if(POLICY CMP0127)
     cmake_policy(SET CMP0127 OLD)
 endif()
+cmake_policy(SET CMP0074 NEW)
```
Hoikas, line 16: Since we have explicitly set `cmake_minimum_required(VERSION 3.14)`, then CMP0074 (introduced in CMake 3.12) should already be set to `NEW`. Was there some error occurring without this explicit policy set?

Brewfile
```diff
@@ -0,0 +1,25 @@
...
+brew "libpng"
+brew "libtool"
+brew "libvorbis"
+brew "libvpx"
+brew "libwebm"
+brew "openal-soft"
+brew "openssl"
+brew "opus"
+brew "physx"
+brew "python@3.9"
```
Hoikas, line 20: Do we have to hardcode the minor version? The Windows build is now on 3.10...

CMakeLists.txt
dpogue: According to https://cmake.org/cmake/help/latest/policy/CMP0074.html:
> This policy was introduced in CMake version 3.12. CMake version 3.22.1 warns when the policy is not set and uses `OLD` behavior. Use the `cmake_policy()` command to set it to `OLD` or `NEW` explicitly.

Brewfile
dpogue: It looks like `python@3.x` might work, I'll give it a shot

CMakeLists.txt
Hoikas: Yes, that is the default behavior. However, [`cmake_minimum_required()`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings) calls `cmake_policy()` with the version specified in the `cmake_minimum_required()` call, which implicitly does `cmake_policy(SET CMP0074 NEW)`.

Brewfile
dpogue: `python@3.x` does not work, but I'm updating it to `python@3.10` now 😞

dpogue:
@Hoikas Rebased and updated to python 3.10

Hoikas (approved):
Much better speed on macOS CI :+1:

PR #1046, opened 2021-12-25
Fix lookups for newer libwebm versions

dpogue:
This should work with both the existing 1.0.0.27 version as well as the newer 1.0.0.28 version, whether the headers are installed under `libwebm/` or in the global include directory.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
@@ -43,8 +43,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plMoviePlayer.h"
 
 #ifdef USE_WEBM
-#   include <libwebm/mkvreader.hpp>
-#   include <libwebm/mkvparser.hpp>
+#   include <mkvreader.hpp>
+#   include <mkvparser.hpp>
 
 #   define VPX_CODEC_DISABLE_COMPAT 1
 #   include <vpx/vpx_decoder.h>
--- a/cmake/Findlibwebm.cmake
+++ b/cmake/Findlibwebm.cmake
@@ -1,7 +1,10 @@
 include(FindPackageHandleStandardArgs)
 include(SelectLibraryConfigurations)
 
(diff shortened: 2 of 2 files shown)


Hoikas:
Where is this broken?

dpogue:
This is currently broken with the libwebm 1.0.0.28 version installed on macOS through Homebrew, but it also breaks when I install libwebm 1.0.0.28 on ArchLinux.

The issue is that in the newer libwebm version, the mkvreader.hpp file is just a compatibility wrapper that does `#include "mkvparser/mkvreader.h"`. If all the headers are installed into `/usr/include/libwebm`, that `mkvparser/mkvreader.h` include will fail because we've set the include path to `/usr/include` rather than `/usr/include/libwebm`.

Hoikas:
Makes sense. If you could fix the indentation in the cmake, I will be +1 on this.

dpogue:
> If you could fix the indentation in the cmake, I will be +1 on this.

Oops, it's been fixed

Hoikas (approved):

zrax:
I realize this was already merged, but the issue is actually slightly deeper than what @dpogue fixed...  Specifically, `mkvparser.hpp` (for example) actually includes `mkvparser/mkvparser.h`, not `mkvparser/mkvparser.hpp`, which means that just adjusting the directory name is not sufficient.

Also of note:  this is broken by the update to 1.0.0.28 in vcpkg as well, since the old patch was replaced by a version that installs the .h files (not .hpp files) directly into the libwebm directory, and does not preserve the additional subdirs: https://github.com/microsoft/vcpkg/blob/3b45774/ports/libwebm/Fix-cmake.patch

dpogue:
This should work as long as all the headers (with the correct directory structure preserved) are installed in either `/usr/include` or `/usr/include/libwebm`.

But keeping the directory structure intact is the key (which vcpkg is totally failing to do, both in that merged PR and in their [attempted fix](https://github.com/microsoft/vcpkg/pull/22189/files)).

zrax:
Ah, ok, so it seems like a problem specific to the vcpkg port...  I've updated the AUR package to keep the libwebm directory structure and also install the new files, based on your findings.

PR #1050, opened 2022-01-11
Cleft Start option added

Hazado:
Asset files - https://github.com/H-uru/moul-assets/pull/99

This modifies python files to allow the Cleft as an alternate start option at character creation.
It uses a chronicle entry "StartPathChosen" for checks as needed. This Chronicle entry is set at character creation or first login (Default is relto).

Also added a check in the cleft imager script to prevent entering the TPOTS code before you have solved the cleft.
This prevents the player from accidentally being sent to K'Veer if randomly entering codes.

PR #1057, opened 2022-01-22
Allow disabling panic linking.

### LINKED ISSUE #1056: Feature Request: Ability to turn on DontPanic flag with python
Currently, if you don't want an avatar to warp back to relto, they would need to type Avatar.Spawn.DontPanic in the console.
There is no way to do this using python.

Hoikas:
Exposes `Avatar.Spawn.DontPanic` to Python. If enabled, instead of using the personal book, the player simply spawns back at the start of the Age. This also fixes some (but not all) longstanding problems related to spawning where the camera would zoom from across the Age when spawning.

Fixes #1056

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
@@ -826,10 +826,10 @@ void plArmatureMod::SpawnAt(int spawnNum, double time)
     w2l.RemoveScale();
     ci->SetTransform(l2w, w2l);
     ci->FlushTransform();
-    
-    if (plVirtualCam1::Instance())
-        plVirtualCam1::Instance()->SetCutNextTrans();
-    
+
+    if (IsLocalAvatar() && plVirtualCam1::Instance())
+        plVirtualCam1::Instance()->SetCutNext();
+
     if (GetFollowerParticleSystemSO())
     {
         // Since particles are in world space, if we've got some surrounding us, we've got to translate them to compensate for our warp.
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.h
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.h
@@ -233,7 +233,9 @@ public:
     virtual void PanicLink(bool playLinkOutAnim = true);
     virtual void PersonalLink();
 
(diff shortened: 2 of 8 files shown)


Hazado (approved):

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
```diff
@@ -826,10 +826,10 @@ void plArmatureMod::SpawnAt(int spawnNum, double time)
     w2l.RemoveScale();
     ci->SetTransform(l2w, w2l);
     ci->FlushTransform();
-    
-    if (plVirtualCam1::Instance())
-        plVirtualCam1::Instance()->SetCutNextTrans();
-    
+
+    if (IsLocalAvatar() && plVirtualCam1::Instance())
```
ZarothYe, line 830: Would you explain for the n00bs why the `IsLocalAvatar()` check was added? And now that the non-transitional cut is being used here, do other places still use the transitional camera cut, or is `SetCutNextTrans` obsolete now?

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.h
```diff
@@ -233,7 +233,9 @@ class plArmatureMod : public plArmatureModBase
     virtual void PanicLink(bool playLinkOutAnim = true);
     virtual void PersonalLink();
 
-    virtual bool ToggleDontPanicLinkFlag() { fDontPanicLink = fDontPanicLink ? false : true; return fDontPanicLink; }
+    bool ToggleDontPanicLinkFlag() { fDontPanicLink = fDontPanicLink ? false : true; return fDontPanicLink; }
```
ZarothYe, line 236: The `virtual` part of the definition was removed, is that intentional? If so, why?

Hoikas: No one is overriding this function in a subclass, so it doesn't need to be `virtual` (improves performance).

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
Hoikas: This ensures that, if a remote player is panic respawned, his/her respawning will not taint your client's camera state. `SetCutNextTrans()` is still used by `plWarpMsg` and the first to third person transition from what I can tell.

Hoikas:
Merging in preparation for #1050.

PR #1058, opened 2022-01-22
Small improvements for /look command and xKIConstants formatting updates

ZarothYe:
- Improve formatting and readability for dictionaries of constants in xKIConstants.py
- Add documentation comments for kEasterEggs dictionary in xKIConstants.py so Age creators can more easily know how to add new values without trial and error or trying to understand the python that uses it in xKIChat.py.
- Improve behavior of /look command so it does not try to get nearby player list when kEasterEggs already defined a value to display for "people".
- Improve formatting of nearby player names by conditionally using "is" or "are" depending on how many players are nearby and ending the concatenated list with ", and {lastPlayerName}" instead of ", {lastPlayerName}"

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -1245,33 +1245,44 @@ class CommandsProcessor:
     ## Look around for exits and informational text.
     def LookAround(self, params):
 
-        # Find the nearby people.
-        playerList = self.chatMgr.GetPlayersInChatDistance(minPlayers=-1)
-        people = "nobody in particular"
-        if len(playerList) > 0:
-            people = ""
-            for player in playerList:
-                people += player.getPlayerName() + ", "
-            people = people[:-2]
-
         # Load the Age-specific text.
         ageInfo = PtGetAgeInfo()
         if ageInfo is None:
             return
+
         currentAge = ageInfo.getAgeFilename()
         see = ""
         exits = " North and West."
+        people = ""
+        peopleVerb = "is"
         if currentAge in kEasterEggs:
             see = kEasterEggs[currentAge]["see"]
+
             if not kEasterEggs[currentAge]["exits"]:
                 exits = "... well, there are no exits."
             else:
-                exits = " " + kEasterEggs[currentAge]["exits"]
+                exits = kEasterEggs[currentAge]["exits"]
+
             if "people" in kEasterEggs[currentAge]:
                 people = kEasterEggs[currentAge]["people"]
 
+        # Find the nearby people if kEasterEggs didn't define people text override for the Age.
+        if not people:
+            playerList = self.chatMgr.GetPlayersInChatDistance(minPlayers=-1)
+            playerListLen = len(playerList)
+            people = " nobody in particular."
+            peopleVerb = "are" if playerListLen > 1 else "is"
+
+            if playerListLen > 0:
+                people = " "
+                for idx, player in enumerate(playerList):
+                    # concatenate player names together with commas (using "and" before the last name)
+                    people += player.getPlayerName() + (", " if idx != playerListLen - 2 else ", and ")
+                people = people[:-2]
+                people += "."
+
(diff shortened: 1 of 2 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -1245,33 +1245,44 @@ def RestartGame(self, params):
...
+                people = " "
+                for idx, player in enumerate(playerList):
+                    # concatenate player names together with commas (using "and" before the last name)
+                    people += player.getPlayerName() + (", " if idx != playerListLen - 2 else ", and ")
+                people = people[:-2]
+                people += "."
+
         ## Display the info.
-        self.chatMgr.AddChatLine(None, "{}: {} Standing near you is {}. There are exits to the{}".format(GetAgeName(), see, people, exits), 0)
+        self.chatMgr.AddChatLine(None, "{}: {} Standing near you {}{} There are exits to the{}".format(GetAgeName(), see, peopleVerb, people, exits), 0)
```
Hoikas, line 1285: All the string concatting and manual formating is performance tanking and tedious. Let's use the magic of Python 3 to make this better. Note that I didn't test this suggestion, so hopefully I didn't bungle the formatting.

```suggestion
        # Find the nearby people if kEasterEggs didn't define people text override for the Age.
        if not people:
            playerList = self.chatMgr.GetPlayersInChatDistance(minPlayers=-1)
            playerListLen = len(playerList)
            peopleVerb = "are" if playerListLen > 1 else "is"

            if playerListLen == 0:
                people = "nobody in particular."
            else:
                people = ", ".join((f"{'and ' if idx + 1 == playerListLen else ''}{player.getPlayerName()}" for idx, player in enumerate(playerList)))

        ## Display the info.
        self.chatMgr.AddChatLine(None, f"{GetAgeName()}: {see} Standing near you {peopleVerb} {people}. There are exits to the{exits}", 0)
```

ZarothYe: I'll try this out and tweak anything that needs to be changed. Not certain the best way to see if this performs better than before, or shall we just assume .join and f-strings will almost certainly be faster?

Hoikas:
The Python documentation explicitly calls out join as being faster than
concatenations. Also, f-strings are noted in the change logs to be faster
than manual formats or and style formats.

Scripts/Python/ki/xKIChat.py
ZarothYe: I think my latest commit properly integrates the use of join and f-strings while keeping the addition of leading space and ending period when creating the nearby player string. Leading and trailing punctuation/spacing are included in the "people" string to ensure Age creators can more fully customize their own "people" value (e.g., using "..." to start or "?" to end).

Scripts/Python/ki/xKIChat.py
```diff
@@ -1245,33 +1245,45 @@ def RestartGame(self, params):
...
+            playerList = self.chatMgr.GetPlayersInChatDistance(minPlayers=-1)
+            playerListLen = len(playerList)
+            peopleVerb = "are" if playerListLen > 1 else "is"
+
+            if playerListLen == 0:
+                people = " nobody in particular."
+            else:
+                # concatenate player names together with commas (using "and" before the last name)
+                people = ", ".join((
+                    f"{' ' if idx == 0 else ''}{'and ' if playerListLen > 1 and idx + 1 == playerListLen else ''}"
```
Hoikas, line 1280:
```suggestion
                    f"{' ' if idx == 0 else ''}{'and ' if playerListLen > 1 and playerListLen == idx + 1 else ''}"
```

Makes the two legs of the if consistent.

ZarothYe: Fair. I'll swap the next line around too, to match?

Hoikas: Good catch 👍

Hoikas (approved):

PR #586, opened 2019-07-02
Warning cleanup

zrax:
Clean up a variety of GCC warnings throughout the codebase.

PR #1059, opened 2022-01-23
Fix #680

### LINKED ISSUE #680: ErrorAssert inappropriately marked [[noreturn]]
This breaks the ability to use the ignore function of the MSVC assertion dialog and enforces crashing on all assertions.

Hoikas:
Imagine you are working on a Python change, such as #1058. You make a minor syntax error and the traceback is longer than 1024. With the current code, `std::abort()` is unconditionally called in debug mode, forcing everything to be murdered immediately. Potentially before you can even get the traceback logged to disk. This fixes things so you can press the "IGNORE" button on the assertion GUI and keep going.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.cpp
+++ b/Sources/Plasma/CoreLib/HeadSpin.cpp
@@ -118,7 +118,10 @@ void ErrorEnableGui(bool enabled)
     s_GuiAsserts = enabled;
 }
 
-NORETURN void ErrorAssert(int line, const char* file, const char* fmt, ...)
+#if !defined(HS_DEBUGGING)
+[[noreturn]]
+#endif // defined(HS_DEBUGGING)
+void ErrorAssert(int line, const char* file, const char* fmt, ...)
 {
 #if defined(HS_DEBUGGING) || !defined(PLASMA_EXTERNAL_RELEASE)
     char msg[1024];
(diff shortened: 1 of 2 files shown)


dpogue:
Is there a reason we want to keep `[[noreturn]]` for non-Windows platforms? I feel like this has been a source of issues for the GL & Metal ports while stubbing stuff out

Hoikas:
I dunno, it was added in #586 for non-Windows in particular, I think.

zrax:
No, `[[noreturn]]` was added for all platforms intentionally (maybe for unsued code checks?  Or maybe for coverity?  I don't remember now).  It is intended to signal that code should not continue executing after calling `ErrorAssert`.  If you are changing that behavior (as you've done for MSVC builds), then `[[noreturn]]` should go away to match, and all callers of `ErrorAssert` need to be prepared to continue executing even after a call to it.  Unfortunately, I don't know of a way to tell C++ `[[noreturn(only if debugger not attached)]]` ;).  Perhaps instead, it could be `[[noreturn]]` only for Release builds?

dpogue:
> Perhaps instead, it could be `[[noreturn]]` only for Release builds?

This seems like a good option for now (whether keyed on `!PLASMA_EXTERNAL_RELEASE` or `HS_DEBUGGING`)

Hoikas:
I've changed this to condition on `!defined(HS_DEBUGGING)` instead of `!defined(_MSC_VER)`.

zrax (approved):
Thanks, I think this makes more sense.

PR #1060, opened 2022-01-23
Prevent empty command '/' or '/  ' from creating a blank message line

ZarothYe:
Calum reported on the OpenUru discord that just sending a chat line with "/" or "/   " would create a prefix-less and message-less line in the KI chat. This PR prevents this bizarre chat line from being added and short-circuits trying to find any emote or other command matches because having an empty command seems clearly invalid and should never match anything.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -893,8 +893,8 @@ class CommandsProcessor:
         if message.startswith("/"):
             words = message.split()
             # get the command word, trimming the / off the front
-            commandWord = str(words[0][1:].casefold())
-            if commandWord == "" or commandWord.isspace():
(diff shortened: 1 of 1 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -892,8 +892,14 @@ def __call__(self, message):
         # Is it an emote, a "/me" or invalid command?
         if message.startswith("/"):
             words = message.split()
+            # get the command word, trimming the / off the front
+            commandWord = str(words[0][1:].casefold())
+            if commandWord == "" or commandWord.isspace():
```
Hoikas, line 897:
```suggestion
            commandWord = words[0][1:].casefold()
            if not commandWord or commandWord.isspace():
```

Hoikas (approved):

PR #1061, opened 2022-01-27
Allow getting PID from name for online buddies or neighbors

ZarothYe:
Previously the name-to-ID conversion would only work for Age Players due to player nodes in Buddies and Neighbors folders not being of type ptPlayer. Primarily for easier /addbuddy usage with neighbor names. Requested by Hazado on OpenUru discord.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -1016,10 +1016,18 @@ class CommandsProcessor:
             return int(params)
         except ValueError:
             for player in self.chatMgr.BKPlayerList:
+                # Age Players
                 if isinstance(player, ptPlayer):
                     plyrName = player.getPlayerName().casefold()
                     if params.casefold() == plyrName:
                         return player.getPlayerID()
+                # Buddies or Neighbors
+                elif isinstance(player, ptVaultNodeRef):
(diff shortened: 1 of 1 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -1016,10 +1016,18 @@ def GetPID(self, params):
...
             for player in self.chatMgr.BKPlayerList:
+                # Age Players
                 if isinstance(player, ptPlayer):
                     plyrName = player.getPlayerName().casefold()
                     if params.casefold() == plyrName:
                         return player.getPlayerID()
+                # Buddies or Neighbors
+                elif isinstance(player, ptVaultNodeRef):
+                    plyrInfoNode = player.getChild().upcastToPlayerInfoNode()
+                    if type(plyrInfoNode) != type(None) and plyrInfoNode.getType() == PtVaultNodeTypes.kPlayerInfoNode:
```
Hoikas, line 1027:
```suggestion
                    if plryInfoNode is not None:
```

`upcastToPlayerInfoNode()` will only ever return `None` objects or `ptVaultPlayerInfoNode` objects whose node type is `kPlayerInfoNode` by definition.

Hoikas (approved):

PR #1062, opened 2022-01-31
Minor Homebrew CI fixes

dpogue:
* OpenSSL 3 is released, and is now the default if you just do `brew install openssl`. It isn't compatible with Plasma, so we need to explicitly install `openssl@1.1` and use that for our builds.  
  (This probably worked on CI because they don't pull homebrew Formula updates on every build, but it will break at some point if we don't do this)

* We were installing `python@3.10` but CMake was finding the system Python 3.9 installation instead. Explicitly tell CMake to use the brew Python.

* Minor tweak to the `qt@5` stuff to make it consistent


zrax (approved):

Hoikas:
Tangentially related, but what are the problems with OpenSSL 3? (See: microsoft/vcpkg#20031)

dpogue:
> Tangentially related, but what are the problems with OpenSSL 3?

That's good question... it's entirely possible I was having issues because I had OpenSSL 1.1 installed and Homebrew was trying to provide that paths to a (non-existent) OpenSSL 3 installation, which is why it couldn't find the libraries. I'll try with OpenSSL 3 tonight and see what happens.

PR #1063, opened 2022-02-05
Add audio.ini EnableSubtitles option and support to display timed subtitles in the KI

ZarothYe:
Tied to [this moul-assets PR](https://github.com/H-uru/moul-assets/pull/101) for supporting SRT files (and upcoming work with  @Hazado to add checkbox to audio option menu GUI).

There is a bug currently in `plWin32StreamingSound::GetActualTimeSec()` that causes this function not to return expected time values as the audio plays, which makes the subtitle timings not work as intended. I aim to fix this in an upcoming PR after doing more testing.


Sources/Plasma/PubUtilLib/plAudio/plWin32StaticSound.cpp
```diff
@@ -171,6 +182,19 @@ bool plWin32StaticSound::LoadSound( bool is3D )
             return false;
         }
 
+        // check if subtitles are enabled and if srcFilename is a localized audio file (e.g., ending in _eng, _fre, etc.)
+        // TODO: surely there is already a function somewhere to do this localization filename check?
+        if (plgAudioSys::IsEnabledSubtitles() && std::regex_match(srcFilename.StripFileExt().AsString().c_str(), std::regex("^.*_(eng|fre|ger|spa|ita|jpn)$", std::regex_constants::icase))) {
```
Hoikas, line 187: Couple of things:
- If I'm reading this correctly, the first leg of this if statement will prevent changing the subtitle on/off state from working until relinking. IMO, the subtitles should always be loaded and we should just check the on/off state before dispatching the subtitle message.
- As your comment notes, there is a better way to do the second leg. Just check to see if the sound buffer filename is localized using `plLocalization::GetLocalized()`.

Hoikas (commented):
It looks like this duplicates some code in both `plWin32StreamingSound` and `plWin32StaticSound`. I think the logic should move down a level to `plWin32Sound` to prevent the duplication, if possible. Further, the subtitle data should probably be read in and held by `plSoundBuffer`, which is responsible for holding data related to sounds.

Sources/Plasma/PubUtilLib/plMessage/plSubtitleMsg.h
```diff
@@ -0,0 +1,76 @@
...
+protected:
+    ST::string fText;
+
+public:
+    plSubtitleMsg()
+    {
+        SetBCastFlag(plMessage::kBCastByExactType);
+    }
+
+    plSubtitleMsg(ST::string msg) : plSubtitleMsg() { fText = std::move(msg); }
```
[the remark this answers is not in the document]
ZarothYe: Can you tell me why suggestion is preferred? I used the delegating constructor (and thus couldn't do the member initialization list shorthand) so I didn't have to have both constructors doing the same SetBCastFlag call duplicated unnecessarily. I also don't understand why you added a call to the empty parent constructor for plMessage().

Scripts/Python/ki/xKIChat.py
```diff
@@ -439,9 +440,20 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
...
+            # Is it subtitles for current audio?
+            if cFlags.subtitle:
+                headerColor = kColors.AudioSubtitleHeader
+                player = None
+
+                PtDebugPrint("xKIChat.AddChatLine(): Adding subtitle {}.".format(message))
+                messageMatches = SUBTITLE_REGEX.match(message)
+                if messageMatches is not None:
+                    pretext = messageMatches.group("speaker")
+                    message = messageMatches.group("subtitle")
```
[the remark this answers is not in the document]
ZarothYe: That seems reasonable, but I'll need to think a bit on what the right way to do that is. AddChatLine takes 3 arguments: player, message, and chatFlags. Maybe I can hack in the ability to use "player" for the speaker name string (if provided) and just know to treat it as a string rather than a ptPlayer when the chatFlags indicate the message is a subtitle? It's a bit non-obvious, I think. If you want, I can try it and push the regex matching up to the plSrtFileReader-level to split out the subtitle lines into speaker and text.

Sources/Plasma/PubUtilLib/plAudio/plSrtFileReader.cpp
```diff
@@ -0,0 +1,166 @@
...
+                if (subtitleNumber == 0) {
+                    subtitleNumber = std::stoi(line);
+                    continue;
+                } else if (subtitleTimings.compare("") == 0) {
+                    subtitleTimings = line;
+                    std::smatch matches;
+
+                    if (std::regex_match(subtitleTimings, matches, std::regex("^(\\d{2}):(\\d{2}):(\\d{2}),(\\d{3}) --> (\\d{2}):(\\d{2}):(\\d{2}),(\\d{3})$"))) {
+                        if (matches.size() < 9) {
+                            // TODO: I dunno, something wasn't formatted right? What should we do?
```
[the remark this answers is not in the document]
ZarothYe: I can add a message to audio.log, but I don't know how to make it red. Is that what you mean by "red log message"?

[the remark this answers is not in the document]
Deledrius: plStatusLog has color options for the output:

https://github.com/H-uru/Plasma/blob/master/Sources/Plasma/PubUtilLib/plStatusLog/plStatusLog.h#L147-L154

Sources/Plasma/PubUtilLib/plAudio/plWin32StaticSound.cpp
ZarothYe: It would no work to turn on/off subtitles in the middle of an already playing audio, but I'm not sure why you're saying it would prevent changing the enabled state via console or options menu until the player links to a different Age? I might be misunderstanding. I guess it would be nicer for it to check at dispatch-time. I was just trying to prevent incurring extra overhead (albeit relatively small overhead, since the SRT files are fairly small) for players who play with the "default" no-subtitle setting.

Sources/Plasma/PubUtilLib/plMessage/plSubtitleMsg.h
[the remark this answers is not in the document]
Hoikas: Static analysis (which we do use!) may complain about `fText` not being initialized. Also, this prevents `fText` from being set twice. Further, my suggestion is how it works everywhere else in the code :wink:

Sources/Plasma/PubUtilLib/plAudio/plWin32StaticSound.cpp
Hoikas: `LoadSound()` bails if the sound is already loaded, so you would need to link out and back in to force `LoadSound()` to proceed. I would favor better UX over optimizing for the handful of kilobytes memory saved by not parsing the subtitle file.

Hoikas (commented):
Just a few minor thoughts for now because I know you're still working on some of the untangling.

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
```diff
@@ -99,20 +100,37 @@ void plSoundPreloader::Run()
...
+                            // same file we were playing before, so start the SRT feed over instead of deleting and reloading
+                            srtReader->StartOver();
+                        } else {
+                            auto newSrtFileReader = new plSrtFileReader(srcFilename);
+                            if (newSrtFileReader->ReadFile()) {
+                                buf->SetSrtReader(newSrtFileReader);
+                            } else {
+                                // nothing could be read, this reader is not useful
+                                delete newSrtFileReader;
+                            }
```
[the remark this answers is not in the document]
ZarothYe: I could do that, but for my own edification, is there anything wrong (other than being less compact/efficient) with handling it the way I'm doing it currently? I will want to port this over to OU and `std::make_unique<T>` is a C++14 construct...

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
```diff
@@ -181,6 +199,9 @@ plSoundBuffer::~plSoundBuffer()
     }
 
     UnLoad();
+
+    delete fSrtReader;
+    fSrtReader = nullptr;
```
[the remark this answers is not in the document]
ZarothYe: Is this suggestion related or independent of the change to use `std::make_unique<T>`?

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -410,6 +410,52 @@ bool hsStream::ReadLn(char *s, uint32_t maxLen, const char beginComment, const c
...
+        ST::string_stream ss;
+        char c;
+        char endCom = endComment;
+
+        while (true) {
+            while (!AtEnd() && strchr("\r\n", c = ReadByte()))
+                /* empty */;
+
+            if (AtEnd())
+                return result;
```
zrax, line 427: This flavor of `ReadLn` has no way to distinguish between this EOF case and an empty line...  It would be better to pass the `ST::string` result by reference and still return a bool to distinguish between the two.

Sources/Plasma/PubUtilLib/plAudio/plWin32StreamingSound.cpp
```diff
@@ -56,6 +56,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pnMessage/plSoundMsg.h"
 #include "pnMessage/plEventCallbackMsg.h"
 #include "plStatusLog/plStatusLog.h"
+#include "plMessage/plSubtitleMsg.h"
+
+#include <regex>
```
zrax, line 61: The other changes in this file are whitespace only...  Are these extra includes really needed?

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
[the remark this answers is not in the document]
zrax: If you want to remain compatible with C++11, you can construct the `unique_ptr` directly (it's just not as pretty):
```c++
std::unique_ptr<plSrtFileReader> newSrtFileReader(new plSrtFileReader(srcFilename));
```

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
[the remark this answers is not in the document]
zrax: This is in the destructor, so we can leave `fSrtReader` in a dirty state (i.e., it's independent of `std::make_unique`)

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,185 @@
...
+                            speakerName = matches[2];
+
+                            // matches[5] is the fourth group (the subtitle text)
+                            subtitleText = matches[5];
+                        }
+                    }
+                } else if (lnCounter % 4 == 3 && subtitleNumber > 0 && subtitleStartTimeMs >= 0 && subtitleEndTimeMs >= 0 && !subtitleText.empty()) {
+                    // entry is complete, add to the queue and reset our temp variables
+                    if (!speakerName.empty())
+                        fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, subtitleText, speakerName);
```
Hoikas, line 122:
```suggestion
                        fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, std::move(subtitleText), std::move(speakerName));
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    else
+                        fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, subtitleText);
```
Hoikas, line 124:
```suggestion
                        fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, std::move(subtitleText));
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    subtitleEndTimeMs = 0;
+                    subtitleText.clear();
+                    speakerName.clear();
+                }
+            }
+
+            if (subtitleNumber > 0 && subtitleStartTimeMs >= 0 && subtitleEndTimeMs >= 0 && !subtitleText.empty()) {
+                // enqueue the last subtitle from the file if we didn't have an extra blank line at the end
+                if (!speakerName.empty())
+                    fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, subtitleText, speakerName);
```
Hoikas, line 137:
```suggestion
                    fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, std::move(subtitleText), std::move(speakerName));
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,185 @@
...
+                else
+                    fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, subtitleText);
```
Hoikas, line 139:
```suggestion
                    fEntries.emplace_back(subtitleNumber, subtitleStartTimeMs, subtitleEndTimeMs, std::move(subtitleText));
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -439,9 +439,17 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
 
         # Is it an object to represent the flags?
         if isinstance(cFlags, ChatFlags):
+            PtDebugPrint("Chat Flags are {}".format(cFlags))
```
Hoikas, line 442:
```suggestion
            PtDebugPrint(f"Chat Flags are {cFlags}", level=kDebugDumpLevel)
```

Scripts/Python/plasma/Plasma.py
```diff
@@ -1367,6 +1375,10 @@ def isEnabled(self):
         """Is the audio enabled? Returns 1 if true otherwise returns 0."""
         pass
 
+    def areSubtitlesEnabled(self):
+        """Are subtitles for the audio enabled? Returns 1 if true otherwise returns 0."""
+        pass
+
```
Hoikas, line 1381: This needs to be alphabetized like the rest of the stubs.

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -99,6 +100,18 @@ void plWin32Sound::IFreeBuffers()
 
 void plWin32Sound::Update()
 {
+    auto buf = this->GetDataBuffer();
+    if (plgAudioSys::AreSubtitlesEnabled() && buf != nullptr) {
+        auto srtReader = buf->GetSrtReader();
```
Hoikas, line 105: Let's use the actual type instead of `auto` here.

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -111,6 +124,17 @@ void plWin32Sound::IActuallyPlay()
     {
         if (fDSoundBuffer && plgAudioSys::Active() )
         {
+            if (!fReallyPlaying && fSynchedStartTimeSec > 0) {
+                // advance past any subtitles that would end before the synched start time
+                // not sure when this actually happens...
+                auto buf = this->GetDataBuffer();
```
Hoikas, line 130: Let's use the actual type instead of `auto` here.

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -111,6 +124,17 @@ void plWin32Sound::IActuallyPlay()
...
+                if (buf != nullptr) {
+                    auto srtReader = buf->GetSrtReader();
```
Hoikas, line 132: Let's use the actual type instead of `auto` here.

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
```diff
@@ -99,20 +100,33 @@ void plSoundPreloader::Run()
...
+
             while (!templist.empty())
             {
                 plSoundBuffer* buf = templist.back();
                 templist.pop_back();
 
                 if (buf->GetData())
                 {
-                    reader = CreateReader(true, buf->GetFileName(), buf->GetAudioReaderType(), buf->GetReaderSelect());
+                    auto srcFilename = buf->GetFileName();
```
Hoikas, line 112: Let's use the actual type instead of `auto` here.

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
```diff
@@ -99,20 +100,33 @@ void plSoundPreloader::Run()
...
+                    reader = CreateReader(true, srcFilename, buf->GetAudioReaderType(), buf->GetReaderSelect());
                     
                     if( reader )
                     {
                         unsigned readLen = buf->GetAsyncLoadLength() ? buf->GetAsyncLoadLength() : buf->GetDataLength();
                         reader->Read( readLen, buf->GetData() );
                         buf->SetAudioReader(reader);     // give sound buffer reader, since we may need it later
+
+                        auto srtReader = buf->GetSrtReader();
```
Hoikas, line 121: Let's use the actual type instead of `auto` here.

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.h
```diff
@@ -120,6 +121,12 @@ class plSoundBuffer : public hsKeyedObject
     plAudioFileReader * GetAudioReader();   // transfers ownership to caller
     void                SetAudioReader(plAudioFileReader *reader);
     void                SetLoaded(bool loaded);
+    plSrtFileReader*    GetSrtReader() const { return fSrtReader; }  // does not transfer ownership
+    void                SetSrtReader(plSrtFileReader* reader)
+    {
+        delete fSrtReader;
+        fSrtReader = reader;
+    }
```
Hoikas, line 129: To improve compile times, let's move this into the .cpp file, remove the `#include "plAudioCore/plSrtFileReader.h"`, and add a forward declaration `class plSrtFileReader;`.

Sources/Plasma/PubUtilLib/plAudioCore/plSoundBuffer.cpp
```diff
@@ -45,6 +45,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsStream.h"
 
 #include "plSoundBuffer.h"
+#include "plAudioCore/plSrtFileReader.h"
```
Hoikas, line 48:
```suggestion
#include "plSrtFileReader.h"
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,184 @@
...
+                            subtitleStartTimeMs += (atoi(matches[4].str().c_str()));
+
+                            // matches[5] is the fifth group    -- the end hour number
+                            subtitleEndTimeMs += (atoi(matches[5].str().c_str()) * 3600000);
+                            // matches[6] is the sixth group    -- the end minute number
+                            subtitleEndTimeMs += (atoi(matches[6].str().c_str()) * 60000);
+                            // matches[7] is the seventh group  -- the end seconds number
+                            subtitleEndTimeMs += (atoi(matches[7].str().c_str()) * 1000);
+                            // matches[8] is the eighth group   -- the end milliseconds number
+                            subtitleEndTimeMs += (atoi(matches[8].str().c_str()));
```
Hoikas, line 100:
```suggestion
                            // matches[0] is the entire match, we don't do anything with it
                            // matches[1] is the first group    -- the start hour number
                            subtitleStartTimeMs += std::stoul(matches[1].str()) * 3600000;
                            // matches[2] is the second group   -- the start minute number
                            subtitleStartTimeMs += std::stoul(matches[2].str()) * 60000;
                            // matches[3] is the third group    -- the start seconds number
                            subtitleStartTimeMs += std::stoul(matches[3].str()) * 1000;
                            // matches[4] is the fourth group   -- the start milliseconds number
                            subtitleStartTimeMs += std::stoul(matches[4].str());

                            // matches[5] is the fifth group    -- the end hour number
                            subtitleEndTimeMs += std::stoul(matches[5].str()) * 3600000;
                            // matches[6] is the sixth group    -- the end minute number
                            subtitleEndTimeMs += std::stoul(matches[6].str()) * 60000);
                            // matches[7] is the seventh group  -- the end seconds number
                            subtitleEndTimeMs += std::stoul(matches[7].str()) * 1000;
                            // matches[8] is the eighth group   -- the end milliseconds number
                            subtitleEndTimeMs += std::stoul(matches[8].str());
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,184 @@
...
+    return false;
+}
+
+void plSrtFileReader::AdvanceToTime(uint32_t timeMs)
+{
+    if (fCurrentEntryIndex >= 0) {
+        while (fCurrentEntryIndex < fEntries.size() && fEntries.at(fCurrentEntryIndex).GetEndTimeMs() <= timeMs) {
+            fCurrentEntryIndex++;
+        }
+    }
```
Hoikas, line 154:
```suggestion
    while (fCurrentEntryIndex < fEntries.size() && fEntries.at(fCurrentEntryIndex).GetEndTimeMs() <= timeMs) {
        fCurrentEntryIndex++;
    }
```

`if (fCurrentEntryIndex >= 0)` will always be true because `fCurrentEntryIdx` is unsigned.

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+
+plSrtEntry* plSrtFileReader::GetNextEntryStartingBeforeTime(uint32_t timeMs)
+{
+    if (fCurrentEntryIndex >= 0 && fCurrentEntryIndex < fEntries.size()) {
```
Hoikas, line 160:
```suggestion
    if (fCurrentEntryIndex < fEntries.size()) {
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,184 @@
...
+            return &nextEntry;
+        }
+    }
+
+    return nullptr;
+}
+
+plSrtEntry* plSrtFileReader::GetNextEntryEndingBeforeTime(uint32_t timeMs)
+{
+    if (fCurrentEntryIndex >= 0 && fCurrentEntryIndex < fEntries.size()) {
```
Hoikas, line 174:
```suggestion
    if (fCurrentEntryIndex < fEntries.size()) {
```

Sources/Plasma/FeatureLib/pfConsole/pfAudioConsoleCommands.cpp
```diff
@@ -58,6 +58,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pnKeyedObject/plFixedKey.h"
 #include "pnMessage/plAudioSysMsg.h"
 #include "pnMessage/plSoundMsg.h"
+#include "plMessage/plSubtitleMsg.h"
 #include "pnNetCommon/plNetApp.h"
 #include "pnSceneObject/plAudioInterface.h"
```
Hoikas, line 63:
```suggestion
#include "pnNetCommon/plNetApp.h"
#include "pnSceneObject/plAudioInterface.h"

#include "plMessage/plSubtitleMsg.h"
```

See #788

Scripts/Python/plasma/Plasma.py
ZarothYe: You're gonna have to explain a bit more for me to understand what you mean. The rest of the stubs in this file are camelCase just like this one, no?

dpogue: The method starts with `a` so it should be near the start of the method list in this class

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -99,6 +102,18 @@ void plWin32Sound::IFreeBuffers()
 
 void plWin32Sound::Update()
 {
+    plSoundBuffer* buf = this->GetDataBuffer();
```
Hoikas, line 105:
```suggestion
    plSoundBuffer* buf = GetDataBuffer();
```

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -99,6 +102,18 @@ void plWin32Sound::IFreeBuffers()
...
+    if (plgAudioSys::AreSubtitlesEnabled() && buf != nullptr) {
+        plSrtFileReader* srtReader = buf->GetSrtReader();
+        if (srtReader != nullptr) {
+            while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(this->GetActualTimeSec() * 1000.0f))) {
```
Hoikas, line 109:
```suggestion
            while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(GetActualTimeSec() * 1000.0f))) {
```

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -111,6 +126,17 @@ void plWin32Sound::IActuallyPlay()
     {
         if (fDSoundBuffer && plgAudioSys::Active() )
         {
+            if (!fReallyPlaying && fSynchedStartTimeSec > 0) {
+                // advance past any subtitles that would end before the synched start time
+                // not sure when this actually happens...
+                plSoundBuffer* buf = this->GetDataBuffer();
```
Hoikas, line 132:
```suggestion
                plSoundBuffer* buf = GetDataBuffer();
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,182 @@
...
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plSrtFileReader.h"
+
+#include "hsStream.h"
+#include "plStatusLog/plStatusLog.h"
```
Hoikas, line 46:
```suggestion

#include "plStatusLog/plStatusLog.h"
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+#include <regex>
+
+static const std::regex speakerTextRegex("^((([\\w.]+(\\s\\w+)?):) )?(.*)");
+static const std::regex timingsRegex("^(\\d{2}):(\\d{2}):(\\d{2}),(\\d{3}) --> (\\d{2}):(\\d{2}):(\\d{2}),(\\d{3})$");
+
+bool plSrtFileReader::ReadFile()
+{
+    plFileName audioSrtPath = plFileName::Join(plFileSystem::GetCWD(), "dat", fAudioFileName.StripFileExt() + ".srt");
```
Hoikas, line 55: Have you given any thought to the suggestion I relayed from @Deledrius about using a format agnostic extension? Also, I think that, for consistency with Myst V, the subtitles should go in the sfx directory.

Hoikas (commented):
I think I'm very close to happy with this. I do wish the format was a bit cleaner to parse, but I suppose that, on the plus side, we'll have support from a common set of subtitling tools.

dpogue:
Thoughts/Opinions on whether the subtitle files should be xTEA/WDYS encrypted like the rest of the data files? On one hand, the audio itself is playable and the encryption offers no security benefits, but it does make it so that "spoiling" speeches requires a bit of effort.

Hoikas:
That's theoretically a good idea, but the loc files are decrypted 😞 - I don't really have an opinion either way.

ZarothYe:
> but it does make it so that "spoiling" speeches requires a bit of effort.

If someone wants to dig through game files to read subtitles of a 15+ year old game... I don't think they should be mad if things are spoiled for them, IMO. :P

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
dpogue: Myst V's subtitles are actually not in the sfx folder. The `.sub` files are in dat, there are `.sgt` files in sfx that are something else.

dpogue:
> If someone wants to dig through game files to read subtitles of a 15+ year old game... I don't think they should be mad if things are spoiled for them, IMO. :P

I was thinking moreso for the case of future fan content, but I don't have strong feelings either way

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
Hoikas: Whoops. Regardless, I think that if we're matching the audio files 1:1, these need to go in the sfx folder.

ZarothYe: I've thought about it, but does anyone have an agnostic extension they wish to put forward as an option? Otherwise I'll lean toward something silly, self-aggrandizing and seemingly-unused by other applications like `.yez` ;) As for the folder, I can go ahead and move them if we have agreement?

dpogue: Myst V uses `.sub` which seems suitably generic, but I don't see a problem with `.srt` if that's the standard extension that's well-supported by existing tools.

Hoikas: Since I didn't specify, @Deledrius's suggestion was to use an agnostic extension so the format could be changed later (without leaving stale files in the client folder) if the srt format is determined to be insufficient in the future.

ZarothYe: Ah, I thought `.sub` was for a specific format of subtitles like `.srt` was, but it looks from The Googles that it can refer to subtitles in many different unspecified formats, so I'm good with that. I think Deledrius suggested an agnostic extension so they could be changed to a different format internally more easily without needing re...manifesting, I guess?

ZarothYe:
> > If someone wants to dig through game files to read subtitles of a 15+ year old game... I don't think they should be mad if things are spoiled for them, IMO. :P
> 
> I was thinking moreso for the case of future fan content, but I don't have strong feelings either way

Perhaps... but is there a reason why we couldn't implement that if/when that situation for sensitive/spoilery fan content audio could come up? I'm not strictly opposed to it, but I don't really see a need right now. Seems more worthwhile to handle it later if it is needed than to try to code for a situation that may never happen, unless there's some reason why changing it later is difficult or impossible. What do you think?

Hoikas (approved):

Hoikas (approved):

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
ZarothYe: I honestly used Notepad++ and a regular audio player app because the format is dirt simple to use, but I tried https://www.happyscribe.com/subtitle-tools/online-subtitle-editor/free out a little bit after I finished the majority of the SRTs and it seems OK if you prefer a more structured way to edit the SRT file. It reads and writes .srt files just fine, and did let me upload the `.sub` file without issue (after selecting "All Files" in the file browser). That tool, and the majority of free online tooling, assumes you're subtitling video, though, so you'd still have to use a separate audio player.

Hoikas:
Blocking on the GUI portion of H-uru/moul-assets#101.

Hoikas:
To keep the ball rolling, I think we need to separate this feature out into its components so it can be merged more atomically (read: easily). Namely, by splitting the core subtitle functionality away from the "user ability to control subtitles via the options menu" feature. In other words, I think these things should go into another PR:
- https://github.com/H-uru/Plasma/pull/1063/commits/48587e452f73cbd587bf8121af65ad2bcaf7f26d and the xOptionsMenu.py part of this commit https://github.com/H-uru/Plasma/pull/1063/commits/e83056423e38c247c854773eb529f16cc8604d90.
-  https://github.com/H-uru/moul-assets/pull/101/commits/2a5e02a4875257df8dba8c452e81368ee737e22f

IMO, until the options menu part is ready to go, we should simply default subtitles to ON. Also, I think there would be bonus points for breaking up the .sub PR into a single "demonstration" subtitle PR to go along with the core functionality and a larger one with the bulk load of the .subs. That way we can merge the functionality itself rather quickly and take time glancing over the subtitles themselves.

ZarothYe:
> To keep the ball rolling, I think we need to separate this feature out into its components so it can be merged more atomically (read: easily). Namely, by splitting the core subtitle functionality away from the "user ability to control subtitles via the options menu" feature. In other words, I think these things should go into another PR:
> 
>     * [48587e4](https://github.com/H-uru/Plasma/commit/48587e452f73cbd587bf8121af65ad2bcaf7f26d) and the xOptionsMenu.py part of this commit [e830564](https://github.com/H-uru/Plasma/commit/e83056423e38c247c854773eb529f16cc8604d90).
> 
>     * [H-uru/moul-assets@2a5e02a](https://github.com/H-uru/moul-assets/commit/2a5e02a4875257df8dba8c452e81368ee737e22f)
> 
> 
> IMO, until the options menu part is ready to go, we should simply default subtitles to ON. Also, I think there would be bonus points for breaking up the .sub PR into a single "demonstration" subtitle PR to go along with the core functionality and a larger one with the bulk load of the .subs. That way we can merge the functionality itself rather quickly and take time glancing over the subtitles themselves.

I'm working on trying to rebase and revert those parts, but it's a little hard for me still. :( Trying to piece out which commits had parts of the options menu and how to revert them; I think there were a couple different commits since I remember I had to fix something about it at a later point than I initially added the menu part. I'll also split out Zandi's intro speech into a single moul-assets PR since that's probably the easiest to test by just creating a fresh toon. Sound ok?

Hoikas:
That sounds good to me :+1:

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.h
```diff
@@ -0,0 +1,104 @@
...
+    ST::string    fSpeakerName;
+
+};
+
+class plSrtFileReader
+{
+public:
+
+    plSrtFileReader(plFileName audioFileName)
+        : fAudioFileName(audioFileName), fEntries(), fCurrentEntryIndex() { }
```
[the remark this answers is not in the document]
ZarothYe: You wanna also remove the `fEntries()` initializer while you're at it?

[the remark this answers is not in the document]
Hoikas: Yes, let's do that :+1:

Hoikas (approved):

Hoikas:
:shipit:

PR #1066, opened 2022-02-09
[WIP - Incomplete] Port Async sockets to ASIO

zrax:
There's still some missing pieces and TODOs here, but it's enough to link in...  At least until you need to queue some data for sending later or shutdown the client cleanly ;)


Hoikas:
Closed in favor of #1205.

PR #1067, opened 2022-02-09
Local Input Config for Keyboard

Hazado:
Allows the keymap to become global and persist through character creations.
If no keymap is found it uses the vault keymap until you go into the keymap settings and change a single keybind.

### DIFF
--- a/Scripts/Python/xIniInput.py
+++ b/Scripts/Python/xIniInput.py
@@ -45,10 +45,12 @@ This module contains the routines to read and write the input.ini file
 """
 
 import xIniHelper
+from PlasmaConstants import *
+from Plasma import *
 
-
-gFilename = "./init/input.fni"
 gIniFile = None
+gFilename = "input.ini"
+gFilenameAndPath = ""
 
 # the different volume commands
 kBindCmd = "Keyboard.BindAction"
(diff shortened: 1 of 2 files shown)


Scripts/Python/xIniInput.py
```diff
@@ -60,19 +62,48 @@
...
+    if gFilenameAndPath == "":
+        if PtIsInternalRelease():
+            # check for local file
+            localNameAndPath = "init/" + gFilename
+            if PtFileExists(localNameAndPath):
+                gFilenameAndPath = localNameAndPath
+                PtDebugPrint("xIniInput::ConstructFilenameAndPath(): Using internal \"" + gFilenameAndPath + "\" file")
+                return
+        # otherwise, use the standard init path
+        gFilenameAndPath = PtGetInitPath() + "/" + gFilename
```
Hoikas, line 94: This should probably return the path instead of stashing it as a global variable. That way, you can get rid of the `gFilenameAndPath` thing and just do `gIniFile.writeFile(ConstructFilenameAndPath())`.

Scripts/Python/xOptionsMenu.py
```diff
@@ -1940,6 +1952,18 @@ def IShowMappedKeys(self,dlg,mapRow1,mapRow2):
...
+        km = ptKeyMap()
+        for control_code in defaultControlCodeBinds.keys():
+            if type(control_code) == type(''):
+                key1 = km.convertVKeyToChar(km.getBindingKeyConsole(control_code), km.getBindingFlagsConsole(control_code))
+                xIniInput.SetConsoleKey('"' + control_code + '"', key1 + ',')
+            else:
+                controlStr = km.convertControlCodeToString(control_code)
+                key1 = km.convertVKeyToChar(km.getBindingKey1(control_code), km.getBindingFlags1(control_code))
+                key2 = km.convertVKeyToChar(km.getBindingKey2(control_code), km.getBindingFlags2(control_code))
+                xIniInput.SetControlKey('"' + controlStr + '"', key1 + ',', key2 + ',')
```
Hoikas, line 1965: These string concats are very difficult for me to read because it uses two different kinds of quotation marks. I'd prefer to see these become f-strings.

PR #1068, opened 2022-02-11
Fix 3dsMax python parameters.

Hoikas:
Only the first parameter was visible in Max 2022 because `MoveWindow()` was no longer resizing the rollup for some reason. Copying and pasting the method from `WM_SIZE_PANEL` fixed it, so here you go.

Before:
image

After:
image


dpogue (approved):

Deledrius:
I'm curious if setting `bRepaint` to TRUE in the call to MoveWindow would fix this for Max 2022.  That appears to be what this code is doing manually.  It would also prevent this manual invalidation from breaking in earlier versions of the plugin (which is currently the case).  Alas, I cannot test the change on Max 2022 myself.

PR #1069, opened 2022-02-11
Fix max crash on exit.

Hoikas:
Fixes crashes observed when closing 3ds Max 2022 when Python files have been loaded.


dpogue (approved):
oooh, I was hitting these on other Max versions too. Nice to see them fixed!

PR #1070, opened 2022-02-11
Squelch script ID warnings in 3ds Max.

Hoikas:
Debounces:
image

### DIFF
--- a/Scripts/Python/Tiam.py
+++ b/Scripts/Python/Tiam.py
@@ -39,11 +39,12 @@
 #       Mead, WA   99021
 #
 # *==LICENSE==*/
+
 from Plasma import *
 from PlasmaTypes import *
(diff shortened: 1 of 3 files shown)


dpogue (approved):
Is there a logic to the ID numbers picked, or just something that doesn't overlap any existing files?

Maybe we want to set a standard involving the sequence prefix?

Hoikas:
These were just random numbers. It would be nice to have some kind of standard, but if we change any of the files that are used in Max files to match that standard, then we may break those Max files 😢. I'm not really sure what we can do?

dpogue:
Yeah, I don't think we can retroactively apply a standard to the existing files.

We should look further into that idea of a PR step that confirms there are no duplicate script IDs...

PR #1072, opened 2022-02-12
Update GitHub CI to VS 2022

dpogue:
GitHub Actions is moving `windows-latest` to point to the `windows-2022` runner, which includes VS 2022 rather than VS 2019.

Alternatively, we could change our CI workflow to point specifically to the `windows-2019` runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,8 +8,8 @@ jobs:
     strategy:
       matrix:
         platform:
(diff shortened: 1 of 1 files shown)


dpogue:
oh, ugh, it's interpreting `windows-latest` differently depending on the fork 🤦🏼‍♂️ 

On my fork, `windows-latest` = `windows-2022` and I have to use VS2022
On H'uru, `window-latest` = `windows-2019` and VS2022 isn't available

Hoikas:
I think we should hardwire windows-2019 for now considering the readme and all other instructions recommend VS 2019.

PR #1075, opened 2022-02-13
Add python/system/lib-dynload to module paths

dpogue:
On non-Windows, some modules with C implementations are compiled as shared libraries under a lib-dynload path. These cannot be found by default, resulting in errors when scripts do things like `import math`.


Hoikas (approved):
We'll need a different solution for external builds, but this looks sufficient to get us going with internal ones.

PR #1076, opened 2022-02-13
Better user data paths on non-Windows

dpogue:
On macOS, this will use a more standard `~/Library/Application Support/Uru Live` path.
On Linux/Solaris/etc., this will be `~/.config/Uru Live` which matches the XDG spec.

~Uses the POSIX standard [`wordexp`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wordexp.html) function to expand the user home path, rather than relying on the `$HOME` environment variable.~

/fyi @colincornaby

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -128,6 +128,9 @@ typedef int32_t   hsError;
 #define _hsMacroJoin_(lhs, rhs) lhs ## rhs
 #define hsMacroJoin(lhs, rhs)   _hsMacroJoin_(lhs, rhs)
 
+// Macro to string-ize its arguments
+#define hsStringize(exp) #exp
+
 // Declare a file-unique identifier without caring what its full name is
 #define hsUniqueIdentifier(prefix) hsMacroJoin(prefix, __LINE__)
 
@@ -434,6 +437,39 @@ void DebugMsg(const char* fmt, ...);
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); __assume(0); break;
 #else
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); break;
+#endif
+
+#if defined(__clang__) || defined(__GNUC__)
+#   define _COMPILER_WARNING_NAME(warning) "-W" warning
+
+    /* Condition is either 1 (true) or 0 (false, do nothing). */
+#   define IGNORE_WARNINGS_BEGIN_IMPL_1(warning) \
+        _Pragma(hsStringize(GCC diagnostic ignored warning))
+#   define IGNORE_WARNINGS_BEGIN_IMPL_0(warning)
+
+#   define IGNORE_WARNINGS_BEGIN_COND(cond, warning) \
+        _Pragma("GCC diagnostic push") \
+        hsMacroJoin(IGNORE_WARNINGS_BEGIN_IMPL_, cond)(warning)
+
+#   if defined(__has_warning)
+#       define IGNORE_WARNINGS_BEGIN_IMPL(warning) \
+            IGNORE_WARNINGS_BEGIN_COND(__has_warning(warning), warning)
+#       define IGNORE_WARNINGS_END \
+            _Pragma("GCC diagnostic pop")
+#   else
+        /* Suppress -Wpragmas to dodge warnings about attempts to suppress unrecognized warnings. */
+#       define IGNORE_WARNINGS_BEGIN_IMPL(warning) \
+            IGNORE_WARNINGS_BEGIN_COND(1, "-Wpragmas") \
+            IGNORE_WARNINGS_BEGIN_COND(1, warning)
+#       define IGNORE_WARNINGS_END \
+            _Pragma("GCC diagnostic pop") \
+            _Pragma("GCC diagnostic pop")
+#   endif
+
+#   define IGNORE_WARNINGS_BEGIN(warning) IGNORE_WARNINGS_BEGIN_IMPL(_COMPILER_WARNING_NAME(warning))
+#else
+#   define IGNORE_WARNINGS_BEGIN(warning)
+#   define IGNORE_WARNINGS_END
 #endif
 
 #endif
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -59,6 +59,17 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   include <sys/types.h>
 #   include <unistd.h>
 #endif
+
+#if HS_BUILD_FOR_APPLE
+#   if defined(__has_include)
+#     if __has_include(<sysdir.h>)
+#       include <sysdir.h>
+#     endif
+#   endif
+#
+#   include <NSSystemDirectories.h>
+#endif
+
 #include <sys/stat.h>
 
 
@@ -459,8 +470,42 @@ plFileName plFileSystem::GetUserDataPath()
             return "";
 
         _userData = plFileName::Join(ST::string::from_wchar(path), plProduct::LongName());
+#elif HS_BUILD_FOR_APPLE
+        char path[PATH_MAX] {};
+        if (__builtin_available(macOS 10.12, *)) {
+            sysdir_search_path_enumeration_state state;
+            state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_APPLICATION_SUPPORT, SYSDIR_DOMAIN_MASK_USER);
+            state = sysdir_get_next_search_path_enumeration(state, path);
+        } else {
+            IGNORE_WARNINGS_BEGIN("deprecated-declarations")
+
+            NSSearchPathEnumerationState state;
+            state = NSStartSearchPathEnumeration(NSApplicationSupportDirectory, NSUserDomainMask);
+            state = NSGetNextSearchPathEnumeration(state, path);
+
+            IGNORE_WARNINGS_END
+        }
+
+        if (path[0] == '~') {
+            char home[PATH_MAX] {};
+            strlcat(home, getenv("HOME"), sizeof(home));
+            strlcat(home, &path[1], sizeof(home));
+
+            _userData = plFileName::Join(home, plProduct::LongName());
+        } else {
+            _userData = plFileName::Join(path, plProduct::LongName());
+        }
 #else
-        _userData = plFileName::Join(getenv("HOME"), "." + plProduct::LongName());
+        const char* homedir = getenv("XDG_CONFIG_HOME");
+        if (homedir) {
+            _userData = plFileName::Join(homedir, plProduct::LongName());
+        } else {
+            homedir = getenv("HOME");
+            if (!homedir)
+                return "";
+
+            _userData = plFileName::Join(homedir, ".config", plProduct::LongName());
+        }
 #endif
         plFileSystem::CreateDir(_userData);
     }
--- a/Sources/Tools/plSystemInfo/main.cpp
+++ b/Sources/Tools/plSystemInfo/main.cpp
@@ -40,6 +40,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
+#include "plFileSystem.h"
 #include "hsSystemInfo.h"
 
 #include <iostream>
@@ -47,6 +48,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 int main()
 {
-    std::cout << hsSystemInfo::AsString() << std::endl;
+    std::cout << hsSystemInfo::AsString() << std::endl << std::endl;
+
+    std::cout << "User Data Path: " << plFileSystem::GetUserDataPath().AsString() << std::endl;
     return 0;
 }
--- a/Sources/Plasma/CoreLib/hsThread_Unix.cpp
+++ b/Sources/Plasma/CoreLib/hsThread_Unix.cpp
@@ -51,7 +51,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #if NO_POSIX_CLOCK
 #include <sys/time.h>
 #include <unistd.h>
-#define CLOCK_REALTIME 0
+
+#ifndef CLOCK_REALTIME
+#   define CLOCK_REALTIME 0
+#endif
 
 //
 // A linux hack b/c we're not quite POSIX
@@ -86,11 +89,15 @@ hsGlobalSemaphore::hsGlobalSemaphore(int initialValue, const ST::string& name)
             throw hsOSException(errno);
         }
     } else {
+        IGNORE_WARNINGS_BEGIN("deprecated-declarations")
+
         /* Anonymous semaphore shared between threads */
         int shared = 0; // 1 if sharing between processes
         fPSema = new sem_t;
         int status = sem_init(fPSema, shared, initialValue);
         hsThrowIfOSErr(status);
+
+        IGNORE_WARNINGS_END
     }
 #else
     int status = ::pthread_mutex_init(&fPMutex, nullptr);
@@ -110,8 +117,12 @@ hsGlobalSemaphore::~hsGlobalSemaphore()
     if (fNamed) {
         status = sem_close(fPSema);
     } else {
+        IGNORE_WARNINGS_BEGIN("deprecated-declarations")
+
         status = sem_destroy(fPSema);
         delete fPSema;
+
+        IGNORE_WARNINGS_END
     }
     hsThrowIfOSErr(status);
 #else


Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
 
         _userData = plFileName::Join(ST::string::from_wchar(path), plProduct::LongName());
 #else
-        _userData = plFileName::Join(getenv("HOME"), "." + plProduct::LongName());
+        ST::string dir;
+#  if HS_BUILD_FOR_APPLE
+        char home[256] = "";
```
dgelessus, line 474: According to the man page:

> The path buffer that is passed to **sysdir_get_next_search_path_enumeration**() will be filled with a null-terminated string, possibly containing as many as PATH_MAX-1 characters.

```suggestion
        char home[PATH_MAX] = "";
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
...
+            #pragma clang diagnostic push
+            #pragma clang diagnostic ignored "-Wdeprecated-declarations"
+            NSSearchPathEnumerationState state;
+            state = NSStartSearchPathEnumeration(NSApplicationSupportDirectory, NSUserDomainMask);
+            state = NSGetNextSearchPathEnumeration(state, home);
+            #pragma clang diagnostic pop
+        }
+        dir = home;
+#  else
+        dir = "~" PATH_SEPARATOR_STR ".config";
```
dgelessus, line 489: To fully follow the XDG spec, this should use the value of the `XDG_CONFIG_HOME` variable, and only if that variable isn't defined fall back to the default `~/.config`.

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
...
+#  endif
+
+        wordexp_t p;
+        wordexp(dir.c_str(), &p, 0);
+        dir = p.we_wordv[0];
+        for (int i = 1; i < p.we_wordc; i++) {
+            dir += " ";
+            dir += p.we_wordv[i];
+        }
+        wordfree(&p);
```
dgelessus, line 499: Looking at the man page, `wordexp` does a lot more than just expanding tildes to the home path - it also expands variables and globs, does shell-style word splitting, and might even try to interpret some other shell syntax. While realistically nobody will have shell metacharacters or weird whitespace in their home path, do we really need to use a complex function like this here?

I don't think we need to worry about the case of `HOME` being unset. The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables) implicitly assumes that `HOME` is always set:

> If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.

And [according to POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_01), expanding a plain tilde when `HOME` is unset has unspecified results:

> If the login name is null (that is, the tilde-prefix contains only the tilde), the tilde-prefix is replaced by the value of the variable *HOME*. If *HOME* is unset, the results are unspecified.

dgelessus: For anyone else reading this and wondering why `wordexp` is also being called on the paths returned by the macOS-specific API - according to the man page:

> Directory paths returned in the user domain will contain "~" to refer to the user’s directory.

So if we don't use `wordexp`, we'd have to handle this case manually - but that should be relatively simple, because the tilde that needs to be expanded should only appear as the first character of the path.

dgelessus (changes requested):

Sources/Plasma/CoreLib/plFileSystem.cpp
Hoikas: Building on the suggestion by @dgelessus 

```suggestion
        char home[PATH_MAX]{};
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -58,7 +58,16 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   include <limits.h>
 #   include <sys/types.h>
 #   include <unistd.h>
+#   include <wordexp.h>
 #endif
+
+#if HS_BUILD_FOR_APPLE
+#   if __has_include(<sysdir.h>)
```
Hoikas, line 65: Is there any case where the `sysdir.h` header does not exist? If so, can we depend on clang supporting `__has_include()` in all clang variants that advertise C++17 support? I'm not super familiar with clang's conformance.

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
...
+        ST::string dir;
+#  if HS_BUILD_FOR_APPLE
+        char home[256] = "";
+        if (__builtin_available(macOS 10.12, *)) {
+            sysdir_search_path_enumeration_state state;
+            state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_APPLICATION_SUPPORT, SYSDIR_DOMAIN_MASK_USER);
+            state = sysdir_get_next_search_path_enumeration(state, home);
+        } else {
+            #pragma clang diagnostic push
+            #pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
Hoikas, line 481:
```suggestion
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
...
+            NSSearchPathEnumerationState state;
+            state = NSStartSearchPathEnumeration(NSApplicationSupportDirectory, NSUserDomainMask);
+            state = NSGetNextSearchPathEnumeration(state, home);
+            #pragma clang diagnostic pop
```
Hoikas, line 485:
```suggestion
#pragma clang diagnostic pop
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -460,7 +469,36 @@ plFileName plFileSystem::GetUserDataPath()
...
+#  else
+        dir = "~" PATH_SEPARATOR_STR ".config";
+#  endif
+
+        wordexp_t p;
+        wordexp(dir.c_str(), &p, 0);
+        dir = p.we_wordv[0];
+        for (int i = 1; i < p.we_wordc; i++) {
+            dir += " ";
+            dir += p.we_wordv[i];
```
Hoikas, line 497: This is non-optimal in terms of performance, FWIW.

Sources/Tools/plSystemInfo/main.cpp
```diff
@@ -41,12 +41,15 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *==LICENSE==*/
 
 #include "hsSystemInfo.h"
+#include "plFileSystem.h"
```
Hoikas, line 44:
```suggestion
#include "plFileSystem.h"
#include "hsSystemInfo.h"
```

Alphabetize core includes

Sources/Plasma/CoreLib/plFileSystem.cpp
dpogue: `sysdir.h` was introduced in macOS 10.12, we need to use `NSSystemDirectories.h` prior to that (which is deprecated, but still exists in current macOS, so we _could_ rely on that entirely and see if Apple ever actually removes it)

Sources/Plasma/CoreLib/plFileSystem.cpp
dpogue: This case only happens when there are spaces in the path. Which is probably rare on Linux, but macOS uses a folder called "Application Support" 🙄

Sources/Plasma/CoreLib/plFileSystem.cpp
colincornaby: Probably a larger conversation - I'm not sure what macOS ranges we're interested in supporting. Metal doesn't start until 10.12, so it would be the OpenGL renderer before that. The gaming input APIs don't exist until 11.0. I'd figured we might have to at runtime figure out if we can use those APIs or fallback to NSEvent. macOS 10.12 seems like a pretty reasonable minimum to me. But I'm not sure what the project goals are.

If Plasma wants to support older OSs, I am guessing NSSystemDirectories probably isn't going anywhere for a long while. It's pretty deeply used in a lot of apps.

dpogue: macOS 11 was released in late 2020, so I'm hesitant to set that as our minimum. I'm fine with using NSSystemDirectories.h if we're reasonably confident it won't disappear.

Testing the existence of sysdir.h with CMake during configuring is probably another option if we really care to do that.

colincornaby: Only thing I'm checking is the status of that API on Apple Silicon. Apple sometimes uses arch transitions to cull deprecated API, but they took a pretty gentle approach on Apple Silicon so I'm betting it's still there.

Hoikas: > Testing the existence of sysdir.h with CMake during configuring is probably another option if we really care to do that.

That still won't help you support < 10.12 - better use only NSSystemDirectories if you want to support those versions of macOS.

colincornaby: NSSearchPathForDirectoriesInDomains is also the other common API used, and it looks like it's not deprecated. But it looks like it requires Obj-C even though it's a C function. The return types are NSString. :(
https://developer.apple.com/documentation/foundation/1414224-nssearchpathfordirectoriesindoma?language=objc

colincornaby: I tried building for an earlier version of macOS, and found there might be issues there anyway. When I try to build back to 10.11 I get a bunch of errors that a lot of the modern C++ features didn't appear until macOS 10.15.

`/Users/colincornaby/Desktop/PlasmaLibs/include/string_theory/st_string.h:1193:26: 'path' is unavailable: introduced in macOS 10.15`

`/Users/colincornaby/Desktop/PlasmaLibs/include/string_theory/st_string.h:1198:37: 'u8path<const char *>' is unavailable: introduced in macOS 10.15`

And more than those.

So far all the errors look like they are in String Theory. I'm not sure how many of these functions we're actually using, but as far as I can tell the client in it's current state won't build further back than 10.15 anyway.

dgelessus: > If so, can we depend on clang supporting `__has_include()` in all clang variants that advertise C++17 support? I'm not super familiar with clang's conformance.

Clang has supported `__has_include` as an extension since ±forever ([at least 3.3](https://releases.llvm.org/3.3/tools/clang/docs/LanguageExtensions.html#include-file-checking-macros), which is from 2013).

> FWIW, this `__has_include()` check is a build-time thing, so it doesn't really help you _run_ on an old macOS, I would assume, unless you build on that old macOS.

It is a build-time check, but the code below uses a [`__builtin_available`](https://clang.llvm.org/docs/LanguageExtensions.html#objective-c-available) check to decide whether to call the new or old API, depending on the runtime system version.

Regarding what should be the minimum macOS target, I would argue that at least macOS 10.14 should be supported - it is the last version to support 32-bit apps, so people are more likely to intentionally stay on that version. But I'm aware that supporting older systems isn't always easy, so if there's a good reason to require a newer version, I won't complain much (or if I do, I'll try to bring patches).

dpogue: So my current thinking is that we should set our minimum macOS compiler version to 10.14 (Xcode 9, last to support targeting 32-bit). That would, in theory, allow us to remove the `__has_include`.

I'd still like to try to use `__builtin_available` where it makes sense to see if we can support **running** on older versions.

Unfortunately, it's impossible to get CI builders that still support Xcode 9, so we aren't going to have a way to watch for breaking changes 😞

colincornaby: We can set the deployment target on current Xcode to 10.14, and it should give most the necessary feedback. Two concerns I'd have:
- The string theory requirement that currently seems to point to 10.15 and higher being a requirement.
- Wider range of supported hardware. I've been trying to ignore Nvidia cards for testing. Metal makes me feel a bit better on older hardware in since OpenGL has usually a wider amount of driver variance. But even in Metal I have seen some vendor only shader issues.

zrax: string_theory definitely supports older versions of macOS (at least as far back as 10.9)...  The trouble here is that using Xcode on newer versions of macOS to target older versions of macOS causes the ST feature detection to enable features it shouldn't (that is, compiling on macOS 10.12 will work perfectly fine for targeting macOS 10.12).  I'm not sure what the proper way to fix this is, since I'm not an experienced macOS developer...  However, you can certainly produce a working build by turning off the `std::filesystem` support via the `-DST_ENABLE_STL_FILESYSTEM=OFF` cmake flag.

dpogue: It might be possible to add an availability attribute to those methods, but I haven't tried it and it's going to be ugly code that hardcodes specific macOS versions 😞 https://clang.llvm.org/docs/AttributeReference.html#availability

We should probably take further discussion of this particular point over to an issue on string_theory.

dgelessus: > It would probably be best to support only the latest Xcode toolchains, and use something like `-DST_ENABLE_STL_FILESYSTEM=OFF`. Supporting older toolchains can be iffy, and as the code gets into Apple Silicon support it may have to find symbols that aren't in older toolchains, or use compiler features that weren't in older toolchains.

I don't quite follow... isn't that usually handled with a configure-time feature check? That is, if the toolchain doesn't support the feature in question, you don't compile the new code path at all. Only if the toolchain is new enough you compile both the old and new code with a runtime check. Official builds would be done with a modern toolchain and so would conditionally support everything, but old toolchains could still compile the code, just without support for features newer than the toolchain.

Also, "the toolchain" doesn't have to be a big monolithic thing. Although Apple ships their SDKs along with specific Clang versions, it's not difficult to use older SDKs with newer Clang versions, which allows you to use modern language features even on older macOSes. Actual OS APIs like the controller input thing are the bigger problem, because you can't easily update/"retrofit" those like the compiler and libc++.

colincornaby: Sorry I've been late to get back to this. It's probably not important to the pull request anyway, but I'll try and be quick. Typically on macOS you dynamically check for features where you can at runtime. So you'd always build with the latest toolchain, and then at runtime check for the presence of the symbol you rely on. It's how we'd handle multiple input methods if we wanted to support NSEvent and the new game controller framework at the same time. There wouldn't be two different builds, just a runtime check. Even though the game controller framework is only present on new OS versions, the linker is smart enough to not link it on older OS versions. For this reason, building against older SDKs is uncommon on macOS. If someone wanted to do a PPC or Intel 32 bit version, thats where older SDKs might appear.

Apple describes this process as weak linking here:
https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html

This is a little different in since it's implicitly compile time, which happens. Unless there was a strong reason we'd need STL filesystem on later versions of macOS, my personal take is I'm fine with disabling the feature across the board. It simplifies things, and if the official build is going to support pre-10.15, it'll always have to be off in the official build anyway.

colincornaby: Also sorry if any of the above is something you already know, and I just misinterpreted what you said. There's both compile time and run time constraints going on here, which makes talking about this in a pull request confused, at least for me.

dgelessus: I think we're mostly talking about the same thing 🙂 Whether you do the runtime check using weak linking or `__builtin_available`/`@available` shouldn't make a big difference - either way you're checking if a certain API is usable at runtime, and both require that the API is supported by the SDK at compile time.

My suggestion was to have a compile-time check *in addition to* the runtime check, so that if the SDK *doesn't* support the new API, only the old code path is compiled and used. This would help when building the source locally on an old macOS - in that case you only have an old SDK, but don't care if your binaries won't use newer APIs (because your macOS doesn't have them anyway).

Getting back to where this discussion started... 🙂 The `__has_include` would only be needed to allow *compiling* on older macOSes, and supporting that would need further changes to the code. As it stands, I think the code further below won't compile if `<sysdir.h>` is not included, so it would be better to make it an unconditional include.

ZarothYe:
Brushing the dust off this one. 😉 It looks really interesting, but some of the discussion is over my head so I'm not sure I follow what the status is. Any thoughts on what is still pending for this to be complete and ready, other than any outstanding code suggestions?

colincornaby:
I'd probably want to manually check this on macOS. But Application Support should be the right place to save this data (and not the invisible home folder being used now.)

colincornaby:
On that note - I manually patched this in to my copy of the Mac client and this change looks good to me. macOS is requiring 10.13 so we'd only follow the 10.12 path, but I don't mind both being there.

dpogue:
> macOS is requiring 10.13 so we'd only follow the 10.12 path, but I don't mind both being there.

If we use the old NSDirectories stuff, we avoid the need to deal with un-tilde-ifying the path... but it's considered deprecated 😕

colincornaby:
> > macOS is requiring 10.13 so we'd only follow the 10.12 path, but I don't mind both being there.
> 
> If we use the old NSDirectories stuff, we avoid the need to deal with un-tilde-ifying the path... but it's considered deprecated 😕

There is a much better version of this API in Obj-C. Unfortunately the Obj-C is also the problem with using it here. :(

dgelessus:
> There is a much better version of this API in Obj-C. Unfortunately the Obj-C is also the problem with using it here. :(

Reading through the conversation again - you're referring to this, right?

> NSSearchPathForDirectoriesInDomains is also the other common API used, and it looks like it's not deprecated. But it looks like it requires Obj-C even though it's a C function. The return types are NSString. :(
https://developer.apple.com/documentation/foundation/1414224-nssearchpathfordirectoriesindoma?language=objc

If the `NSArray`/`NSString` return types are the only issue, you can work around that by using Core Foundation to manipulate the return value (`NSArray *` is compatible with `CFArrayRef` and `NSString *` with `CFStringRef`). You'd have to manually declare the function though, because the Foundation headers require Objective-C.

```c
#include <CoreFoundation/CoreFoundation.h> // for CFArrayRef
#include <objc/NSObjCRuntime.h> // for BOOL, NSUInteger

extern CFArrayRef NSSearchPathForDirectoriesInDomains(NSUInteger directory, NSUInteger domainMask, BOOL expandTilde);
```

colincornaby:
> Reading through the conversation again - you're referring to this, right?

I'd forgotten about NSSearchPathForDirectoriesInDomains but it would also work. This was the function I was thinking of:
https://developer.apple.com/documentation/foundation/nsfilemanager/1407693-urlfordirectory

> If the `NSArray`/`NSString` return types are the only issue, you can work around that by using Core Foundation to manipulate the return value (`NSArray *` is compatible with `CFArrayRef` and `NSString *` with `CFStringRef`). You'd have to manually declare the function though, because the Foundation headers require Objective-C.

That could be one way to do it. I'm always a little leery of manually declaring a function.

It might be simpler to just drop a small Obj-C++ file into the project that wraps that function and provides C++ friendly types. Comes with the complexity of adding a Mac specific source file to track, but reduces the risk of declaring a function that shadows Apple's definition. That's an option I'm already looking at for some of the Headspin utilities like dialogs that also need to call Obj-C functions from C++.

colincornaby:
Also another complication is I think we'd technically need to wrap that call with an autorelease pool. It's not a pure CoreFoundation function, even if we bridge to CF types. So it follows Obj-C memory management rules, not just C. And the autorelease pool wrapping would probably require the Obj-C headers.

On Mac we're using ARC, which would mean using (at)autorelasepool.

dpogue:
> Also another complication is I think we'd technically need to wrap that call with an autorelease pool. It's not a pure CoreFoundation function, even if we bridge to CF types. So it follows Obj-C memory management rules, not just C. And the autorelease pool wrapping would probably require the Obj-C headers.

Hypothetically we could use the autorelease pool stuff from metal-cpp, but I don't love spreading a dependency on that through the whole engine.

I'll try to play around with this over the weekend and find something that works without needing to contaminate anything with Objective C.

colincornaby:
I think another option would be to use a compiler flag to force this file to compile with Objective-C++ on the Mac. Then you could embed Obj-C directly in. I'm not a fan of unexpected compile flags being injected per platform by CMake, but that could also work.

Could we just have a Mac specific file that defines the implementation of GetUserDataPath() instead of using the preprocessor?

dpogue:
Okay, bunch of updates here.

1. On Linux, this will try to use `$XDG_CONFIG_DIR` before falling back to `$HOME/.config`
2. On macOS, this will find the tilde-based Application Support folder, and then replace the tilde with `$HOME`
3. Since we're using `$HOME`, we don't need the wordexp stuff anymore.

I also added some helper macros for disabling specific warnings in blocks of code. I'm going to need these for GL Pipeline too (since macOS throws deprecation warnings if you use any GL functions now). The implementation is heavily inspired by how these sorts of warnings are handled by Apple in WebKit.

Hoikas (commented):
This looks like an improvement to me. I'm not quite comfortable with singing off on this due to lack of knowledge, though. I'd appreciate comments from @Deledrius and/or @zrax.

Sources/Tools/plSystemInfo/main.cpp
ZarothYe: This suggestion is still relevant.

Deledrius (approved):
This looks good to me as well.  I'm not familiar with the nuances on those platforms to offer anything additional, but the discussion on it here points to this being the right direction.  Further refinement can always follow if needed.

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -436,4 +436,40 @@ void DebugMsg(const char* fmt, ...);
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); break;
 #endif
 
+#if defined(__clang__) || defined(__GNUC__)
+#   define _COMPILER_WARNING_NAME(warning) "-W" warning
+#   define _COMPILER_STRINGIZE(exp) #exp
+#   define _COMPILER_CONCAT_I(a, b) a ## b
+#   define _COMPILER_CONCAT(a, b) _COMPILER_CONCAT_I(a, b)
```
zrax, line 443: We already have `hsMacroJoin()` above to do the same thing...

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -436,4 +436,40 @@ void DebugMsg(const char* fmt, ...);
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); break;
 #endif
 
+#if defined(__clang__) || defined(__GNUC__)
+#   define _COMPILER_WARNING_NAME(warning) "-W" warning
+#   define _COMPILER_STRINGIZE(exp) #exp
```
zrax, line 441: This should probably be moved outside the `__clang__ || __GNUC__` block, perhaps above with `hsMacroJoin()`?

dpogue:
@zrax Thanks for catching the duplicate macros, I've adjusted it to use hsMacroJoin and hsStringize now

zrax (approved):

PR #1077, opened 2022-02-14
hsG3DDeviceSelector support for OpenGL

dpogue:
This allows us to run `plClient.exe /Renderer=gl` and find a device and immediately explode due to a pipeline consisting of empty functions, but it does properly find an OpenGL driver on Windows, macOS, and Linux.

I'm being very lax about looping through resolutions and buffer sizes and stuff like that because it's not clear to me that those are still as relevant as they were back in the early 2000s, but I will probably want to filter the modes a bit more when we get to actually initializing the pipeline.

The Windows side should in theory also create a second window and use the relevant ARB extensions to request a more specific newer OpenGL context.

Finally, I'd like to move the DirectX stuff to a `plDXEnumerate` function rather than having hsG3DDeviceSelector split across multiple modules, but it's messy and it works as-is for now.


Sources/Plasma/FeatureLib/pfGLPipeline/CMakeLists.txt
```diff
@@ -19,10 +20,17 @@ target_link_libraries(pfGLPipeline
...
         pnFactory
 )
 
+if(APPLE AND CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    # On macOS, we need to link against OpenGL.framework for access to the CGL
+    # functions. On other platforms, the OpenGL library is loaded dynamically
+    # by epoxy.
+    target_link_libraries(pfGLPipeline PRIVATE "-framework OpenGL")
+endif()
+
```
Hoikas, line 34:
```suggestion
# On macOS, we need to link against OpenGL.framework for access to the CGL
# functions. On other platforms, the OpenGL library is loaded dynamically
# by epoxy.
target_link_libraries(pfGLPipeline PRIVATE "$<$<PLATFORM_ID:Darwin>:-framework OpenGL>")
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,282 @@
...
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "HeadSpin.h"
+
+#include <vector>
+#include <epoxy/gl.h>
```
Hoikas, line 46: Alphabetize includes

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,282 @@
...
+}
+#endif // USE_EGL
+
+
+#ifdef HS_BUILD_FOR_WIN32
+#include "hsWindows.h"
+#include <tchar.h>
+#include <epoxy/wgl.h>
+
+void cleanupWGL(LPCTSTR cls, HWND wnd = 0, HDC dc = 0, HGLRC ctx = 0)
```
Hoikas, line 138:
```suggestion
static void cleanupWGL(LPCTSTR cls, HWND wnd = 0, HDC dc = 0, HGLRC ctx = 0)
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,282 @@
...
+        DestroyWindow(wnd);
+    }
+
+    UnregisterClass(cls, GetModuleHandle(nullptr));
+}
+
+void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
+{
+    WNDCLASS tempClass;
+    memset(&tempClass, 0, sizeof(tempClass));
```
Hoikas, line 159:
```suggestion
    WNDCLASS tempClass{};
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    if (!SetPixelFormat(dc, format, &pfd))
+        return cleanupWGL(clsName, wnd, dc);
+
+    HGLRC ctx = wglCreateContext(dc);
+    if (!ctx)
+        return cleanupWGL(clsName, wnd, dc);
+
+    if (!wglMakeCurrent(dc, ctx))
+        return cleanupWGL(clsName, wnd, dc, ctx);
```
Hoikas, line 204: This repeated cleanup logic could probably be cleaned up with smart pointers or do... while(0)

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
```diff
@@ -56,6 +56,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #define hsG3DDeviceSelector_inc
 
 #include <vector>
+#include <list>
```
Hoikas, line 59: Alphabetize includes

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.cpp
```diff
@@ -349,23 +351,12 @@ void hsG3DDeviceSelector::Enumerate(hsWindowHndl winRef)
...
-    uint16_t ret = RegisterClass(&tempClass);
-    hsAssert(ret, "Cannot create temporary window class to test for device modes" );
-
     ITryDirect3DTnL(winRef);
-
-    /// Get rid of the class
-    UnregisterClass(fTempWinClass, GetModuleHandle(nullptr));
 #endif
+
+    for (auto enumerator : sEnumerators) {
```
Hoikas, line 357:
```suggestion
    for (const auto& enumerator : sEnumerators) {
```

Avoid copying a function object

cmake/Findepoxy.cmake
```diff
@@ -0,0 +1,90 @@
...
+set(epoxy_DEFINITIONS ${PKG_epoxy_CFLAGS})
+
+find_path(epoxy_INCLUDE_DIRS
+    NAMES epoxy/gl.h
+    HINTS ${PKG_epoxy_INCLUDEDIR} ${PKG_epoxy_INCLUDE_DIRS}
+)
+find_library(epoxy_LIBRARIES
+    NAMES epoxy
+    HINTS ${PKG_epoxy_LIBDIR} ${PKG_epoxy_LIBRARY_DIRS}
+)
```
Hoikas, line 50: Convention in CMake find modules is that  `_LIBRARIES` is an output variable, not an in/out, FWIW, Ideally this would be finding the `_LIBRARY_RELEASE` and the `_LIBRARY_DEBUG` separately, then using `select_library_configurations()` to swap them as appropriate. But I see that libepoxy is a bad citizen in that is doesn't suffix its debug libraries and uses an inferior build system.

Sources/Plasma/Apps/plClient/main.cpp
```diff
@@ -59,5 +64,9 @@ static plClientLoader gClient;
 // Stub main function so it compiles on non-Windows
 int main(int argc, const char** argv)
 {
+#ifdef PLASMA_PIPELINE_GL
+    hsG3DDeviceSelector::AddDeviceEnumerator(&plGLEnumerate);
```
Hoikas, line 68: I think this would be better to do in some static object defined in `pfGLPipeline`, preventing this bit of boilerplate from needing to be duplicated into all client main implementations.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,306 @@
...
+    devRec.SetCap(hsG3DDeviceSelector::kCapsDoesSmallTextures);
+
+    devRec.SetLayersAtOnce(8);
+
+    // Just make a fake mode so the device selector will let it through
+    hsG3DDeviceMode devMode;
+    devMode.SetWidth(hsG3DDeviceSelector::kDefaultWidth);
+    devMode.SetHeight(hsG3DDeviceSelector::kDefaultHeight);
+    devMode.SetColorDepth(hsG3DDeviceSelector::kDefaultDepth);
+    devRec.GetModes().emplace_back(devMode);
```
Hoikas, line 76:
```suggestion
    devRec.GetModes().push_back(devMode);
```

This more accurately models the behavior you're using. It would be nice if `hsG3DDeviceMode` had a default constructor so you could simply `devRec.GetModes().emplace_back();`.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+        if (!eglMakeCurrent(display, surface, surface, context))
+            break;
+
+        hsG3DDeviceRecord devRec;
+        devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
+        devRec.SetDriverName("EGL");
+
+        if (fillDeviceRecord(devRec))
+            records.emplace_back(devRec);
```
Hoikas, line 132:
```suggestion
            records.push_back(devRec);
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,306 @@
...
+        tempClass.hbrBackground = (HBRUSH)GetStockObject(BLACK_BRUSH);
+        tempClass.lpszClassName = _T("GLTestClass");
+        tempClass.style = CS_OWNDC;
+
+        cls = RegisterClass(&tempClass);
+        if (!cls)
+            break;
+
+        wnd = CreateWindowEx(0, reinterpret_cast<LPCTSTR>(cls),
+                _T("OpenGL Test Window"),
```
Hoikas, line 175:
```suggestion
        wnd = CreateWindowExW(0, reinterpret_cast<LPCWSTR>(cls),
                L"OpenGL Test Window",
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+        if (!wglMakeCurrent(dc, ctx))
+            break;
+
+        hsG3DDeviceRecord devRec;
+        devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
+        devRec.SetDriverName("opengl32.dll");
+
+        if (fillDeviceRecord(devRec))
+            records.emplace_back(devRec);
```
Hoikas, line 214:
```suggestion
            records.push_back(devRec);
```

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -0,0 +1,306 @@
...
+    }
+#endif
+
+#ifdef HS_BUILD_FOR_WIN32
+    plWGLEnumerate(records);
+#endif
+
+#ifdef HS_BUILD_FOR_MACOS
+    plCGLEnumerate(records);
+#endif
```
Hoikas, line 305: Should there be handling for a not supported platform?

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
dpogue: `emplace_back` matches what the DX side is doing: https://github.com/H-uru/Plasma/blob/f7811a2f3728eb9206d194f3615d876a82400448/Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.cpp#L670

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
dpogue: No, in that case there will be no DeviceRecord added to the records vector, and hsG3DDeviceSelector will report that OpenGL is not available (and either fail over to something else, or (more likely, since GL is always going to be the second choice when other options are supported) cause plClient to show an error about no appropriate rendering devices)

Sources/Plasma/Apps/plClient/main.cpp
dpogue: yeah, I definitely don't like the duplication.

I wish C++ would do something like Java's `static { /* static init stuff here */ }` blocks rather than having to fake it with constructors of static objects 😞

colincornaby:
Just peeking into this pull request in since it got posted in IRC.

I'm thinking through Macs with multiple GPUs and eGPUs. Metal also has a concept of a default device that I think should follow eGPU attach, and most multi GPU concepts. Didn't know if it might be possible to pass a display id into the enumeration in the future. For multi display/multi GPU systems the default GPU might not be the right choice.

Display modes will continue to be interesting. On modern macOS there is a scaler, so really you just get to pick any resolution you want independent of what the physical display is capable of. Its more like a modern console.

When this change goes in, I will try to get it into the Mac client. But I'm realizing since I added networking/login support that client now depends on ASIO.

dpogue:
> Display modes will continue to be interesting. On modern macOS there is a scaler, so really you just get to pick any resolution you want independent of what the physical display is capable of. Its more like a modern console.

I'm pretty unclear on if the display mode stuff is actually used by the pipeline (or rather, I'm sure that it's not for GLPipeline). I think it dates back to the days of full-screen DirectX, which we've now replaced with a fake-fullscreen.
It does need to be set for the device selector to consider the device valid, but maybe that's an assumption that should be revisited?

colincornaby:
> > Display modes will continue to be interesting. On modern macOS there is a scaler, so really you just get to pick any resolution you want independent of what the physical display is capable of. Its more like a modern console.
> 
> I'm pretty unclear on if the display mode stuff is actually used by the pipeline (or rather, I'm sure that it's not for GLPipeline). I think it dates back to the days of full-screen DirectX, which we've now replaced with a fake-fullscreen. It does need to be set for the device selector to consider the device valid, but maybe that's an assumption that should be revisited?

The pipeline needs to know the display mode so it can set the correct projection matrix. I'd have to look at the implementation but I think pl3DPipeline might handle that for the implementation. Beyond that I don't think it's used much by the pipeline.

Windowed mode gets complicated because the window is resizable by the user on Mac, it's not locked. But I don't remember if the DX engine bypasses the current display mode for windowed. I know there is some special handling for windowed mode. Looking to see what modern engines do with that, and it looks like World of Warcraft locks the windows aspect ratio so it can pretend it's in the same display mode even if the window is a different size.

Nothing to solve now, but stuff I'll be thinking about. I think there will need to be some sort of display mode selector for older hardware with integrated GPUs rendering to higher resolution displays. A MacBook Air connected to a 4k or 5k display could be... bad.

Hoikas:
At the moment, DX/Windows locks the window size to a specific set of display resolutions returned by DirectX. I think there are many users who would appreciate this restriction being lifted, but it hasn't happened yet.

dpogue:
I think this is ready to go.

The DX enumerate uses `emplace_back`, so I've matched that for consistency, but I'm open to changing both if that's desired.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLPlateManager.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+plGLPlateManager::plGLPlateManager(plGLPipeline* pipe)
+    : plPlateManager(pipe)
+{}
+
+plGLPlateManager::~plGLPlateManager()
+{}
+
+void plGLPlateManager::IDrawToDevice(plPipeline* pipe)
+{}
```
Hoikas, line 54: Newline at EOF :wink:

Sources/Plasma/FeatureLib/pfGLPipeline/plGLPlateManager.h
```diff
@@ -0,0 +1,64 @@
...
+public:
+    virtual ~plGLPlateManager();
+
+protected:
+    plGLPlateManager(plGLPipeline* pipe);
+
+    void IDrawToDevice(plPipeline* pipe) override;
+};
+
+#endif //_plGLPlateManager_inc_
```
Hoikas, line 64: Newline at EOF

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
```diff
@@ -317,10 +318,12 @@ class hsG3DDeviceSelector : public hsRefCnt
         kDefaultDepth   = 32
     };
 
+    typedef std::function<void(std::vector<hsG3DDeviceRecord>&)> DeviceEnumerator;
```
Hoikas, line 321: We probably need an include for `<functional>`.

Hoikas (commented):
Just a few minor stragglers.

Hoikas (approved):

PR #1078, opened 2022-02-14
Use `char16_t` over the wire protocols.

zrax:
... instead of assuming the size of `wchar_t` is 16-bit (hint: it isn't on most *nix platforms).

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglFile.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglFile.cpp
@@ -144,7 +144,7 @@ struct BuildIdRequestTrans : NetFileTrans {
 struct ManifestRequestTrans : NetFileTrans {
     FNetCliFileManifestRequestCallback  m_callback;
     void *                              m_param;
-    wchar_t                             m_group[MAX_PATH];
+    char16_t                            m_group[MAX_PATH];
     unsigned                            m_buildId;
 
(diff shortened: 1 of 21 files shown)


Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglFile.cpp
```diff
@@ -1112,7 +1120,8 @@ bool DownloadRequestTrans::Send () {
         return false;
 
     Cli2File_FileDownloadRequest filedownloadReq;
-    StrCopy(filedownloadReq.filename, m_filename.WideString().data(), std::size(filedownloadReq.filename));
+    const ST::utf16_buffer buffer = m_filename.AsString().to_utf16();
+    StrCopy(filedownloadReq.filename, buffer.data(), std::size(filedownloadReq.filename));
```
Hoikas, line 1124: Why not the std::char_traits copy method?

zrax: Because `st::char_traits<T>::copy` is more equivalent to `memcpy()`, wheras `StrCopy()` is closer to `strncpy()`.  Or more specifically, `char_traits` copies exactly the specified number of characters (regardless of content) whereas `StrCopy` copies up to either the first `'\0'` or the specified maximum.

Hoikas (approved):

PR #1080, opened 2022-02-15
Fix compatibility with CMake older than 3.18

dgelessus:
On those CMake versions, the target of an `ALIAS` cannot be a non-`GLOBAL` imported target, which `EXPAT::EXPAT` and `Freetype::Freetype` are. As a workaround, set them as `IMPORTED_GLOBAL` on the affected CMake versions.


Hoikas:
Thanks for pointing this out. I think it might be better to use the approach from #1019 to avoid changing targets we didn't create. For example:
```cmake
add_library(expat::expat INTERFACE IMPORTED)
set_target_properties(expat::expat PROPERTIES
    INTERFACE_LINK_LIBRARIES EXPAT::EXPAT
)
```

dgelessus:
Makes sense - and that approach seems to work just as well on CMake 3.17.

cmake/Findexpat.cmake
```diff
@@ -8,7 +8,12 @@ if(NOT TARGET expat::expat)
 
     if(EXPAT_FOUND)
         if(NOT TARGET expat::expat)
-            add_library(expat::expat ALIAS EXPAT::EXPAT)
+            if(${CMAKE_VERSION} VERSION_LESS 3.18)
```
Hoikas, line 11: This might be a bit pedantic, but I think we should just drop the CMake 3.18+ behavior and only use the code that works on older CMakes.

Hoikas (approved):

PR #1091, opened 2022-02-25
Add Python Scripts Verification to CI.

### LINKED ISSUE #939: Validate Magic Numbers
Currently, there are two disjoint set of magic numbers in our scripts: age sequence prefixes and `ptModifier` IDs. Currently, little to no validation occurs to ensure these values are unique. Indeed, Korman actually exports AgeSDLHook python files with an id of `-1` - and we now have two of those in repo... trebivdil.py and Vothol.py. When opening the Max plugin with these data, it displays an error about dupe IDs and drops both scripts on the floor. We need to add CI checks to ensure all of these supposedly unique integers are actually unique.

Hoikas:
These workflows post an automatic review of the Python code to all pull requests. Currently, the code tests if any PFM ids are duplicated (or missing) and if all Python scripts import correctly without exceptions. ~~To see an example of this in action, see Hoikas/Plasma#5. The bot will not post in this PR because the trusted workflow will only execute from the main branch nor will it post in any pull request until after this pull request is merged.~~ (This outdated method was replaced with annotations in the action.)

Fixes #939.


dpogue:
There is a way with GitHub Actions to add "annotations" to specific lines of the files in the PR, which might be a better experience than comments with filenames and line numbers. I can try to dig up some (probably very lacking) documentation on this.

Hoikas:
I would do that if I knew how to do more than copy and paste JS ;)

dpogue:
You might not need JS actually... You can use "problem matchers" which will run regex on the stdout of your tool:
https://github.com/actions/toolkit/blob/4a49e1bb389dabe514e49d31b8c7e313c156df92/docs/problem-matchers.md

Here is an example problem matcher for the `setup-python` action: https://github.com/actions/setup-python/blob/main/.github/python.json

Here's an example to surface MSVC compiler errors/warnings: https://github.com/ammaraskar/msvc-problem-matcher

Hoikas:
I've pushed an example of the annotation dealio. It's a little bit less obvious than the bot posting to the PR and the annotations can be difficult to find compared to a bot comment. That being said, getting rid of the whole trusted/untrusted workflow (and the limitation of the trusted workflow only working on main) is nice. This example just shows a module load error. I can push some more errors, if desired.

Scripts/Tests/xGamePythonReview.py
```diff
@@ -0,0 +1,190 @@
...
+
+    for id, dupe_files in get_duplicated_ids().items():
+        exitcode = 1
+        for py_file in dupe_files:
+            other_dupes = dupe_files - set((py_file,))
+            output.write(
+                result_str.format(
+                    status="WARNING",
+                    file=py_file.relative_to(repo_path),
+                    line="",
```
dpogue, line 143: You might need to specify a line for the annotation to work... it might be safe to hardcode 1 here, or (ideally) figure out which line actually has the `self.id` declaration on it

Hoikas: I'll push a break commit tomorrow to see if this works.

Hoikas: It looks like this works correctly. I am wondering, though, if it would make sense to change the regex to try to parse the actual Python tracebacks. I've got some additional work I want to do around the infodumps now that I see how this all looks in GHA, so this isn't ready yet.

.github/workflows/ci.yml
```diff
@@ -249,3 +249,21 @@ jobs:
       - name: Install
         run: |
           cmake --build build --target install -j 2
+
+
+  scripts:
```
dpogue, line 254: If you pull this out into its own workflow, you could use a file restriction to only run it when there are edits to Scripts/Python files: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

Hoikas: Good idea. I've split it out into a new workflow that only runs when either the python files or the workflow files change.

Hoikas:
@dpogue Is there anything else you think I need to do here?

dpogue (approved):
This looks good to me. We can, at the very least, get it merged as a start and address any improvements as they arise

Hoikas:
Had to rebase due to conflicts 😭

dpogue (approved):

PR #1095, opened 2022-03-01
Cancel running CI jobs when a PR is updated

dpogue:
Ref https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: CI
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   windows:
     runs-on: windows-2019


Hoikas (commented):
This seems like a good idea and looks correct. It's always annoying to have to cancel runs manually after a force-push. Have you tested that it works as expected?

dpogue:
This appears to work for the PR runs, but not the runs on branches of the fork itself. So it's probably an improvement, but less useful for the particular case that actually bothered me.

PR runs: https://github.com/H-uru/Plasma/actions/workflows/ci.yml?query=branch%3Aepoxy
Branch runs (in my fork): https://github.com/dpogue/Plasma/actions?query=branch%3Aepoxy

Hoikas (approved):
That looks like what I'd expect for this change to do.

PR #1097, opened 2022-03-02
Prevent deferred link race conditions.

Hoikas:
Per reports, it seems to be possible for deferred links that trigger multiple Age registrations from code can sometimes result in the player linking to the wrong Age. So, we add some validation to the deferred linking logic to ensure they go to where they expect.


Hoikas:
This passed testing on Minkata and appears to fix the issue it targeted. I'm inclined to merge it barring any objections in the next few days.

Hoikas:
Note that these changes will be included with the next MOULa update. If no one has any concerns, I'll merge this when that update lands.

PR #1098, opened 2022-03-04
Add margins to KI chat display to fix shadow clipping.

Hoikas:
When chat lines did not have a leading space character, you could sometimes see that shadows were being clipped due to drawing 2 pixels out of bounds. While I generally think managing the shadow unclipping would be better in the text rendering code, Plasma's text rendering is not that smart and works on the idea that text is rendered exactly where you specify it. So, we're largely obligated to handle this kind of text positioning logic in higher level code.

This issue became very apparent when testing #1063 due to the subtitles being drawn at x=0. To illustrate...
Before:
image

After:
image

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -5451,6 +5451,7 @@ class xKI(ptModifier):
             chatArea.lock()         # Make the chat display immutable.
             chatArea.moveCursor(PtGUIMultiLineDirection.kBufferEnd)
             chatArea.disableScrollControl()
+            chatArea.setMargins(left=2, right=2)
             btnUp = ptGUIControlButton(KIMicro.dialog.getControlFromTag(kGUI.miniChatScrollUp))
             btnUp.show()
             btnUp.hide()
@@ -5516,6 +5517,7 @@ class xKI(ptModifier):
             chatArea.moveCursor(PtGUIMultiLineDirection.kBufferEnd)
             # Hide the chat scroll buttons (should be nothing in chat area yet anyhow).
             chatArea.disableScrollControl()
+            chatArea.setMargins(left=2, right=2)
             btnUp = ptGUIControlButton(KIMini.dialog.getControlFromTag(kGUI.miniChatScrollUp))
             btnUp.show()
             privateChbox = ptGUIControlCheckBox(KIMini.dialog.getControlFromTag(kGUI.miniPrivateToggle))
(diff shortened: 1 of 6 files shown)


Scripts/Python/plasma/Plasma.py
```diff
@@ -4271,6 +4271,10 @@ def getKey(self):
         """Returns the ptKey for this GUI control"""
         pass
 
+    def getMargins(self) -> Tuple[int, int, int, int]:
+        """Returns a tuple of (top, left, right, bottom) margins"""
```
ZarothYe, line 4275: This comment doesn't match the ordering in `setMargins`: (top, left, bottom, right). I think the order in `setMargins` is or probably should be the correct one, although if your intention was to match something like CSS's margin/position descriptor rules, it'd be (top, right, bottom, left), IIRC.

Scripts/Python/ki/__init__.py
```diff
@@ -5516,6 +5517,7 @@ def ProcessNotifyMini(self, control, event):
             chatArea.moveCursor(PtGUIMultiLineDirection.kBufferEnd)
             # Hide the chat scroll buttons (should be nothing in chat area yet anyhow).
             chatArea.disableScrollControl()
+            chatArea.setMargins(left=2, right=2)
```
ZarothYe, line 5520: These additions look good to me, but can you double check whether this needs to also be done in `def RemoveKILevel`? It's possible that it's not needed, but that process does a lot of the same `chatArea` initialization shenanigans that these notification-based initializations are doing.

Scripts/Python/plasma/Plasma.py
Hoikas: Plasma's convention here matches the Win32 coordinate system.

ZarothYe: Understandable, I've also seen the Win32 coordinate system as well. I just wanted to make sure I understood your intention. In any case, the comments should match the reality, though. ;)

Hoikas: Agreed. I'm going through and fixing things now.

Scripts/Python/ki/__init__.py
Hoikas: Interesting. I don't think we need to re-do this because nothing else should be changing the KI chat area's margins.

ZarothYe (approved):

Hoikas:
Any objections here?

Deledrius (approved):

PR #1101, opened 2022-03-09
Fix moar Max `char*` being reinterpreted as `wchar_t*`.

Hoikas:


zrax (approved):

Hazado:
tested on xKIGUI.max and no more chinese character. Seems like it did its job.

PR #1102, opened 2022-03-10
Fix moar Max plugin crashes.

Hoikas:
Fixes Max plugin crashes reported by @Hazado:
- Hardens the pick node dialog setters against null strings, which are passed in by design.
- Fixed a crash when destroying the GUI Konstant Tag rollout in the GUI Dialog component. Likely related to 64-bit.
- Fixed more UTF-8 being reinterpreted as UCS-2 in note track related UI items.


zrax (approved):

Hazado (approved):
Used to make changes to xKIGUI.max and no more crashes

PR #1096, opened 2022-03-02
Options menu support for enabling/disabling subtitles

ZarothYe:
Separating out options menu script for enabling/disabling subtitles to be able to be merged later than #1063. Opening as draft so I can merge/rebase on top of master and set subtitles to OFF by default once the other PR is merged and the supporting max/PRP assets PR is ready to be merged as well.

PR #1103, opened 2022-03-10
Don't feed remaining subtitles when audio skips to past the end of last subtitle

ZarothYe:
In the Yeesha Cleft speech, it appears that when you press the imager button to stop the hologram in the middle of her speech, the audio skips close to the ending position and stays there for several `Update` cycles before actually ending, which was causing the updater to feed all the remaining subtitles up to that point. This doesn't seem to be the case when Kirel podium audio is stopped, so I don't exactly know why the Yeesha audio does this. Added a check to make sure the current position/time isn't past the end of the last subtitle before requesting to display entries up to that time.

This will collide with #1096 , so please let me rebase and merge/fix before merging this one.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
+++ b/Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
@@ -106,11 +106,14 @@ void plWin32Sound::Update()
     if (buf != nullptr) {
         plSrtFileReader* srtReader = buf->GetSrtReader();
         if (srtReader != nullptr) {
-            while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(GetActualTimeSec() * 1000.0f))) {
-                if (plgAudioSys::AreSubtitlesEnabled()) {
-                    // add a plSubtitleMsg to go... to whoever is listening (probably the KI)
-                    plSubtitleMsg* msg = new plSubtitleMsg(nextEntry->GetSubtitleText(), nextEntry->GetSpeakerName());
-                    msg->Send();
+            uint32_t currentTimeMs = (uint32_t)(GetActualTimeSec() * 1000.0f);
+            if (currentTimeMs <= srtReader->GetLastEntryEndTime()) {
+                while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(GetActualTimeSec() * 1000.0f))) {
+                    if (plgAudioSys::AreSubtitlesEnabled()) {
+                        // add a plSubtitleMsg to go... to whoever is listening (probably the KI)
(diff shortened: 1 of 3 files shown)


Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.h
```diff
@@ -92,6 +92,7 @@ class plSrtFileReader
     void            AdvanceToTime(uint32_t timeMs);
     plSrtEntry*     GetNextEntryStartingBeforeTime(uint32_t timeMs);
     plSrtEntry*     GetNextEntryEndingBeforeTime(uint32_t timeMs);
+    uint32_t        GetLastEntryEndTime();
```
Hoikas, line 95:
```suggestion
    uint32_t        GetLastEntryEndTime() const;
```

Sources/Plasma/PubUtilLib/plAudioCore/plSrtFileReader.cpp
```diff
@@ -181,3 +181,12 @@ plSrtEntry* plSrtFileReader::GetNextEntryEndingBeforeTime(uint32_t timeMs)
 
     return nullptr;
 }
+
+uint32_t plSrtFileReader::GetLastEntryEndTime()
```
Hoikas, line 185:
```suggestion
uint32_t plSrtFileReader::GetLastEntryEndTime() const
```

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -106,11 +106,14 @@ void plWin32Sound::Update()
...
         plSrtFileReader* srtReader = buf->GetSrtReader();
         if (srtReader != nullptr) {
-            while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(GetActualTimeSec() * 1000.0f))) {
-                if (plgAudioSys::AreSubtitlesEnabled()) {
-                    // add a plSubtitleMsg to go... to whoever is listening (probably the KI)
-                    plSubtitleMsg* msg = new plSubtitleMsg(nextEntry->GetSubtitleText(), nextEntry->GetSpeakerName());
-                    msg->Send();
+            uint32_t currentTimeMs = (uint32_t)(GetActualTimeSec() * 1000.0f);
+            if (currentTimeMs <= srtReader->GetLastEntryEndTime()) {
+                while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime((uint32_t)(GetActualTimeSec() * 1000.0f))) {
```
Hoikas, line 111: Is `(uint32_t)(GetActualTimeSec() * 1000.0f)` the same value as `currentTime` here?

ZarothYe: Well dangit, I think I borked that in the merge. Don't do merges late at night, kids. Yes, the `currentTime` variable was supposed to be used in both places. I'll fix.

Hoikas (approved):

PR #1105, opened 2022-03-11
Re-add stale Python attributes.

Hoikas:
Unfortunately, the attributes seem to be stored in max Python File Components by the order they appear in the python file rather than by their integer `id`. So, if you remove them, the max file can no longer be exported correctly. These changes fix the export of xKIGUI.max.

Original xKI.py:
image

Current ki module:
image

Error:
image


Hoikas:
I'm going to go ahead and merge this because it's blocking being about to export xKIGUI.max with modern versions of max (read: 64-bit). Hope no one minds!

PR #1106, opened 2022-03-12
Fix crash when `plClipboard::SetClipboardText()` is called twice.

Hoikas:
Found when experimenting with #1024.

Windows does not like it when you delete the HGLOBAL that you hand off to the clipboard. Subsequent calls to `GlobalLock()` will return nullptr and give the error "Invalid handle." - as in - you deleted my handle! As a bonus, cleanup some of the more C-ish aspects of this fxn.

### DIFF
--- a/Sources/Plasma/CoreLib/hsWindows.h
+++ b/Sources/Plasma/CoreLib/hsWindows.h
@@ -107,7 +107,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
         HRESULT fResult;
 
         hsCOMError() : fResult() { }
-        hsCOMError(hsLastWin32Error_Type) : fResult(HRESULT_FROM_WIN32(GetLastError())) { }
+        hsCOMError(hsLastWin32Error_Type, DWORD error) : fResult(HRESULT_FROM_WIN32(error)) { }
         hsCOMError(HRESULT r) : fResult(r) { }
         hsCOMError& operator =(const hsCOMError&) = delete;
         hsCOMError& operator =(HRESULT r) { fResult = r; return *this; }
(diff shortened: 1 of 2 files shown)


Sources/Plasma/CoreLib/hsWindows.h
```diff
@@ -97,12 +97,17 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    struct hsLastWin32Error_Type {};
+    constexpr hsLastWin32Error_Type hsLastWin32Error;
+
     /** COM Result holder used for formatting to log. */
     struct hsCOMError
     {
         HRESULT fResult;
 
         hsCOMError() : fResult() { }
+        hsCOMError(hsLastWin32Error_Type) : fResult(HRESULT_FROM_WIN32(GetLastError())) { }
```
zrax, line 110: This must be used carefully...  For example, if you construct an `hsCOMError()` as a parameter to a function call (or constructor), you need to make sure none of the other parameters will set the last error and cause this to get the wrong value...

Personally, I would prefer to see `GetLastError()` called explicitly by the calling code, and then an `hsCOMError` constructed with the result (presumably still with a tag type to distinguish it from the `HRESULT` overload)

Sources/Plasma/PubUtilLib/plClipboard/plClipboard.cpp
```diff
@@ -89,29 +91,38 @@ void plClipboard::SetClipboardText(const ST::string& text)
...
 
-    ::EmptyClipboard();
-
-    wchar_t* target = (wchar_t*)::GlobalLock(copy.get());
+    wchar_t* target = reinterpret_cast<wchar_t*>(GlobalLock(copy.get()));
+    if (target == nullptr) {
+        hsAssert(0, ST::format("GlobalLock() failed:\n{}", hsCOMError(hsLastWin32Error)).c_str());
+        return;
+    }
     memcpy(target, buf.data(), (len + 1) * sizeof(wchar_t));
```
zrax, line 115:
```suggestion
    memcpy(target, buf.data(), len * sizeof(wchar_t));
```
We explicitly initialize `target[len]` below, so no need to write it twice here...

zrax (approved):

PR #1107, opened 2022-03-12
Allow clickable URLs to begin with "www."

Hoikas:
The Kirel 2022 Q1 audio has subtitles of URLs that begin with only www. as the prefix. This will enable that common URL anti-pattern to also be highlighted and opened directly from the KI.

See H-uru/moul-assets#113.

### DIFF
--- a/Scripts/Python/xGUILinkHandler.py
+++ b/Scripts/Python/xGUILinkHandler.py
@@ -54,10 +54,10 @@ from PlasmaTypes import *
 from xCensor import xCensor
 
 # These regexes match URLs in the format of:
-# https://(anything not a space
+# https://(anything not a space)
 # http://(anything not a space)
 # www.(anything not a space or a dot).(anything not a space)
-_HOST_REGEX = re.compile(R"(?:https?:\/\/)?(?:www\.)?([^\/]+).*", re.IGNORECASE)
+_HOST_REGEX = re.compile(R"(?:https?:\/\/)?(?:www\.)?([^\/?#]+).*", re.IGNORECASE)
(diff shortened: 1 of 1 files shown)


Scripts/Python/xGUILinkHandler.py
```diff
@@ -53,8 +53,12 @@
 
 from xCensor import xCensor
 
-_HOST_REGEX = re.compile(R"https?:\/\/(?:www\.)?([^\/]+).*", re.IGNORECASE)
-_URL_REGEX = re.compile(R"https?:\/\/[\S]+", re.IGNORECASE)
+# These regexes match URLs in the format of:
+# https://(anything not a space
```
zrax, line 57:
```suggestion
# https://(anything not a space)
```

Scripts/Python/xGUILinkHandler.py
```diff
@@ -53,8 +53,12 @@
...
+# http://(anything not a space)
+# www.(anything not a space or a dot).(anything not a space)
+_HOST_REGEX = re.compile(R"(?:https?:\/\/)?(?:www\.)?([^\/]+).*", re.IGNORECASE)
```
zrax, line 60: You may also want to strip of fragments and queries, if they appear before the first `/`
```suggestion
_HOST_REGEX = re.compile(R"(?:https?:\/\/)?(?:www\.)?([^\/?#]+).*", re.IGNORECASE)
```

Also, I don't see a problem with leaving `www.` on the prompted hostname, unless we're worried it will cause the message to get too long.

zrax (approved):

PR #1108, opened 2022-03-16
Remove foreign swear words from the censor.

Hoikas:
Per Zaroth, `aujourd'hui` is being censored in the French Q1 2022 Kirel speech subtitles. "hui" is "today" in French but is vulgar slang for penis in Russian. Given the English-centric nature of the list and the possibility of false-friends, it seems best to simply remove foreign words from the censor list. Future work would be to add proper censored words for each language.

In other words, French speakers may now say "today" in KI chat without being censored. Also, VERGE!!!

### DIFF
--- a/Scripts/Python/xCensorFilters.py
+++ b/Scripts/Python/xCensorFilters.py
@@ -92,7 +92,6 @@ xSentenceFilters = [\
             "anus"          : Rating(xRatedR),\
             "apeshit"          : Rating(xRatedR),\
             "ar5e"          : Rating(xRatedR),\
-            "arschloch"          : Rating(xRatedR),\
             "arse"          : Rating(xRatedR),\
             "arrse"          : Rating(xRatedR),\
             "arsehole"          : Rating(xRatedR),\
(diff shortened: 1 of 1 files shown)


Scripts/Python/xCensorFilters.py
```diff
@@ -369,14 +367,8 @@
...
             "kike"          : Rating(xRatedX),\
-            "klootzak"          : Rating(xRatedR),\
             "koon"          : Rating(xRatedX),\
-            "knulle"          : Rating(xRatedX),\
-            "kuk"          : Rating(xRatedR),\
-            "kuksuger"          : Rating(xRatedX),\
             "kunt"          : Rating(xRatedX),\
-            "kurac"          : Rating(xRatedR),\
-            "kurwa"          : Rating(xRatedX),\
             "kyrpa"          : Rating(xRatedR),\
```
ZarothYe, line 372: You missed the Finnish penis...es. https://en.wiktionary.org/wiki/kyrp%C3%A4

Hoikas: I was wondering what that meant 🤣

ZarothYe:
Reviewed again because I needed a good laugh. You appear to still have `pula`, `pule`, and `suka`, which are Romanian, Norwegian, and Russian swears, respectively. Additionally, "pule" is a legitimate English verb, albeit quite uncommon. Otherwise the pared down list looks good to me. 👍

Hoikas:
Removed more bad words.

Hoikas:
This will be needed for the next MOULa update content to function correctly with our client. I intend to merge this when MOULa is next updated unless someone has any objections.

ZarothYe (approved):

PR #1110, opened 2022-03-19
Fix errors when HAVE_CPUID is false

dpogue:
If `HAVE_CPUID` was not defined or falsey, then we would not define a value for `cpuid_t` but we would also not guard any of the code that used `cpuid_t`, leading to compilation errors from the undefined type.


Hoikas (commented):
I would prefer to see an `#else` added at line 92 to define  `cpuid_t` to something like `int`. Then, all the other `#ifdefs` are irrelevant - the code already handles "wait, CPUID doesn't exist," I just goofed the type logic.

dpogue:
@Hoikas sounds good, PR updated

Hoikas (approved):

PR #1112, opened 2022-03-20
By popular demand, (re-)enable subtitles by default

ZarothYe:
Several more people requested this with quite reasonable logic, so by default subtitles will be ON for players who haven't disabled them yet. Calum said they'll argue the point with anyone who complains so I don't have to, yay! And yes, @Hoikas, you were right and I was wrong. 😜


Deledrius:
Seems sensible to me.  Can someone provide a summary of the supporting argument, for the sake of documentation?

Deledrius (approved):

ZarothYe:
My only reason for initially choosing off-by-default is that it changed the "original" or "pure" experience of the game and I thought (imagined?) that some people would get upset by that change. However, here are the various reasons given by others for why on-by-default is preferred:

- In general, it makes sense to have the game be more accessible/inclusive by default rather than less.
- People often don't pay attention to release notes, so it may be a long time before they know or realize the new feature is available.
- People may not encounter others who could help them find the option to enable subtitles until after the Cleft, where a large bulk of story-introductory speeches occur.
- Some people often play with their headphones off, being used to only needing them for experiencing ambiance or the rare player voice chat, and may not realize speech is happening in some areas where the speaker is not visible or audio is being added (e.g., Gahreesen Wall, Kirel auditorium, Bahro caves)
- Non-native English speakers might find it nice to be able to listen and read at the same time or look up words after the speech is finished, which would be possible since the KI does not clear the subtitles like closed captioning would. Enabling by default means that they wouldn't have to dig through an English-language options menu to find the right setting.
- It may enhance immersion for some people, as it draws a similarity between NPCs talking to the player and players chatting to each other. The KI is then more of a "universal interaction mechanic".

Hoikas (approved):

Deledrius:
@ZarothYe Thanks!

PR #1114, opened 2022-03-25
Adds new Linking Point to Gahreesen

Hazado:
Adds the definitions needed for linking panels to work correctly.


ZarothYe:
I know this has been out here for a while... so I need someone to refresh my memory. Can you define what these 3 additions are and how they are used, @Hazado ? Is there something about the Wall game that is broken without these?

Hazado:
This adds an additional linking spot in the upper area above the prison cells. It also puts in the linking spot for bahro stone to the gahreesen wall return spots from the maintainer nexi. There is no bahro stones in game yet that do this though.

PR #1117, opened 2022-04-01
Short-circuit setting KI font size when new size is same as current size

ZarothYe:
The Big KI's fontsize control is a weird continuous slider with only 5 distinct values. When sliding the slider, the SetFontSize method is constantly firing and presumably doing all the work and refreshing and whatnot that it does, even when the font size is exactly the same as before. At least for me, this can make the slider usage quite chunky and jerky, especially when there is a lot of text in the KI chat areas (e.g., subtitles from speeches). So... let's make it only do all the work when the new fontsize from the slider is actually different from the old value, yeah? It seems much smoother to me now, but I don't have timings or evidence to back up my claim or anything. 🤷

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -2421,11 +2421,6 @@ class xKI(ptModifier):
 
         PtDebugPrint("xKI.SetFontSize(): Setting font size to {} from old size of {}.".format(fontSize, currentSize), level=kWarningLevel)
 
-        if self.KILevel < kNormalKI:
-            mKIdialog = KIMicro.dialog
-        else:
-            mKIdialog = KIMini.dialog
-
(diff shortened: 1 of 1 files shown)


Deledrius:
Seems like a reasonable improvement to me.

Hoikas:
I think this check should be implemented in the engine-side C++ instead. That way, if someone else adds a similar feature to another GUI, they get this fix for free.

ZarothYe:
> 



> I think this check should be implemented in the engine-side C++ instead. That way, if someone else adds a similar feature to another GUI, they get this fix for free.

I am not sure of a good way to do this. The expensive part (the refreshing of ~4 controls) is called by the Python code, not the C++. The setFontSize parts in the C++ are extremely cheap operations and the refreshes are triggered independently.

Hoikas (approved):

PR #1118, opened 2022-04-01
Fix pfJournalBook parsing where wchar_t != 2 bytes

colincornaby:
wchar_t length can vary by compiler. On Clang/64 bit/macOS it's 4 bytes, not 2. This was causing parsing to fail.

With this change linking books are now rendering properly on Mac/Metal and are linkable. Other books that weren't rendering correctly (like the welcome book in Great Zero) are now rendering in the correct font. I haven't done a full comparison with the Windows version yet so I don't want to label those books as correctly rendering yet.

Also possible more wchar_t issues could be lurking. There is still some funny behavior in places like key mapping that may or may not be related. I haven't really investigated yet, but if the assumption was made wchar_t is always two bytes, that could show up elsewhere.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
+++ b/Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
@@ -2210,7 +2210,7 @@ bool    pfJournalBook::IGetNextOption( const wchar_t *&string, wchar_t *name, wc
         return false;
 
     // Copy name
-    size_t len = ((uintptr_t)string - (uintptr_t)c)/sizeof(wchar_t);
+    size_t len = string - c;
     wcsncpy( name, c, len );
     name[len] = L'\0';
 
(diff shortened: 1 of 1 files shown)


dpogue:
wchar_t is 2 bytes on Windows, 4 bytes pretty much everywhere else. Since these are coming in from Python, I think this makes sense...

We just have to be careful around network code, because that expects 16-bit wchars (which I think were all converted to char16_t in https://github.com/H-uru/Plasma/pull/1078)

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -2210,7 +2210,7 @@ bool    pfJournalBook::IGetNextOption( const wchar_t *&string, wchar_t *name, wc
         return false;
 
     // Copy name
-    size_t len = ((uintptr_t)string - (uintptr_t)c)/2; // divide length by 2 because each character is two bytes
+    size_t len = ((uintptr_t)string - (uintptr_t)c)/sizeof(wchar_t);
```
zrax, line 2213: `string` and `c` are already `wchar_t` pointers...  This can (and should) be simplified to just:
```suggestion
    size_t len = string - c;
```

colincornaby: That’s a really great point. I’ll make the change.

zrax (approved):

PR #1123, opened 2022-04-10
Add clipboard support to the console.

### LINKED ISSUE #1113: Feature request: Paste functionality in console
A helpful addition to the plasma/python consoles would be the ability to paste into it. Currently I am creating python files when I need longer commands at the ready, but pasting directly into the console would save a lot of time.

Hoikas:
A lot of the changes are just churn to avoid duplicating console logic. Unfortunately, this does not attempt to make the console unicode friendly. The console still uses the ACP.

If you paste multi-line text into the console, you will get this prompt:
image
and all intervening lines (ie all but the last) are executed.

Fixes #1113.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -557,8 +557,6 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
     else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_V)
     {
         ST::string text = plClipboard::GetInstance().GetClipboardText();
-        if (text.empty())
-            return;
 
         // Chop off leading or trailing newlines, which are probably not intended.
         text = text.trim("\r\n");
@@ -630,9 +628,11 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
         {
             // FIXME: make the console unicode friendly.
             IHandleCharacter((char)key);
-            findAgain = false;
-            findCounter = 0;
-            IUpdateTooltip();
+            if (strnlen(fWorkingLine, std::size(fWorkingLine)) < kMaxCharsWide - 2 && key != 0) {
+                findAgain = false;
+                findCounter = 0;
+                IUpdateTooltip();
+            }
         }
     }
 }
@@ -650,7 +650,7 @@ void    pfConsole::IHandleCharacter(const char c)
     size_t i = strnlen(fWorkingLine, std::size(fWorkingLine));
 
     // The current working line is too long. Ignore this character.
-    if (!(i < kMaxCharsWide - 2 && c != '\0'))
+    if (i >= kMaxCharsWide - 2 || c == '\0')
         return;
 
     // Advance any trailing characters one space to make room for the new character.
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -64,11 +64,13 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pnKeyedObject/plFixedKey.h"
 #include "pnNetCommon/plNetApp.h"
 
+#include "plClipboard/plClipboard.h"
 #include "plGImage/plPNG.h"
 #include "plInputCore/plInputDevice.h"
 #include "plInputCore/plInputInterface.h"
 #include "plInputCore/plInputInterfaceMgr.h"
 #include "plMessage/plCaptureRenderMsg.h"
+#include "plMessage/plConfirmationMsg.h"
 #include "plMessage/plConsoleMsg.h"
 #include "plMessage/plInputEventMsg.h"
 #include "plMessage/plInputIfaceMgrMsg.h"
@@ -357,7 +359,7 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
 {
     char            *c;
     wchar_t         key;
-    int             i,eol;
+    size_t          i,eol;
     static bool     findAgain = false;
     static uint32_t   findCounter = 0;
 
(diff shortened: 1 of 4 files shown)


ZarothYe:
Leaving a comment/review here as well for transparency:

The only thing I wonder if we should do is if the content ends with an EOL, maybe we strip that off the end before we decide that it's multi-line. I could be copying and pasting just one command line from a multi-line larger file, and if I'm not careful I might copy `import Plasma<EOL>` instead of `import Plasma` and that gets treated like a multi-line command instead of the single command that it is.

Hoikas:
OK, I've pushed a change to trim newline-ish characters from both ends of the string.

ZarothYe:
> OK, I've pushed a change to trim newline-ish characters from both ends of the string.

Have confirmed this is working after today's commits. Yay!

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -525,145 +527,76 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
...
-        fWorkingLine[ fWorkingCursor = 0 ] = 0;
+        IExecuteWorkingLine();
         findAgain = false;
         findCounter = 0;
         IUpdateTooltip();
     }
     else if( msg->GetKeyCode() == KEY_END )
     {
-        fWorkingCursor = strlen( fWorkingLine );
+        fWorkingCursor = strnlen(fWorkingLine, std::size(fWorkingLine));
```
ZarothYe, line 537: I did some reading up on these functions (they are C functions, not C++, right?) and wondered if `strnlen_s` might be better, as it also returns 0 if `fWorkingLine` is null and `strnlen` will not do so? I was reading from: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strnlen-strnlen-s?view=msvc-170

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -525,145 +527,76 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
...
+    else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_V)
+    {
+        ST::string text = plClipboard::GetInstance().GetClipboardText();
+        if (text.empty())
+            return;
+
+        // Chop off leading or trailing newlines, which are probably not intended.
+        text = text.trim("\r\n");
+        if (text.empty())
+            return;
```
ZarothYe, line 566: Since `ST::string` instances are never null (right?), is it worth condensing these to do the `GetClipboardText()` and then trim before doing the if-empty-then-return bit? So then we're only checking once and not repeating?

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -693,21 +626,39 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
...
+        return;
+    if (c == '\n') {
+        IExecuteWorkingLine();
+        return;
+    }
+
+    size_t i = strnlen(fWorkingLine, std::size(fWorkingLine));
+
+    // The current working line is too long. Ignore this character.
+    if (!(i < kMaxCharsWide - 2 && c != '\0'))
```
ZarothYe, line 653: Super minor quibble, but I feel like De-Morgan-ing this makes it clearer and match the comment better.
```suggestion
    if (i >= kMaxCharsWide - 2 || c == '\0'))
```

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -1002,6 +953,120 @@ void    pfConsole::IUpdateTooltip()
...
+    if (fWorkingLine[0] == 0 && !fHelpMode && !fPythonMode) {
+        // Blank line--just print a blank line to the console and skip
+        IAddLine("");
+        return;
+    }
+
+    // only save history line if there is something there
+    if (fWorkingLine[0] != 0) {
+        // Save to history
+        strcpy(fHistory[fPythonMode].fData[fHistory[fPythonMode].fCursor], fWorkingLine);
```
ZarothYe, line 982: The instances of `strlen` were changed to `strnlen`. Is it worth similarly changing this instance of `strcpy` to `strncpy`?

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -1002,6 +953,120 @@ void    pfConsole::IUpdateTooltip()
...
+        fHistory[fPythonMode].fRecallCursor = fHistory[fPythonMode].fCursor;
+    }
+
+    // EXECUTE!!! (warning: DESTROYS fWorkingLine)
+    if (fHelpMode) {
+        if (fWorkingLine[0] == 0) {
+            IPrintSomeHelp();
+        } else if (stricmp(fWorkingLine, "commands") == 0) {
+            char empty[] = "";
+            fEngine->PrintCmdHelp(empty, IAddLineCallback);
```
ZarothYe, line 993: I get that this is just refactored code from elsewhere, but is there a reason NOT to get rid of this first line and just use `fEngine->PrintCmdHelp("", IAddLineCallback)`? It feels strange to create a char array for the empty string, but maybe there's a C/C++/pointer-y reason that I'm not thinking of. The `empty` variable is never used again...

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -1002,6 +953,120 @@ void    pfConsole::IUpdateTooltip()
...
+            // else we are just doing single lines
+            // was there actually anything in the input buffer?
+            if (fWorkingLine[0] != 0) {
+                AddLineF(">>> %s", fWorkingLine);
+                // check to see if this is going to be a multi line mode ( a ':' at the end)
+                if (fWorkingLine[eol - 1] == ':') {
+                    fPythonMultiLines = 1;
+                } else
+                    // else if not the start of a multi-line then execute it
+                {
```
ZarothYe, line 1044:
```suggestion
                } else {
                    // else if not the start of a multi-line then execute it
```

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -1002,6 +953,120 @@ void    pfConsole::IUpdateTooltip()
...
+        }
+        // find the end of the line
+        for (c = fWorkingLine, eol = 0; *c != 0; eol++, c++);
+    } else {
+        if (!fEngine->RunCommand(fWorkingLine, IAddLineCallback)) {
+            c = (char*)fEngine->GetErrorMsg();
+            if (c[0] != 0)
+                AddLine(c);
+        }
+    }
```
ZarothYe, line 1064:
```suggestion
    } else if (!fEngine->RunCommand(fWorkingLine, IAddLineCallback)) {
        c = (char*)fEngine->GetErrorMsg();
        if (c[0] != 0)
            AddLine(c);
    }
```

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Hoikas: `strlen_s` is a Microsoft-only function in the MSVCRT and cannot be used on non-Windows platforms (unless you build with something esoteric, like winelib).

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -693,21 +626,39 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
...
+
+    size_t i = strnlen(fWorkingLine, std::size(fWorkingLine));
+
+    // The current working line is too long. Ignore this character.
+    if (!(i < kMaxCharsWide - 2 && c != '\0'))
+        return;
+
+    // Advance any trailing characters one space to make room for the new character.
+    for (i += 1; i > fWorkingCursor; --i)
+        fWorkingLine[i] = fWorkingLine[i - 1];
```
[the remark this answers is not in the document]
Hoikas: I think it's safe to assume any code operating on the end of the string has null-terminated it.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Hoikas: This is code that was copied and pasted from the `KEY_ENTER` case, and I would prefer to not change it in this PR.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Hoikas: See above, I didn't want to go down the rabbit hole of refactoring code that I was just moving to a new location. That would greatly increase the scope and complexity of this changeset.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -693,21 +626,39 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
...
-                fWorkingLine[ i ] = fWorkingLine[ i - 1 ];
-
-            // TODO: What about keys outside of `char` range?
-            fWorkingLine[fWorkingCursor++] = char(key);
-
+            // FIXME: make the console unicode friendly.
+            IHandleCharacter((char)key);
             findAgain = false;
             findCounter = 0;
             IUpdateTooltip();
```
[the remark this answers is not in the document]
Hoikas: It should be fine - the difference will be that any time someone types in a null character (can you actually do this?) or the line is too long, the tooltip will get regenerated. I can guard against that happening to prevent odd semantics, however.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Hoikas: See above

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Hoikas: See above

ZarothYe (approved):

Hoikas:
Anyone have any ohbjections to this?

dpogue (approved):
LGTM

PR #1124, opened 2022-04-10
Allow usage of /textcolor command to override chat area text colors

ZarothYe:
Needs to be merged with its paired assets (LOCs) PR: https://github.com/H-uru/moul-assets/pull/119

URL/Link color is not affected. Changing that will be a hairier deal than the overriding the other colors and I was rather doubtful that making the clickable links blend into the other text was The Right Thing To Do. Open to opinions.

Also need to decide whether to make the `kColorNames` constant a dictionary with localized keys for translated color names and dictionary values as the color names required to use with `ptColor`. Or something like that... to support non-English color naming.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -44,6 +44,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 import re
 import time
 import random
+from inspect import signature
 
 # Plasma Engine.
 from Plasma import *
@@ -79,6 +80,7 @@ class xKIChat(object):
         self.onlyAllowBuddiesOnRequest = False
         self.privateChatChannel = 0
         self.toReplyToLastPrivatePlayerID = None
+        self.chatTextColor = None
 
         # Fading & blinking globals.
         self.currentFadeTick = 0
@@ -601,6 +603,11 @@ class xKIChat(object):
         else:
             chatMentions = []
 
+        # if we have an override chat color set, use it instead of the various default chat colors
+        headerColor = self.chatTextColor if self.chatTextColor is not None else headerColor
+        bodyColor = self.chatTextColor if self.chatTextColor is not None else bodyColor
+        mentionColor = self.chatTextColor if self.chatTextColor is not None else mentionColor
+
         for chatArea in (self.miniChatArea, self.microChatArea):
             with PtBeginGUIUpdate(chatArea):
                 savedPosition = chatArea.getScrollPosition()
@@ -884,12 +891,14 @@ class CommandsProcessor:
         # Does the message contain a standard command?
         for command, function in commands.items():
             if msg.startswith(command):
-                theMessage = message.split(" ", 1)
+                callableCommandFn = getattr(self, function)
+                numParams = len(signature(callableCommandFn).parameters)
+                PtDebugPrint(f"xKI.CommandsProcessor: Processing {command} function with {numParams} parameters", level=kDebugDumpLevel)
+                theMessage = message.split(" ", numParams)
                 if len(theMessage) > 1 and theMessage[1]:
-                    params = theMessage[1]
+                    callableCommandFn(*theMessage[1:])
                 else:
-                    params = None
-                getattr(self, function)(params)
+                    callableCommandFn(None)
                 return None
 
         # Is it a simple text-based command?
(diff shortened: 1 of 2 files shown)


Scripts/Python/ki/xKIConstants.py
```diff
@@ -69,6 +69,10 @@
 ## The name of the KI light scene object.
 kKILightObjectName = "RTOmniKILight"
 
+## The color names usable to set the KI text color override
+## TODO: Can/should these be localized somehow? Would need conversion back to English before used as the ptColor method names.
+kColorNames = set(["black", "blue", "brown", "cyan", "darkbrown", "darkgreen", "darkpurple", "gray", "green", "magenta", "maroon", "navyblue", "orange", "pink", "red", "slateblue", "steelblue", "tan", "white", "yellow"])
```
Hoikas, line 74:
```suggestion
kColorNames = {"black", "blue", "brown", "cyan", "darkbrown", "darkgreen", "darkpurple", "gray", "green", "magenta", "maroon", "navyblue", "orange", "pink", "red", "slateblue", "steelblue", "tan", "white", "yellow"}
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1208,6 +1217,63 @@ def ChangePassword(self, newPassword):
...
+        if not arg1:
+            self.chatMgr.AddChatLine(None, PtGetLocalizedString("KI.Errors.MalformedChatSetTextColorCmd"), kChat.SystemMessage)
+            return
+
+        try:
+            colorRed = float(arg1)
+            colorBlue = float(arg2)
+            colorGreen = float(arg3)
+        except ValueError:
+            pass # arguments weren't valid float numbers, so we will try treating arg1 as a string below
```
Hoikas, line 1234: PEP8 says comments shall be on their own line.

Scripts/Python/ki/xKIChat.py
```diff
@@ -44,6 +44,7 @@
 import re
 import time
 import random
+from inspect import signature
```
Hoikas, line 47: I have a slight preference for importing just `inspect` and calling `inspect.signature()` for clarity.

Hoikas (commented):
I think it might also be useful to accept int triplets in addition to float triplets.

Scripts/Python/ki/xKIChat.py
ZarothYe: Fair, I'll modify. I haven't had any insomnia lately that would drive me to read the PEP8 documentation. 😛

ZarothYe:
> I think it might also be useful to accept int triplets in addition to float triplets.

What range are you picturing for this? From 0 to 255, similar to the hex numbers?

Hoikas:
Yeah, 0 to 255, just in non-hex 😉

Scripts/Python/ki/xKIChat.py
```diff
@@ -1208,6 +1217,80 @@ def ChangePassword(self, newPassword):
...
+            self.chatMgr.DisplayStatusMessage(PtGetLocalizedString("KI.Chat.TextColorSetValue", [newColor]))
+        elif newColor == PtGetLocalizedString("KI.Commands.ChatSetTextColorDefaultArg"):
+            # user requested resetting text colors to defaults
+            self.chatMgr.chatTextColor = None
+            self.chatMgr.DisplayStatusMessage(PtGetLocalizedString("KI.Chat.TextColorSetDefault"))
+        else:
+            hexColor = newColor.lstrip("#")
+            if len(hexColor) == 3:
+                # turn RGB into RRGGBB
+                hexColor = f"{hexColor[0]}{hexColor[0]}{hexColor[1]}{hexColor[1]}{hexColor[2]}{hexColor[2]}"
```
Hoikas, line 1275: Am I reading that int triplets are expected to be in hexadecimal format? I would expect to be able to provide them in decimal, eg `/textcolor 255 255 255` because that's how most programs typically deal with int triplets.

Scripts/Python/ki/xKIConstants.py
```diff
@@ -286,29 +310,30 @@ class kColors:
...
+        "/look in pocket" : "LookForFeathers"}
     Text = {"/go" : "Put one foot in front of the other and eventually you will get there.",
             "/fly" : "You close your eyes, you feel light headed and the ground slips away from your feet... Then you open your eyes and WAKE UP! (Ha, you can only dream about flying.)"}
     Other = {"/party" : "PartyTime",
-             "/saveclothing" : "SaveClothing",
-             "/loadclothing" : "LoadClothing",
-             "/threaten" : "CoopExample",
+        "/saveclothing" : "SaveClothing",
+        "/loadclothing" : "LoadClothing",
+        "/threaten" : "CoopExample",
```
Hoikas, line 336: Can we back out these changes and defer to another PR?

ZarothYe: Dangit, I thought I had reverted those properly. Sorry.

Scripts/Python/ki/xKIChat.py
ZarothYe: You are reading it wrong. Look at the first `try-except-else` block for 0-255 int parsing.

Hoikas: Ok. Note to self, this is for something like #000.

Hoikas (approved):

PR #1125, opened 2022-04-10
Customize the libepoxy vcpkg port

dpogue:
- Resolve multithreading issues
- Enable building as a static library

Unfortunately, using it as a static library causes linking errors, so we're keeping the line that forces it to be a dynamic library on Windows.

Ref: https://github.com/anholt/libepoxy/pull/265


Hoikas:
I wonder if the patch file should contain authorship information for https://github.com/anholt/libepoxy/pull/265?

dpogue:
> I wonder if the patch file should contain authorship information for [anholt/libepoxy#265](https://github.com/anholt/libepoxy/pull/265)?

Yeah, probably a good plan. It's not that diff directly because there are conflicts against the latest libepoxy, so I had to make the changes manually locally and generate a new diff file, but there no changes that did not originate in that PR

PR #1126, opened 2022-04-11
Adds ability to reward clothing items on visit to hood

Hazado:
Request was put to me to setup a clothing reward system when visiting your Hood.

Since we already did this with the Relto, just copied the code from that.

### DIFF
--- a/Scripts/Python/Neighborhood.py
+++ b/Scripts/Python/Neighborhood.py
@@ -81,3 +81,54 @@ class Neighborhood(ptResponder):
             
     def OnNotify(self,state,id,events):
         pass
+
+    def _AddClothingFromAgeSDL(self, clothingList, *sdls):
+        ageSDL = PtGetAgeSDL()
+        for i in sdls:
+            if not ageSDL[i][0]:
+                PtDebugPrint(f"Neighborhood._AddClothingFromAgeSDL(): You don't have the SDL '{i}' so no '{clothingList}' for you", level=kWarningLevel)
+                return
+
+        if isinstance(clothingList, str):
+            clothingList = [clothingList]
+
+        avatar = PtGetLocalAvatar()
+        if avatar.avatar.getAvatarClothingGroup() == kFemaleClothingGroup:
+            clothingPrefix = "FReward"
+        else:
+            clothingPrefix = "MReward"
+
+        entireClothingList = avatar.avatar.getEntireClothingList(0)
+        for i in filter(None, clothingList):
+            desiredName = "_".join((clothingPrefix, i))
+            if desiredName not in entireClothingList:
+                PtDebugPrint(f"ERROR: Neighborhood._AddClothingFromAgeSDL(): Invalid clothing item '{desiredName}' requested!")
+                continue
+            if not any((j[0] == desiredName for j in avatar.avatar.getWardrobeClothingList())):
+                PtDebugPrint(f"Neighborhood._AddClothingFromAgeSDL(): Adding '{desiredName}' clothing item to your closet! Aren't you lucky?", level=kWarningLevel)
+                avatar.avatar.addWardrobeClothingItem(desiredName, ptColor().white(), ptColor().white())
+            else:
+                PtDebugPrint(f"Neighborhood._AddClothingFromAgeSDL(): You already have {desiredName} so I'm not going to add it again.", level=kDebugDumpLevel)
+
+    def OnServerInitComplete(self):
+        if not ptVault().amOwnerOfCurrentAge():
+            PtDebugPrint("Neighborhood.OnServerInitComplete(): This isn't my Hood, so no reward clothing grants.", level=kWarningLevel)
+            return
+
+        ageSDL = PtGetAgeSDL()
+        try:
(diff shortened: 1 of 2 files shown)


ZarothYe:
Can you explain the desired use case or example a little more? I don't think I fully understand why this is wanted/needed, or what the difference is between the local and global clothing lists in the SDL are.

Scripts/SDL/Neighborhood.sdl
```diff
@@ -1966,3 +1966,140 @@ STATEDESC Neighborhood
...
+    VAR BOOL            nb01BahroBoatsEnabled[1]                DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01BahroBoatsChance[1]                 DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BOOL            nb01BahroBoatsProximity[1]              DEFAULT=0   DEFAULTOPTION=VAULT
+
+# HoodClothing is a list of names of reward clothing to add to the player's closet, separated by semicolins
+# (do NOT use this for clothing items to be given to everyone)
+    VAR STRING32    HoodClothing[1]                           DEFAULTOPTION=VAULT
+
+# GlobalHoodClothing is a list of names of reward clothing to add to the player's closet, separated by semicolins
+# (do NOT use this for clothing items to be given to a select group of people)
```
ZarothYe, line 2102:
```suggestion
# HoodClothing is a list of names of reward clothing to add to the player's closet, separated by semicolons
# (do NOT use this for clothing items to be given to everyone)
    VAR STRING32    HoodClothing[1]                           DEFAULTOPTION=VAULT

# GlobalHoodClothing is a list of names of reward clothing to add to the player's closet, separated by semicolons
# (do NOT use this for clothing items to be given to a select group of people)
```

Hazado:
Well guild of greeters has a shirt, only way to get the shirt is for it to be rewarded by shard owner.
With this code, shard owner could add the shirt to the HoodClothing SDL for guild of greeters Hood and save themselves time having to assign it to every person manually.

GlobalHoodReward is if you want to reward a shirt for every hood in the game.

Scripts/Python/Neighborhood.py
```diff
@@ -81,3 +81,54 @@ def OnFirstUpdate(self):
...
+
+        avatar = PtGetLocalAvatar()
+        if avatar.avatar.getAvatarClothingGroup() == kFemaleClothingGroup:
+            clothingPrefix = "FReward"
+        else:
+            clothingPrefix = "MReward"
+
+        entireClothingList = avatar.avatar.getEntireClothingList(0)
+        for i in filter(None, clothingList):
+            desiredName = "_".join((clothingPrefix, i))
```
ZarothYe, line 103:
```suggestion
            desiredName = f"{clothingPrefix}_{i}"
```
I could be wrong, but I think generally f-strings will be preferable and perform better than `.join` or `+` in Python.

Scripts/SDL/Neighborhood.sdl
Hazado: fixed

Scripts/Python/Neighborhood.py
```diff
@@ -81,3 +81,54 @@ def OnFirstUpdate(self):
...
+            if desiredName not in entireClothingList:
+                PtDebugPrint(f"ERROR: Neighborhood._AddClothingFromAgeSDL(): Invalid clothing item '{desiredName}' requested!")
+                continue
+            if not any((j[0] == desiredName for j in avatar.avatar.getWardrobeClothingList())):
+                PtDebugPrint(f"Neighborhood._AddClothingFromAgeSDL(): Adding '{desiredName}' clothing item to your closet! Aren't you lucky?", level=kWarningLevel)
+                avatar.avatar.addWardrobeClothingItem(desiredName, ptColor().white(), ptColor().white())
```
ZarothYe, line 109: Will there be a KI system message to notify the user of the new wardrobe item(s)?

Hazado: There isn't currently for the relto age.
I could add it, but it requires a lot of extra code.
Feel like we should probably make the extra code a helper function instead.

ZarothYe: I didn't realize it wasn't already a shared/helper function. I am 👍 on refactoring and using shared code. Probably in a separate PR would be best, though.

Hazado:
> I am disinclined to accept this because it adds a trivial avenue for reward clothing granting control to be taken away from the shard owner in favor of an easily manipulable in game mechanic. This facility belongs to the shard owner to do specific, targeted grants, IMO.

I was approached because a hood wanted to have a custom shirt for the hood members.
It doesn't mean that the GoG shirt will be done this way.
This doesn't stop the Shard Owner from still using the targeted rewards as before.
It adds an additional way to reward clothing for the shard owner to use as appropriate.
Yes it could be abused. Do we want to also put in a removal mechanism as well like the guild shirts have?

Hoikas:
That sounds even worse - we’re copying and pasting in code, increasing our maintenance burden, because a small group wants to be an exclusive club. I really don’t think this is a good idea.

PR #1133, opened 2022-04-23
Fix 3dsMax crash caused by Garrison Wall.

Hoikas:
Plasma functions should never be executed in the constructor. This is because you may not be executing in the client... for example, it might be a type checker or test tool. In this case, the network client is not initialized, so the call to `PtGetDniTime()` explodes.


ZarothYe (approved):
It compiles, ship it!

Hazado (approved):

PR #1131, opened 2022-04-22
Update xLinkingBookDefs.py for Elonin

DoobesURU:
Adds Book definitions for the Elonin cavern location.

PR #1136, opened 2022-04-28
Add Elonin

DoobesURU:
Adds script files for Elonin cavern location by Dulcamara and maintained by ametist

### DIFF
--- a/Scripts/Python/Elonin.py
+++ b/Scripts/Python/Elonin.py
@@ -1,8 +1,50 @@
+""" *==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
+PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+licensors of this Program grant you additional
+permission to convey the resulting work. Corresponding Source for a
+non-source form of such a combination shall include the source code for
+the parts of OpenSSL and IJG JPEG Library used as well as that of the covered
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+ *==LICENSE==* """
+
 from Plasma import *
 from PlasmaTypes import *
 
 class Elonin(ptResponder):
     def __init__(self):
         ptResponder.__init__(self)
-        self.id = -1
+        self.id = 481
         self.version = 0
--- a/Scripts/Python/Elonin.py
+++ b/Scripts/Python/Elonin.py
@@ -1,0 +1,8 @@
+from Plasma import *
+from PlasmaTypes import *
+ 
+class Elonin(ptResponder):
+    def __init__(self):
+        ptResponder.__init__(self)
+        self.id = -1
+        self.version = 0
--- a/Scripts/Python/Elonin.py
+++ b/Scripts/Python/Elonin.py
@@ -1,6 +1,6 @@
 from Plasma import *
 from PlasmaTypes import *
- 
+
 class Elonin(ptResponder):
     def __init__(self):
         ptResponder.__init__(self)
--- a/Scripts/SDL/Elonin.sdl
+++ b/Scripts/SDL/Elonin.sdl
@@ -1,0 +1,80 @@
+# /*==LICENSE==*
+# 
+# CyanWorlds.com Engine - MMOG client, server and tools
+# Copyright (C) 2011 Cyan Worlds, Inc.
+# 
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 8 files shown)


DoobesURU:
So yeah...no idea why the scripts review blew up. Any ideas on how to fix it?

ZarothYe:
> So yeah...no idea why the scripts review blew up. Any ideas on how to fix it?
> 
> EDIT: Looks like removing that whitespace in the Python file didn't do the trick 🙁

I don't think it's your fault. It's choking on something in `xGUILinkHander.py` (you can see the failure details in the "Files changed" tab). It looks like the recently-merged python checker is syntactically checking against 3.8 and maybe something in that script file is requiring something newer? I don't know. @Hoikas will have to diagnose. Just didn't want you to feel too bad or tear your hair out.

Scripts/Python/Elonin.py
```diff
@@ -0,0 +1,8 @@
+from Plasma import *
+from PlasmaTypes import *
+
+class Elonin(ptResponder):
+    def __init__(self):
+        ptResponder.__init__(self)
+        self.id = -1
```
ZarothYe, line 7: Is it expected that this `id` is -1? All the other Age python files that I've found have a positive (usually in the thousands or more) `id` value...

Scripts/Python/Elonin.py
```diff
@@ -0,0 +1,8 @@
+from Plasma import *
```
Hoikas, line 1: License block is missing.

Scripts/Python/Elonin.py
Hoikas: We'll want to manually assign an ID to this. Korman automatically exports this file with `self.id = -1` to ensure that AgeSDL works correctly.

Scripts/SDL/Elonin.sdl
```diff
@@ -0,0 +1,80 @@
...
+    VAR BOOL    flwrPuzSolved[1] DEFAULT=0
+    VAR BOOL    flwrPuzInUse[1] DEFAULT=0
+    VAR INT     keyPzlCorrectPushes[1] DEFAULT=0
+    VAR BOOL    keyPzlSolved[1] DEFAULT=0
+    VAR BOOL    keyPzlInUse[1] DEFAULT=0
+    VAR BOOL    fireOne[1] DEFAULT=0
+    VAR BOOL    fireTwo[1] DEFAULT=0
+ 
+ 
+}
```
Hoikas, line 60: Is this version used on any production shards? If not, I think we should omit it.

Scripts/SDL/Elonin.sdl
```diff
@@ -0,0 +1,80 @@
...
+ 
+STATEDESC Elonin
+ 
+{
+   VERSION 2
+ 
+    VAR BOOL    fntOnOff[1] DEFAULT=0
+    VAR BOOL     birdie[1] DEFAULT=0
```
Hoikas, line 68:
```suggestion
    VAR BOOL    birdie[1] DEFAULT=0
```

Hoikas (commented):
To encourage an organized history, I think the changes from #1131 should be moved to this PR.

Hoikas:
Please rebase this on top of the latest master to resolve the failures.

Scripts/SDL/Elonin.sdl
DoobesURU: > Is this version used on any production shards? If not, I think we should omit it.

This is the exact version used on Minkata/MOULa. /shrug

Scripts/Python/Elonin.py
DoobesURU: Should the ID be the same as the sequence prefix or something else?

Scripts/Python/Elonin.py
DoobesURU: Added. Once we decide on a `self.id`, I'll push that along with this fix.

Scripts/SDL/Elonin.sdl
DoobesURU: Fixed. Just need advice on the Elonin.py fix before I push a new commit.

Scripts/SDL/Elonin.sdl
Hoikas: So version 2 is unused?

Scripts/Python/Elonin.py
Hoikas: Any random 32-bit integer that doesn't cause the scripts review to detect a conflict will work.

Scripts/SDL/Elonin.sdl
DoobesURU: > So version 2 is unused?

I'm not sure. Just carrying over what's on OU, FWIW. Ametist would know more about the finer details.

ametist365: I don't know what you mean...I have in Elonin.sdl two versions, the latest VERSION 2 has two more variables than version 1.

Hoikas: I would like to avoid accumulating "development churn" into mainline. If version 2 was the first SDL used on production shards, (eg Gehn or MOULa), then I think we should consider omitting version 1 in the repository. In that way, we don't accumulate more SDL files with 50+ versions that have to be scrolled through for a shard admin to figure out what's going on.

DoobesURU: > I would like to avoid accumulating "development churn" into mainline. If version 2 was the first SDL used on production shards, (eg Gehn or MOULa), then I think we should consider omitting version 1 in the repository. In that way, we don't accumulate more SDL files with 50+ versions that have to be scrolled through for a shard admin to figure out what's going on.

Fair enough. I just committed a new version of the SDL with `VERSION 2` as `VERSION 1`.

DamnBriggsy: Won't this cause problems if used on MOULa? The version there should be updated as well to avoid conflicts.

Hoikas:
This PR has broken the Elonin state blobs on TrollLand.

Scripts/SDL/Elonin.sdl
ZarothYe: > Won't this cause problems if used on MOULa? The version there should be updated as well to avoid conflicts.

To answer this question and also elaborate on Adam's comment regarding the state blobs on TrollLand... yes, this change will cause major problems if it is used on MOULa. You cannot re-number a version of an SDL as @DoobesURU did with version 2 becoming version 1. What Adam was suggesting originally was to omit the version 1 block entirely and keep only version 2 **as version 2**.

Please do not take this PR with your self-built clients to MOULa.

dpogue:
There's a Nexus.sdl change in here that I think is live on MOULa now, so we probably need to either merge this (since Elonin is also live on MOULa) or pull the Nexus.sdl into its own PR for merging

DamnBriggsy:
> There's a Nexus.sdl change in here that I think is live on MOULa now, so we probably need to either merge this (since Elonin is also live on MOULa) or pull the Nexus.sdl into its own PR for merging

I think it needs the updated islmRegisterNexusLink.py, nxusBookMachine.py, and xLinkingBookDefs.py to be compatible with MOULa, too. Just replacing the nexus.sdl allows you to link to the nexus, but there are still some bugs.

Hoikas:
Yeah, we probably need to split out the Nexus parts into a separate PR. Also, we really need to fix islmRegisterThingy.py to have a way to avoid the whole special-casing nonsense.

Hoikas:
Please fix the conflicts - this PR is now breaking TL deployment.

Hoikas:
PR is in conflict with master.

DoobesURU:
Given some conflicts this PR is having, going to do a fresh PR for this as well. Closing.

PR #1137, opened 2022-04-28
(Python Checker Test) Turn /sneeze into a looping emote because spring allergies!

ZarothYe:
Sneezing is never normal. I never sneeze! 😉 

This PR is actually to check if the python checker is failing just for Doobes's GoMePubNew/Elonin PRs or for all PRs that include python.


Hoikas:
OK to close in light of #1138?

PR #1139, opened 2022-04-30
Fix for divide by zero breaking Mac/Vee Tsah

colincornaby:
This divide by zero was resolving to a INF value, which was causing the shader to generate garbage. Changing the divide by zero to non IEEE behavior which aligns with what HLSL 1.1 seems to expect. Metal inherits the IEEE behavior which causes problems.

More background - 

I've been trying to figure out why this hasn't caused problems in HLSL. My best guess is that HLSL treats divide by zero as a FLT_MAX result. The documentation for the rcp instruction in HLSL seems to make this explicit:
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/rcp---vs

This also differs a bit from IEEE behavior where the sign of the denominator should be carried to the result. I'm mostly guessing at how DX Classic would interpret an IEEE float when loading into a shader that doesn't support IEEE. Following the HLSL 1.1 rules, I'm setting the value to FLT_MAX when the denominator is zero. If anyone has insights, that would be appreciated.

I haven't tested this change on Windows, but it fixes the water on Mac/Metal.

This likely is not the only divide by zero/shader issue lurking. But this one was breaking Vee Tsah on Mac. Vee Tsah also won't run under VMWare with a Windows guest, and I'm guessing this might have something to do with that. This would be tricky nuance for a translation layer to deal with.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plSurface/plShader.cpp
+++ b/Sources/Plasma/PubUtilLib/plSurface/plShader.cpp
@@ -139,17 +139,17 @@ void plShader::SetColor(size_t i, const hsColorRGBA& col)
 void plShader::SetVector(size_t i, const hsScalarTriple& vec)
 {
     /* Doesn't touch .fW */
-    fConsts[i].fX = vec.fX;
-    fConsts[i].fY = vec.fY;
-    fConsts[i].fZ = vec.fZ;
+    fConsts[i].fX = std::clamp(vec.fX, -FLT_MAX, FLT_MAX);
+    fConsts[i].fY = std::clamp(vec.fY, -FLT_MAX, FLT_MAX);
+    fConsts[i].fZ = std::clamp(vec.fZ, -FLT_MAX, FLT_MAX);
 }
 
 void plShader::SetVector(size_t i, float x, float y, float z, float w)
 {
-    fConsts[i].x = x;
-    fConsts[i].y = y;
-    fConsts[i].z = z;
-    fConsts[i].w = w;
+    fConsts[i].x = std::clamp(x, -FLT_MAX, FLT_MAX);
+    fConsts[i].y = std::clamp(y, -FLT_MAX, FLT_MAX);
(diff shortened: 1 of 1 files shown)


colincornaby:
I'm still seeing a few other issues so there likely will be another commit to this pull request.

colincornaby:
FLT_MAX is causing it's own issues. If something in Plasma multiples the value, the float can overflow - resulting in another INF that gets into the shader and mucks everything up.

I could step back and change the shader to handle INF values instead of guarding against them. This would probably require me to clamp values in the shader between min and max.

colincornaby:
Ok, I've changed how this is done a bit. The WaveSet7 algorithm is left alone - but I'm sanitizing all the float input on the way into the shaders and clamp it into the correct range. This should allow non-HLSL 1.1 environments to better inherit the intended behavior.

Within the shader, FLT_MAX is still dangerous and risks overflow. But that should be true in HLSL 1.1 as well - and this seems to kick the Metal shaders back into the right execution flow.

Hoikas:
This looks good to me. On Windows, would I need to simply observe the water in Veelay to verify that there are no regressions?

colincornaby:
Yeah - on Windows everything should be unchanged. I don't know if there will be a difference between INF and FLT_MAX when the values are passed into the shader - but in theory even if there is a difference it should be unnoticeable. I haven't been able to get PIX attached in that age so I don't know what exactly the shader is doing.

I think I'll also compile under Windows here and give it a shot under VMWare. I'm curious if this changes fixes VMWare's DirectX -> Metal translation. But I wouldn't consider that a must have for this change.

colincornaby:
This changes does seem to fix Vee Tsah in VMWare as well. Before the render would just die mid frame and I'd never get complete output out. I haven't gone through with PIX, but it looks acceptable to me.

Screen Shot 2022-05-01 at 2 38 19 PM

Hoikas (approved):

dpogue (approved):
This seems totally reasonable to me

PR #1140, opened 2022-05-01
Ensure emotes and afk message are shown in receiver's localization

ZarothYe:
Previously, these messages were localized by the sender and sent via net propagation. By Adam's request for a return to UU capabilities, these messages now send one or more LOC key strings along with a flag indicating for the receiver to process the message localization themselves.

As a side effect of my digging and the use of the new chat flag, I have also fixed at least one cause of AFK response spam.

Needs https://github.com/H-uru/moul-assets/pull/131.

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -997,21 +997,24 @@ class xKI(ptModifier):
             if cFlags.broadcast and cFlags.channel != self.chatMgr.privateChatChannel:
                 return
 
-            # Is the message from an ignored plaer?
+            # Is the message from an ignored player?
             vault = ptVault()
             ignores = vault.getIgnoreListFolder()
             if ignores is not None and ignores.playerlistHasPlayer(player.getPlayerID()):
                 return
+
+            if cFlags.lockey:
+                keys = message.split(":")
+                message = PtGetLocalizedString(keys[0], [player.getPlayerName(), PtGetLocalizedString(keys[1] if len(keys) > 1 else "KI.EmoteStrings.Their")])
 
             # Display the message if it passed all the above checks.
             self.chatMgr.AddChatLine(player, message, cFlags, forceKI=not self.sawTheKIAtLeastOnce)
 
             # If they are AFK and the message was directly to them, send back their state to sender.
             try:
-                if self.KIDisabled or PtGetLocalAvatar().avatar.getCurrentMode() == PtBrainModes.kAFK and cFlags.private and not cFlags.toSelf:
-                    myself = PtGetLocalPlayer()
-                    AFKSelf = ptPlayer(myself.getPlayerName() + PtGetLocalizedString("KI.Chat.AFK"), myself.getPlayerID())
-                    PtSendRTChat(AFKSelf, [player], " ", cFlags.flags)
+                if (self.KIDisabled or PtGetLocalAvatar().avatar.getCurrentMode() == PtBrainModes.kAFK) and cFlags.private and not cFlags.lockey:
+                    cFlags.lockey = True
+                    PtSendRTChat(PtGetLocalPlayer(), [player], "KI.Chat.AFK", cFlags.flags)
             except NameError:
                 pass
 
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1004,8 +1004,8 @@ class xKI(ptModifier):
                 return
 
             if cFlags.lockey:
-                keys = message.split(":")
-                message = PtGetLocalizedString(keys[0], [player.getPlayerName(), PtGetLocalizedString(keys[1] if len(keys) > 1 else "KI.EmoteStrings.Their")])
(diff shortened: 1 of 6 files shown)


Scripts/Python/xLocTools.py
```diff
@@ -42,6 +42,7 @@
  *==LICENSE==* """
 from Plasma import *
 from PlasmaVaultConstants import *
+from PlasmaTypes import *
```
Hoikas, line 45: Please alphabetize your imports 😉

ZarothYe: Whoops, sorry. Done.

Scripts/Python/ki/xKIChat.py
```diff
@@ -664,16 +664,26 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
         self.ResetFadeState()
 
     ## Display a status message to the player (or players if net-propagated).
-    def DisplayStatusMessage(self, statusMessage, netPropagate=0):
+    def DisplayStatusMessage(self, message, netPropagate=0, isMessageLocKey=False):
```
Hoikas, line 667: I think it would be better to, instead of having an explicit flag `isMessageLocKey` to instead allow `message` to be a [`NamedTuple`](https://docs.python.org/3.8/library/typing.html#typing.NamedTuple), eg:
```python
class LocKey(NamedTuple):
    message: str
    pronoun: str = "KI.EmoteStrings.Their"
```

You could then use `isinstance(message, LocKey)` to implicitly decide if the message is a loc key message.

Scripts/Python/ki/xKIChat.py
```diff
@@ -664,16 +664,26 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
...
             plyrList = self.GetPlayersInChatDistance()
             if len(plyrList) > 0:
-                PtSendRTChat(PtGetLocalPlayer(), plyrList, statusMessage, cFlags.flags)
-        self.AddChatLine(None, statusMessage, cFlags)
+                PtSendRTChat(localPlayer, plyrList, message, cFlags.flags)
+
+        # the message is just a localization key and needs processing before display
+        if isMessageLocKey:
+            keys = message.split(":")
+            message = PtGetLocalizedString(keys[0], [localPlayer.getPlayerName(), PtGetLocalizedString(keys[1] if len(keys) > 1 else "KI.EmoteStrings.Their")])
```
Hoikas, line 684: If you use my `NamedTuple` from above, you can remove the inline if statement.

Scripts/Python/ki/__init__.py
```diff
@@ -997,21 +997,24 @@ def OnRTChat(self, player, message, flags):
...
-            # Is the message from an ignored plaer?
+            # Is the message from an ignored player?
             vault = ptVault()
             ignores = vault.getIgnoreListFolder()
             if ignores is not None and ignores.playerlistHasPlayer(player.getPlayerID()):
                 return
 
+            if cFlags.lockey:
+                keys = message.split(":")
+                message = PtGetLocalizedString(keys[0], [player.getPlayerName(), PtGetLocalizedString(keys[1] if len(keys) > 1 else "KI.EmoteStrings.Their")])
```
Hoikas, line 1008: If you had `LocKey` available here, this could be simplified to:
```suggestion
            if cFlags.lockey:
                keys = LocKey(*message.split(":"))
                message = PtGetLocalizedString(keys.message [player.getPlayerName(), PtGetLocalizedString(keys.pronoun)])
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -940,18 +965,9 @@ def __call__(self, message):
...
-                    hisHer = PtGetLocalizedString("KI.EmoteStrings.His")
-                    if gender == kFemaleClothingGroup:
-                        hisHer = PtGetLocalizedString("KI.EmoteStrings.Her")
-                    statusMsg = PtGetLocalizedString(emote[1], [PtGetLocalPlayer().getPlayerName(), hisHer])
-                else:
-                    statusMsg = PtGetLocalizedString(emote[1], [PtGetLocalPlayer().getPlayerName()])
-                self.chatMgr.DisplayStatusMessage(statusMsg, 1)
+
+                pronounKey = xLocTools.GetLocalAvatarPossessivePronounLocKey()
+                self.chatMgr.DisplayStatusMessage(LocKey(emote[1], pronounKey), netPropagate=1)
```
Hoikas, line 970:
```suggestion
                self.chatMgr.DisplayStatusMessage(LocKey(emote[1], pronounKey), netPropagate=True)
```

Scripts/Python/ki/__init__.py
ZarothYe: I'm confused how this code is here. This is not the latest version from my commit last night...

ZarothYe: I made this go. Thanks for the suggestion.

Hoikas (approved):

PR #1142, opened 2022-05-03
Fixes wrong Ahnonay instance bug

Hazado:
Steps to trigger bug:
Visit your personal Ahnonay Cathedral, leave without going to your Ahnonay instance
Visit someone else's Ahnonay
Your Ahnonay book on the Relto Bookshelf is now tied to the Ahnonay you visited instead of your personal Ahnonay

Basically the previous code had no way of checking that you were in the correct Ahnonay sub age.
If you had visited your Ahnonay Cathedral, the next Ahnonay you visited became the one your Relto bookshelf book was tied to.

So added a check in Ahnonay to verify the Cathedral subage was your personal Cathedral and not someone elses.
This check allows the owner variable to be correctly set.

Also added a check in Ahnonay Cathedral that the GUID that is set for your AhnonayLink is the correct GUID.
If it isn't the correct one, it sets it to the correct one.

### DIFF
--- a/Scripts/Python/Ahnonay.py
+++ b/Scripts/Python/Ahnonay.py
@@ -71,6 +71,7 @@ class Ahnonay(ptResponder):
         pass
 
     def OnServerInitComplete(self):
+        vault = ptVault()
         agevault = ptAgeVault()
         ageinfo = agevault.getAgeInfo()
         guid = ageinfo.getAgeInstanceGuid()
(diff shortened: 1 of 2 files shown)


Scripts/Python/AhnonayCathedral.py
```diff
@@ -60,32 +60,47 @@ def OnFirstUpdate(self):
...
+                            elif chron and chron.getName() == "AhnonayOwner":
+                                owner = chron
+                        break
+
+            if owner == None:
+                PtDebugPrint("I own this Cathedral, but I haven't set myself as Ahnonay owner yet.")
+                newNode = ptVaultChronicleNode(0)
+                newNode.chronicleSetName("AhnonayOwner")
+                newNode.chronicleSetValue(str(PtGetClientIDFromAvatarKey(PtGetLocalAvatar().getKey())))
+                ageDataFolder.addNode(newNode)
```
Hoikas, line 103: Just so I understand what's going on here... All you did was move the `vault.amOwnerOfCurrentAge()` check up to the very top level, correct?

Scripts/Python/Ahnonay.py
```diff
@@ -81,10 +82,26 @@ def OnServerInitComplete(self):
...
+        folder = vault.getAgesIOwnFolder()
+        contents = folder.getChildNodeRefList()
+        for content in contents:
+            link = content.getChild()
+            link = link.upcastToAgeLinkNode()
+            if link:
+                info = link.getAgeInfo()
+                if info.getAgeFilename() == "AhnonayCathedral":
+                    personalCathedralGuid = info.getAgeInstanceGuid()
+                    break
```
Hoikas, line 100:
```suggestion
        cathedralInfoTemplate = ptVaultAgeInfoNode(0)
        cathedralInfoTemplate.setAgeFilename("AhnonayCathedral")
        if cathedralInfo := folder.findNode(cathedralInfoTemplate):
            personalCathedralGuid = cathedralInfo.upcastToAgeInfoNode().getAgeInstanceGuid()
        else:
            personalCathedralGuid = None
```

Yeah, Cyan's python scripters really didn't know how to write good code for their own engine. Sad.

Scripts/Python/AhnonayCathedral.py
Hazado: I moved it up and added a check for the AhnonayLink chronicle, Line 69 and Line 90 are the relevant starts to the code for that check.

Scripts/Python/Ahnonay.py
Hoikas: Wait, don't use this suggestion. Unfortunately, `ptVaultNode.findNode()` won't search down the tree the same way the C++ version will. I'm going to leave this suggestion in, though, for future reference.

Hoikas: I've updated this suggestion to function properly when #1143 is merged.

Hazado: Changes made.

PR #1143, opened 2022-05-03
Allow `ptVaultNode.findNode()` to search deeper than one level.

Hoikas:
#1142 revealed inferior AgeInfo node finding code is required due to `ptVaultNode.findNode()` not allowing searches deeper than one level. This allows them.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
@@ -181,6 +181,10 @@ ST::string pyVaultAgeInfoNode::GetAgeFilename() const
 
 void pyVaultAgeInfoNode::SetAgeFilename(const ST::string& v)
 {
+    if (fNode) {
+        VaultAgeInfoNode access(fNode);
+        access.SetAgeFilename(v);
+    }
 }
 
 ST::string pyVaultAgeInfoNode::GetAgeInstanceName() const
@@ -194,6 +198,10 @@ ST::string pyVaultAgeInfoNode::GetAgeInstanceName() const
 
 void pyVaultAgeInfoNode::SetAgeInstanceName(const ST::string& v)
 {
+    if (fNode) {
+        VaultAgeInfoNode access(fNode);
+        access.SetAgeInstanceName(v);
+    }
 }
 
 ST::string pyVaultAgeInfoNode::GetAgeUserDefinedName() const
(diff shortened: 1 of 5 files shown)


Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
```diff
@@ -181,6 +181,10 @@ ST::string pyVaultAgeInfoNode::GetAgeFilename() const
 
 void pyVaultAgeInfoNode::SetAgeFilename(const ST::string& v)
 {
+    if (fNode) {
+        VaultAgeInfoNode access(fNode);
+        access.SetAgeFilename(v);
+    }
```
ZarothYe, line 187: Are any of the changes in this file relevant to the `findNode` search functionality, or are they just "while I've got this open, lemme just..." changes?

Sources/Plasma/FeatureLib/pfPython/pyVaultNode.cpp
```diff
@@ -580,13 +580,13 @@ PyObject * pyVaultNode::GetNode2( uint32_t nodeID ) const
     PYTHON_RETURN_NONE;
 }
 
-PyObject* pyVaultNode::FindNode( pyVaultNode * templateNode )
+PyObject* pyVaultNode::FindNode( pyVaultNode * templateNode, unsigned int maxDepth )
```
ZarothYe, line 583: Is there a reason this doesn't also use `unsigned int maxDepth = 1` like the header file does? I'm not sure what the rules or convention around optional parameters in C++ are, but in C# I think I would normally write it in both places, e.g., in a method declaration in an interface and in the method definition in the class implementing the interface.

Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
Hoikas: Per the second commit message:
> Add setters to ptVaultAgeInfoNode.
>
> These are critically important to supporting the case of searching for arbitrary Ages in the player vault.

Sources/Plasma/FeatureLib/pfPython/pyVaultNode.cpp
Hoikas: In C and C++, the default argument should only be listed in the declaration (the header file with the class definition in this case). Including it in the method definition is actually an error on some compilers (eg gcc).

Sources/Plasma/FeatureLib/pfPython/pyVaultAgeInfoNode.cpp
ZarothYe: Oh... I didn't notice the little `...` button that indicated the commit messages had more information. I apologize. Everything else seemed to make sense to me!

ZarothYe (approved):

PR #1150, opened 2022-05-13
Fix issue with TPOTS code and Relto start

Hazado:
If you did the relto start, you would be prevented from entering the TPOTS imager code until you solved the cleft age.

Before the cleft start, you could do both Ahnonay and Er'cana, acquire the TPOTS imager code to solve that puzzle before you ever solved the cleft.

This fixes that issue.

It also now prevents you from brute-force solving the first journey code.

### DIFF
--- a/Scripts/Python/clftImager.py
+++ b/Scripts/Python/clftImager.py
@@ -441,7 +441,7 @@ class clftImager(ptResponder):
                 PlayFull = 1
                 PlayFinal = 0
                 PtDebugPrint("play full opening speech")
-            elif (IsCleftSolved() and self.TPOTSolved()):
+            elif (StartInRelto() and self.TPOTSolved()):
                 PlayFull = 0
                 PlayFinal = 0
                 PlayTPOT = 1


Deledrius:
"It also now prevents you from brute-force solving the first journey code."
Does this mean that even if you enter the correct code, it won't be accepted?  That's a break from how Myst games have worked from the very beginning.

Hoikas:
> Does this mean that even if you enter the correct code, it won't be accepted? That's a break from how Myst games have worked from the very beginning.

It looks like Uru was previously enforcing this inconsistent position. This changeset just seems to be fixing logic related to what kind of start path was selected.

Hazado:
> >It also now prevents you from brute-force solving the first journey code.
> 
> 
> 
> Does this mean that even if you enter the correct code, it won't be accepted?  That's a break from how Myst games have worked from the very beginning.

it would only apply to cleft start players, but yes if a cleft start player tried to enter the correct code to finish the first journey it wouldnt work

ZarothYe (approved):

Scripts/Python/clftImager.py
```diff
@@ -433,15 +433,15 @@ def OnNotify(self,state,id,events):
             cam.enableFirstPersonOverride()
             #PtEnableForwardMovement()
 
-            if (self.EndgameSolved()):
+            if (StartInRelto() and self.EndgameSolved()):
```
Hoikas, line 436:
```suggestion
            if (self.EndgameSolved()):
```

This is a puzzle design change that I think @Deledrius is taking issue with. Let's split this out into a separate PR so the bug fix portion can be accepted immediately.

Hazado: Ive reverted the change,

Scripts/Python/clftImager.py
```diff
@@ -441,7 +441,7 @@ def OnNotify(self,state,id,events):
                 PlayFull = 1
                 PlayFinal = 0
                 PtDebugPrint("play full opening speech")
-            elif (IsCleftSolved() and self.TPOTSolved()):
+            elif (StartInRelto() and self.TPOTSolved()):
```
Hoikas, line 444: So, if I am reading this correctly, players who start on the cleft path can never see the TPotS Yeesha speech?

ZarothYe: `StartInRelto()` is true once they solve the Cleft (and then start in Relto for subsequent logins)

Hazado: StartInRelto() is false if you have solved the cleft.
This check is here because TPOTSolved() will error due the code not being generated until you make it to relto.

Hoikas (approved):

PR #1152, opened 2022-05-13
Fix my dumb so /textcolor 999 won't try to convert None to int and die

ZarothYe:
*sigh* I missed a test case. It exploded. Reported by Calum on OU discord.

Previously, `/textcolor <integer>` (integer may be valid hex or invalid altogether) would explode because it would try to go through the (int, int, int) path and cast the `None` value of `arg2` as an int. So... let's check if `arg2` and `arg3` are None before we try to process things as an int or float triplet.


Hoikas:
I think it would be simpler to change the `except ValueError:` lines to `except (TypeError, ValueError):`

ZarothYe:
> I think it would be simpler to change the `except ValueError:` lines to `except (TypeError, ValueError):`

Hmm, yes probably simpler. I kind of like that. Do you think it would still be clear enough what cases are being handled?

Hoikas (approved):

PR #1153, opened 2022-05-14
Fix pod portal responder for multiplayer

Hazado:
Pod portals animation and sounds would stick around for any players in the age if the age owner went through the portal or left the age when the animation was active.

This issue is caused by the responder being netforced and controlling the experience for anyone in the age.
When the person controlling it leaves the age, the responder doesn't finish its commands and stalls at whatever step it was left at.

In this case you get left with portal sparkles for anyone that missed the portal.

### DIFF
--- a/Scripts/Python/xPodBahroSymbol.py
+++ b/Scripts/Python/xPodBahroSymbol.py
@@ -102,30 +102,20 @@ class xPodBahroSymbol(ptResponder):
     def OnNotify(self,state,id,events):
         PtDebugPrint("xPodBahroSymbol.OnNotify:  state=%f id=%d events=" % (state,id),events)
 
-        if (id == -1 and events[0][1] == 'beginning'):
-            respBahroSymbol.run(self.key, state="beginning", netPropagate=False)
-            respSFX.run(self.key, state="play", netPropagate=False)
-            
-        if (id == -1 and events[0][1] == 'end'):
-            respBahroSymbol.run(self.key, state="end", netPropagate=False)
-            respSFX.run(self.key, state="stop", netPropagate=False)
-            
-        if (id == -1 and events[0][1] == 'reset'):
-            self.ISetTimers()
-
         if id == respBahroSymbol.id:
             PtAtTimeCallback(self.key, 32, 3)
 
     ###########################
     def OnTimer(self,TimerID):
         PtDebugPrint("xPodBahroSymbol.OnTimer: callback id=%d" % (TimerID))
-        if self.sceneobject.isLocallyOwned():
-            if TimerID == 1:
-                self.SendNote('beginning')
-            elif TimerID == 2:
-                self.SendNote('reset')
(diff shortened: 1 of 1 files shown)


Hoikas:
Since you're changing the portals to be started locally, I don't really understand why the timers need to send out notifications. Why not close the portals locally when the local timer expires?

Hoikas:
I'm not sure that the time drift would be enough to justify the additional complexity of the indirection through notification sending.

Hazado:
Okay removed the new code and the isLocallyOwned() calls

Scripts/Python/xPodBahroSymbol.py
```diff
@@ -108,15 +108,14 @@ def OnNotify(self,state,id,events):
...
-                respBahroSymbol.run(self.key, state="beginning")
-                respSFX.run(self.key, state="play")
-            elif TimerID == 2:
-                self.ISetTimers()
-            elif TimerID == 3:
-                respBahroSymbol.run(self.key, state="end")
-                respSFX.run(self.key, state="stop")
+        if TimerID == 1:
+            respBahroSymbol.run(self.key, state="beginning")
+            respSFX.run(self.key, state="play")
```
Hoikas, line 113: One minor thought after re-reading your message:
> My thinking was this is sort of a multiplayer experience and you want everyone to see the portal at the same time.

To get the best of both worlds, you could retain the ownership test here and netForce these responders, which would cause the portal to appear for everyone at the same time.

```suggestion
        if TimerID == 1 and self.sceneobject.isLocallyOwned():
            respBahroSymbol.run(self.key, state="beginning", netForce=True)
            respSFX.run(self.key, state="play", netForce=True)
```

I'm :+1: on merging this as-is, however, and will leave the decision up to you.

Hoikas (approved):

Scripts/Python/xPodBahroSymbol.py
Hazado: Netforcing the responders was the issue.
The age owner would leave and the responders wouldnt continue.

Hazado:
Tested this code on OU with two others.
I was there 4 hours before, one person 15 minutes before, third was 5 minutes before.
We all saw the portal at the same time.

Scripts/Python/xPodBahroSymbol.py
Hoikas: This suggestion would still let the close responder run on each individual client - the broken behavior was the netforcing of the close case 😉

Hazado: Ahh I see, yeah that would work.
You were correct about the timing not being an issue though.
At least for around 4 hours.

Hazado: Im okay with the current implementation, ingame testing was succesful and the closing issue didnt happen.

PR #1155, opened 2022-05-14
Fix method name collisions in `ptVaultTextNoteNode`.

Hoikas:
These method names collided with the vault node type getters and setters in `ptVaultNode`. In previous experiments with removing the manual upcasting junk, this collision broke the KI and imagers. Fix it now to match the variant found in `ptVaultFolderNode`. In my grepping, the text note variants are currently unused.

### DIFF
--- a/Scripts/Python/plasma/Plasma.py
+++ b/Scripts/Python/plasma/Plasma.py
@@ -9445,16 +9445,20 @@ class ptVaultTextNoteNode(ptVaultNode):
         """Returns ptVaultNodeRef if is a child node, or None"""
         pass
 
+    def getNoteSubType(self):
+        """Returns the subtype of this text note node."""
+        pass
+
+    def getNoteType(self):
+        """Returns the type of text note for this text note node."""
+        pass
+
     def getOwnerNode(self):
         """Returns a ptVaultNode of the owner of this node"""
         pass
 
     def getOwnerNodeID(self):
         """Returns the node ID of the owner of this node"""
-        pass
-
-    def getSubType(self):
-        """Returns the subtype of this text note node."""
         pass
 
     def getText(self):
@@ -9471,10 +9475,6 @@ class ptVaultTextNoteNode(ptVaultNode):
 
     def getTitleW(self):
         """Unicode version of getTitle"""
-        pass
-
-    def getType(self):
-        """Returns the type of text note for this text note node."""
         pass
 
     def hasNode(self,id):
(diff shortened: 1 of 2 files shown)


ZarothYe:
I don't entirely understand the purpose of this, but I believe you that it's something that should be fixed. I can see that these `pyVaultTextNoteNode` methods were possibly overrides of the parent `pyVaultNode` methods, and maybe that's not ideal? But with your changes it now appears that you are duplicating existing methods. If they're not the same, can you help me understand the difference? If they are the same, are we able to eliminate one of the varieties?

`setNoteType` vs `noteSetType`
`getNoteType` vs `noteGetType`
`setNoteSubType` vs `noteSetSubType`
`getNoteSubType` vs `noteGetSubType`

ZarothYe:
GitHub won't let me comment on lines of code that aren't deemed "close enough" to your changed lines, but your `PyErr_SetString` lines will need to reflect the new method names for the `set` methods. They're giving error messages with the old `setType` and `setSubType` method names.

Hoikas:
> I don't entirely understand the purpose of this, but I believe you that it's something that should be fixed. I can see that these `pyVaultTextNoteNode` methods were possibly overrides of the parent `pyVaultNode` methods, and maybe that's not ideal?

Correct. About 10 years ago, I tried to fix the current vault node anti-pattern of:
```python
node = somethingThatReturnsNode()
if node.getType() == PtVaultNodeTypes.kFoo:
    node = node.upcastToFooNode()
    if node:
        # finally the useful code here
```

to actually return an instance of the final node subclass and therefore be the more Pythonic:
```python
node = somethingThatReturnsNode()
if isinstance(node, ptVaultFooNode):
    # finally the useful code here
```

The problem that I ran in to was that Imagers would break because these functions were hiding the same named version on the `ptVaultNode` base. Therefore, this change will enable better code on down the line.

> But with your changes it now appears that you are duplicating existing methods. If they're not the same, can you help me understand the difference? If they are the same, are we able to eliminate one of the varieties?

At some point between TPotS and MOUL, the vault node subclass getters and setters were duplicated, and the variants following the general pattern of `noteGetText()` were declared "LEGACY". If you look at the other node types, you'll see the same problem. I wanted this pull request to be very small, fixing only the problem of base class hiding. De-duplicating is a larger task that will require checking lots of function calls, so I'd prefer to handle that in another PR.

dpogue (approved):

PR #1156, opened 2022-05-15
Improvements to plMoviePlayer

colincornaby:
- Freeing frames after we’re done with them to keep memory use flatter
- Caching the last frame so if the surface needs to be recreated we can populate it with image data

The first change frees frames after we don't need them. I don't think there is a leak here as everything was freed at the end of the movie. But memory use previously climbed during playback.

The second change is a little more nitpicky. On Mac, if the client changed resolutions during playback (or swapped between windowed/fullscreen) display would be blank momentarily. This is because the movie player only draws when it has a new frame. By retaining the last frame, the movie player can also immediately draw when a new surface is created in response to a reconfiguration.

I'm guessing Windows theoretically has similar issues but has tighter control over when a user can change modes. Haven't decided what to do with the Mac client yet, but trying to keep at least full screen/windowed transitions working nicely and at all times.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
@@ -186,7 +186,8 @@ plMoviePlayer::plMoviePlayer()
       fLastFrameTime(),
       fPosition(),
       fPlaying(),
-      fPaused()
+      fPaused(),
+      fLastImg()
 {
     fScale.Set(1.0f, 1.0f);
 }
@@ -327,13 +328,21 @@ void plMoviePlayer::IProcessVideoFrame(const std::vector<blkbuf_t>& frames)
     }
 
     if (img) {
+        if (fLastImg) {
+            vpx_img_free(fLastImg);
+        }
+        fLastImg = img;
+    }
+    
+    // We need to refill the buffer if there is a new image or a new buffer
+    if (img || !fTexture->GetDeviceRef()) {
         // According to VideoLAN[1], I420 is the most common image format in videos. I am inclined to believe this as our
         // attemps to convert the common Uru videos use I420 image data. So, as a shortcut, we will only implement that format.
         // If for some reason we need other formats, please, be my guest!
         // [1] = http://wiki.videolan.org/YUV#YUV_4:2:0_.28I420.2FJ420.2FYV12.29
-        switch (img->fmt) {
+        switch (fLastImg->fmt) {
         case VPX_IMG_FMT_I420:
-            plPlanarImage::Yuv420ToRgba(img->d_w, img->d_h, img->stride, img->planes, reinterpret_cast<uint8_t*>(fTexture->GetImage()));
+            plPlanarImage::Yuv420ToRgba(fLastImg->d_w, fLastImg->d_h, fLastImg->stride, fLastImg->planes, reinterpret_cast<uint8_t*>(fTexture->GetImage()));
             break;
 
         DEFAULT_FATAL("image format");
@@ -460,6 +469,9 @@ bool plMoviePlayer::Stop()
         fAudioSound->Stop();
     if (fPlate)
         fPlate->SetVisible(false);
+    if (fLastImg) {
+        vpx_img_free(fLastImg);
+    }
 
     for (auto cb : fCallbacks)
         cb->Send();
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
@@ -471,6 +471,7 @@ bool plMoviePlayer::Stop()
         fPlate->SetVisible(false);
     if (fLastImg) {
         vpx_img_free(fLastImg);
+        fLastImg = nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -55,6 +55,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 class plMessage;
 
+#ifdef USE_WEBM
+#   include <vpx/vpx_decoder.h>
+#endif
+
```
Hoikas, line 61: Can we use a forward declaration instead of exposing libvpx to consumers of `plMoviePlayer`?

colincornaby: I tried but vpx_image_t is a struct, which AFAIK can't be forward declared. I could anonymize it as a void * though.

Hoikas: `typedef struct vpx_image vpx_image_t;` is probably what you're looking for.

colincornaby: That worked!

Hoikas (dismissed):

Hoikas:
This PR appears to introduce a crash when playing the intro video on Windows.

Hoikas:
For reference:

```
UruLive.1.918 - 883f2a6 - Internal.Release
OS: Windows 10 Professional (Build 19044)
CPU: Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
RAM: 32768 MiB

Stack Trace:
Level 00: plClient!plMoviePlayer::IProcessVideoFrame (File: plMoviePlayer.cpp:345)
Level 01: plClient!plMoviePlayer::NextFrame (File: plMoviePlayer.cpp:441)
Level 02: plClient!plClient::BeginGame (File: plClient.cpp:1425)
Level 03: plClient!WinMain (File: winmain.cpp:1259)
Level 04: plClient!__scrt_common_main_seh (File: exe_common.inl:288)
Level 05: KERNEL32!BaseThreadInitThunk
Level 06: ntdll!RtlGetAppContainerNamedObjectPath
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -460,6 +469,9 @@ bool plMoviePlayer::Stop()
         fAudioSound->Stop();
     if (fPlate)
         fPlate->SetVisible(false);
+    if (fLastImg) {
+        vpx_img_free(fLastImg);
```
dpogue, line 473: Do we need to assign `nullptr` to `fLastImg` here to prevent it from potentially getting double freed during the frame loop?

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -327,13 +328,21 @@ void plMoviePlayer::IProcessVideoFrame(const std::vector<blkbuf_t>& frames)
...
 
     if (img) {
+        if (fLastImg) {
+            vpx_img_free(fLastImg);
+        }
+        fLastImg = img;
+    }
+    
+    // We need to refill the buffer if there is a new image or a new buffer
+    if (img || !fTexture->GetDeviceRef()) {
```
dpogue, line 338: Should this use `fLastImg` instead of `img`? In theory it shouldn't matter?

colincornaby: It needs to use img to know if a new frame was actually loaded. fLastImg would be populated either way.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
colincornaby: I think so. Good catch! Wondering is this is the source of the crash on Windows but the project is still configuring.

colincornaby: Actually this probably isn't the Windows crash. That happens in normal frame playback.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -335,7 +335,7 @@ void plMoviePlayer::IProcessVideoFrame(const std::vector<blkbuf_t>& frames)
     }
     
     // We need to refill the buffer if there is a new image or a new buffer
-    if (img || !fTexture->GetDeviceRef()) {
+    if (img || (!fTexture->GetDeviceRef() && fLastImg)) {
```
colincornaby, line 338: The crash on Windows was here. Windows may enter the draw loop before a frame is decompressed (not sure why Mac doesn't.) That means we need to make sure fLastImage isn't null before we try to use it.

@Hoikas - give this a try. I'm away from my home setup for the next few weeks, so I'm just getting my Windows laptop stood up. I'm not set up to get all the way into an age, but the intro movie does now complete successfully.

Hoikas (approved):
Crash is now fixed.

PR #1159, opened 2022-05-16
Code for static imager image to show in imager slides

Hazado:
This would be used for showing a puzzle image or any static image for imagers at a set interval.

It does require an additional imager object setup to function properly.
You create another imager object, and set the final texture to use the static image you want.
When the timer for the static image is up, it will hide the normal imager object and show the static imager object instead.
That imager goes for the standard image time, then resumes the standard imager images.

### DIFF
--- a/Scripts/Python/xSimpleImager.py
+++ b/Scripts/Python/xSimpleImager.py
@@ -339,11 +339,11 @@ class xSimpleImager(ptModifier):
                             self.IShowCurrentContent()
                     elif event[1][:9] == "Uploaded=":
                         newID = int(event[1][9:])
-                        if newID == CurrentDisplayedElementID:
-                            if self.clueHandler:
-                                # force the clue imager to turn off so the new upload is visible
-                                self.clueHandler.UpdateClueState(forceOff=True)
-                            self.IShowCurrentContent()
+                        if self.clueHandler:
+                            # force the clue imager to turn off so the new upload is visible
+                            self.clueHandler.UpdateClueState(forceOff=True)
+                        CurrentDisplayedElementID = newID
+                        self.IShowCurrentContent()
                     elif event[1][:7] == "Upload=":
                         deviceName = event[1][7:]
                         nodeId = int(event[3])
--- a/Scripts/Python/xSimpleImager.py
+++ b/Scripts/Python/xSimpleImager.py
@@ -54,8 +54,8 @@ from PlasmaTypes import *
 from PlasmaKITypes import *
 from PlasmaVaultConstants import *
 
-import random
 import xCensor
+from xSimpleImagerClueHandler import *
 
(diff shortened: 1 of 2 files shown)


Scripts/Python/xSimpleImager.py
```diff
@@ -73,6 +73,8 @@
 ImagerButtonResp = ptAttribResponder(8,"start or stop the button animation",['buttonOn','buttonOff'])
 ImagerInboxVariable = ptAttribString(9,"Inbox SDL variable (optional)")
 ImagerPelletUpload = ptAttribBoolean(10, "Pellet Score Imager?", 0)
+ImagerClueObject = ptAttribSceneobject(11, "Imager Object (for puzzle clue)")
+ImagerClueTime = ptAttribInt(12, "Number of seconds until Clue Image Shows",default=870)
```
MareinK, line 77: Is 'Shows' capitalization intentional?

Hazado: fixed

Hazado:
Honestly, I would love for this code to use the existing imager object instead of having to create a new object.
Im just not sure how to accomplish that.

Hoikas:
I'm thinking that it might make more sense to split this functionality out into a second script so as to keep the simple imager simple. What you could do is, in this new script, supply the imager plane object, imager name, and clue stuff. When the clue timer expires, hide the imager plane in the new script and show your clue(s). You'll probably want to hide the clue and restore the imager if someone uploads something while the clue is active.

Scripts/Python/xSimpleImagerClueHandler.py
```diff
@@ -0,0 +1,109 @@
...
+
+            # set alarm to show the clue again if it was on before we turned it off (versus forcing while already off)
+            # otherwise multiple alarms will be active at the same time
+            if self.isShowing:
+                self.shouldShow = False
+                self.SetClueAlarm()
+
+            self.isShowing = False
+        elif self.shouldShow:
+            # only the owner of the imager changes the images
```
Hoikas, line 94: Be sure to net force the changes. Otherwise, this action will only happen on the owner's local client.

ZarothYe: Sorry, I missed this comment from a couple weeks ago. Which changes are you referring to? The enable/disable/suppress things below?

Hoikas: Yes and those found above. I can't remember if those changes are already net-forced from the xSI end of things, so please check.

ZarothYe: This should be resolved when @Hazado pushes to this PR with my updates.

Hoikas (approved):

PR #1161, opened 2022-05-17
Blue Flowers relto Page

Hazado:
Assets - https://github.com/H-uru/moul-assets/pull/145

Scripts needed for new relto page

Cleft SDL has new variable - clftYeeshaPage28Vis

### DIFF
--- a/Scripts/Python/xLinkingBookDefs.py
+++ b/Scripts/Python/xLinkingBookDefs.py
@@ -99,11 +99,14 @@ xYeeshaPage23 = ( "YeeshaPage23", '<pb><img src="xYeeshaPageAlphaSketchErcaPlant
 xYeeshaPage24 = ( "YeeshaPage24", '<pb><img src="xYeeshaPageAlphaSketchStorm*1#0.hsm" align=center check=00ff18,00800c,%d link=25>' )
 xYeeshaPage25 = ( "YeeshaPage25", '<pb><img src="xYeeshaPageAlphaSketch14*1#0.hsm" align=center check=00ff18,00800c,%d link=26>' )
 xYeeshaPage26 = ( "YeeshaPage26", '<pb><img src="xYeeshaPageVeelay7.dds" align=center check=00ff18,00800c,%d link=27>' )
+xYeeshaPage27 = ( "YeeshaPage27", '<pb><img src="xYeeshaPageOcean.dds" align=center check=00ff18,00800c,%d link=28>' )
+xYeeshaPage28 = ( "YeeshaPage28", '<pb><img src="xYeeshaPageBlueFlower.dds" align=center check=00ff18,00800c,%d link=29>' )
 
 
 xYeeshaPages = [ xYeeshaPage1, xYeeshaPage2, xYeeshaPage3, xYeeshaPage4, xYeeshaPage5, xYeeshaPage6, xYeeshaPage7,\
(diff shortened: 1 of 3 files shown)


Hoikas:
It might be worth separating the cleft and Relto parts of these changes, that way the Relto parts can be reviewed and merged more quickly and independently of the Cleft parts, which are blocking on a corresponding MOULa update due to SDL changes.

ZarothYe:
I could be wrong, but it looks like this PR is for Blue Flowers **and** the Ocean page? Maybe those need to be split out because Ocean isn't ready yet.

Scripts/Python/psnlYeeshaPageChanges.py
```diff
@@ -237,6 +238,9 @@ def EnableDisable(self, val):
 
             respAudioStop.run(self.key,avatar=None,fastforward=0)
             respDisable.run(self.key,avatar=None,fastforward=1)
+            if len(bushDistrib.value) > 0:
```
Hoikas, line 241:
```suggestion
            if bushDistrib.value:
```

Scripts/Python/psnlYeeshaPageChanges.py
```diff
@@ -237,6 +238,9 @@ def EnableDisable(self, val):
...
+                for x in bushDistrib.value:
+                    x.setVisible(0)
```
Hoikas, line 243:
```suggestion
                    x.setVisible(False)
```

Scripts/Python/clftYeeshaPage08.py
```diff
@@ -147,32 +150,26 @@ def OnGUINotify(self,id,control,event):
...
             else:
                 PtDebugPrint ("DEBUG: clftYeeshaPage08.py: Yeesha Page #8 is new to you.")       
-                PtDebugPrint ("DEBUG: clftYeeshaPage08.py: Trying to update the value of the SDL variable %s to 1" % ("YeeshaPage8"))
+                PtDebugPrint ("DEBUG: clftYeeshaPage08.py: Trying to update the value of the SDL variable %s to 4" % ("YeeshaPage8"))
                 vault = ptVault()  
                 psnlSDL = vault.getPsnlAgeSDL()
                 if psnlSDL:
                     YeeshaPageVar = psnlSDL.findVar("YeeshaPage8")    
-                    YeeshaPageVar.setInt(1)
+                    YeeshaPageVar.setInt(4)
```
ZarothYe, line 172: It's very unclear to me what this code is doing. It does not appear to be related to the page additions, as page 8 is a very old page. Could you please 1) explain what this is for and 2) split these out to a separate PR so they can be tracked as a fix for whatever this is fixing and merged independently of these new pages?

DoobesURU: I believe it has to do with this? #1147 
Not sure why it's repeated in this PR though since the other was merged.

ZarothYe: Ah, that makes sense, thanks for tracking that down, Doobes! @Hazado then I think it would make sense for you to rebase all the new Relto page PRs to make sure they don't include all these unrelated changes, right?

DoobesURU: Heh, that's something I try to make a habit, but I'll admit, I forget sometimes too, particularly for existing branches. 😛

PR #1162, opened 2022-05-18
Add prefix to chat logs, plus to chat headers when /textcolor is enabled

ZarothYe:
Since /textcolor overrides all chat colors (except clickable URLs) we automatically add the chat prefixes to specify what context a message is from for non-broadcast/status messages. Also added to chat logs all the time because those never have colors to indicate where a message came from and things can get confusing. Requires https://github.com/H-uru/moul-assets/pull/147.

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -447,7 +447,7 @@ class xKIChat(object):
                 contextPrefix = PtGetLocalizedString("KI.Chat.SubtitleContextPrefix")
                 if player is not None:
                     # add subtitle speaker's name if it was provided
-                    # add leading space to match indent for broadcast player messages
+                    # add any leading pretext to match broadcast player messages
                     pretext = f"{PtGetLocalizedString('KI.Chat.BroadcastMsgRecvd')}{player}"
                 player = None
 
@@ -509,6 +509,7 @@ class xKIChat(object):
                     # Are we mentioned in the message?
                     elif self._chatMentionRegex.search(message) is not None:
                         hasMention = True
+                        contextPrefix = PtGetLocalizedString("KI.Chat.MentionContextPrefix")
                         PtFlashWindow()
 
             # Is it a ccr broadcast?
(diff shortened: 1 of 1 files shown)


Scripts/Python/ki/xKIChat.py
```diff
@@ -627,8 +641,9 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
                         chatArea.insertStringW(chatMessageFormatted[lastInsert:start], censorLevel=censorLevel)
 
                     lastInsert = end
+                    
                     chatArea.insertColor(mentionColor)
-                    chatArea.insertStringW(mention, censorLevel=censorLevel, urlDetection=False)
+                    chatArea.insertStringW(f"{'*' if self.chatTextColor else ''}{mention}{'*' if self.chatTextColor else ''}", censorLevel=censorLevel, urlDetection=False)
```
Hoikas, line 646: What if the player's name starts and ends with asterisks?

ZarothYe: Now with 100% more asterisks! 🤷 

More seriously, the other options I could think of were to either ignore mentions entirely or to have a prefix of like "[PING]" or whatever... neither of which I liked very much. Putting some kind of noticeable markers (like asterisks) around the name seemed like the most reasonable thing I could do that was similar to the existing functionality that used color difference.

Hoikas: I kind of like the [PING] prefix lol

ZarothYe: I was thinking about adding mention support for messages coming from Buddies and Neighbors. That seemed like an oversight that it doesn't currently search for mentions in those sources. In that case, though, you'd end up doubling up on prefixes "[PING][BUDDIES]", which seems... rather unfortunate.

ZarothYe: I'd love some more perspectives and opinions on the mention/ping topic and the PR as a whole, if possible.

Hoikas (approved):

PR #1170, opened 2022-05-20
Fix so marker list panel's scroll buttons are appropriately displayed or not

ZarothYe:
Marker list panel's scroll buttons won't be shown when there are no markers in the list when editing or playing the game. Also appropriately set the scroll position (and thus recalculate scroll up/down button visibility) when displaying an in-progress game where one or more markers have been found.

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -4948,7 +4948,7 @@ class xKI(ptModifier):
                 mlbMarkerList.unlock()
 
                 # Refresh the scroll position
-                self.BigKIMarkerListScrollVis(True)
+                self.BigKIMarkerListScrollVis(bool(mgr.markers))
                 mlbMarkerList.setScrollPos(self.MFScrollPos)
 
                 mlbMarkerTextTB.hide()
@@ -5030,7 +5030,6 @@ class xKI(ptModifier):
             mtbPlayEnd.show()
             mlbMarkerList.clearAllElements()
             mlbMarkerList.show()
-            self.BigKIMarkerListScrollVis(True)
 
             # Assume that the game is finished, unless an unseen Marker is still left.
             questGameFinished = True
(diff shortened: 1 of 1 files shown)


Scripts/Python/ki/__init__.py
```diff
@@ -4936,10 +4937,11 @@ def BigKIDisplayMarkerGame(self):
                     hSpans = coord.getHSpans()
                     vSpans = coord.getVSpans()
                     mlbMarkerList.addStringW("[{}:{},{},{}] {}".format(FilterAgeName(age), torans, hSpans, vSpans, xCensor.xCensor(desc, self.censorLevel)))
+                    showScroll = True
                 mlbMarkerList.unlock()
 
                 # Refresh the scroll position
-                self.BigKIMarkerListScrollVis(True)
+                self.BigKIMarkerListScrollVis(showScroll)
```
Hoikas, line 4944: This could be simplified with something like:

```suggestion
                self.BigKIMarkerListScrollVis(bool(mgr.marker_total))
```

Scripts/Python/ki/__init__.py
```diff
@@ -5035,8 +5037,13 @@ def BigKIDisplayMarkerGame(self):
                     hSpans = coord.getHSpans()
                     vSpans = coord.getVSpans()
                     mlbMarkerList.addStringW("[{}:{},{},{}] {}".format(FilterAgeName(age), torans, hSpans, vSpans, xCensor.xCensor(desc, self.censorLevel)))
+                    showScroll = True
                 else:
                     questGameFinished = False
+
+            self.BigKIMarkerListScrollVis(showScroll)
```
Hoikas, line 5044: Same idea:

```suggestion
            self.BigKIMarkerListScrollVis(bool(mgr.markers_captured))
```

Hoikas (approved):

PR #1174, opened 2022-05-22
Archive the macOS build output as an artifact

dpogue:
The only thing in here right now is a handful of the tools, but it's worth bundling it up so we can see that it's building properly and actually outputting usable binaries.

Since the homebrew-based macOS builds run in a reasonable timeframe, I've also set it up to run the same matrix of configurations as the other platforms.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,13 +21,13 @@ jobs:
 
     steps:
       - name: Checkout Plasma
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: true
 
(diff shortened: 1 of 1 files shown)


Hoikas (approved):

Hoikas:
This is only tangentially related, but is there a way for us to purchase more concurrent runners?

dpogue:
> This is only tangentially related, but is there a way for us to purchase more concurrent runners?

I don't believe so... we could add our own self-hosted runners, but that requires someone to host them. I'm looking into that as an option for older macOS support for a few projects (including Plasma).

PR #1175, opened 2022-05-24
Sort clothing in closet and fix clothing resetting when clicking on arrows

Hazado:
Clothing and Clothing Categories are sorted alphabetically by the name (not the description)
This mostly sorts everything in a semi-correct order.
I tried sorting by description and the item that clothing category represented was not usually the first thing shown.

Fixed a bug that would cause the clothing to reset when you were clicking the right/left arrows and it would disappear.
This fix avoids the issue by always showing the arrows, and never hiding them.
The bug can still happen if you click on anything inside the "Additional Options" that is not a piece of clothing.
The code currently tells the section to refresh itself whenever you click on it and when you click on something that isn't a piece of clothing it just resets everything in that section.

### DIFF
--- a/Scripts/Python/xAvatarCustomization.py
+++ b/Scripts/Python/xAvatarCustomization.py
@@ -1247,7 +1247,7 @@ class xAvatarCustomization(ptModifier):
                 avatar.avatar.tintClothingItem(matchingItem[0], lastcolor1, 0)
                 avatar.avatar.tintClothingItemLayer(matchingItem[0], lastcolor2, 2, 0)
 
-        PtAtTimeCallback(self.key, 1, 666)
+        PtAtTimeCallback(self.key, 0.1, 666)
 
     def IRemoveWornSet(self):
         setToRemove = ''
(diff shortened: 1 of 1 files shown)


Scripts/Python/xAvatarCustomization.py
```diff
@@ -2561,6 +2557,8 @@ def UpdateListbox(self):
         listbox.show()
         listbox.enable()
         displayItems = []
+        # sort clothing alphabetically
+        self.clothingList = sorted(self.clothingList, key=lambda x: getattr(x, 'name').lower().replace("*",""))
```
Hoikas, line 2561:
```suggestion
        self.clothingList = sorted(self.clothingList, key=lambda x: x.name.lower().replace("*",""))
```

Hazado:
Figured out why the clothing sets would cause the client to freeze for a few seconds.
The timer callback was stacking on itself and causing issues.
Fix pushed :)

Hoikas:
One thing we need to check *before* this is merged... Cyan has a special sorting field in the Clothing PRPs themselves. It's basically only used to to set the defautl T-shirt (🙄), but we'll want to make sure that this doesn't somehow impact that. Lest we get accused again of malfeasance because we ~~accidentally~~dared to change the default shirt option (gasp).

Hazado:
> One thing we need to check _before_ this is merged... Cyan has a special sorting field in the Clothing PRPs themselves. It's basically only used to to set the defautl T-shirt (🙄), but we'll want to make sure that this doesn't somehow impact that. Lest we get accused again of malfeasance because we ~accidentally~dared to change the default shirt option (gasp).

I created a new male and female character and they had the default clothing on.
Changing clothing sets to normal will still set the rest of the clothing set to the default clothing.

Not sure how else to test this...

Hoikas (approved):

PR #1178, opened 2022-05-30
Remove unused (and terrible) NetComm callback stuff.

Hoikas:
It looks like, during the transition from the OG networking code to the new stuff, this stuff was used with our friend PropagateBuffer to send the old plNetTasks to the AuthSrv. This is long defunct now, so remove it.


dpogue (approved):
I miss the old plNetTasks 😞

PR #1179, opened 2022-05-30
Use PEP 590 Vectorcall

Hoikas:
As of Python 3.8, a new calling convention was added to CPython. The new convention is faster, per the the PSF, so we now use it. The major benefit to it on our end is that we don't have to construct intermediate Python tuples holding function call arguments anymore. Instead, we simply pass arrays of `PyObject*`. As part of this, I added scaffolding for more generic C++ -> PyObject* conversion.

There is still lots of room for improvement in the Python API in a large variety of topics, for example:
- There are many places in `plPythonFileMod` where an `ST::string` is being `c_str()`'d and the result being sent to Python.
- All of the glue should probably be updated to use [`METH_FASTCALL`](https://docs.python.org/3.8/c-api/structures.html#METH_FASTCALL) to remove the overhead of parsing argument tuples.

These items are outside the scope of this PR.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
+++ b/Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
@@ -54,102 +54,24 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plProfile.h"
 
 #include "cyPythonInterface.h"
+#include "plPythonConvert.h"
 #include "pyGlueHelpers.h"
 #include "pyObjectRef.h"
 
 plProfile_Extern(PythonUpdate);
 
-namespace plPythonCallable
+namespace plPython
 {
-    template<typename ArgT>
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, ArgT value) = delete;
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, bool value)
+    template<typename... Args>
+    inline pyObjectRef CallObject(const pyObjectRef& callable, Args&&... args)
     {
-        PyTuple_SET_ITEM(tuple, idx, PyBool_FromLong(value ? 1 : 0));
-    }
+        hsAssert(PyCallable_Check(callable.Get()), "Trying to call a non-callable, eh?");
 
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, char value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyUnicode_FromFormat("%c", (int)value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, const char* value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyUnicode_FromString(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, double value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyFloat_FromDouble(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, float value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyFloat_FromDouble(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, int8_t value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyLong_FromLong(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, int16_t value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyLong_FromLong(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, int32_t value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyLong_FromLong(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, PyObject* value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, value);
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, pyObjectRef& value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, value.Release());
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, const ST::string& value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyUnicode_FromSTString(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, uint8_t value)
-    {
-        PyTuple_SET_ITEM(tuple, idx, PyLong_FromSize_t(value));
-    }
-
-    inline void IBuildTupleArg(PyObject* tuple, size_t idx, uint16_t value)
-    {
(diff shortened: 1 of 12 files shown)


Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
```diff
@@ -0,0 +1,186 @@
...
+    {
+        return PyFloat_FromDouble(value);
+    }
+
+    inline PyObject* ConvertFrom(float value)
+    {
+        return PyFloat_FromDouble(value);
+    }
+
+    inline PyObject* ConvertFrom(int8_t value)
```
zrax, line 86: This will cause an issue on systems (read: Solaris) which alias `int8_t` as just `char` instead of `signed char`.  It would probably be better to use the base types directly (`char`, `signed`/`unsigned` `char`, `short`, `int`, `long`, etc), as was done for String Theory

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
```diff
@@ -0,0 +1,186 @@
...
+
+    inline PyObject* ConvertFrom(wchar_t value)
+    {
+        return PyUnicode_FromFormat("%c", (int)value);
+    }
+
+    inline PyObject* ConvertFrom(std::nullptr_t)
+    {
+        Py_INCREF(Py_None);
+        return Py_None;
```
zrax, line 129:
```suggestion
        Py_RETURN_NONE;
```

Hoikas:
The CI failure on 9da3b2c76a383a1ce712f896d589f6a01bc70705 appears to be spurious.

Hoikas:
Rebased on top of latest master.

ZarothYe:
I have a lot of reading to do in order to understand this, but I'll try to review this weekend since I know this is a prerequisite for the marker game restoration work.

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
Hoikas: Just to check, I resolved this correctly in 69b3a4821e84fe833ac12071d9cac842383c1848, right?

zrax: Yep, looks good to me

Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
```diff
@@ -54,102 +55,90 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
 #include "cyPythonInterface.h"
+#include "plPythonConvert.h"
 #include "pyGlueHelpers.h"
 #include "pyObjectRef.h"
 
 plProfile_Extern(PythonUpdate);
 
-namespace plPythonCallable
+namespace plPython
```
ZarothYe, line 64: Why was this namespace changed? The file name is still `plPythonCallable.h`. Is it OK for the filename and namespace/class in the file not to match?

Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
```diff
@@ -54,102 +55,90 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        if constexpr (sizeof...(args) == 1) {
+            // Use Python's built-in vectorcall optimization for one argument.
+            pyObjectRef arg = ConvertFrom(std::forward<Args>(args)...);
+            plProfile_BeginTiming(PythonUpdate);
+            pyObjectRef result = _PyObject_CallOneArg(callable, arg.Get());
+            plProfile_EndTiming(PythonUpdate);
+            return result;
+        } else if constexpr (sizeof...(args) == 0) {
+            plProfile_BeginTiming(PythonUpdate);
+#if PY_VERSION_HEX >= 0x03090000
```
ZarothYe, line 118: I'm not familiar with the hex comparison here. Is this 3.9 or greater, so you can use the built-in versus the provisional vectorcall methods?

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    inline PyObject* ConvertFrom(float value)
+    {
+        return PyFloat_FromDouble(value);
+    }
+
+    inline PyObject* ConvertFrom(signed char value)
+    {
+        return PyLong_FromLong(value);
```
ZarothYe, line 88: I am highly perplexed at signed and unsigned chars being treated as longs rather than like... line#68 `PyUnicode_FromFormat("%c", (int)value)`. This might just be a C++ type quirk that I'm not familiar with. I can't say that I knew chars could be signed or unsigned until now.

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
```diff
@@ -0,0 +1,205 @@
...
+                    plPython::ConvertFrom(
+                        std::forward<_ElementT>(
+                            std::get<_IdxT>(tuple)
+                        )
+                    )
+                );
+                ConvertToTuple<_IdxT + 1, _TupleArgsT...>(result, std::forward<_TupleT>(tuple));
+            }
+        }
+    };
```
ZarothYe, line 188: I think I need some comments for this stuff. I don't expect that I'll understand it fully, but just... something.

Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
Hoikas: Yeah, there's no requirement for namespaces to match filenames (or have a name at all, which is another barrel of monkeys). I changed the name because, previously, all of the machinery I had written was specifically for calling Python callables. Now, there is a batch of general object conversion functions in plPythonConvert.h, so I wanted the namespace name to be more general.

Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
Hoikas: IIRC the zero argument specialization for vectorcall was added in Python 3.9, so I had to add manual handling for 3.8.

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
Hoikas: `signed char` and `unsigned char` are single byte integers, which are distinct from `char`acatars :wink:

Sources/Plasma/FeatureLib/pfPython/plPythonConvert.h
```diff
@@ -164,40 +164,15 @@ namespace plPython
...
-                ConvertToTuple<_IdxT + 1, _TupleArgsT...>(result, std::forward<_TupleT>(tuple));
-            }
-        }
-    };
-
     struct ToTuple_Type {};
     constexpr ToTuple_Type ToTuple;
 
     template<typename... _TupleArgsT>
     inline PyObject* ConvertFrom(ToTuple_Type, _TupleArgsT... args)
```
ZarothYe, line 171:
```suggestion
    inline PyObject* ConvertFrom(ToTuple_Type, _TupleArgsT&&... args)
```

Hoikas:
Rebased to get the fix for homebrew breaking libjpeg detection on macOS.

ZarothYe (approved):

Hoikas:
Merging per previous.

PR #1180, opened 2022-06-01
Fix journal GUI variable access before declaration error for some cases

ZarothYe:
The `journalContents` variable was not being declared and set in all branches of the if/else block, so trying to retrieve a Dynamic text block from the vault where the vault didn't have the item caused a somewhat unhelpful python error rather than a blank journal and a warning message in the python log.

Initialize it to empty string to make sure it always has been declared before it is used.


Hoikas (approved):
There does appear to be a debug print that will fire an error message when this happens. This change just suppresses the traceback.

PR #1190, opened 2022-06-05
Clean up autogenerated Xcode schemes

dpogue:
Instead of cluttering up Xcode with a scheme for every library, let's tell it to only include schemes for the executable targets.

Also add a macOS-specific option to tell it to do universal multi-architecture builds (disabled by default).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,11 @@ if(APPLE)
 
     if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
         add_definitions(-DHS_BUILD_FOR_MACOS)
+
+        option(PLASMA_MAC_UNIVERSAL "Build macOS multi-architecture universal builds?" OFF)
+        if (PLASMA_MAC_UNIVERSAL)
+            set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
+        endif()
     endif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
 endif(APPLE)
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -45,9 +45,15 @@ endif(UNIX)
 
 if(APPLE)
     add_definitions(-DHS_BUILD_FOR_APPLE)
+    set(CMAKE_XCODE_GENERATE_SCHEME FALSE)
```
Hoikas, line 48: This overrides any use setting, which is bad practice in general. Let's set the cache instead.

CMakeLists.txt
```diff
@@ -45,9 +45,15 @@ endif(UNIX)
...
 
     if(CMAKE_SYSTEM_NAME MATCHES "Darwin")
         add_definitions(-DHS_BUILD_FOR_MACOS)
+
+        option(PLASMA_MAC_UNIVERSAL "Build macOS multi-architecture universal builds?" OFF)
+        if (PLASMA_MAC_UNIVERSAL)
+            set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
```
Hoikas, line 55: See above.

CMakeLists.txt
dpogue: So that would be this instead?

`set(CMAKE_XCODE_GENERATE_SCHEME FALSE CACHE BOOL)`

Hoikas: `set(CMAKE_XCODE_GENERATE_SCHEME FALSE CACHE BOOL "")`

CMake gets mad if you leave out the docstring :(

dpogue:
@Hoikas Thanks, updated

Hoikas (approved):

PR #1194, opened 2022-06-13
Add capability to use a sliding solution window for puzzle buttons

ZarothYe:
In response to a report by @EhrenCG on the OU discord and comment that @ametist365 would like to try the sliding window variation. When the puzzle button python file mod is used without providing feedback to the user that the combination is reset on a wrong click and prompting them to start over from the beginning of the sequence, it can seem that entering the right combination will not work correctly. With this new attribute added for puzzles, the script will check a sliding window of the last N entries so that, for example, entering 1,1,2,3 will match a combination of 1,2,3 rather than failing after 1,1 and expecting the user to start the sequence over.

**Original report by Ehren (re: flower button sequence in Elonin):**
Guys, I feel like I've found a major programming oversight and it can be understood pretty well just looking at the Python log, in the red and blue squares I show that push the same sequence of buttons in both, but the one in red registers as all wrong, and the one in blue as all right.

The problem lays in the fact that the first button in the sequence counts as wrong if you hit it a second time (rather then considering that you are starting the sequence over again from the start).

This could lead to confusing situations where someone messes with the buttons goes to check on something, comes back to try entering the code again, and having the code that SHOULD work end up not working...
image

This issue affects simple puzzle in Tiam too, you know the three buttons you have to keep using each time  you visit?  Try pushing the first one two times, then the rest of the code, you now have a situation where you have entered the correct code, but the initial first press canceled it out.  This would not have happened if you had pressed it three times though, just even numbers where you keep canceling the first code button out.

### DIFF
--- a/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
+++ b/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
@@ -60,8 +60,6 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
         self.version = 1
         self.id = 719
 
-        self._attempt = []
-
         # Defer setting up the handlers until PFM is initialized by the engine
         self._NotifyHandlers = {}
         self._HitActId = None
@@ -84,6 +82,8 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
         ageSDL.setNotify(self.key, butsInUseVariableName.value, 0.0)
         ageSDL.setFlags(numCorrectVariableName.value, True, True)
         ageSDL.sendToClients(numCorrectVariableName.value)
+        self.SDL.setDefault("attemptCombo", ())
+        self.SDL.sendToClients("attemptCombo")
 
         if not PtGetPlayerList():
             self._butsInUse = False
@@ -128,6 +128,11 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
         ageSDL.setTagString(attr.value, hint)
         ageSDL.setIndexNow(attr.value, 0, state)
 
+    def _GetScriptSDL(self, name):
+        return self.SDL[name]
+    def _SetScriptSDL(self, name, value):
+        self.SDL[name] = tuple(value)
+
     @property
     def _butsInUse(self):
         return self._GetAgeSDL(butsInUseVariableName)
@@ -151,6 +156,14 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
     @_numCorrect.setter
     def _numCorrect(self, value):
         self._SetAgeSDL(numCorrectVariableName, value)
+
+    @property
+    def _attemptCombo(self):
+        return list(self._GetScriptSDL("attemptCombo"))
+
+    @_attemptCombo.setter
+    def _attemptCombo(self, value):
+        self._SetScriptSDL("attemptCombo", value)
 
     # ======================================================================
 
@@ -202,11 +215,14 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
         if self._solved:
             PtDebugPrint("xAgeSDLBoolActComboSet._TriggerButton():\tYou just unsolved it, moron.", level=kWarningLevel)
             self._solved = False
+            self._numCorrect = 0
+            self._attemptCombo = [actId]
             return
 
         if allowSlidingSolution.value:
-            self._numCorrect = next((len(self._attempt) - i for i in range(len(self._attempt)) if self._IsAttemptCorrectAtIndex(i)), 0)
-            PtDebugPrint(f"xAgeSDLBoolActComboSet._TriggerButton():\t Sliding attempt {self._attempt} has {self._numCorrect} correct.", level=kWarningLevel)
+            attempt = self._attemptCombo
+            self._numCorrect = next((len(attempt) - i for i in range(len(attempt)) if self._IsAttemptCorrectAtIndex(i)), 0)
+            PtDebugPrint(f"xAgeSDLBoolActComboSet._TriggerButton():\t Sliding attempt {attempt} has {self._numCorrect} correct.", level=kWarningLevel)
             self._CheckSolved()
             return
 
@@ -240,9 +256,13 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
                 self._butsInUse = True
 
     def _AddToAttempt(self, actId):
-        self._attempt.append(actId)
-        if len(self._attempt) > len(self._combination):
-            self._attempt = self._attempt[1:]
+        attempt = self._attemptCombo
+        attempt.append(actId)
+        if len(attempt) > len(self._combination):
+            attempt = attempt[1:]
+        PtDebugPrint(f"xAgeSDLBoolActComboSet._AddToAttempt():\t New attempt value: {attempt}", level=kWarningLevel)
+        self._attemptCombo = attempt
 
     def _IsAttemptCorrectAtIndex(self, i):
-        return self._attempt[i:] == self._combination[0:len(self._attempt)-i]
+        attempt = self._attemptCombo
+        return attempt[i:] == self._combination[0:len(attempt)-i]
--- a/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
+++ b/Scripts/Python/xAgeSDLBoolActivatorComboSet.py
@@ -215,12 +215,13 @@ class xAgeSDLBoolActivatorComboSet(ptResponder):
         if self._solved:
             PtDebugPrint("xAgeSDLBoolActComboSet._TriggerButton():\tYou just unsolved it, moron.", level=kWarningLevel)
             self._solved = False
-            self._numCorrect = 0
             self._attemptCombo = [actId]
             return
 
         if allowSlidingSolution.value:
             attempt = self._attemptCombo
+            # find the subset of the attempt array that matches the start of the combination, if it exists.
+            # the length of this matching attempt subset is the current number of correct values, otherwise 0 are correct.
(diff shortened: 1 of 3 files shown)


Deledrius:
It depends on the design (in-game) of the puzzle itself, but sliding window solutions are generally preferable whenever appropriate.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -224,3 +231,18 @@ def _ValidateCombination(self):
...
+            if disableOnSolve.value:
+                self._butsInUse = True
+
+    def _AddToAttempt(self, actId):
+        self._attempt.append(actId)
+        if len(self._attempt) > len(self._combination):
+            self._attempt = self._attempt[1:]
+
+    def _IsAttemptCorrectAtIndex(self, i):
+        return self._attempt[i:] == self._combination[0:len(self._attempt)-i]
```
Hoikas, line 248: Newline at EOF

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -52,13 +52,16 @@
...
 respButtonPush = ptAttribResponder(8, "Resp: Button Push")
+allowSlidingSolution = ptAttribBoolean(9, "Allow solving by sliding solution rather than needing feedback on failure", default=False)
 
 class xAgeSDLBoolActivatorComboSet(ptResponder):
     def __init__(self):
         ptResponder.__init__(self)
         self.version = 1
         self.id = 719
 
+        self._attempt = []
```
Hoikas, line 63: The good thing about this is that it allows Age creators to trivially opt into sliding behaviors without changing anything in their already existing Age using this script. The bad news is that it exposes curious logic in multiplayer. The built in assumption here is that the game master will be present in the Age for every single button press because the window is being cached in Python code rather than persisting on the server. This means that, depending on linking order, two players could perform the same button presses and get different answers for whether or not the puzzle is solved. I think this is arguably worse than the default behavior, which is always consistent in terms of how it accepts and rejects combinations.

I think the best way to solve this would be to change attribute 9 above to an optional SDL variable name in which the combination window is stored. If the window is not specified, fall back to the original behavior. The downside to this is that SDL changes are required, which are always a pain.

dpogue: At risk of opening some sort of can of worms... you can have per-script SDL files rather than per-Age SDL files (see tldnWRCC as an example). Maybe that would be an option here to make it transparent to Age builders?

Hoikas: Perhaps - my request on that would be to not remove the age-wide solved and number correct variables. I especially wanted those visible for Age creators to be able to attach things like xAgeSDLBoolShowHide and xAgeSDLIntRespond. 😄

Hoikas: Related: the addition of a `STATEDESC xAgeSDLBoolActivatorComboSet` would necessitate the application of the `Blocking on MOULa` tag.

ZarothYe: I don't plan to touch anything about the existing solved and numCorrect SDL storage. That seems to be working fine. I would like to try the script SDL solution, but I feel a little out of my dept on how to get started with that.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -194,9 +210,20 @@ def _RespPushComplete(self, events):
         self._HitActId = None
 
     def _TriggerButton(self, actId):
+        self._AddToAttempt(actId)
+
         if self._solved:
             PtDebugPrint("xAgeSDLBoolActComboSet._TriggerButton():\tYou just unsolved it, moron.", level=kWarningLevel)
             self._solved = False
+            self._numCorrect = 0
```
Hoikas, line 218:
```suggestion
```

This is done implicitly in `OnSDLNotify()`.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -224,3 +247,22 @@ def _ValidateCombination(self):
...
+            PtDebugPrint("xAgeSDLBoolActComboSet._CheckSolved():\tWoohoo! You have solved the puzzle!", level=kWarningLevel)
+            self._solved = True
+            if disableOnSolve.value:
+                self._butsInUse = True
+
+    def _AddToAttempt(self, actId):
+        attempt = self._attemptCombo
+        attempt.append(actId)
+        if len(attempt) > len(self._combination):
+            attempt = attempt[1:]
```
Hoikas, line 262:
```suggestion
        attempt = [-len(self._combination):]
```

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -224,3 +247,22 @@ def _ValidateCombination(self):
...
+        PtDebugPrint(f"xAgeSDLBoolActComboSet._AddToAttempt():\t New attempt value: {attempt}", level=kWarningLevel)
+        self._attemptCombo = attempt
+
+    def _IsAttemptCorrectAtIndex(self, i):
+        attempt = self._attemptCombo
+        return attempt[i:] == self._combination[0:len(attempt)-i]
```
Hoikas, line 268:
```suggestion
        return attempt[i:] == self._combination[:-i]
```

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
```diff
@@ -194,9 +210,20 @@ def _RespPushComplete(self, events):
...
         if self._solved:
             PtDebugPrint("xAgeSDLBoolActComboSet._TriggerButton():\tYou just unsolved it, moron.", level=kWarningLevel)
             self._solved = False
+            self._numCorrect = 0
+            self._attemptCombo = [actId]
+            return
+
+        if allowSlidingSolution.value:
+            attempt = self._attemptCombo
+            self._numCorrect = next((len(attempt) - i for i in range(len(attempt)) if self._IsAttemptCorrectAtIndex(i)), 0)
```
Hoikas, line 224: This looks a little weird to me. I think I could use an explanation either here on in the comments about the desired process.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
ZarothYe: I added it because it was also being done down on line 242: https://github.com/H-uru/Plasma/pull/1194/files#diff-675dbaccdec1487a34621a04230d97bd6bae77cd670d6f99d50593ad5c29713aR242

It seemed like it should be consistent, unless there's something different about the context that necessitates one being set explicitly. I'll remove it here, but should I also remove it there as well?

Hoikas: On line 242, the SDL solved is probably already false, so the `OnSDLNotify()` won't be fired (no-op changes don't fire notifications), so the explicit set is needed.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
ZarothYe: I'll try to put a comment in the code, but basically this is looping through partial slices of the attempt array and comparing against the *start* of the combination array (up to the length of the attempt so far). For example, an attempt [3, 2, 1] for a solution of [1, 2, 3, 4, 5] would do the following comparison:
First -- index 0: [3, 2, 1] == [1, 2, 3]
Second -- index 1: [2, 1] == [1, 2]
Third -- index 2: [1] == [1]

and then return 1 (length of 3 minus index value of 2) as the number of correct values in the attempt. If there is no match, it uses the default value of 0 returned from the `next` iterator.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
ZarothYe: I don't think that's the same thing, is it? I think I should be able to omit the 0 as the start of the range, but the end of the range is `len(attempt)-i`, not `len(self._combination)-i`, which I think is what your suggestion would be doing, right?

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
ZarothYe: I had to try this out in the REPL because I wasn't sure what this would do when len(self._combination) was greater than len(attempt) but it looks like it plays nice. Thanks for the suggestion!

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
Hoikas: Ah, sorry, I missed that we were indexing from the end of a different list.

Scripts/Python/xAgeSDLBoolActivatorComboSet.py
ZarothYe: I'm currently stuck and need some assistance from @dpogue or @Hoikas to help. I have re-implemented things using the script SDL, and it works as before, but that SDL value isn't appearing to be propagated or shared between multiple users in the Age. Each person interacting with the puzzle gets their own "attemptCombo" SDL that changes independently, which is no better than the original implementation using a local array in Python without involving SDL.

I don't know how to figure out why this isn't getting shared. Maybe this really will need a new Age SDL variable, which will make this whole thing almost entirely unhelpful for existing Ages.

Hoikas (approved):

PR #1196, opened 2022-06-17
Fix for infinite BK refresh due to addition to Recents with inbox item senders

ZarothYe:
Bug introduced in https://github.com/H-uru/Plasma/pull/1129 by me. Could infinitely recurse when opening the Big KI Incoming folder and it added a player to a full Recents list, triggering an addition and removal, which triggered the Big KI refresh again, which tried to add/remove another player to Recents, etc. With this fix, it will recurse to a maximum of 50 (max Recents list size) refreshes if the Recents folder was empty and never try to add more Recents from Inbox Savers if the Recents list is already full.

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -4314,12 +4314,11 @@ class xKI(ptModifier):
             del self.BKContentList[removeidx]
 
         if self.BKFolderListOrder[self.BKFolderSelected] == xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kInboxFolder):
-            PIKAFolder = ptVault().getPeopleIKnowAboutFolder()
-            if PIKAFolder and kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList()) > 0:
-                PtDebugPrint(f"xKI.BigKIProcessContentList(): {kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList())} remaining recents slots.", level=kWarningLevel)
+            recentsList = ptVault().getPeopleIKnowAboutFolder()
+            if recentsList and (remRecents := kLimits.MaxRecentPlayerListSize - len(recentsList.getChildNodeRefList())) > 0:
+                PtDebugPrint(f"xKI.BigKIProcessContentList(): {remRecents} remaining recents slots.", level=kWarningLevel)
                 # add Inbox senders (sorted by most recent) to Recents folder
                 # but don't overload Recents or refresh triggers infinitely due to Recents node additions/removals
-                recentsList = PIKAFolder.upcastToPlayerInfoListNode()
                 cannotAdd = lambda ref: not (s := xKIHelpers.GetSaverID(ref)) or recentsList.playerlistHasPlayer(s) or not self.chatMgr.AddPlayerToRecents(s)
                 # add only one then stop because addition triggers BK refresh again that will add next one, etc.
                 inboxRef = next(itertools.dropwhile(cannotAdd, self.BKContentList), None)
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -64,6 +64,7 @@ import os    # Used for saving pictures locally.
 import glob  # Used for saving pictures locally.
 import math
 import functools
+import itertools
 
 from xGUILinkHandler import xGUILinkHandler
 import xLocTools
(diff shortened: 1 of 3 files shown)


Scripts/Python/ki/__init__.py
```diff
@@ -4315,6 +4313,18 @@ def BigKIProcessContentList(self, removeInboxStuff=False):
         for removeidx in removeList:
             del self.BKContentList[removeidx]
 
+        if self.BKFolderListOrder[self.BKFolderSelected] == xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kInboxFolder):
+            PIKAFolder = ptVault().getPeopleIKnowAboutFolder()
+            if PIKAFolder and kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList()) > 0:
+                PtDebugPrint(f"xKI.BigKIProcessContentList(): {kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList())} remaining recents slots.", level=kWarningLevel)
+                # add Inbox senders (sorted by most recent) to Recents folder
+                # but don't overload Recents or refresh triggers infinitely due to Recents node additions/removals
+                recentsList = PIKAFolder.upcastToPlayerInfoListNode()
```
Hoikas, line 4322: `PIKA` is already a `ptVaultPlayerInfoListNode`.

Scripts/Python/ki/__init__.py
```diff
@@ -4315,6 +4313,18 @@ def BigKIProcessContentList(self, removeInboxStuff=False):
         for removeidx in removeList:
             del self.BKContentList[removeidx]
 
+        if self.BKFolderListOrder[self.BKFolderSelected] == xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kInboxFolder):
+            PIKAFolder = ptVault().getPeopleIKnowAboutFolder()
+            if PIKAFolder and kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList()) > 0:
+                PtDebugPrint(f"xKI.BigKIProcessContentList(): {kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList())} remaining recents slots.", level=kWarningLevel)
```
Hoikas, line 4319: We should probably only calculate `kLimits.MaxRecentPlayerListSize - len(PIKAFolder.getChildNodeRefList())` once.

Scripts/Python/ki/__init__.py
```diff
@@ -4315,6 +4313,18 @@ def BigKIProcessContentList(self, removeInboxStuff=False):
         for removeidx in removeList:
             del self.BKContentList[removeidx]
 
+        if self.BKFolderListOrder[self.BKFolderSelected] == xLocTools.FolderIDToFolderName(PtVaultStandardNodes.kInboxFolder):
```
Hoikas, line 4316: I'm not looking at the code right now, but I wonder if there is a way to compare integer folder types instead of the folder names?

Scripts/Python/ki/xKIChat.py
```diff
@@ -707,13 +707,14 @@ def DisplayStatusMessage(self, message, netPropagate=0):
     ## Adds a player to the recent players folder.
     def AddPlayerToRecents(self, playerID):
 
-        vault = ptVault()
-        PIKAFolder = vault.getPeopleIKnowAboutFolder()
+        isAdded = False
+        PIKAFolder = ptVault().getPeopleIKnowAboutFolder()
         if PIKAFolder:
             PIKA = PIKAFolder.upcastToPlayerInfoListNode()
```
Hoikas, line 713: I see now where the redundant upcast came from -_-

Scripts/Python/ki/xKIChat.py
```diff
@@ -707,13 +707,14 @@ def DisplayStatusMessage(self, message, netPropagate=0):
     ## Adds a player to the recent players folder.
     def AddPlayerToRecents(self, playerID):
 
-        vault = ptVault()
-        PIKAFolder = vault.getPeopleIKnowAboutFolder()
+        isAdded = False
```
Hoikas, line 710: IMO the code would be cleaner if the add-path returned `True` directly and the function ended with `return False .

Scripts/Python/ki/xKIHelpers.py
```diff
@@ -227,6 +227,11 @@ def CMPNodeDate(nodeA, nodeB):
             return 1
     return 0
 
+def GetSaverID(ref: Optional[ptVaultNodeRef]) -> int:
+    if ref and ref.getSaver():
+        return ref.getSaverID()
+    return 0
```
Hoikas, line 233:
```suggestion
    return ref.getSaverID() if ref else 0
```

Scripts/Python/ki/xKIChat.py
ZarothYe: Yeah, sorry. I was taking from what I had seen used before and I didn't know to be skeptical... :(

Scripts/Python/ki/__init__.py
ZarothYe: I'm not sure about that. Perhaps? I was doing what was consistent with other comparisons in the same method and throughout the KI code.

Hoikas (approved):

PR #1199, opened 2022-06-19
Endianness test & detection fixes

dpogue:
Somehow, `LITTLE_ENDIAN` was ending up defined (probably from another header file somewhere) on a big endian machine and that causes all sort of issues with trying to read data.

It's less likely that any other headers will set `HS_LITTLE_ENDIAN`, so let's prefix our definition.

Also added some test cases for the byte swapping, which is what allowed me to narrow down the problem.


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -196,7 +196,7 @@ inline double hsSwapEndianDouble(double dvalue)
     return value.f;
 }
 
-#if LITTLE_ENDIAN
+#if HS_LITTLE_ENDIAN
```
Hoikas, line 199:
```suggestion
#ifdef HS_LITTLE_ENDIAN
```

Pedantic suggestion

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,146 @@
...
+    double i;
+    uint8_t c[8];
+} endianDouble_t;
+
+
+TEST(endianSwap, detection_accuracy)
+{
+    endian32_t val = { 0x01020304 };
+
+#if HS_LITTLE_ENDIAN
```
Hoikas, line 77:
```suggestion
#ifdef HS_LITTLE_ENDIAN
```

Deledrius (approved):

dpogue:
While I'm touching this, should we make it so that Big Endian is the special case and Little Endian is the default?

Hoikas:
In what way? The CMake `TestBigEndian` module tests only for big endianness, and there isn't a corresponding LE test module AFAIK. If we were on CMake 3.20+, we could go `if(CMAKE_CXX_BYTE_ORDER STREQUAL LITTLE_ENDIAN)`.

dpogue:
I mean changing it so the check is `#ifdef HS_BIG_ENDIAN`

dpogue:
If we had C++20 there's probably some super fancy stuff we could do with `std::conditional` and `std::endian`

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+TEST(endianSwap, detection_accuracy)
+{
+    endian32_t val = { 0x01020304 };
+
+#ifdef HS_BIG_ENDIAN
+    EXPECT_EQ(val.c[0], 0x01);
+#else
+    EXPECT_EQ(val.c[0], 0x04);
+#endif
```
zrax, line 81: Unfortunately, this sort of type punning (even using a union) is UB in C++ (even though it's perfectly valid and legal in C).  It may work on supported compilers for now, but it's not guaranteed to in the future...  The "valid" ways to do this are either with memcpy or `std::bit_cast` (C++20).

dpogue: _grumble grumble_ Okay, I've moved everything to using memcpy

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+    memcpy(c, &s, sizeof(s));
+
+    EXPECT_EQ(c[0], 0x02);
+}
+
+TEST(endianSwap, toBE16)
+{
+    uint16_t s = hsToBE16(0x0102);
+    uint8_t c[sizeof(s)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 76:
```suggestion
    uint8_t c[sizeof(s)] {};
```

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+    memcpy(c, &i, sizeof(i));
+
+    EXPECT_EQ(c[0], 0x01);
+}
+
+TEST(endianSwap, toLE64)
+{
+    uint64_t l = hsToLE64(0x0102030405060708);
+    uint8_t c[sizeof(l)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 106:
```suggestion
    uint8_t c[sizeof(l)] {};
```

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    EXPECT_EQ(c[0], 0x01);
+}
+
+TEST(endianSwap, toLEFloat)
+{
+    // Float value of PI is 0x40490fdb
+    float f = hsToLEFloat(hsConstants::pi<float>);
+    uint8_t c[sizeof(f)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 127:
```suggestion
    uint8_t c[sizeof(f)] {};
```

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    EXPECT_EQ(c[0], 0xdb);
+}
+
+TEST(endianSwap, toBEFloat)
+{
+    // Float value of PI is 0x40490fdb
+    float f = hsToBEFloat(hsConstants::pi<float>);
+    uint8_t c[sizeof(f)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 138:
```suggestion
    uint8_t c[sizeof(f)] {};
```

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    EXPECT_EQ(c[0], 0x40);
+}
+
+TEST(endianSwap, toLEDouble)
+{
+    // Double value of PI is 0x400921fb54442d18
+    double d = hsToLEDouble(hsConstants::pi<double>);
+    uint8_t c[sizeof(d)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 149:
```suggestion
    uint8_t c[sizeof(d)] {};
```

Sources/Tests/CoreTests/test_endianSwap.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    EXPECT_EQ(c[0], 0x18);
+}
+
+TEST(endianSwap, toBEDouble)
+{
+    // Double value of PI is 0x400921fb54442d18
+    double d = hsToBEDouble(hsConstants::pi<double>);
+    uint8_t c[sizeof(d)];
+    std::fill_n(c, sizeof(c), 0);
```
zrax, line 160:
```suggestion
    uint8_t c[sizeof(d)] {};
```

zrax (approved):

PR #1204, opened 2022-06-28
Fixing parser to not split tokens containing flags

colincornaby:
The tokenizer as it was written would split tokens midway if they contained a flag character. I made changes so that the initial tokenization is done by whitespace and separators, and then flags are removed from the front instead of a split being used.

My assumption is that all arguments have to be separated by whitespace and separators.

This should fix things like -Age=GuildPub-Writers

### DIFF
--- a/Sources/Tests/CoreTests/test_plCmdParser.cpp
+++ b/Sources/Tests/CoreTests/test_plCmdParser.cpp
@@ -548,7 +548,7 @@ TEST(plCmdParser, case_sensitive_match)
     EXPECT_EQ(specified, true);
 }
 
-TEST(plCmdParser, seperator_in_value)
+TEST(plCmdParser, separator_in_value)
 {
     const plCmdArgDef cmds[] = {
         { kCmdTypeString, "color", 0}
--- a/Sources/Tests/CoreTests/test_plCmdParser.cpp
+++ b/Sources/Tests/CoreTests/test_plCmdParser.cpp
@@ -548,7 +548,7 @@ TEST(plCmdParser, case_sensitive_match)
     EXPECT_EQ(specified, true);
 }
 
-TEST(plCmdParser, separator_in_value)
+TEST(plCmdParser, separator_in_value_nix)
 {
     const plCmdArgDef cmds[] = {
         { kCmdTypeString, "color", 0}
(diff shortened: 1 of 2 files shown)


dpogue:
Because this is one of the 3 classes with actual unit tests, I'm going to ask you add a test case to https://github.com/H-uru/Plasma/blob/master/Sources/Tests/CoreTests/test_plCmdParser.cpp that tests for arguments containing flag characters

dpogue (approved):
This _seems_ fine to me, but I'd feel more confident if @Hoikas confirmed (to his knowledge) that no other flags should be broken by this change

Sources/Tests/CoreTests/test_plCmdParser.cpp
```diff
@@ -547,3 +547,17 @@ TEST(plCmdParser, case_sensitive_match)
 
     EXPECT_EQ(specified, true);
 }
+
+TEST(plCmdParser, seperator_in_value)
```
ZarothYe, line 551:
```suggestion
TEST(plCmdParser, separator_in_value)
```
Extremely pedantic spelling correction, but this one's a bit of a pet peeve. Sorry. ;)

colincornaby: Fixed. (Should have just hit the commit suggestion button but noticed it too late!)

Hoikas (commented):
This looks fine to me, but I wonder if we should add some tests to make sure that `/` arguments still function as expected?

colincornaby:
> This looks fine to me, but I wonder if we should add some tests to make sure that `/` arguments still function as expected?

Would you be ok with me just adding a variant of the same test I added, but with a Windows style flag?

Sources/Tests/CoreTests/test_plCmdParser.cpp
```diff
@@ -547,3 +547,31 @@ TEST(plCmdParser, case_sensitive_match)
...
+}
+
+TEST(plCmdParser, separator_in_value_windows)
+{
+    const plCmdArgDef cmds[] = {
+        { kCmdTypeString, "color", 0}
+    };
+
+    plCmdParser parser(cmds, std::size(cmds));
+    parser.Parse("plCmdParser /color reddish-blue");
```
colincornaby, line 572: I'm not exactly sure what the specific syntax supported by Uru is here. Starting with the Windows "/" and no space syntax since that seems to be what most older Windows apps expect. (Newer ones are using the *nix convention which would also work on Windows.)

dpogue: I know for sure it does stuff like `/Renderer=gl` but not sure about `/Renderer gl`

colincornaby: I can throw the = in. It should all be ok according to the parser, but probably best to stick to what is being documented or typically used.

Hoikas (approved):

PR #1205, opened 2022-07-01
Asio Changes

colincornaby:
Pull request to cut networking over from NT sockets to ASIO. This change affects all operating systems including Windows.

This pull request is necessary to move forward on native macOS and Linux versions, but should be tested and considered carefully. It's working reliably on macOS - but the NT sockets version had a lot of handling for edge cases and timeouts. There are open questions on if the NT sockets versions could reach all the defined code paths. It's not clear that the Windows version could do soft disconnects which a lot of this code is for.

Once approved and merged, next steps will be pull requests to start pulling in the native macOS client. Thats not going to be a quick process though. I need to do more review on the Mac code to make sure it meets code standards, and has any clear rough edges figured out. This will allow the Metal branch to pull in ASIO from master and fully build.


Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -125,7 +125,7 @@ struct WriteOperation
 
     asio::const_buffer AsBuffer() const
     {
-        return asio::buffer(fNotify.buffer, fNotify.bytes);
+        return asio::buffer(fNotify.buffer + fNotify.bytesCommitted, fNotify.bytes - fNotify.bytesCommitted);
```
colincornaby, line 128: One issue with reusing buffers is it's not clear what the start and endpoints of the new data is, so it can cause the data length sent to the server to be wrong. bytes is the total size of the buffer, but that may include the length of data already sent.

It seems like Windows had a Win32 structure called OVERLAPPED that was part of socket I/O to deal with this problem and track which parts of the buffer were already committed. Don't need anything fancy, just need a `bytesCommitted` field to track how much of `bytes` has already been submitted.

`bytes` is still important because it's compared against `bytesProcessed`

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,658 @@
...
+            WriteOperation *op = conn->fWriteOps.front();
+            
+            size_t opBytesWritten = std::min(bytes, (size_t) op->fNotify.bytes - op->fNotify.bytesProcessed);
+            op->fNotify.bytesProcessed += opBytesWritten;
+            bytes -= opBytesWritten;
+            if  (op->fNotify.bytes == op->fNotify.bytesProcessed) {
+                conn->fWriteOps.pop_front();
+                
+                // callback notification procedure if requested
+                if (0) {
```
colincornaby, line 607: This is the weird part. The async path checks if it needs to call a notification callback, and then in response to that callback, can soft disconnect.

In Windows, there is a op->notify boolean that controls if the callback should be called. The weird thing is - I can't find any place that is set to true. SocketQueueAsyncWrite even directly sets it to false.

I could be very wrong here - and something I haven't seen could be going down the soft disconnect path. The ASIO version doesn't have the notify boolean right now, so I'm just disabling the entire path.

dgelessus: I'm absolutely not an expert on this low-level IO code, but it looks like most uses of `Nt::Operation.notify` were in `pnAceNtFile`, which H'uru removed way back in 2012 (d1da655fb2524fa6d45f8afc6df39da5e71cd4ee). One more dead use in `pnAceNtSocket` was removed in 2021 (e8b651d13a54c97a13f59f01a95f17231fd0e8f9).

In the current code, `op->notify` it's only set to `true` once at the start in `AsyncSocketConnect`, and then (presumably always) changed to `false` in `SocketQueueAsyncWrite`, as you say. So there should be no way for the callback to get called here. Even if it does get called, all code that handles the `kNotifySocketWrite` event looks like this:

```cpp
        case kNotifySocketWrite:
            // No action
        break;
```

So in my non-expert opinion, this code looks quite dead.

colincornaby: I can remove this section that looks for a callback, and then add an assert on the disconnect function if anything out there tries a soft disconnect. That would invalidate other large sections of code in this pull request that could be removed.

Not necessarily on topic - how many of these network changes have hit the OpenUru client? One risk of removing these sections is breaking integration paths with OpenUru. But looking at the previous structure, it also seems unlikely any of that would survive a transition to ASIO.

Hoikas: None, and all indication is that none ever will.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,658 @@
...
+        auto membuf = new uint8_t[sizeof(WriteOperation) + bytesAlloc];
+        WriteOperation* op = new (membuf) WriteOperation;
+        conn->fWriteOps.emplace_back(op);
+
+        // TODO: asyncId
+        op->fAllocSize = bytesAlloc;
+        op->fNotify.asyncId = /* TODO */nullptr;
+        op->fNotify.buffer = membuf + sizeof(WriteOperation);
+        op->fNotify.bytes = bytes;
+        op->fNotify.bytesProcessed = 0;
```
colincornaby, line 579: The NT socket sets `bytesProcessed` to `bytes` instead of 0. Having slightly different record keeping isn't the biggest problem, but looking at this difference is on my list of things to do. Maybe the NT socket has some trick that would simplify record keeping a bit on the transaction.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,575 @@
...
+    if (bytes) {
+        // Allocate storage alongside the operation structure itself
+        unsigned bytesAlloc = std::max(bytes, kMinBacklogBytes);
+        auto membuf = new uint8_t[sizeof(WriteOperation) + bytesAlloc];
+        WriteOperation* op = new (membuf) WriteOperation;
+        conn->fWriteOps.emplace_back(op);
+
+        // TODO: asyncId
+        op->fAllocSize = bytesAlloc;
+        op->fNotify.asyncId = /* TODO */nullptr;
```
colincornaby, line 499: Looking through the NT source trying to figure out the role of asyncId.

My understanding is that the NT sockets kept a master operation list for all read/write operations within a socket. So each op needs an id by which it can be tracked in since read/writes can happen in parallel.

The ASIO version has a dedicated array for write operations - so I'm not sure how the NT sockets work exactly maps. Buffers should always be written in order, so I should always be able to refer to the head of the write ops and assume it is the operation that just called back.

I'm still trying to better understand the NT code though. `INtOpDispatch` seems like the function that deals the most with the asyncId and the operation list.

Hoikas:
Just checking in -- is this ready for a review?

colincornaby:
Gave this PR a try with two players on the same server. Everything went well except for this asserts which would get triggered sometimes on link out:

> ASSERTION FAILED:
> File: /Users/colincornaby/Documents/Plasma/PlasmaSrc/Sources/Plasma/PubUtilLib/plNetClient/plNetClientMgr.cpp   > Line: 391
> Message: Still remote players left when linking out

I'm not sure if this is a known issue or unique to this socket. And I'm don't have a clear enough idea of the networking to quite know where to look.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -192,7 +192,19 @@ static void SocketStartAsyncRead(AsyncSocket sock)
     sock->fSock.async_read_some(asio::buffer(start, count),
                                 [sock](const asio::error_code& err, size_t bytes) {
         if (err) {
-            LogMsg(kLogError, "Failed to read from socket: {}", err.message());
+            if(err.value() == asio::error::operation_aborted) {
```
Hoikas, line 205:
```suggestion
            if (err.value() == asio::error::operation_aborted) {
```

style nitpicking

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -471,18 +472,31 @@ static bool SocketQueueAsyncWrite(AsyncSocket conn, const void* data, unsigned b
...
     std::vector<asio::const_buffer> allWrites;
     allWrites.reserve(conn->fWriteOps.size());
-    for (const WriteOperation* op : conn->fWriteOps)
+    for (WriteOperation* op : conn->fWriteOps) {
         allWrites.emplace_back(op->AsBuffer());
-    conn->fSock.async_write_some(allWrites, [conn](const asio::error_code& err, size_t bytes) {
-        //...
+        op->fNotify.bytesCommitted = op->fNotify.bytes;
+    }
+    async_write(conn->fSock, allWrites, [conn, allWrites](const asio::error_code& err, size_t bytes) {
```
Hoikas, line 517: Do `conn` and `allWrites` need to be copied by value, or can we use references or move them into the lamdba? By move, I mean use `std::,move()` to move the contents in the capture.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -471,18 +472,31 @@ static bool SocketQueueAsyncWrite(AsyncSocket conn, const void* data, unsigned b
...
+        hsLockGuard(s_connectCrit);
+        while(bytes != 0) {
+            hsAssert(conn->fWriteOps.size() > 0, "buffer mismatch");
+            WriteOperation *op = conn->fWriteOps.front();
```
Hoikas, line 521:
```suggestion
            WriteOperation* op = conn->fWriteOps.front();
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -471,18 +472,31 @@ static bool SocketQueueAsyncWrite(AsyncSocket conn, const void* data, unsigned b
...
+            
+            size_t opBytesWritten = std::min(bytes, (size_t) op->fNotify.bytes - op->fNotify.bytesProcessed);
+            op->fNotify.bytesProcessed += opBytesWritten;
+            bytes -= opBytesWritten;
+            if  (op->fNotify.bytes == op->fNotify.bytesProcessed) {
```
Hoikas, line 526:
```suggestion
            if (op->fNotify.bytes == op->fNotify.bytesProcessed) {
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -50,6 +50,12 @@ static constexpr unsigned int   kMaxWorkerThreads   = 32;
 static constexpr int            kTcpSndBufSize      = 64*1024-1;
 static constexpr int            kTcpRcvBufSize      = 64*1024-1;
 
+// wait before checking for backlog problems
+static const unsigned   kBacklogInitMs      = 3*60*1000;
```
Hoikas, line 54:
```suggestion
static constexpr unsigned   kBacklogInitMs      = 3*60*1000;
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -50,6 +50,12 @@ static constexpr unsigned int   kMaxWorkerThreads   = 32;
 static constexpr int            kTcpSndBufSize      = 64*1024-1;
 static constexpr int            kTcpRcvBufSize      = 64*1024-1;
 
+// wait before checking for backlog problems
+static const unsigned   kBacklogInitMs      = 3*60*1000;
+
+// destroy a connection if it has a backlog "problem"
+static const unsigned   kBacklogFailMs      = 2*60*1000;
```
Hoikas, line 57:
```suggestion
static constexpr unsigned   kBacklogFailMs      = 2*60*1000;
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -444,8 +454,30 @@ void AsyncSocketDisconnect(AsyncSocket conn, bool hardClose)
...
-    // TODO: Backlog shit
     hsLockGuard(s_connectCrit);
+    
+    // check for data backlog
+    WriteOperation * firstQueuedWrite = conn->fWriteOps.front();
+    if  (firstQueuedWrite) {
+        unsigned currTimeMs = TimeGetMs();
+        if (((long) (currTimeMs - firstQueuedWrite->queueTimeMs) >= (long) kBacklogFailMs)
+        &&  ((long) (currTimeMs - conn->initTimeMs) >= (long) kBacklogInitMs)
+        ) {
```
Hoikas, line 460: The indentation here is a bit baffling IMO.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -444,8 +454,30 @@ void AsyncSocketDisconnect(AsyncSocket conn, bool hardClose)
...
+            PerfAddCounter(kAsyncPerfSocketDisconnectBacklog, 1);
+
+            if (firstQueuedWrite->fNotify.bytes) {
+                LogMsg(
+                    kLogPerf,
+                    "Backlog, c:{} q:{}, i:{}",
+                       firstQueuedWrite->fNotify.buffer[0],
```
Hoikas:
```suggestion
                   firstQueuedWrite->fNotify.buffer[0],
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -116,6 +116,7 @@ struct ConnectOperation
     void*                   fParam;
     std::vector<uint8_t>    fConnectBuffer;
     unsigned int            fConnectionType;
+    std::recursive_mutex    critsect;
```
Hoikas, line 119: Needs the f prefix

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -162,14 +164,24 @@ static std::recursive_mutex         s_connectCrit;
 static std::list<ConnectOperation*> s_connectList;
 static unsigned                     s_nextConnectCancelId = 1;
 
+const unsigned kCloseTimeoutMs = 8*1000;
```
Hoikas:
```suggestion
static constexpr unsigned kCloseTimeoutMs = 8*1000;
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -430,6 +442,46 @@ static void HardCloseSocket(AsyncSocket conn)
...
+            if (0 < (signed) (sleepMs = sock->closeTimeMs - currTimeMs))
+                break;
+
+            if (sock->fSock.is_open())
+                sockets.push_back(sock);
+
+            // To avoid a race condition, this unlink must occur
+            // after the socket handle has been cleared
+            s_socketList.pop_front();
+        }
```
Hoikas: Should this have a sleep somewhere so that we aren't getting into a loop that consumes an entire CPU core?

Hoikas (commented):
This is just a brief glance at the work done on top of #1066. I haven't reviewed the #1066 parts at all nor is this review necessarily complete.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
Hoikas: Also, there is a brace spacing problem just above this.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: This code is part of the timer path, which was part of the soft disconnect path, which was removed. This function implementation isn't part of the PR anymore. I did find a declaration of this function that should have been removed, but this implementation shouldn't be/isn't part of the current head.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: I think this is more dead code from the timer/soft disconnect path. Removing.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: allWrites looks like more dead code. conn could simply be passed by reference using & in since the lifetime of conn should not exceed the lambda. At least - if it does exceed the lambdas lifetime - something else is very wrong.

colincornaby: I committed a change to make conn pass by reference, but conn is actually a pointer type so that may have been kind of silly.

Hoikas: Yeah, if conn is a pointer, copying by value is fine. I was primarily worried about copying the vector `allWrites`, which could be expensive (eg require an extra allocation).

colincornaby: I've removed the commit that pass `conn` by reference. `allWrites` is no longer passed through at all.

Hoikas:
This has been in use on the TL client for several weeks to no adverse effect, FWIW. Full review still pending.

dpogue:
I've seen some weirdness on Linux when I tried to run my plFilePatcher PR with this, aborting on exit. Entirely possible I'm doing something wrong, but probably worth investigating.

colincornaby:
I can investigate the combination but you might be the best person to take a first swing at it @dpogue. I'm juggling a few things and it would make me feel better to have someone crawling through this code besides me since it's so critical.

dpogue:
I'll try to investigate, but it might be into September before I get a chance to really dig into it

colincornaby:
Totally understood. If you could write up repo steps in since I haven't used that tool yet - I can probably look in the next few weeks. I don't have all the environments available to me (and my Windows setup is still kind of a mess) - so it might take me some time to get to the bug if it's not happening on macOS.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,572 @@
...
+{
+    hsLockGuard(s_connectCrit);
+    
+    // check for data backlog
+    WriteOperation * firstQueuedWrite = conn->fWriteOps.front();
+    if  (firstQueuedWrite) {
+        unsigned currTimeMs = TimeGetMs();
+        if (((long) (currTimeMs - firstQueuedWrite->queueTimeMs) >= (long) kBacklogFailMs)
+        &&  ((long) (currTimeMs - conn->initTimeMs) >= (long) kBacklogInitMs)
+        ) {
```
zrax, line 457:
```suggestion
        if (((long) (currTimeMs - firstQueuedWrite->queueTimeMs) >= (long) kBacklogFailMs)
            && ((long) (currTimeMs - conn->initTimeMs) >= (long) kBacklogInitMs))
        {

```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,572 @@
...
+    allWrites.reserve(conn->fWriteOps.size());
+    for (WriteOperation* op : conn->fWriteOps) {
+        if (op->fNotify.bytes - op->fNotify.bytesCommitted > 0) {
+            allWrites.emplace_back(op->AsBuffer());
+            op->fNotify.bytesCommitted = op->fNotify.bytes;
+        }
+    }
+    async_write(conn->fSock, allWrites, [conn](const asio::error_code& err, size_t bytes) {
+        hsLockGuard(s_connectCrit);
+        while(bytes != 0) {
```
zrax, line 516:
```suggestion
        while (bytes != 0) {

```

Hoikas:
Hmmm... Looks like we have a conflict, and (for some reason) AppVeyor has returned???

colincornaby:
> Hmmm... Looks like we have a conflict, and (for some reason) AppVeyor has returned???

I'll start with rebase and see if I can get anywhere.

colincornaby:
The rebase seems to have solved the conflict (there was a bug fix in the NT sockets) and AppVeyor seems to be gone again.

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
```diff
@@ -54,6 +54,33 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    }
+    
+private:
+    void operator()(_Args&&... args) {
+        fFuncPtr(args...);
+        fSem.Signal();
+    }
+    _Fp&& fFuncPtr;
+    hsSemaphore fSem = hsSemaphore();
+};
```
colincornaby, line 83: Proposing something here - haven't wired it in yet.

This is a subclass of std::thread that implements a timed join. It has a few benefits:
- Just needs a semaphore to function. Semaphore is signaled when the function completes - so no need to rely on functionality in the underlying threading library. And no breaking std::thread by calling underlying function.
- Implements the same basic thread functionality we're using on Windows, and works everywhere.
- Properly implements timeout or join, no risk of a join firing later and messing things up if the thread was detached.

Downside: We'd need to use this subclass instead of std::thread where we need to use TimedJoin.

colincornaby: Trying this out in practice. This implementation won't be sufficient in since in implementation threads are copied. That also means mutex ownership needs to move. Still working through that.

I guess I'm also starting to wonder what we're trying to accomplish here as well. The timeout based join will prevent deadlocks, but may also cause bad behavior by terminating threads mid execution. I'd be worried it we're relying on that timeout.

colincornaby:
I've reverted the change that added the subclass of std::thread that tried to patch in a timed join via a condition. There were a lot of issues trying to patch things over the std::thread constructor. Another developer might have more luck than me - but I haven't seen any other implementations adding the functionality to std::thread via a subclass yet.

To review the issue:
std::thread does not like the underlying thread state being changed, and will throw/log errors if the underlying thread state does not match its own bookkeeping. This can cause problems with the platform specific code that's trying to implement a timed thread join.
- On Windows there isn't a problem because the Windows code is doing a timed observation of thread completion, it's not actually trying to join the thread. So no underlying thread state is being changed.
- The Linux code is a problem because it's actually attempting a join, which will change the thread state.

A few solutions have been proposed.

### Conditions
- I've been looking at putting a condition into the thread. Conditions can be waited on with a timeout. When the thread is done it can signal the condition, and signal any observer that is waiting on thread completion. This is the approach Boost is using.
- Trying to implement this directly into std::thread is troublesome. I have to make sure both the thread's body and the thread object have access to the same condition, and the thread body is created in std::thread's constructor. The condition also has to survive object moves.
- Writing our own thread wrapper around std::thread seems more plausible, but requires deeper surgery in the code.

### Futures
- @dpogue has proposed a change that spawns a second thread using a promise, and that promise joins the thread. The promise's state can be observed with a timeout.
- This behavior isn't quite the same as a timed join. If a timed join times out, the join is cancelled. If waiting on the promise times out, the promise will continue executing in the background, and a join will eventually happen.
- There's other side effects as well - the behavior being implemented is that the thread will be detached after a timeout. If the promise ever does execute its join, this will cause an exception. It's illegal to join a thread after it's been detached. We could technically catch this exception.

### Do we need to worry about this?

I'm a little worried about this class of bug in general. The threads are being force detached so they can be killed mid execution. This could be very bad and we should possibly be weary of this. I'm not sure of the entire range of things these threads do - but they are attached to different types of I/O. We may want to be careful about just killing I/O midway through execution.

Because this code also touches critical functions, the possibility of a deadlock should be concerning on its own. We may not want to silently kill deadlocks in code that should not be deadlocking. Hiding the deadlocks could be hiding more serious issues and underlying I/O failures.

I'm not an expert though at everything these threads are doing. Someone else on the thread might have a better idea of how much danger is possibly lurking here. And maybe there are known deadlocks here that I'm not familiar with.

Personally - I'm still in favor of shipping with a non-timed-out join on *nix, for the reasons above. I'm not eager to cover over deadlock issues with this fallback, or cut off slow or deadlocked threads that may be doing real I/O. But, I've mostly been focusing on the render portions of this engine so there may be things I'm not seeing or understanding here. We also could continue to deal with the timed join on *nix in follow up PRs.

zrax:
I assume the whole timed join stuff was originally put in to prevent the game from hanging indefinitely on shutdown when a deadlock does occur...  What if, instead of trying to solve the problem of timed joins on individual threads, we just use a normal join everywhere, but spawn a single thread during shutdown that waits for a while and then `std::terminate`s the whole process (or debug break in debug builds) if it doesn't get cancelled before its timeout expires?  That way, we still prevent the possibility of indefinite hangs at shutdown, but don't have to deal with individual threads being left in a bad state, or with trying to make timed joins work across different platforms.

colincornaby:
Would `std::terminate` solve the problem of threads dying mid work? I think it would still cut off threads mid execution (especially if the intention is to kill deadlocked threads.)

I'm doing some research into the source history, but I'm also still concerned about covering up deadlocked threads in general. Deadlocking threads would indicate some serious problem in the engine and it feels a little weird to plan for them. Obviously the client deadlocking on exit is not a great experience either. But if threads are having deadlocking issues, the deadlock could have happened halfway through the users play session. Timed joins can be useful, just worried about the implications here. Threads are complicated and there is potential for deadlock that wasn't understood when the client shipped. But thread deadlocks seem like something to fix directly to me because of the implications of a network thread deadlocking.

Another option is to go back to (I know I know) something like abcbba7a45d0a20efecdb6b8fd0b2ed25eb42065. The structure there would be usable for adding auxiliary things like a condition member necessary for cross platform timed joins.

But again - still worried about covering up deadlocks in general. Maybe we could log a warning message in debug builds so it at least gets some attention during development. Still worried about implications of deadlocking threads in a released build though.

colincornaby:
This is a pretty good reference on Stack Overflow - and it's noted that if main exits all threads will be instakilled mid execution. So if there is a deadlock - I don't think there is a safe way to exit. (Although a user killing the process may just end up being the same amount of unsafe.)

https://stackoverflow.com/questions/22803600/when-should-i-use-stdthreaddetach

This answer is also pretty good on the utility of detach:

> This answer is aimed at answering question in the title, rather than explaining the difference between join and detach. So when should std::thread::detach be used?
> 
> In properly maintained C++ code std::thread::detach should not be used at all. Programmer must ensure that all the created threads gracefully exit releasing all the acquired resources and performing other necessary cleanup actions. This implies that giving up ownership of threads by invoking detach is not an option and therefore join should be used in all scenarios.
> 
> However some applications rely on old and often not well designed and supported APIs that may contain indefinitely blocking functions. Moving invocations of these functions into a dedicated thread to avoid blocking other stuff is a common practice. There is no way to make such a thread to exit gracefully so use of join will just lead to primary thread blocking. That's a situation when using detach would be a less evil alternative to, say, allocating thread object with dynamic storage duration and then purposely leaking it.

I have noticed Plasma is using things like global mutexes that when not properly cleaned up will cause the next session to fail to launch. I'm not sure if that's happening in any of these paths, but there is certainly state that can be a problem when not cleaned up.

Again - if the user does the force quit themselves though it may end up being the same thing. So I'm not sure if we should be putting ourselves in a bad state or if we should wait for the user to do it.

colincornaby:
On iOS I'm noticing the ASIO networking in max'ing out at about 800 kb/s. I thought maybe it was a bandwidth restriction of Trollland, but I'm not seeing the same issues with the NT sockets on Windows. I'll dig a bit more.

colincornaby:
Issue seems to be Mac (and maybe Linux) specific - Windows version is getting the expected download speeds with ASIO.

colincornaby:
I can get macOS performance to improve to match Windows if I increase the size of the network receive buffer. But Linux is showing the same issues and the buffer size change doesn't work there.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,572 @@
...
+    if (err)
+        LogMsg(kLogError, "Failed to set socket to non-blocking: {}", err.message());
+
+    sock->fSock.set_option(asio::socket_base::send_buffer_size(kTcpSndBufSize), err);
+    if (err)
+        LogMsg(kLogError, "Failed to set socket send buffer size: {}", err.message());
+
+    sock->fSock.set_option(asio::socket_base::receive_buffer_size(kTcpRcvBufSize), err);
+    if (err)
+        LogMsg(kLogError, "Failed to set socket recv buffer size: {}", err.message());
```
colincornaby, line 302: Disabling the two set_option calls seems to resolve the download bandwidth issues on both Linux and macOS. Using the file patcher I'm seeing download speeds of up to 30 MiB/s, improved from 800 KiB/s.

There are some threads hinting that leaving these unset will allow ASIO and the underlying network stack to automatically scale buffer size up and down. I saw these values are inherited from the NT sockets implementation. I'm not sure the side effects of not setting these sizes - and I haven't retested on Windows.

Generally I can't find any discussion online yet of why POSIX sockets and NT sockets would behave so differently. So I'm unsure why this buffer size is working fine on Windows but not elsewhere.

colincornaby:
Some of this conversation happened in IRC - but repeating here.

I've removed the manual buffer size configuration - which has resolved the issues on macOS and Linux.

Furthermore - Microsoft recommends not setting these options. The network stack can more intelligently configure these buffers:
https://techcommunity.microsoft.com/t5/networking-blog/windows-network-performance-suffering-from-bad-buffering/ba-p/339668

The reasoning makes sense. When the buffer fills, there is an additional interaction between the client and the server as TCP/IP manages the overflow. The latency on my fiber connection is 5-10 ms, so at a buffer size of 64k, I'd incur 5-10 ms of latency each 64k. This would constrain any download speeds to approx 800 k/sec, which matches what I'm seeing.

There is still the mystery of why Windows is unaffected. Microsoft's documentation says you can suggest a buffer size, but they may quietly not honor the requested buffer size. That could be whats happening here - especially if MS feels there is a problem with many applications abusing this configuration option.

One open question is how far back Windows intelligently manages buffer sizes. The MS tech note says Windows 10 - but it was written during the Windows 10 era. I've seen some hints that Windows 10's modern network stack was introduced in Vista. But that doesn't mean XP's network stack didn't make the same automatic adjustments to the network buffer. That said - I have not tested this change on XP.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,561 @@
...
+          fParam(op->fParam), fConnectionType(op->fConnectionType),
+          fUserState(), fBuffer(), fBytesLeft(), initTimeMs(), closeTimeMs()
+    { }
+};
+
+static AsyncIoPool*                 s_ioPool;
+
+static std::recursive_mutex         s_connectCrit;
+static std::list<ConnectOperation*> s_connectList;
+static unsigned                     s_nextConnectCancelId = 1;
```
dpogue, line 159: This should be `uintptr_t` rather than `unsigned`. I'm seeing warnings about casting it to a `AsyncCancelId` (which is effectively a `void*`) because of size differences.

colincornaby: Fixed!

colincornaby:
Going through and looking more at the timeout stuff. I tried to unwind timeouts but then realized how deeply they're implemented.

DNS resolution, for example, uses timeouts in its destructors. This may have made sense at one time... you wouldn't want a destructor held up by a stalled network request. But the network stack has built in cancellation support that can cause an immediate return from network functions. And those cancellation functions are being invoked.

I'm also looking at ThreadDestroy in pnAcThread.cpp (https://github.com/colincornaby/Plasma/blob/5d535f580a897ea0d74631aacbff9456d3acd1ca/Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp#L62). I think this is a function that is meant to ensure all threads are destroyed. But even though it's called destroy, it doesn't actually destroy anything, it just waits.

The one new issue I see around ThreadDestroy is it relies on a PerfCounter to decide when all threads have been destroyed. When AceThread used to be a wrapper around std::thread, that wrapper would automatically increment and decrement the counter. Now that std::thread is being used without a wrapper, each thread is responsible to incrementing and decrementing its own PerfCounters. This creates room for a lot of possible bugs. If a thread forgets to increment/decrement a counter or enters into some unexpected control flow that causes one of those things not to happen - the TheadDestroy function will unnecessarily timeout.

Nothing here is outright broken. But it is kind of code-smelly. Again, most of this was inherited from the original source.

I think I'm leaning in favor again of re-wrapping std::thread with the AceThread wrapper, and enforcing things like the increment/decrement of the PerfCounter. Re-wrapping std::thread would also give more space for properly implementing cross platform async join (even if I'm skeptical of async-join-or-die in general.)

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
```diff
@@ -67,6 +84,36 @@ void ThreadDestroy (unsigned exitThreadWaitMs) {
...
+{
+    bool threadFinished = ref.impl->completion->try_lock_for(std::chrono::milliseconds(timeoutMs));
+    if (threadFinished) {
+        //thread is finished, safe to join with no deadlock risk
+        ref.impl->completion->unlock();
+        ref.impl->handle->join();
+    } else {
+        ref.impl->handle->detach();
+    }
+}
```
colincornaby, line 115: I've added a test implementation of restoring the original thread wrapper - more or less. Few advantages:
- Real, cross platform async join is working.
- Stuff like PerfAddCounter goes back into the wrapper, each thread body doesn't need to manage that itself.

I've only implemented the thread wrapper so far on the async socket. It seems like it's working fine, although I still need to try Linux/Windows. On macOS async join is now working.

Mostly looking for feedback here - I know this code was removed, but I think there are advantages to wrapping (like async join.) I don't know if stylistically it's all correct.

Looks like there may also be a build issue, I'll look.

dpogue:
I've tested this on Linux with plFilePatcher, and it was all good. Previous issues around threads aborting on exit are gone.

colincornaby:
I'm going to bake the wrapper down into the other parts of the code that use async join. I still think the wrapper itself can use some review. It exists in this awkward place in between C and C++. But if it unblocks Linux and keeps moving things forward that's good. It is a code cleanliness regression, using std::thread would be cleaner. But I don't think a bare std::thread is workable here. There are some changes coming in future versions of C++ threads that may obsolete this wrapper. But that would be in C++20 and beyond.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,553 @@
...
+        // get cancel id; we can avoid checking for zero by always using an odd number
+        ASSERT(s_nextConnectCancelId & 1);
+        s_nextConnectCancelId += 2;
+
+        *cancelId = op->fCancelId = (AsyncCancelId)s_nextConnectCancelId;
+    }
+
+    s_connectList.emplace_back(op);
+
+    tcp::endpoint remoteAddr(asio::ip::address_v4(netAddr.GetHostBytes()),
```
Hoikas, line 345: This could probably just use the old `GetHost()` method with a `reinterpret_cast<uint8_t*>(&host);` instead of requiring the addition of an entirely new method in `plNetAddress`.

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
```diff
@@ -65,10 +66,40 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 // for IoWaitId/TimerCreate/TimerUpdate
 constexpr unsigned kAsyncTimeInfinite = (unsigned) -1;
 
+struct AsyncThread;
+
+struct AsyncThreadRef {
+    std::shared_ptr<AsyncThread> impl;
```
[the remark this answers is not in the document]
colincornaby: I think that change can happen. It couldn't in earlier iterations that are irrelevant now. std::thread will invalidate any old value once moved, so the shared pointer was to insulate against that issue when the value gets stored, or gets provided to a thread. The AsyncThreadRef structure provides that insulation now in since impl is a shared_ptr.

colincornaby:
I mentioned on IRC I was having a weird issue - it was actually a known issue. plFilePatcher deadlocks when it's run against a working directory with no server.ini. There are underlying issues in the engine if the servers aren't configured. It's an edge case because the client will just give up if the server.ini isn't detected. But plFilePatcher doesn't have similar catches.

I did figure out the deadlock and fixed it - but that's a bug in the core networking code and not the ASIO code. I'll leave that for a different pull requests. But just letting everyone know the issue I was working on is not related to ASIO. I just forgot to set my working directory...

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: I generally agree with this comment, but I'll let this sit for a bit in case anyone has any other feedback. Adding a function where one wasn't seems more intentional and I didn't know if there was a deeper architectural reasoning here.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,553 @@
...
+        : fSock(std::move(op->fSock)), fNotifyProc(op->fNotifyProc),
+          fParam(op->fParam), fConnectionType(op->fConnectionType),
+          fUserState(), fBuffer(), fBytesLeft(), initTimeMs(), closeTimeMs()
+    { }
+};
+
+static AsyncIoPool*                 s_ioPool;
+
+static std::recursive_mutex         s_connectCrit;
+static std::list<ConnectOperation*> s_connectList;
```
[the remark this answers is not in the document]
colincornaby: It's not clear to me this change should be made. I don't know if I have the expertise in C++ or ASIO to make the call. This struct contains values that will be made invalid in the source struct the struct is copied. Strict passing by reference could sidestep this issue, but passing by pointer could be a way to make sure the struct is not unintentionally copied.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,553 @@
...
+{
+    tcp::socket             fSock;
+    plNetAddress            fRemoteAddr;
+    AsyncCancelId           fCancelId;
+    FAsyncNotifySocketProc  fNotifyProc;
+    void*                   fParam;
+    std::vector<uint8_t>    fConnectBuffer;
+    unsigned int            fConnectionType;
+
+    ConnectOperation(asio::io_context& context)
```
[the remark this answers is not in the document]
colincornaby: A const io_context& seems to be disallowed.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,553 @@
...
+{
+    std::recursive_mutex        fCritsect;
+    tcp::socket                 fSock;
+    FAsyncNotifySocketProc      fNotifyProc;
+    void*                       fParam;
+    unsigned int                fConnectionType;
+    void*                       fUserState;
+    uint8_t                     fBuffer[kAsyncSocketBufferSize];
+    size_t                      fBytesLeft;
+    std::list<WriteOperation *> fWriteOps;
```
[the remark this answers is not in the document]
colincornaby: WriteOperation structures seem to have a variable size (see pnAceSocket:471), so I'm not sure that would be viable.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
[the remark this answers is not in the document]
Hoikas: You can delete the copy constructor to make sure that the struct isn't copied accidentally. In the struct's decalration:

```C++
struct ConnectOperation
{
    ConnectOperation(const ConnectOperation&) = delete;
};
```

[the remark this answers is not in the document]
colincornaby: Oh right, I'll take another look then.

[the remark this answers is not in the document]
colincornaby: Ok, took another look. In pnAceSocket.cpp:350 there is this:
```
        // This operation is no longer a candidate for cleanup or cancellation
        // by the time we get a callback from async_connect
        {
            hsLockGuard(s_connectCrit);
            auto iter = std::find(s_connectList.cbegin(), s_connectList.cend(), op);
            if (iter != s_connectList.cend())
                s_connectList.erase(iter);
        }
```

A comparison operator could be written. But I'm starting to think maybe this is just best left as a pointer... Unless you have any other tricks! :)

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: Looking through this one again. Here's what I'll propose: I think this change should be implemented, but that GetHostBytes should stay. plNetAddress is a nice utility class and GetHostBytes could be useful for debugging. But the current implementation does a bit of unnecessary work.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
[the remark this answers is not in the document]
Hoikas: That's a good point in that I missed something in my code review. The snippet that you've referenced could be done in constant time instead of by re-looking up the connection in the list by having `std::list` return an iterator to the newly created item. `std::list` iterators are guaranteed to be stable (read: valid) until the item is erased from the container. This will also remove extra pointer indirection, potentially improving cache utilization. For example, I would rewrite https://github.com/H-uru/Plasma/pull/1205/files/21db57551b4b0cc60cbe1c2d09b1f37884d0aca7#diff-ac6773cf2946916e561937a5d6a0b1fafd0245b5412722d5b1a5933531e51542R311-R378 to look like:

```c++
void AsyncSocketConnect(AsyncCancelId* cancelId, const plNetAddress& netAddr,
                        FAsyncNotifySocketProc notifyProc, void* param,
                        const void* sendData, unsigned sendBytes)
{
    ASSERT(notifyProc);
    ASSERT(s_ioPool);

    PerfAddCounter(kAsyncPerfSocketConnAttemptsOutCurr, 1);
    PerfAddCounter(kAsyncPerfSocketConnAttemptsOutTotal, 1);

    // Locking the critical section here will not deadlock because `asio::basic_socket::async_connect`
    // will return immediately. Therefore, the lock in the callback cannot deadlock us.
    hsLockGuard(s_connectCrit);
    auto op = s_connectList.emplace(s_connectList.end(), s_ioPool->fContext);
    op->fRemoteAddr = netAddr;
    op->fNotifyProc = std::move(notifyProc);
    op->fParam = param;
    if (sendBytes) {
        auto sendBuffer = reinterpret_cast<const uint8_t*>(sendData);
        op->fConnectBuffer.assign(sendBuffer, sendBuffer + sendBytes);
        op->fConnectionType = sendBuffer[0];
    } else {
        op->fConnectionType = kConnTypeNil;
    }

    if (cancelId) {
        // get cancel id; we can avoid checking for zero by always using an odd number
        ASSERT(s_nextConnectCancelId & 1);
        s_nextConnectCancelId += 2;

        *cancelId = op->fCancelId = (AsyncCancelId)s_nextConnectCancelId;
    }

    tcp::endpoint remoteAddr(asio::ip::address_v4(netAddr.GetHostBytes()),
                             netAddr.GetPort());
    op->fSock.async_connect(remoteAddr, [op](const asio::error_code& connError) {
        // This operation is no longer a candidate for cleanup or cancellation
        // by the time we get a callback from async_connect
        bool success;
        if (connError) {
            LogMsg(kLogError, "Failed to connect: {}", connError.message());
            success = false;
        } else {
            success = SocketInitConnect(op);
        }

        if (!success) {
            AsyncNotifySocketConnect failed;
            failed.param      = op->fParam;
            failed.connType   = op->fConnectBuffer[0];
            failed.remoteAddr = op->fRemoteAddr;
            failed.localAddr.Clear();
            op->fNotifyProc(nullptr, kNotifySocketConnectFailed, &failed, nullptr);
        }

        {
            hsLockGuard(s_connectCrit);
            s_connectList.erase(op);
        }

        PerfSubCounter(kAsyncPerfSocketConnAttemptsOutCurr, 1);
    });
}
```

I also added a comment about why what appeared to me to be a potential deadlock is not actually a deadlock :)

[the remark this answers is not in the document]
colincornaby: Let me take a look at implementing your snippet and the other required changes. Did not know that was a feature of C++ iterators!

colincornaby:
There's some warnings about unsigned int and size_t mismatches. I'm going to standardize on size_t where I can in since that's best practice for buffer sizes.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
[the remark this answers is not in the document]
Hoikas: Each type of container has slightly different iterator rules. It's generally best to assume that iterators are not stable, especially for containers like std::vector. The list iterator is IIUC basically a fancy reference to a node in the linked list, so this is fine :)

Hoikas:
I don't think I understand why `AsyncThreadRef` exists. Why can't we pass around references to the `AsyncThread` itself instead?

colincornaby:
> I don't think I understand why `AsyncThreadRef` exists. Why can't we pass around references to the `AsyncThread` itself instead?

We can. That would be a similar change to what was done for passing references to the other structure. I can make that change.

colincornaby:
Ok, coming back around to `AsyncThreadRef` again.

One of the issues with `AsyncThread` is it has to exist in two threads at once. There's the main thread that's going to add it to an array to be tracked, and then the thread needs its own access to get to the mutex. Both threads need to own the `AsyncThread` simultaneously. If it goes out of scope in one thread, it still needs to be alive in the other. This is fairly incompatible with reference semantics.

`shared_ptr` does solve this problem rather well. The `AsyncThreadRef` covers that implementation detail. But it could be removed and all the code could be tweaked to work with `shared_ptr` directly.

Hoikas:
Thanks for the summary!

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
```diff
@@ -65,10 +66,39 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+
+// Threads are also allowed to set the workTimeMs field of their
+// structure to a nonzero value for "on", and IO_TIME_INFINITE for
+// "off" to avoid the overhead of calling these functions. Note
+// that this function may not be called for the main thread. I
+// suggest that application code not worry that timeMs might
+// "accidentally" equal the IO_TIME_INFINITE value, as it only
+// happens for one millisecond every 49 days.
+struct AsyncThread {
+    LINK(AsyncThread)                    link;
```
Hoikas, line 86: Are threads still placed into an intrusilvely linked list as implemented by `pnUtList`?

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
```diff
@@ -67,30 +84,41 @@ void ThreadDestroy (unsigned exitThreadWaitMs) {
 }
 
 //============================================================================
-void AsyncThreadTimedJoin(std::thread& thread, unsigned timeoutMs)
+AsyncThreadRef AsyncThreadCreate (
+    std::function<void()>    threadProc
+                               ) {
+    AsyncThreadRef ref;
+    ref.impl                  = std::make_shared<AsyncThread>();
+    ref.impl->proc            = threadProc;
```
Hoikas, line 92:
```suggestion
    ref.impl->proc            = std::move(threadProc);
```

Hoikas (commented):
I think we are getting very close to extensive testing and merge taim. It would be great if you could take care of the code style issues in the interim :smile:

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
colincornaby: No, I'll eliminate that struct member.

colincornaby:
> I think we are getting very close to extensive testing and merge taim. It would be great if you could take care of the code style issues in the interim 😄

I'll give the new code a run-through with Clang Format and see if that catches the format issues.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+static void SocketStartAsyncRead(AsyncSocket sock)
+{
+    hsLockGuard(sock->fCritsect);
+    uint8_t* start = sock->fBuffer + sock->fBytesLeft;
+    size_t   count = sizeof(sock->fBuffer) - sock->fBytesLeft;
+    sock->fSock.async_read_some(asio::buffer(start, count),
+                                [sock](const asio::error_code& err, size_t bytes) {
+        if (err) {
+            if (err.value() == asio::error::operation_aborted) {
```
colincornaby, line 190: There's a gap here that caused the deadlock when the UUID was wrong on the Mac client. When the server disconnects, we get an EOF error. But EOFs aren't handled here. My thought is an error should be handled like a disconnect?

Hoikas: I would consider an EOF a disconnect.

colincornaby: It's in. plFilePatcher still deadlocks when the Cyan Worlds Gatekeeper hangs up on it, but the network code is at least still routing the message out correctly. I think somewhere deeper in the code there is a while loop that doesn't do error checking. Likely beyond the scope of this PR.

Hoikas: That seems likely. We had to add special handling for Cyan's hangup to the Windows login dialog.

colincornaby:
It seems like the testing tasks we've wanted to do so far have been completed. Just checking in to see if there is more testing people would like to do.

colincornaby:
> > It seems like the testing tasks we've wanted to do so far have been completed. Just checking in to see if there is more testing people would like to do.
> 
> It hasn't (to my knowledge) been tested against a MOSS server (such as OU's Minkata), but if it works on all the other implementations then I'd be inclined to say any bugs are MOSS bugs rather than ASIO bugs.
> 
> I'm pretty eager to see this merged so that Linux work can be unblocked 😁

Yeah, as soon as this is done, the macOS client would likely go into a (likely long running) draft PR as well. On macOS I haven't noticed any issues with this code. The only thing I'd possibly be worried about are error or edge cases that are not commonly encountered.

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcIo.h
```diff
@@ -132,20 +132,24 @@ struct AsyncNotifySocketListen : AsyncNotifySocketConnect {
...
+    size_t          bytesCommitted;
+    
+    AsyncNotifySocketWrite() : AsyncNotifySocketRead(), bytesCommitted() { }
+};
+
+// return false to disconnect
+typedef std::function<bool (AsyncSocket /* sock */,
+                            EAsyncNotifySocket /* code */,
+                            AsyncNotifySocket* /* notify */,
+                            void** /* userState */)> FAsyncNotifySocketProc;
```
Hoikas, line 152:
```suggestion
using FAsyncNotifySocketProc = std::function<bool(AsyncSocket, EAsyncNotifySocket, AsyncNotifySocket*, void**)> ;
```

The arugment names should probably go into a doxygen comment, IMO.

Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcThread.h
```diff
@@ -51,6 +51,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #define PLASMA20_SOURCES_PLASMA_NUCLEUSLIB_PNASYNCCORE_PRIVATE_PNACTHREAD_H
 
 #include <thread>
+#include <mutex>
```
Hoikas, line 54: Alphabetize includes per #788

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+    }
+}
+
+static void SocketStartAsyncRead(AsyncSocket sock)
+{
+    hsLockGuard(sock->fCritsect);
+    uint8_t* start = sock->fBuffer + sock->fBytesLeft;
+    size_t   count = sizeof(sock->fBuffer) - sock->fBytesLeft;
+    sock->fSock.async_read_some(asio::buffer(start, count),
+                                [sock](const asio::error_code& err, size_t bytes) {
```
Hoikas, line 188:
```suggestion
        [sock](const asio::error_code& err, size_t bytes) {
```

indentation

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+            return;
+
+        sock->fBytesLeft += bytes;
+
+        AsyncNotifySocketRead notifyRead;
+        notifyRead.buffer = sock->fBuffer;
+        notifyRead.bytes = sock->fBytesLeft;
+
+        if (!sock->fNotifyProc || !sock->fNotifyProc(sock, kNotifySocketRead, &notifyRead, &sock->fUserState))
+        {
```
Hoikas, line 218:
```suggestion
        if (!sock->fNotifyProc || !sock->fNotifyProc(sock, kNotifySocketRead, &notifyRead, &sock->fUserState)) {
```

braces on same line

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+            // No callback, or the callback told us to stop reading
+            return;
+        }
+
+        // if only some of the bytes were used, then shift
+        // remaining bytes down.  Otherwise, clear the buffer.
+        sock->fBytesLeft -= notifyRead.bytesProcessed;
+        if (sock->fBytesLeft != 0) {
+            if ((sock->fBytesLeft > sizeof(sock->fBuffer)) || ((notifyRead.bytesProcessed + sock->fBytesLeft) > sizeof(sock->fBuffer)))
+            {
```
Hoikas, line 228:
```suggestion
            if ((sock->fBytesLeft > sizeof(sock->fBuffer)) || ((notifyRead.bytesProcessed + sock->fBytesLeft) > sizeof(sock->fBuffer))) {
```

braces on same line

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+    // Send initial data
+    if (!op.fConnectBuffer.empty()) {
+        if (!AsyncSocketSend(sock.get(), op.fConnectBuffer.data(), op.fConnectBuffer.size()))
+            return false;
+    }
+
+    if (!IS_TEXT_CONNTYPE(op.fConnectionType)
+        && op.fConnectBuffer.size() < sizeof(AsyncSocketConnectPacket))
+    {
```
Hoikas, line 294:
```suggestion
    if (!IS_TEXT_CONNTYPE(op.fConnectionType)
        && op.fConnectBuffer.size() < sizeof(AsyncSocketConnectPacket)) {
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+static bool SocketQueueAsyncWrite(AsyncSocket conn, const void* data, size_t bytes)
+{
+    hsLockGuard(s_connectCrit);
+
+    // check for data backlog
+    WriteOperation* firstQueuedWrite = conn->fWriteOps.front();
+    if (firstQueuedWrite) {
+        unsigned currTimeMs = TimeGetMs();
+        if (((long)(currTimeMs - firstQueuedWrite->queueTimeMs) >= (long)kBacklogFailMs) && ((long)(currTimeMs - conn->initTimeMs) >= (long)kBacklogInitMs))
+        {
```
Hoikas, line 431:
```suggestion
        if (((long)(currTimeMs - firstQueuedWrite->queueTimeMs) >= (long)kBacklogFailMs) && ((long)(currTimeMs - conn->initTimeMs) >= (long)kBacklogInitMs)) {
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+            }
+            AsyncSocketDisconnect((AsyncSocket)conn, true);
+            return false;
+        }
+    }
+
+    // If the last buffer still has space available then add data to it
+    if (!conn->fWriteOps.empty()) {
+        WriteOperation* op = conn->fWriteOps.back();
+        size_t          bytesLeft = std::min(op->fAllocSize - op->fNotify.bytes, bytes);
```
Hoikas, line 450:
```suggestion
        size_t bytesLeft = std::min(op->fAllocSize - op->fNotify.bytes, bytes);
```

errant whitespace

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+            op->fNotify.bytes += bytesLeft;
+            data = (const uint8_t*)data + bytesLeft;
+            bytes -= bytesLeft;
+        }
+    }
+
+    if (bytes) {
+        // Allocate storage alongside the operation structure itself
+        size_t          bytesAlloc = std::max(bytes, kMinBacklogBytes);
+        auto            membuf = new uint8_t[sizeof(WriteOperation) + bytesAlloc];
```
Hoikas, line 463:
```suggestion
        size_t bytesAlloc = std::max(bytes, kMinBacklogBytes);
        auto membuf = new uint8_t[sizeof(WriteOperation) + bytesAlloc];

```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+        WriteOperation* op = new (membuf) WriteOperation;
+        conn->fWriteOps.emplace_back(op);
+
+        // TODO: asyncId
```
Hoikas, line 467: Is this still `TODO`?

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
```diff
@@ -0,0 +1,545 @@
...
+        op->fAllocSize = bytesAlloc;
+        op->fNotify.asyncId = /* TODO */ nullptr;
```
Hoikas, line 469: same as above

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
```diff
@@ -52,45 +52,75 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+{
     unsigned bailAt = TimeGetMs() + exitThreadWaitMs;
     while (AsyncPerfGetCounter(kAsyncPerfThreadsCurr) && signed(bailAt - TimeGetMs()) > 0)
         std::this_thread::sleep_for(std::chrono::milliseconds(10));
 }
 
 //============================================================================
-void AsyncThreadTimedJoin(std::thread& thread, unsigned timeoutMs)
+AsyncThreadRef AsyncThreadCreate(
+    std::function<void()> threadProc)
```
Hoikas, line 88:
```suggestion
AsyncThreadRef AsyncThreadCreate(std::function<void()> threadProc)
```

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceSocket.cpp
colincornaby: Let me look. Async ID is not actually used anywhere. Let me see what it was for originally in context though.

colincornaby: So... here's my take on Async ID. And someone might know more than me.

It looks like all Async ID was used for was tracking write operations sequentially. And they went in a linked list, so it may have just been ended up using to identify what was a write operation.

The new code has a dedicated write operation vector, so we don't need the Async ID to determine what was a write operation. Beyond that - I can't find any use for Async ID - and it probably should be removed.

I'm not sure if maybe back in the day people thought this should be connected to logging? Maybe there was some diagnostic tool that used it?

colincornaby:
Hey, I know every is busy. But just a gentle nudge for review in since this is a critical branch for other things.

Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceDns.cpp
```diff
@@ -125,7 +117,7 @@ static void AddressResolved(const asio::error_code& err,
     PerfSubCounter(kAsyncPerfNameLookupAttemptsCurr, 1);
 }
 
-void AsyncAddressLookupName(FAsyncLookupProc lookupProc,
+void AsyncAddressLookupName(FAsyncLookupProc  lookupProc,
```
Hoikas, line 120:
```suggestion
void AsyncAddressLookupName(FAsyncLookupProc lookupProc,

```

Looks like a bit of spurious whitespace here.

Hoikas (approved):

Hoikas:
💥 :boom: :boom:

zrax:
🎉

PR #1206, opened 2022-07-02
Convert Cyan docs from Word to Markdown

dgelessus:
This way they can be read and (theoretically) edited more easily. No changes to the actual content, only a few adjustments where the original formatting cannot be faithfully reproduced in Markdown.

The original .doc files have modification times from 2002 (Distributor.doc) and 2001 (Resource Manager.doc), so the content is going to be a bit outdated. I don't know the engine well enough to tell *how* outdated they are exactly, but I'm assuming that the general concepts have remained somewhat similar, so these docs hopefully aren't entirely useless.

I also went ahead and deleted ~~ReleaseNotes.txt (which has no real content) and~~ "sound cones.htm" (which seems to be an old copy of a DirectX doc page that can be easily found online).


dgelessus:
Oh, the ReleaseNotes.txt is used by the build? I didn't know that - I assumed that Docs was really just documentation and not data files for the game. In that case, perhaps the ReleaseNotes.txt should be moved under Scripts instead?

Docs/Resource Manager.md
```diff
@@ -0,0 +1,308 @@
...
+* Receive from the Resource Manager a plKey that other objects will know how to request.
+* Interact with non-procedural assets normally.
+* UnRegister from the Resource Manager at any time.
+* If still registered, delete at shutdown via the normal dependency mechanism.
+
+#### Identifiers Across Source Asset Files
+
+Again because the UOID’s are generated at scene conversion time,
+there is currently no way for an object in one source Max file to reference an object in another Max file.
+The requirement is that the Age/Chapter/Page/ClassID/&lt;string identifier> of an object may be efficiently converted at run-time into a plKey for the object.
```
Hoikas, line 167:
```suggestion
The requirement is that the Age/Chapter/Page/ClassID/<string identifier> of an object may be efficiently converted at run-time into a plKey for the object.
```

Docs/Resource Manager.md
```diff
@@ -0,0 +1,308 @@
+## Resource Manager
+
+### Functional Overview
+
+The Resource Manager processes requests for assets,
+finds those assets in memory, disk cache, or from network servers,
+and notifies the requestor when that asset is available.
+It is also responsible for notifying users of an asset when that asset is no longer available.
```
Hoikas, line 8: What's the deal with the odd newline placement?

Docs/Resource Manager.md
dgelessus: Escaping the `<` as `&lt;` is necessary because Markdown supports inline HTML. Perhaps not all Markdown parsers implement this consistently, but at least GitHub and IntelliJ's parsers recognize `<string identifier>` as an HTML tag, so it becomes invisible if the `<` isn't escaped like this.

Docs/Resource Manager.md
dgelessus: For freetext documentation like this, I try to put the line breaks at the end of each sentence (and after commas if the sentence is long). Compared to always wrapping at a fixed column, this works better with version control - changing a few words doesn't require you to rewrap the entire remaining paragraph (which leads to unnecessary changes in merges and `git blame`). In the rendered output, it looks the same either way, because single newlines are ignored by Markdown.

But if you prefer, I can rewrap everything to a fixed column width. (How many columns do we use here usually?)

Docs/Resource Manager.md
Hoikas: In that case, let's escape both brackets to make it clearer that it's not an error 😄

Docs/Resource Manager.md
Hoikas: I think you have a good point. I'm not sure what the answer is here. README.md seems to break at 80 columns while CONTRIBUTING.md doesn't break, ever. Any thoughts @Deledrius, @dpogue?

dpogue: 80 columns is really nice for reading in the terminal, and &mdash; as @dgelessus mentioned &mdash; is absolutely terrible for rewording or editing anything because you have to redo all the line breaks.
These docs are mostly historical and unlikely to change, so maybe that's fine?

Having each paragraph on its own line is the other alternative I usually tend towards. Most reasonable editors will wrap it, so you can just size your editor to how wide you want it to be for comfortable reading.

Hoikas: Based on the discussion, I think this would be better if the line breaks were changed to occur around 80 columns. Thanks for converting these documents to markdown for us!

Hoikas (approved):
Thanks again for converting these documents!

PR #1207, opened 2022-07-03
Fix platform-dependent path problems in the patcher.

colincornaby:
Fixes for the patcher running on macOS (and likely Linux). The netcode linked some string lengths to MAX_PATH, but this is variable for each platform and can cause problems in network transactions.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -550,11 +550,11 @@ void pfPatcherWorker::Run()
 void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
 {
     // Check to see if ours matches
-    plFileInfo mine(file.fClientPath);
+    plFileInfo mine(file.fClientPath.Normalize());
     if (mine.FileSize() == file.fFileSize) {
-        plMD5Checksum cliMD5(file.fClientPath);
+        plMD5Checksum cliMD5(file.fClientPath.Normalize());
         if (cliMD5 == file.fChecksum) {
-            WhitelistFile(file.fClientPath, false);
+            WhitelistFile(file.fClientPath.Normalize(), false);
             return;
         }
     }
@@ -569,12 +569,12 @@ void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
 
     // If you got here, they're different and we want it.
     PatcherLogYellow("\tEnqueuing '{}'", file.fServerPath);
-    plFileSystem::CreateDir(file.fClientPath.StripFileName());
+    plFileSystem::CreateDir(file.fClientPath.Normalize().StripFileName());
 
     // If someone registered for SelfPatch notifications, then we should probably
     // let them handle the gruntwork... Otherwise, go nuts!
     if (fSelfPatch) {
-        if (file.fClientPath == plFileSystem::GetCurrentAppPath().GetFileName()) {
+        if (file.fClientPath.Normalize() == plFileSystem::GetCurrentAppPath().GetFileName()) {
             file.fClientPath += ".tmp"; // don't overwrite myself!
             file.fFlags |= kSelfPatch;
         }
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -124,7 +124,7 @@ struct pfPatcherQueuedFile
     uint32_t fFlags;
 
     pfPatcherQueuedFile(Type t, const NetCliFileManifestEntry& file)
-        : fType(t), fClientPath(ST::string::from_utf16(file.clientName)),
+    : fType(t), fClientPath(plFileName(ST::string::from_utf16(file.clientName)).Normalize()),
           fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(),
           fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
     {
@@ -550,11 +550,11 @@ void pfPatcherWorker::Run()
 void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
 {
     // Check to see if ours matches
-    plFileInfo mine(file.fClientPath.Normalize());
+    plFileInfo mine(file.fClientPath);
     if (mine.FileSize() == file.fFileSize) {
-        plMD5Checksum cliMD5(file.fClientPath.Normalize());
+        plMD5Checksum cliMD5(file.fClientPath);
         if (cliMD5 == file.fChecksum) {
-            WhitelistFile(file.fClientPath.Normalize(), false);
+            WhitelistFile(file.fClientPath, false);
             return;
         }
     }
@@ -569,12 +569,12 @@ void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
 
     // If you got here, they're different and we want it.
     PatcherLogYellow("\tEnqueuing '{}'", file.fServerPath);
-    plFileSystem::CreateDir(file.fClientPath.Normalize().StripFileName());
+    plFileSystem::CreateDir(file.fClientPath.StripFileName());
 
     // If someone registered for SelfPatch notifications, then we should probably
     // let them handle the gruntwork... Otherwise, go nuts!
     if (fSelfPatch) {
-        if (file.fClientPath.Normalize() == plFileSystem::GetCurrentAppPath().GetFileName()) {
+        if (file.fClientPath == plFileSystem::GetCurrentAppPath().GetFileName()) {
             file.fClientPath += ".tmp"; // don't overwrite myself!
             file.fFlags |= kSelfPatch;
         }
--- a/Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
+++ b/Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
@@ -49,15 +49,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *
 ***/
 
-// FIXME: Remove all Windows assumptions in the net code
-#ifndef MAX_PATH
-#   define MAX_PATH 260
-#endif
-
 //============================================================================
 // Network constants
 //============================================================================
 const unsigned kMaxTcpPacketSize                = 1460;
+const unsigned kCli2File_FilenameSize           = 260;
 
 //============================================================================
 // Crypto constants
--- a/Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
+++ b/Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
@@ -53,7 +53,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 // Network constants
 //============================================================================
 const unsigned kMaxTcpPacketSize                = 1460;
-const unsigned kCli2File_FilenameSize           = 260;
+const unsigned kNetDefaultStringSize            = 260;
 
 //============================================================================
 // Crypto constants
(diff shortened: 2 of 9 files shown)


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -368,7 +368,6 @@ int hsMessageBoxWithOwner(hsWindowHndl owner, const wchar_t* message, const wcha
 #   define hsWFopen(name, mode)     fopen(hsWStringToString(name), hsWStringToString(mode))
 
 #   include <limits.h>
-#   define MAX_PATH PATH_MAX
```
colincornaby, line 371: We should actually get long path support for local files in, but for now having a working patcher seems like the bigger bug.

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -569,12 +569,12 @@ void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
...
     // If you got here, they're different and we want it.
     PatcherLogYellow("\tEnqueuing '{}'", file.fServerPath);
-    plFileSystem::CreateDir(file.fClientPath.StripFileName());
+    plFileSystem::CreateDir(file.fClientPath.Normalize().StripFileName());
 
     // If someone registered for SelfPatch notifications, then we should probably
     // let them handle the gruntwork... Otherwise, go nuts!
     if (fSelfPatch) {
-        if (file.fClientPath == plFileSystem::GetCurrentAppPath().GetFileName()) {
+        if (file.fClientPath.Normalize() == plFileSystem::GetCurrentAppPath().GetFileName()) {
```
colincornaby, line 577: This is probably the wrong way to do this. I wasn't entirely sure which functions needed the Windows path and which ones need the local path. I'm getting more comfortable with this code - but am intentionally leaving this messy so I can get feedback on which path should go where.

I considered just editing these functions to normalize paths but that could be risky or extra overhead where it isn't needed. But leaves open the possibility that other code paths may use Windows paths on a Mac or Linux host.

Sources/Plasma/CoreLib/HeadSpin.h
dpogue: My preference is definitely to deal with this at the network protocol level and not at the HeadSpin.h level. What we're dealing with is an implementation requirement of the network protocol, not an issue with file paths overall.

We've already got some weird implementation requirements for the network protocol, like using `char16_t` for strings. Let's just define a `kCli2File_FilenameSize = 260` constant in NucleusLib/pnNetProtocol/Private/Protocols/Cli2File/pnNpCli2File.h

colincornaby: It looks like pnNbConst.h already has a FIXME around this behavior.

I'm fine splitting things but MAX_PATH is also used a lot of places so it would have to be a careful audit. The MAX_PATH define maybe should also be moved out of pnNbConst.h for Windows as well and into the Headspin header. Then a proper constant for all platforms could be defined in pnNbConst.h instead.

dpogue: `MAX_PATH` is always defined on Windows by system headers, so that one in plNbConst.h should only be affecting Linux/macOS

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
Hoikas: I would prefer for te queued file to store a normalized copy of fClientPath. Needing to explicitly normalize something already called a "client path" is surprising and therefore bugprone.

colincornaby: Done. Lemme know if it's sufficiently modern C++. I don't know any tricks besides brute forcing the plFileName constructor inside of the file constructor to get to the Normalize function.

Sources/Plasma/CoreLib/HeadSpin.h
colincornaby: I'm going to try to get the network code shifted over to it's own constant. I'll roll that in later today.

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
Hoikas: The "modern C++" way would probably be to change `fClientPath` to a `plFileName`, but that's outside the scope of what you're trying to accomplish, I think 😄

Sources/Plasma/CoreLib/HeadSpin.h
colincornaby: Ok, the network code is changed over to a kCli2File_FilenameSize constant. I did a one by one find/replace in all network classes to move it over to the new constant. This work should probably be double checked.

Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
```diff
@@ -49,15 +49,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *
 ***/
 
-// FIXME: Remove all Windows assumptions in the net code
-#ifndef MAX_PATH
-#   define MAX_PATH 260
-#endif
-
```
colincornaby, line 56: Tossed this. This define conflicted with the other one, and we don't use MAX_PATH in the netcode anymore.

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -380,7 +380,9 @@ constexpr float hsInvert(float a) { return 1.f / a; }
 #ifdef _MSC_VER
 #   define ALIGN(n) __declspec(align(n))
 #else
-#   define ALIGN(n) __attribute__((aligned(n)))
+#   ifndef ALIGN
+#       define ALIGN(n) __attribute__((aligned(n)))
+#   endif
```
colincornaby, line 385: Only my Mac it's saying ALIGN is already defined. Only redefining it if it isn't defined already.

colincornaby:
The build is failing on something that seems to be related to the 3DS Max plugin. Not quite sure.

Sources/Plasma/CoreLib/HeadSpin.h
dpogue: This seems like it's going to be a problem... the existing `ALIGN` macro in param.h on macOS does not set the alignment attribute...

```cpp
// param.h
#define ALIGN(p)        __DARWIN_ALIGN(p)
```

```cpp
// _param.h
#define __DARWIN_ALIGNBYTES     (sizeof(__darwin_size_t) - 1)
#define __DARWIN_ALIGN(p)       ((__darwin_size_t)((__darwin_size_t)(p) + __DARWIN_ALIGNBYTES) &~ __DARWIN_ALIGNBYTES)
```

We probably (in another PR) want to rename `ALIGN` to something like `HS_ALIGNED`

Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
```diff
@@ -49,15 +49,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 //============================================================================
 // Network constants
 //============================================================================
 const unsigned kMaxTcpPacketSize                = 1460;
+const unsigned kCli2File_FilenameSize           = 260;
```
dpogue, line 56: I didn't realize MAX_PATH was being used for a bunch of string outside of File Server stuff... would `kNetDefaultStringSize` be a better name?

(Would it have killed them to use Pascal-style strings in their network messages? This is so wasteful 🤦🏼)

Sources/Plasma/CoreLib/HeadSpin.h
dgelessus: Wouldn't it be better to switch to C++11 standard [`alignas`](https://en.cppreference.com/w/cpp/language/alignas) instead?

dpogue:
The CI failure is due to a missed MAX_PATH:
NucleusLib\pnNetProtocol\Private\Protocols\Cli2Auth\pnNpCli2Auth.cpp(159,5): error C2065: 'MAX_PATH': undeclared identifier

Sources/Plasma/CoreLib/HeadSpin.h
dpogue: yeah, that sounds like a good solution :+1:

Sources/Plasma/NucleusLib/pnNetProtocol/Private/Protocols/Cli2Auth/pnNpCli2Auth.cpp
```diff
@@ -149,9 +149,9 @@ static const NetMsgField kAcctActivateRequestFields[] = {
 };
 
 static const NetMsgField kFileListRequestFields[] = {
-    kNetMsgFieldTransId,                // transId
-    NET_MSG_FIELD_STRING(MAX_PATH),     // directory
-    NET_MSG_FIELD_STRING(MAX_EXT),      // ext
```
dgelessus, line 154: This is not the same as `MAX_PATH` btw - `MAX_EXT` is defined to 256, not 260. Also, `MAX_EXT` comes from Plasma itself and not Windows, so it's less of a problem.

Sources/Plasma/NucleusLib/pnNetProtocol/Private/Protocols/Cli2Auth/pnNpCli2Auth.cpp
```diff
@@ -149,9 +149,9 @@ static const NetMsgField kAcctActivateRequestFields[] = {
...
+    kNetMsgFieldTransId,                                // transId
+    NET_MSG_FIELD_STRING(kCli2File_FilenameSize),       // directory
+    NET_MSG_FIELD_STRING(kCli2File_FilenameSize),       // ext
```
dgelessus, line 154: This is not the same as `MAX_PATH` btw - `MAX_EXT` is defined to 256, not 260. Also, `MAX_EXT` comes from Plasma itself and not Windows, so it's less of a problem.

colincornaby: Fixed

Sources/Plasma/CoreLib/HeadSpin.h
colincornaby: Is the ALIGN macro even used anywhere? I did a quick text search but couldn't find anything.

dpogue: I'm just looking at this now, and it looks like the only user is plDXPipeline for setting up SSE skinning. So that code maybe got copied into plGLPipeline and plMetalPipeline, but should be easy enough to clean up

colincornaby: Metal is using Apple's Accelerate library for that math, so that wouldn't have impacted Mac. But that explains why my Xcode project isn't seeing it.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -8871,8 +8871,8 @@ static void IBlendVertBuffer(plSpan* span, hsMatrix44* matrixPalette, int numMat
         src = inlExtract<hsVector3>(src, vec);
 
         // Destination buffers (float4 for SSE alignment)
-        ALIGN(16) float destNorm_buf[] = { 0.f, 0.f, 0.f, 0.f };
-        ALIGN(16) float destPt_buf[] = { 0.f, 0.f, 0.f, 1.f };
+        alignas(16) float destNorm_buf[] = { 0.f, 0.f, 0.f, 0.f };
+        alignas(16) float destPt_buf[] = { 0.f, 0.f, 0.f, 1.f };
```
colincornaby, line 8875: It's building now, and I'm pretty sure this is the right syntax. This work should be double checked though.

Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
colincornaby: Probably would be a better name. It is not great that the Windows maximum path length got baked into so many things.

Hoikas:
Once you're happy with this, I'd appreciate separating the concerns of the align macro and the path fix into separate PRs with good descriptions. Otherwise, we'll probably be weeping in 15 years when we want to try to understand the history behind these fixes 😉

colincornaby:
Ok, I've renamed the constant to kNetDefaultStringSize per @dpogue's suggestion, and pushed the align stuff out into it's own PR.

dpogue (approved):
Looks good to me

Hoikas (approved):

PR #1213, opened 2022-07-07
Update gtest to 1.12.1

dpogue:
https://github.com/google/googletest/releases/tag/release-1.12.1


dpogue:
I was hoping this would get rid of all those CMake policy warnings, but apparently not 🙃

Hoikas (approved):

PR #1218, opened 2022-07-10
Fixing plSplineEaseCurve asserts on Mac

colincornaby:
This one is gonna be a bit weird. macOS (and maybe Linux in the future) is tripping the assert on 331 repeatedly. There is a key difference between macOS and Windows on the line I changed.

M_PI as a float is 3.14159274. hsConstants::pi<float> is also 3.14159274.

acos(-1.0f) should be Pi. And on Windows acos(-1.0f) is:
3.14159274.

Makes sense. But on macOS, acos(-1.0f) is instead:
3.1415925

I don't know why. It's a big enough change it moves around this function and trips the assert.

Basically the best solution seems to be to force the acos to be done as a double, and then downcast that into a float. That's the same way hsConstants::pi<float> converts it's value down to float.

`(float)acos(double(R / sqrt(Q3)))` properly resolves to 3.14159274. Even though it's stored in a double, we need the actual values precision to be consistent for things to line up with the float values.

I kind of know roughly what this function is for - but I'm not familiar with the algorithm. I tried increasing the precision of more things to double and it only made the assert trip more. Perhaps with a more careful set of changes to this code everything could be made consistent and we wouldn't have to worry about weird float and double inconsistencies.

I also don't know why macOS is different. I looked at a whole bunch of compile flags, but I don't know why it's output is different.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plInterp/plATCEaseCurves.cpp
+++ b/Sources/Plasma/PubUtilLib/plInterp/plATCEaseCurves.cpp
@@ -305,7 +305,7 @@ float plSplineEaseCurve::TimeGivenVelocity(float velocity) const
     if (D >= 0) 
     {   
         // 3 roots, find the one in the range [0, 1]
-        double theta = acos(R / sqrt(Q3));
+        double theta = (float)acos(double(R / sqrt(Q3)));
         double sqrtQ = sqrt(Q);
 
         root = (float)(-2 * sqrtQ * cos((theta + 4 * hsConstants::pi<float>) / 3) - c / 3); // Middle root, most likely to match


dpogue (approved):
This seems fine to me... weird, but fine

colincornaby:
> This seems fine to me... weird, but fine

I was hoping someone here would make this feel less weird to me but so far no such luck!

Hoikas:
That's why I haven't commented, but nothing is really coming to mind.

colincornaby:
Reviewing this again - I still don't have a great explanation on why Clang is acting the way it is. It seems like a straightforward operation. I'm almost curious enough to open up a stack overflow and see if anyone can come up with a better explanation.

Deledrius:
That's not a bad idea!

colincornaby:
Ok, I asked Stack Overflow and got a lot of replies.

https://stackoverflow.com/questions/73202732/slight-acos-precision-difference-between-clang-and-visual-c?noredirect=1#comment129287864_73202732

I don't think anything was definitive but the issue might be better understood now.

Visual C++'s output is more accurate. But technically acos is never supposed to return a value great than Pi. The value Visual C++ is returning is greater than Pi.

Clang's return value is the largest float value that is not larger than Pi. So while Clang is less accurate - it's within spec.

This change is basically re-forcing the closest behavior by doing the acos operation on a double, and then casting to a float, which is rounded to the closest float value.

Long term - I'm not sure what to do with this function. It's very sensitive to a particular acos behavior that debatably is not correct. I think this was based on another GPU Gems algorithm - but I can't easily find the original reference. Online it seems like there are alternative approaches to this algorithm, but I don't understand the algorithm deeply enough to implement them yet.

I also looked at setting compiler rounding modes - but I don't recall that solving anything. Likely because the rounding is already baked inside of the Math library.

I feel like this change is probably better understood - although I'm less comfortable with the way this function is implemented in general. It seems like it's relying on observed behaviors instead of guaranteed behaviors.

Hoikas:
Are we all happy with this and ready to merge?

dpogue:
I think we're all slightly unhappy with this, but ready to merge 😛

colincornaby:
I'm slightly unhappy with the algorithm in general - but I think it's understood what this fix does and why it works. Replacing this algorithm might be a longer term project.

Hoikas:
Ok, so let's get this incremental improvement in and worry about the algorithm itself later.

Deledrius:
Might be a good idea to open an Issue on this topic so the remaining work isn't forgotten.

PR #1221, opened 2022-07-12
Mac/Quartz version of plTextFont

colincornaby:
Initial work. Rendering well - might need some extremely fine tuning in the future.

I would totally be open to this being a FreeType implementation in the future. Unfortunately, I do not know FreeType. So for now we get a CoreText/Quartz implementation. 😀 This will get the Mac working until something more properly cross platform can be done.

This is also such a small chunk of code it may be fine to keep in platform specific versions. But there might be greater questions about the future of text in Plasma that go into a FreeType sort of decision.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
+++ b/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
@@ -231,6 +231,7 @@ uint16_t  *plTextFont::IInitFontTexture()
     
     bitmapBits = (uint32_t *)CGBitmapContextGetData(bitmapContext);
 
+    //TEMPORARY: Mac client should vend through better data to make this decision
     nHeight = (fSize * 192.0f) / 72.0f;
     fFontHeight = nHeight;
 
--- a/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
+++ b/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
@@ -57,6 +57,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plDebugText.h"
 
+#if defined(HS_BUILD_FOR_MACOS)
+#import <CoreGraphics/CoreGraphics.h>
+#import <CoreText/CoreText.h>
+#endif
+
 #define DisplayableChar(c) (c >= 0 && c <= 128)
 
 //// Constructor & Destructor /////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -191,6 +196,117 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+
+    // Special case the tab key
+    fCharInfo[ '\t' ].fUVs[ 1 ].fX = fCharInfo[ '\t' ].fUVs[ 0 ].fX = fCharInfo[ 32 ].fUVs[ 0 ].fX;
+    fCharInfo[ '\t' ].fUVs[ 1 ].fY = fCharInfo[ '\t' ].fUVs[ 0 ].fY = 0;
+    fCharInfo[ '\t' ].fUVs[ 0 ].fZ = fCharInfo[ '\t' ].fUVs[ 1 ].fZ = 0;
+    fCharInfo[ '\t' ].fW = fCharInfo[ 32 ].fW * 4;
+    fCharInfo[ '\t' ].fH = fCharInfo[ 32 ].fH;
+
+    /// Now create the data block
+    uint16_t  *data = new uint16_t[ fTextureWidth * fTextureHeight ];
```
colincornaby, line 289: This code outputs a 4-bits-per-channel bitmap, and the conversion is here.

Unfortunately, Metal requires a minimum of 8 bits per channel on most hardware. So I immediately up convert it again in the Metal renderer.

Tempted to just handle it as a single channel 16 bit image in it writes the same value across all channels.

I'm also not sure what bAlpha is doing.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -191,6 +196,117 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+                                                colorSpace,
+                                                kCGImageAlphaPremultipliedLast);
+    CFRelease(colorSpace);
+    CGContextSetTextMatrix(bitmapContext, CGAffineTransformMakeScale(1, -1));
+    CGContextTranslateCTM(bitmapContext, 0, fTextureHeight);
+    CGContextScaleCTM(bitmapContext, 1, -1);
+    
+    bitmapBits = (uint32_t *)CGBitmapContextGetData(bitmapContext);
+
+    nHeight = (fSize * 192.0f) / 72.0f;
```
colincornaby, line 234: I remembered I hardcoded a DPI of 192. I probably should not do that.

- [ ] Fix DPI check

colincornaby: I'm going to leave this at what it's been hardcoded to for now. I think better integration could be done once the actual Mac client is in place. I need to really pass the display device through the pipeline - which the pipeline can do. Just not without the Mac client in place.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -191,6 +196,118 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+        fTextureWidth = fTextureHeight = 1024;
+    else if( fSize > 20 )
+        fTextureWidth = fTextureHeight = 512;
+    else
+        fTextureWidth = fTextureHeight = 256;
+
+
+    // Create a new DC and bitmap that we can draw characters to
+    CGColorSpaceRef colorSpace = CGColorSpaceCreateWithName(kCGColorSpaceGenericRGB);
+    CGContextRef bitmapContext = CGBitmapContextCreate(NULL,
```
Hoikas, line 220:
```suggestion
    CGContextRef bitmapContext = CGBitmapContextCreate(nullptr,
```

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -57,6 +57,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plDebugText.h"
 
+#if defined(HS_BUILD_FOR_MACOS)
```
Hoikas, line 60: Hmm... I thought this define was changed to `HS_BUILD_FOR_APPLE`, or am I misremembering things?

dpogue: `HS_BUILD_FOR_APPLE` for stuff that's Apple-y, `HS_BUILD_FOR_MACOS` for stuff that is strictly macOS.

CoreGraphics, I _think_ would fall under Apple, since it's also available on iOS?

colincornaby: Yeah, you're right. If we ever did an iOS version (which I'd love to at least hack on someday), this code would work. I'll update the check.

Hoikas (commented):
This LGTM. I'd like to have another concurring review as well before merging just because I'm somewhat alseep at the wheel.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
dpogue: This macro is still checking `HS_BUILD_FOR_MACOS` instead of `HS_BUILD_FOR_APPLE`

colincornaby:
> This looks fine to me.
> 
> 
> 
> Admittedly, I'm not at all familiar with the CoreText API, so I can't judge whether that code is "correct", but it seems fine to me.

Long term it might need some fine tuning, but it's getting output. I'm kind of assuming these platform implementations might get replaced anyway when Linux enters the picture. But this is mostly just to get something intelligible out.

PR #1217, opened 2022-07-10
Restore support for GameMgr-based marker games.

Hoikas:
This re-adds support for the game manager to our codebase in a way that allows the client to use the "new" style marker games, garden doors, etc. on servers that do not support the game manager and falls back to the game managers on servers that do (eg MOULa). In this way, our client can be more safely used by individuals who do not understand the implications of our changes.

This pull request only addresses the core game manager functionality and marker games. Per discussion in IRC, it seems the desired approach is to perform one pull request per game type.

Depends on #1179, so go review that.

PR #1225, opened 2022-07-30
Add object library support.

Hoikas:
The benefit to this change is that for IDE-based generators (eg Visual Studio), you can now see the headers in a dedicated project instead of either having ghost headers that appear only as includes (or, worse, the headers duplicated into all consuming projects).

Required by #1217.

### DIFF
--- a/Sources/Plasma/FeatureLib/inc/CMakeLists.txt
+++ b/Sources/Plasma/FeatureLib/inc/CMakeLists.txt
@@ -1,9 +1,9 @@
-# Because there are no CPP sources, this needs to be an interface library. Change to
-# a static library if you ever add any real C++ code.
-add_library(pfFeatureInc INTERFACE)
-set_property(
-    TARGET pfFeatureInc
-    PROPERTY INTERFACE_LINK_LIBRARIES
+set(pfFeatureInc_HEADERS
+    pfAllCreatables.h
+)
+plasma_library(pfFeatureInc OBJECT SOURCES ${pfFeatureInc_HEADERS})
+target_link_libraries(pfFeatureInc
+    PUBLIC
         pfAnimation
         pfAudio
         pfCamera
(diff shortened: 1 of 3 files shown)


Sources/Plasma/PubUtilLib/inc/CMakeLists.txt
```diff
@@ -33,7 +34,4 @@ set_property(
         plSurface
         plVault
 )
-set_property(
-    TARGET plPubUtilInc
-    PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_LIST_DIR}"
-)
+target_include_directories(plPubUtilInc INTERFACE "${CMAKE_CURRENT_LIST_DIR}")
```
dpogue, line 37: Do we want `source_group("Header Files" FILES ${plPubUtilInt_HEADERS})` for folders in VS & Xcode?

Sources/Plasma/FeatureLib/inc/CMakeLists.txt
```diff
@@ -17,7 +17,4 @@ set_property(
         pfPython
         pfSurface
 )
-set_property(
-    TARGET pfFeatureInc
-    PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_LIST_DIR}"
-)
+target_include_directories(pfFeatureInc INTERFACE "${CMAKE_CURRENT_LIST_DIR}")
```
dpogue, line 20: Same question about `source_group` here

Sources/Plasma/PubUtilLib/inc/CMakeLists.txt
Hoikas: CMake automatically puts headers into that group if not groups are specified.

image

dpogue: Well that's not obvious behaviour, but okay...

dpogue (approved):

PR #1231, opened 2022-08-09
Fix for plFileSystem not finding self on Mac

colincornaby:
The code as written should have worked - but Apple seems to have changed the way they do things. There are other reports of /proc/curproc/file disappearing from macOS. I'm not sure when that happened.

This code gets called as part of the patching process - which is why I wrote this change. Patching did not work without this change.

I defined a new implementation - but I realized this will only work if the current process is an app bundle. For a flat executable (like your command line patcher @dpogue), this probably isn't going to work. I'm open to other suggestions - this take is very AppKit centric.

### DIFF
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -56,8 +56,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   include <fnmatch.h>
 #   include <dirent.h>
 #   include <limits.h>
-#   include <sys/types.h>
 #   include <sys/param.h>
+#   include <sys/types.h>
 #   include <unistd.h>
 #endif
 
@@ -68,9 +68,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #     endif
 #   endif
 #
-#   include <NSSystemDirectories.h>
 #   include <CoreFoundation/CoreFoundation.h>
 #   include <mach-o/dyld.h>
+#   include <NSSystemDirectories.h>
 #endif
 
 #include <sys/stat.h>
@@ -572,11 +572,11 @@ plFileName plFileSystem::GetCurrentAppPath()
     return appPath;
 #elif HS_BUILD_FOR_MACOS
     CFBundleRef myBundle = CFBundleGetMainBundle();
-    if(!myBundle) {
+    if (!myBundle) {
         char path[MAXPATHLEN];
         uint32_t pathLen = MAXPATHLEN;
         _NSGetExecutablePath(path, &pathLen);
-        appPath = ST::string::from_utf8(path);
+        appPath = ST::string::from_utf8(path, pathLen);
     } else {
         CFURLRef url = CFBundleCopyBundleURL(myBundle);
         CFStringRef path = CFURLCopyPath(url);
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -68,6 +68,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   endif
 #
 #   include <NSSystemDirectories.h>
+#   include <CoreFoundation/CoreFoundation.h>
 #endif
 
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


dpogue:
This returns the path to the bundle, so I'm guessing that's the `.app` "file" and not the actual executable within it?

dpogue:
How cursed is `_NSGetExecutablePath`? https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html

colincornaby:
> How cursed is `_NSGetExecutablePath`? https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html

Seems workable to me, I wasn't aware it existed.

I'm not sure if it should be the executable path or the bundle path in the case of the Mac client. That might be a question we're not ready for. In since Mac apps are folders that have to be replaced completely in order to not break code signing - we probably don't want the patcher crawling into the bundle. So we'd have to get the patcher to treat the bundle like it was a file, which might require some rethink and extensions to the patcher. I don't know if we could even do something fancy like store the app as a zipfile on the server and have the patcher decompress it into place.

dpogue:
> I'm not sure if it should be the executable path or the bundle path in the case of the Mac client.

It _might_ make sense to use the bundle path if it's a bundle, and the executable path if not?

colincornaby:
> > I'm not sure if it should be the executable path or the bundle path in the case of the Mac client.
> 
> It _might_ make sense to use the bundle path if it's a bundle, and the executable path if not?

That could be a good path forward. I can implement that. If it doesn't work out when it comes time to actually do patching of the Mac client that could always be changed.

Hoikas:
That sounds like a good path forward to me.

colincornaby:
I just remembered this one is still open. I'll get back around to implementing the changes.

colincornaby:
FYI - this code is no longer being called in the latest version of the Mac patcher. Likely because I've opted out of self patching. But that means it'll likely come back into use. I manually called the function for testing.

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -566,6 +569,24 @@ plFileName plFileSystem::GetCurrentAppPath()
         appPath = ST::string::from_wchar(path);
     }
 
+    return appPath;
+#elif HS_BUILD_FOR_MACOS
+    //Not sure how we'd ever patch the executable on iOS,
+    //this is really Mac specific
+    //This will only work if this is an app bundle
```
dpogue, line 576: This comment doesn't seem accurate anymore?

colincornaby: Yep, removed!

dpogue (approved):
LGTM

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -57,6 +57,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   include <dirent.h>
 #   include <limits.h>
 #   include <sys/types.h>
+#   include <sys/param.h>
```
Hoikas, line 60: Alphabetize

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -68,6 +69,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   endif
 #
 #   include <NSSystemDirectories.h>
+#   include <CoreFoundation/CoreFoundation.h>
+#   include <mach-o/dyld.h>
```
Hoikas, line 73: Alphabetize

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -566,6 +569,21 @@ plFileName plFileSystem::GetCurrentAppPath()
         appPath = ST::string::from_wchar(path);
     }
 
+    return appPath;
+#elif HS_BUILD_FOR_MACOS
+    CFBundleRef myBundle = CFBundleGetMainBundle();
+    if(!myBundle) {
```
Hoikas, line 575:
```suggestion
    if (!myBundle) {
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -566,6 +569,21 @@ plFileName plFileSystem::GetCurrentAppPath()
...
+        char path[MAXPATHLEN];
+        uint32_t pathLen = MAXPATHLEN;
+        _NSGetExecutablePath(path, &pathLen);
+        appPath = ST::string::from_utf8(path);
```
Hoikas, line 579:
```suggestion
        appPath = ST::string::from_utf8(path, pathLen);
```

colincornaby:
Code suggestions have been applied. Waiting to see if they build on CI in since it's difficult for me to build these on Mac without a full client exercising things.

Hoikas (approved):

Hoikas:
@colincornaby It looks good to me - are you ready for a merge?

colincornaby:
Yep! This is ready for merge.

PR #1233, opened 2022-08-14
Fixing chat crash on macOS (and likely Linux)

colincornaby:
Long is 8 byte outside of Windows, not 4. Hardcoded uint32_t is causing pointer corruption on the stack because Python expects an 8 byte type.

I did a quick check and didn't see any other issues that looked like this in cyMiscGlue - doesn't mean they aren't lurking.

Intel has a guide for long size here:
https://www.intel.com/content/www/us/en/developer/articles/technical/size-of-long-integer-type-on-different-architecture-and-os.html

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue.cpp
@@ -232,10 +232,10 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtSendRTChat, args, "Params: fromPlayer,toPlayer
     PyObject* fromPlayerObj = nullptr;
     PyObject* toPlayerListObj = nullptr;
     ST::string message;
-    long msgFlags = 0;
+    unsigned int msgFlags = 0;
     const char* err = "PtSendRTChat expects a ptPlayer, a sequence of ptPlayers, a string, and an optional long";
 
-    if (!PyArg_ParseTuple(args, "OOO&|l", &fromPlayerObj, &toPlayerListObj,
+    if (!PyArg_ParseTuple(args, "OOO&|I", &fromPlayerObj, &toPlayerListObj,
                           PyUnicode_STStringConverter, &message, &msgFlags))
     {
         PyErr_SetString(PyExc_TypeError, err);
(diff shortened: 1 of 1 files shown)


colincornaby:
I wrote bit in the comment, I meant byte. I'll update the commit notes.

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue.cpp
```diff
@@ -232,7 +232,7 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtSendRTChat, args, "Params: fromPlayer,toPlayer
     PyObject* fromPlayerObj = nullptr;
     PyObject* toPlayerListObj = nullptr;
     ST::string message;
-    uint32_t msgFlags = 0;
+    long msgFlags = 0;
     const char* err = "PtSendRTChat expects a ptPlayer, a sequence of ptPlayers, a string, and an optional long";
 
     if (!PyArg_ParseTuple(args, "OOO&|l", &fromPlayerObj, &toPlayerListObj,
```
Hoikas, line 238:
```suggestion
    unsigned int msgFlags = 0;
    const char* err = "PtSendRTChat expects a ptPlayer, a sequence of ptPlayers, a string, and an optional long";

    if (!PyArg_ParseTuple(args, "OOO&|I", &fromPlayerObj, &toPlayerListObj,
```

`cyMisc::SendRTChat()` accepts a `uint32_t` for the flags field, so it would be better to instruct Python to use the correct type. In this case, `unsigned int` is the best match, so update the initial declaration and the format string.

colincornaby: I'll apply the suggestion. Lemme validate the change on macOS real quick afterwards.

Hoikas:
It is likely, IMO, that there are additional problems like this lurking in the Python glue code. We'll have to keep our eyes out until the glue is updated to use METH_FAST as proposed in #1179, which will implicitly fix all problems due to poor type safety.

colincornaby:
The suggested change is also working on macOS

Hoikas (approved):

PR #1236, opened 2022-08-17
Fix macOS compiler errors when sysdir.h is not available

dpogue:
Technically, the issue here is the `__builtin_available` on compilers that don't support it (such as gcc), but we can effectively mask it by only running the `__builtin_available` check if we actually need to (when sysdir.h is found).

### DIFF
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -60,12 +60,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   include <unistd.h>
 #endif
 
-#if HS_BUILD_FOR_APPLE
-#   if defined(__has_include)
-#     if __has_include(<sysdir.h>)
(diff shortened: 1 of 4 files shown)


Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -64,6 +64,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #   if defined(__has_include)
 #     if __has_include(<sysdir.h>)
 #       include <sysdir.h>
+#       define HAS_SYSDIR
```
Hoikas, line 67: CoreLib is using unity build, so this define may leak out into other TUs. It might be better to move the sysdir detection out to CMake or test for if `__builtin_available()` functions in CMake.

dpogue: Testing for `__builtin_available()` with CMake is probably the right approach

Hoikas (approved):

PR #1237, opened 2022-08-17
Try to detect EFX support at compile time

dpogue:
This should fix compiling with the OpenAL.framework on macOS that doesn't include the efx.h header file. We always encourage the use of openal-soft instead of system OpenAL versions, but this allows for less dependency wrangling and support for platforms where openal-soft isn't available.


Deledrius (approved):

cmake/FindOpenAL.cmake
```diff
@@ -42,4 +42,10 @@ if(TARGET OpenAL::OpenAL)
             MAP_IMPORTED_CONFIG_RELWITHDEBINFO Release
         )
     endif()
+
+    get_target_property(_INCLUDE_DIRS OpenAL::OpenAL INTERFACE_INCLUDE_DIRECTORIES)
+    find_file(_EFX_H NAMES efx.h PATHS ${_INCLUDE_DIRS} NO_CACHE NO_DEFAULT_PATH)
+    if(_EFX_H)
+        set(OPENAL_HAS_EFX TRUE CACHE BOOL "Whether OpenAL includes EFX support")
+    endif()
```
Hoikas, line 50: This should use [CheckIncludeFile](https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html) to conform with modern CMake standards.

```suggestion
    include(CheckIncludeFile)
    include(CMakePushCheckState)
    cmake_push_check_state()
    get_target_property(CMAKE_REQUIRED_INCLUDES OpenAL::OpenAL INTERFACE_INCLUDE_DIRECTORIES)
    check_include_file(efx.h OPENAL_HAS_EFX)
    cmake_pop_check_state()
```

Hoikas: On second thought: this is probably a better suggestion:

```suggestion
    include(CheckIncludeFile)
    include(CMakePushCheckState)
    cmake_push_check_state()
    set(CMAKE_REQUIRED_LIBRARIES OpenAL::OpenAL)
    check_include_file(efx.h OPENAL_HAS_EFX)
    cmake_pop_check_state()
```

Hoikas (approved):

PR #1238, opened 2022-08-18
Add xPodium.py script

DoobesURU:
Adds a global version of the `krelPodium.py` script for use in future Ages. Modified version of Cyan's and @Hazado 's work.


Scripts/Python/xPodium.py
```diff
@@ -0,0 +1,105 @@
...
+respButtonOneshot = ptAttribResponder(2, "resp: Push Podium oneshot")
+respList = ptAttribResponderList(3, "ResponderList", byObject=1)
+stringFormat = ptAttribString(4, "Responder name format string")
+stringVarName = ptAttribString(5, "Age SDL variable name")
+
+class xPodium(ptResponder):
+
+    def __init__(self):
+        ptResponder.__init__(self)
+        self.id = 5245
```
ZarothYe, line 56: This `self.id` value needs to be updated so it doesn't collide with `krelPodium.py`. I think it would be nice if `krelPodium.py` actually used this new script rather than having two nearly-identical script files... which would also make the `id` collision moot. ;)

Scripts/Python/xPodium.py
```diff
@@ -0,0 +1,105 @@
...
+
+        elif id == respButtonOneshot.id and self.sceneobject.isLocallyOwned():
+            xPodiumSpeech -= 1
+            if xPodiumSpeech < 0:
+                xPodiumSpeech = len(respList.byObject) - 1
+            respName = (stringFormat.value % xPodiumSpeech)
+            if respName in respList.byObject:
+                ageSDL[stringVarName.value] = (xPodiumSpeech,)
+            else:
+                PtDebugPrint("ERROR: chisoPodium Invalid speech %d selected!" % xPodiumSpeech)
```
ZarothYe, line 92:
```suggestion
                PtDebugPrint(f"ERROR: xPodium Invalid speech {xPodiumSpeech} selected!")
```
f-strings for great justice! Also missed the generic naming in this message.

Scripts/Python/xPodium.py
Hazado: Would need to be updated in the hood as well, since it uses it for the podium there.

Scripts/Python/xPodium.py
DoobesURU: Oops! looks like I missed that one. Can ya tell I was playing around with the original a bit? 😝

Scripts/Python/xPodium.py
DoobesURU: > Would need to be updated in the hood as well, since it uses it for the podium there.

Hopefully it'd only just require updating the Python mod a bit...as in pointing it to xPodium instead of krelPodium and adding the SDL string.

DoobesURU:
Pushed the fixes that @ZarothYe suggested. Do we have a list of `self.id`s for the scripts to avoid future overlap or...?

Hoikas:
Not at this time. I think it would be best to pair this PR with a removal of krelPodium.py as @ZarothYe suggested so we don't have any weird duplication.

DoobesURU:
> Not at this time. I think it would be best to pair this PR with a removal of krelPodium.py as @ZarothYe suggested so we don't have any weird duplication.

That's certainly doable. We'd only need to slightly modify the Kirel Python Mod and add a small Kirel asset update to complete the removal, which I can certainly do.

Scripts/Python/xPodium.py
```diff
@@ -48,27 +47,32 @@
...
-class krelPodium(ptResponder):
+class xPodium(ptResponder):
 
     def __init__(self):
         ptResponder.__init__(self)
         self.id = 5245
         version = 2
         self.version = version
-        PtDebugPrint("__init__krelPodium v.", version, ".0")
+        PtDebugPrint("__init__xPodium v.", version, ".0")
```
Hoikas, line 59:
```suggestion
        PtDebugPrint(f"__init__xPodium v.{version}.0")
```

Scripts/Python/xPodium.py
```diff
@@ -78,24 +82,24 @@ def OnNotify(self, state, id, events):
...
+            if xPodiumSpeech < 0:
+                xPodiumSpeech = len(respList.byObject) - 1
+            respName = (stringFormat.value % xPodiumSpeech)
             if respName in respList.byObject:
-                ageSDL["nb01CmnRmSpeech"] = (nb01CmnRmSpeech,)
+                ageSDL[stringVarName.value] = (xPodiumSpeech,)
             else:
-                PtDebugPrint(f"ERROR: krelPodium Invalid speech {nb01CmnRmSpeech} selected!")
-                ageSDL["nb01CmnRmSpeech"] = (0,)
+                PtDebugPrint("ERROR: xPodium Invalid speech %d selected!" % xPodiumSpeech)
```
Hoikas, line 92:
```suggestion
                PtDebugPrint(f"ERROR: xPodium Invalid speech {xPodiumSpeech} selected!")
```

Scripts/Python/xPodium.py
```diff
@@ -78,24 +82,24 @@ def OnNotify(self, state, id, events):
...
         ageSDL = PtGetAgeSDL()
-        nb01CmnRmSpeech = ageSDL["nb01CmnRmSpeech"][0]
-        respName = (stringFormat.value % nb01CmnRmSpeech)
+        xPodiumSpeech = ageSDL[stringVarName.value][0]
+        respName = (stringFormat.value % xPodiumSpeech)
 
-        if nb01CmnRmSpeech >= 0 and respName in respList.byObject:
-            PtDebugPrint("krelPodium: Playing speech", respName)
+        if xPodiumSpeech >= 0 and respName in respList.byObject:
+            PtDebugPrint("xPodium: Playing speech", respName)
```
Hoikas, line 104:
```suggestion
            PtDebugPrint(f"xPodium: Playing speech {respName=}")
```

Hoikas:
We've still got some erroneous whitespace changes in the license block 😞

Hoikas:
Any updates on the licensing block thing?

Hoikas:
Pushed fix for whitespace problems.

Hoikas (approved):

PR #1241, opened 2022-09-05
Improving Mac Font Rendering

colincornaby:
Mac font rendering has been improved and I'm trying to stream the changes in incrementally as best I can.

I've made a few improvements:

- Antialiasing has been disabled. The font textures go through a 1 bit pass which isn't compatible with antialiasing and produces a really bold look.
- macOS supports resolution independent font rendering. In theory - Windows on a HiDPI display would double the font size for double DPI. On macOS I can keep the font size constant - but double the resolution of the graphics context. I can then scale the x/y units down by half so the font will render the same - with double the pixels.

There probably still needs to be a fix so that the scale is passed through properly  - and fonts know if they need to render at a higher DPI.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
+++ b/Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
@@ -207,9 +207,9 @@ uint16_t  *plTextFont::IInitFontTexture()
 
     
     // Figure out our texture size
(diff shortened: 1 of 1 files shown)


Hoikas (commented):
Seems fair to me. Is this ready to go?

colincornaby:
Yep! It's ready to go.

PR #1245, opened 2022-09-21
Fixing possible deadlock in plPythonCallable

colincornaby:
PyObject varient of CallObject is not declared before pyObjectRef CallObject varient. This could lead to pyObjectRef CallObject calling itself repeatedly in since the desired PyObject varient is not technically in scope.

I took the approach of declaring the function instead of moving the order - as it might be a bit safer and more clear for future developers. I'm assuming there may be some discussion on the approach (or style of where that declaration should go.)


Hoikas (commented):
Good find! I think it would be perfectly acceptable move the actual function definition instead of using a forward declaration.

Hoikas (approved):

PR #1248, opened 2022-10-04
Fix some truncation warnings.

Hoikas:
Most of these are fairly noisy because they're in header files.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPipeline/plPlates.h
+++ b/Sources/Plasma/PubUtilLib/plPipeline/plPlates.h
@@ -185,8 +185,8 @@ class plGraphPlate : public plPlate
 
         const char      *GetLabelText( int i ) { return fLabelText[ i ].c_str(); }
         uint32_t        GetDataColor( int i ) { return fDataHexColors[ i ]; }
-        uint32_t        GetNumLabels() { return fLabelText.size(); }
-        uint32_t        GetNumColors() { return fDataHexColors.size(); }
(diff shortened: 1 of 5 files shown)


zrax (approved):

Deledrius (approved):

dpogue (commented):
Shouldn't these be using `const_cast<T>()` instead of C-style casts? :trollface:

Sources/Plasma/PubUtilLib/plPipeline/plPlates.h
```diff
@@ -185,8 +185,8 @@ class plGraphPlate : public plPlate
 
         const char      *GetLabelText( int i ) { return fLabelText[ i ].c_str(); }
         uint32_t        GetDataColor( int i ) { return fDataHexColors[ i ]; }
-        uint32_t        GetNumLabels() { return fLabelText.size(); }
-        uint32_t        GetNumColors() { return fDataHexColors.size(); }
+        uint32_t        GetNumLabels() { return (uint32_t)fLabelText.size(); }
+        uint32_t        GetNumColors() { return (uint32_t)fDataHexColors.size(); }
```
dpogue, line 189: Is there a downside to making these return the `size_t` instead of truncating?

Hoikas: I chose the most expedient option for silencing warnings - I was worried that bringing `size_t` out to the public interface might trigger more of them.

PR #1253, opened 2022-10-07
Port plNetObjectDebugger to ST::string

zrax:
This allows us to get rid of the final user of `hsStrLower()`, which can now be removed from the code.  As a bonus, we also don't need to keep around copies of strings converted to lower-case any more, since ST already enables all the case-insensitive comparisons we need to perform.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
@@ -52,11 +52,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plResMgr/plKeyFinder.h"
 #include "plStatusLog/plStatusLog.h"
 
-plNetObjectDebugger::DebugObject::DebugObject(const ST::string& objName, plLocation& loc, uint32_t flags)
-    : fObjName(objName.to_lower()), fLoc(loc), fFlags(flags)
-{
-}
-
 //
 // return true if string matches objName according to flags
 //
@@ -201,7 +196,7 @@ bool plNetObjectDebugger::AddDebugObject(ST::string objName, const ST::string& p
         flags |= kPageMatch;
     }
 
-    fDebugObjects.push_back(new DebugObject(objName, loc, flags));
+    fDebugObjects.push_back(new DebugObject(std::move(objName), loc, flags));
 
     ICreateStatusLog();
 
--- a/Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
@@ -277,11 +277,11 @@ void plNetObjectDebugger::LogMsgIfMatch(const ST::string& msg) const
     }
 }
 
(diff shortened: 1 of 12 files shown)


Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
```diff
@@ -52,48 +52,25 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        return str.compare(fObjName, ST::case_insensitive) == 0;
 
     if (fFlags & kEndStringMatch)
-    {
-        int len=strlen(str);
-        if (len>fObjName.size())
-            return false;
-        return !stricmp(str, fObjName.c_str()+fObjName.size()-len);
-    }
+        return str.ends_with(fObjName, ST::case_insensitive);
```
zrax, line 67: The observant reviewer may notice that the old code actually performed the equivalent of `fObjName.ends_with(str, ...)`.  However, since the pattern specified in `fObjName` was created as "*foo", this didn't make much sense (and the SubStringMatch case below also fits that `fObjName` should be the search criteria, not the text to search in).  Therefore, I've corrected it to match what I believe was the intended behavior.

Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
```diff
@@ -52,48 +52,25 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
     if (fFlags & kStartStringMatch)
-    {
-        int len=strlen(str);
-        if (len>fObjName.size())
-            return false;
-        return !strnicmp(str, fObjName.c_str(), strlen(str));
-    }
+        return str.starts_with(fObjName, ST::case_insensitive);
```
zrax, line 70: Likewise, this was updated to match what I believe was the intended behavior, not the coded behavior.

Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
```diff
@@ -265,31 +238,27 @@ void plNetObjectDebugger::ClearAllDebugObjects()
...
         return;
 
     // extract object name from msg, expects '...object:foo,...'
-    std::string tmp = msg;
-    hsStrLower((char*)tmp.c_str());
-    std::string objTag="object";
-    const char* c=strstr(tmp.c_str(), objTag.c_str());
-    if (c && c != tmp.c_str())
+    const auto objTag = ST_LITERAL("object");
+    ST_ssize_t pos = msg.find(objTag, ST::case_insensitive);
```
Hoikas, line 248: If I read this correctly, we would previously not match `ObJeCt`, but we now do. Minor change of behavior, but I think it's fine.

Hoikas (approved):

Sources/Plasma/PubUtilLib/plNetCommon/plNetObjectDebugger.cpp
zrax: Actually, we would have matched it before too, since the whole message was converted to lower-case before searching.

PR #1254, opened 2022-10-09
Remove duplicate copies of core engine SDL files

dgelessus:
I noticed that a few internal engine SDL files (animation.sdl, avatar.sdl, etc.) exist in two separate copies in the repo: once under Scripts/SDL and once under Sources/Plasma/PubUtilLib/plSDL/SDL. It's been this way since the original open-source release. Thankfully, the two copies haven't diverged (except for formatting).

AFAICT, the files under plSDL are completely unused at the moment. In the plSDL/CMakeLists.txt, the SDL files are added as source files to the plSDL target, but that seems to do precisely nothing, because CMake has no idea what to do with .sdl files. At runtime the game uses the other copy from Scripts/SDL.

I asked in the OpenUru Discord and @Hoikas suggested:

> The duplication is probably an artifact of how the open sourcing was initially handled. The Plasma repo has the engine copy that Cyan was probably developing with. The moul-scripts repo (Scripts/SDL) is just a copy of scripts assets required to run the game. We should probably delete the Scripts/SDL copy and install from the engine directory. Maybe a good issue to open.

So this PR does that.

The only downside to this approach is that you can no longer grab a complete set of SDL files straight from one directory in the repo, e. g. when setting up a server. (For example, I've set my local DIRTSAND test server to read the SDL files straight from Plasma/Scripts/SDL, so I can update them simply by pulling my clone of the Plasma repo.) The alternative would be to keep the core engine SDL files under Scripts/SDL and just delete the plSDL/SDL directory.


Hoikas:
I like it! Thoughts, @Deledrius, @dpogue, @zrax.

dpogue:
In general, I like this.

I don wonder if we want to standardize on tabs vs spaces for SDL files. We seem to have a mix right now (the ones in PubUtilLib were using spaces, the ones in Scripts are a mix)

Deledrius:
What's the reasoning for these being needed specifically inside of the engine code?

dgelessus:
AFAIK, these are the only `STATEDESC`s that are used directly from the engine code - see plSDLTypes.h. All the other `STATEDESC`s are only used from Python and PRPs.

dgelessus:
Rebased after #1477.

Hoikas:
I think that, before we merge this, we need to make sure that [DockerSand](https://github.com/H-uru/dirtsand/blob/ccc7b6e8cd2224d941768fdc56f5eefc974c59d7/Dockerfile#L68) is equipped to deal with not all SDL files being in the Scripts/SDL directory. UruManifest (strictly speaking my responsibility) should be fine because it seems to simply ingest every SDL file that it can lay its hands on.

dgelessus:
Alternatively, we could just put all .sdl files in Scripts/SDL instead of trying to keep the core engine SDLs separate. It would make my life a bit easier too, because then I can continue pointing DIRTSAND directly at the Plasma/Scripts/SDL directory 😉

Mystler:
My 2cts: let SDL files live with their friends and family in the SDL folder. Splitting them up and requiring everyone working with SDLs to track two locations just because they're referenced in code seems silly, especially since we now have the files all live in the same repo anyway and there is no moul-scripts.

PR #1256, opened 2022-10-09
Document all available server.ini configuration settings

dgelessus:
The only existing documentation I could find is the example_server.ini, but it doesn't demonstrate all supported settings. So I went through the code and wrote down a (hopefully) complete list.


Docs/server.ini.md
```diff
@@ -0,0 +1,111 @@
...
+**Example:**
+
+```
+Server.Gate.Host "moula.example.net"
+Server.File.Host "moula.example.net"
+Server.Auth.Host "moula.example.net"
+```
+
+The addresses of the gatekeeper, file, and auth servers, respectively. Each
+address may be a domain nam or a plain IP addresses, optionally including a port
```
Deledrius, line 60: domain *name*

Deledrius (changes requested):

Docs/server.ini.md
Deledrius, line 60:
```suggestion
address may be a domain name or a plain IP addresses, optionally including a port
```

Deledrius (changes requested):

Docs/server.ini.md
```diff
@@ -0,0 +1,111 @@
+# Server configuration
+
+This is a complete list of all supported settings for the server.ini
+configuration file. See [example_server.ini](../example_server.ini) for a more
+practical example.
+
+The server.ini actually uses Plasma console script syntax, which is also used in
```
Hoikas, line 7:
```suggestion
The server.ini actually uses Plasma console commands, which are also used by
```

example_server.ini
```diff
@@ -3,19 +3,20 @@
 # won't parse them correctly!  Also note that the decoded keys are in OpenSSL
 # byte order, which is opposite from the byte order stored in the pnNetBase
 # .hpp files in Cyan's original source.
+# See Docs/server.ini.md for more details.
 
 # The following keys and addresses represent the official MOULagain server
-Server.Game.N "kEso0EnSJNkgyQtVyUO8IGAzy9cVgtEZG3Dy/s0urdYgu+omUP0/3sQihN5EOJqqXQl2ahI7cifHptdDafekOA=="
-Server.Game.X "Pds9j7NYWE9o6WV6o88EhGczzxR/NpNimHn9yLq2X9xoAdiEH1x2R1w/O5nwWv3CPmVEPgm9IalIcr0CvZOy+Q=="
+Server.Game.N "kEso0EnSJNkgyQtVyUO8IGAzy9cVgtEZG3Dy/s0urdYgu+omUP0/3sQihN5EOJqqXQl2ahI7cifHptdDafekOA=="
+Server.Game.X "Pds9j7NYWE9o6WV6o88EhGczzxR/NpNimHn9yLq2X9xoAdiEH1x2R1w/O5nwWv3CPmVEPgm9IalIcr0CvZOy+Q=="
```
Hoikas, line 10: What's with the (per GitHub) no-op changes? I'm assuming you changed the line endings from CRLF to LF, but I just want to make sure.

dgelessus: I was wondering what was up with that - IntelliJ also complained about the line endings in a way that didn't quite make sense... I checked again, and apparently the file used *mixed* CRLF and LF line separators - ugh! Fixed now in a separate commit, so the individual commit diffs should be clean.

Hoikas (approved):

Deledrius (approved):

PR #1258, opened 2022-10-14
Remove incorrectly "fixed" obfuscation check in `ReadSafeWString`

dgelessus:
This was accidentally broken in #617. See commit message for details.


zrax (commented):
Good catch!  The same fix was made to `ReadSafeWStringLong()`, so presumably it should be re-fixed as well.

dgelessus:
Good point, fixed that too. (Although the SafeWStringLong methods seem to be completely unused.)

zrax (approved):

PR #1260, opened 2022-10-15
Remove near-unused `hsStream` `SafeStringLong`/`SafeWStringLong` methods

dgelessus:


Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -109,83 +109,10 @@ void hsStream::CopyToMem(void* mem)
...
-    }
-
-    return ST::string::from_utf16(retVal);
-}
-
 uint32_t hsStream::WriteSafeString(const ST::string &string)
 {
     size_t len = string.size();
-    hsAssert(len<0xf000, ST::format("string len of {} is too long for WriteSafeString {}, use WriteSafeStringLong",
+    hsAssert(!(len & 0xf000), ST::format("string len of {} is too long for WriteSafeString {}",
```
zrax, line 115: This would now fail if a string was even more excessively long (e.g. 0x10000 bytes)...  I imagine the original assertion was intended to be `len < 0x1000` rather than `0xf000`.

Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
```diff
@@ -87,7 +87,10 @@ void plClientResMgr::ILoadResources(const plFileName& resfile)
                 for (int i = 0; i < num_resources; i++) {
                     plMipmap* res_data = nullptr;
                     uint32_t res_size = 0;
-                    ST::string res_name = in.ReadSafeStringLong();
+                    ST::char_buffer res_name;
+                    uint32_t numChars = in.ReadLE32();
+                    res_name.allocate(numChars);
+                    in.Read(numChars, res_name.data());
```
zrax, line 93: This doesn't account for the Safe String bit flipping...

Sources/Plasma/CoreLib/hsStream.cpp
dgelessus: Ah, you're right. I tried to make the condition very similar to the actual bit mask, to avoid any more mismatches between the length check and the masking, but you're right that this doesn't work for very long string lengths.

Sources/Plasma/PubUtilLib/plClientResMgr/plClientResMgr.cpp
dgelessus: That's on purpose. AFAICT, the resource.dat names are always stored in plain text and never use the bit-flipping obfuscation.

Deledrius: I believe that's only true for ones we produce.  If you check the one used in Myst V, it has obfuscated strings.

dgelessus: Hmm, I checked the resource.dat of every MOULa shard I have installed (Cyan, the Minkatas, Gehn, TOC, TL), and the file names are all plain strings, even the shards that ship non-H'uru clients. In my copy of Myst V (from GOG), the Cursors.dat also has plain file names. That's why I assumed that they were never obfuscated at any point.

(Besides, since this is Plasma and not libHSPlasma, is compatibility with pre-MOULa data files even important?)

Deledrius: Oh, I must have misremembered that about Myst V.  I could have sworn it did...

zrax (approved):

PR #1261, opened 2022-10-18
Add volume threshold for subtitled audio before sending subtitles

ZarothYe:
This should prevent subtitles from being sent to the KI when audio is muted or the source is so far away that it is not audible.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
+++ b/Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
@@ -110,7 +110,7 @@ void plWin32Sound::Update()
             if (currentTimeMs <= srtReader->GetLastEntryEndTime()) {
                 while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime(currentTimeMs)) {
                     float currVol = plSound::QueryCurrVolume();
-                    if (plgAudioSys::AreSubtitlesEnabled() && currVol > 0.05f) {
+                    if (plgAudioSys::AreSubtitlesEnabled() && currVol > 0.01f) {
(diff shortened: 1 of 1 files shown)


Deledrius (approved):

Sources/Plasma/PubUtilLib/plAudio/plWin32Sound.cpp
```diff
@@ -109,7 +109,8 @@ void plWin32Sound::Update()
             uint32_t currentTimeMs = (uint32_t)(GetActualTimeSec() * 1000.0f);
             if (currentTimeMs <= srtReader->GetLastEntryEndTime()) {
                 while (plSrtEntry* nextEntry = srtReader->GetNextEntryStartingBeforeTime(currentTimeMs)) {
-                    if (plgAudioSys::AreSubtitlesEnabled()) {
+                    float currVol = plSound::QueryCurrVolume();
```
dpogue, line 112: Does this need the `plSound::` prefix? It's calling a method on the current class (or a parent class), right?

ZarothYe: I think you're right that it's not necessary, my bad. I'll commit with the prefix removed later today.

dpogue: for the record, this is totally valid C++ and probably does what you want in this case, but it has some non-obvious behaviour... this will always call `plSound`'s `QueryCurrVolume()` method with the current instance, rather than any overridden implementation further down the inheritance chain

Hoikas (approved):

dpogue (approved):

PR #1262, opened 2022-10-19
Fix for GitHub's set-output deprecation & workflow warnings

dpogue:
Ref: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This also adjusts the workflow so that it always checks for the 3DS Max token before trying to clone the Max SDK repo, otherwise forks will get `Missing input 'token'` errors (which are harmless but annoying)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # HACK to enable/disable the max CI based on presence/absence of secret. See also:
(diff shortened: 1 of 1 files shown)


Hoikas (approved):

dpogue:
From the PR build (which has no Max SDK), it looks like it's skipping all the Windows builds, so I'll need to look into that.

dpogue:
Sorted out the conditionals that were causing it to not run Windows builds. This run should be all good

Hoikas (approved):

PR #1263, opened 2022-10-19
Use more pythonic ways of looping in various files

ZarothYe:
Get rid of manually-incremented indexes in for-loops in several places. I ran a couple of these past @Hoikas and he provided better/faster replacements than I was going to use. (Thanks!)

### DIFF
--- a/Scripts/Python/xIniHelper.py
+++ b/Scripts/Python/xIniHelper.py
@@ -153,40 +153,32 @@ class iniFile:
         except IndexError:
             pass
 
-    def findByCommand(self,cmd,idx=0):
-        idx = 0
-        for entry in self.entries[idx:]:
+    def findByCommand(self, cmd, idx=0):
+        for index, entry in enumerate(self.entries):
             if entry.command == cmd:
-                return entry,idx
-            idx += 1
-        return None,-1
+                return entry, index
+        return None, -1
 
-    def findByFirstValue(self,value,idx=0):
-        idx = 0
-        for entry in self.entries[idx:]:
+    def findByFirstValue(self, value, idx=0):
+        for index, entry in enumerate(self.entries):
             if len(entry.values) > 0 and entry.values[0] == value:
-                return entry,idx
-            idx += 1
-        return None,-1
+                return entry, index
+        return None, -1
 
-    def findByLastValue(self,value,idx=0):
-        idx = 0
-        for entry in self.entries[idx:]:
+    def findByLastValue(self, value, idx=0):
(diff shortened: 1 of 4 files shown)


zrax (approved):

Deledrius (approved):

Scripts/Python/xIniHelper.py
```diff
@@ -153,40 +153,32 @@ def removeEntry(self,idx):
...
-        return None,-1
+                return entry, index
+        return None, -1
 
-    def findByFirstValue(self,value,idx=0):
-        idx = 0
-        for entry in self.entries[idx:]:
+    def findByFirstValue(self, value, idx=0):
+        for index, entry in enumerate(self.entries):
             if len(entry.values) > 0 and entry.values[0] == value:
```
Hoikas, line 164:
```suggestion
            if entry.values and entry.values[0] == value:
```

Scripts/Python/xIniHelper.py
```diff
@@ -153,40 +153,32 @@ def removeEntry(self,idx):
...
-        for entry in self.entries[idx:]:
+    def findByLastValue(self, value, idx=0):
+        for index, entry in enumerate(self.entries):
             vlist = entry.values[-1:]
             if len(vlist) > 0:
                 if entry.values[-1:][0] == value:
-                    return entry,idx
-            idx += 1
-        return None,-1
+                    return entry, index
```
Hoikas, line 173: 😬 

```suggestion
            vlist = reversed(entry.values)
            if vlist and vlist[0] == value:
                return entry, index
```

zrax: Why not just `entry and entry.values[-1]` ?

Hoikas: I was being extra, I guess :trollface: 

```suggestion
            if entry.values and entry.values[-1] == value:
                return entry, index
 ```

File: Scripts/Python/xIniHelper.py (hunk 1 above)
Comment (ZarothYe) [reply Scripts/Python/xIniHelper.py:164]: Would you mind fixing this suggestion so it can be applied from the GitHub UI? I think you accidentally made the suggestion on the "old" side of the changeset and it won't let me apply it. I could apply it by hand locally and push, but I'm a little lazy. ;)

Comment (Hoikas) [inline Scripts/Python/xIniHelper.py:164]:
```suggestion
            if entry.values and entry.values[0] == value:
```

Hoikas (approved):

PR #1265, opened 2022-11-05
Various Mingw fixes

dpogue:
Opening this as a draft because I suspect some of these might not be the correct approach, but I was able to (mostly) compile with MinGW on Linux after these changes.


zrax (commented):
These changes all look fine to me

zrax (approved):

Hoikas (approved):
These changes look good to me as well. I'd like to see the PR rebased onto master now that the macOS Python problem has been solved so that we can get green checks on CI. Not that I expect for CI to fail :).

dpogue:
Will do. Still sorting out a few issues with Wchar string comparisons in pfJournalBook and header casings. I guess MinGW wants all the Windows headers to be lowercase :\

Hoikas (approved):

PR #1266, opened 2022-11-14
Future-based timed thread joins

dpogue:
Fixes some crashes with the pthread timed joins and also adds the ability to do a timed join on macOS.

I left Windows using the existing implementation because Hoikas said futures had major performance issues in MSVC.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
+++ b/Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
@@ -47,9 +47,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "Pch.h"
 
-#if defined(HAVE_PTHREAD_TIMEDJOIN_NP)
-#include <pthread.h>
-#include <time.h>
+#if HS_BUILD_FOR_UNIX
+#   include <future>
 #endif
 
 /*****************************************************************************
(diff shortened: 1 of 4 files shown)


Sources/Plasma/NucleusLib/pnAsyncCoreExe/pnAceThread.cpp
```diff
@@ -69,28 +68,16 @@ void ThreadDestroy (unsigned exitThreadWaitMs) {
...
-    thread.detach();
 #else
-    LogMsg(kLogDebug, "No timed thread join support for this system... "
-                      "Performing a blocking join instead.");
-    thread.join();
+    auto future = std::async(std::launch::async, &std::thread::join, &thread);
+    if (future.wait_for(std::chrono::milliseconds(timeoutMs)) == std::future_status::timeout) {
+        LogMsg(kLogDebug, "Thread did not terminate after {} ms", timeoutMs);
+        thread.detach();
+    }
```
colincornaby, line 81: One possible issue is that the same thread can't be both joined and detached. I'm not sure if C++ will quietly handle that (possible it might.) But this isn't a true timeout. Even if the future times out it may still eventually execute before or after the thread has been detached.

Hoikas:
I don't remember why the pthread timed join was problematic on Linux. I'm assuming that the problem on macOS is that the call is not available. Can you recap the issue for me so that I can better understand the goal of these changes?

colincornaby:
> I don't remember why the pthread timed join was problematic on Linux. I'm assuming that the problem on macOS is that the call is not available. Can you recap the issue for me so that I can better understand the goal of these changes?

Two issues here:
- pthread_timedjoin_np is non portable and is only available on Linux (hence the np suffix.)
- std::thread doesn't really allow native functions to be used on the thread even though it has a native handle. It introduces issues in the underlying tracking and causes an error when the process terminates. The error isn't harmful, but it still is an unclean exit. So we also can't use pthread_timedjoin_np anyway.

The standard solution seems to be adding some sort of wrapping std::thread to add a lock to the end of execution that can be waited on - because locking has a timeout capability. Basically implementing the functionality using locks instead of trying to rely on anything built in to threads - which is also portable.

dpogue:
The ASIO branch has examined this problem in a lot more detail and taken a different approach (which seems to be working), so I'm going to close this

Issue #1214, opened 2022-07-08
Garbled Text when System Scaling >= 300%


patmauro:
When system scaling is set to a value >= 300% (ie a high resolution display such as a 4K display is connected), lowercase characters **t-z** stop displaying correctly in a few key contexts in-game, resulting in bizarre garbled text. This is seen most noticeably when mousing over other player characters, but can also be seen in loading screens and the plasma console itself. This has been reproduced on multiple machines and on both OU and H'uru clients.

![image](https://user-images.githubusercontent.com/6642931/177893860-21ed5eec-8954-4229-ae27-7a2de06f34d6.png)
![image](https://user-images.githubusercontent.com/6642931/177893693-11a22218-ba1f-496a-94fd-3b271324d013.png)
![image](https://user-images.githubusercontent.com/6642931/177893709-7815f1a2-fa65-49c3-bde0-c7af91d3024d.png)
![image](https://user-images.githubusercontent.com/6642931/177893734-47b75161-57f0-4d73-897c-125b00bb0e1d.png)

Note that while this is similar to issue https://github.com/H-uru/Plasma/issues/636, it does not appear to be related. Note also that in that case, the "pixellation" issues can be fixed by setting UruExplorer.exe specifically to override system scaling settings and setting to "Scaling Performed By: Application." In this case, the text garbling occurs regardless of what scaling method is set here and what resolution the game is set to run in - as long as system scaling is set to a value >= 300%, this issue will kick in.

Hoikas:
Confirmed. This seems to be another case of #636. In my testing, the issue did not appear until I forced an override for "Scaling Performed By: Application." This indicates that when #636 is fixed completely and correctly, this problem will also be fixed. Of note: Windows doesn't offer above 225% scaling on my system and gives an ominous warning about custom scaling.

dgelessus:
> Of note: Windows doesn't offer above 225% scaling on my system and gives an ominous warning about custom scaling.

I think the maximum scale shown in the dropdown menu depends on the resolution of your screen. For example, at 1024x768 I get only 125%, at 1920x1080 it allows up to 175%, and at 3000x2000 (MS Surface Go built-in screen) up to 200%. Maybe with a 4K screen it displays the 300% option.

Issue #636, opened 2020-03-02
Game is pixelated when using HiDPI Scaling on Windows


Hoikas:
Conversation with @cwalther on IRC 2020-02-29 indicates problems with playing the game when Windows scaling is enabled. He reports the image is improperly scaled using nearest neighbor interpolation when playing in full screen. There is a [previous report](https://forum.guildofwriters.org/viewtopic.php?f=117&t=7062) of chicanery as well.

```[17:16:20] <CWalther> Hi everyone! I’m just trying out Gehn 27, and it’s not looking as good as it should on my 4K display when I choose the full 3840x2160 resolution. Things are scaled as if it was rendering for 3840x2160 (e.g. the cursor is tiny), but the view is rendered at 1920x2160 and pixel-doubled, so it looks blocky. Is that the way it’s supposed to be? It may have been that way before, I don’t remember exactly. Do I need to
[17:16:21] <CWalther> tweak anything in my Windows display settings? It’s set to 200%, “fix scaling” off and no custom scaling.
[17:16:53] <CWalther> *rendered at 1920x1080
[17:26:27] <Doobes> Hoikas would probably have the best answer for that as he's been able to grab high res screenshots of things.
[17:27:24] <Doobes> He may currently be AFK or busy with something though.
[17:27:27] <CWalther> Yeah, I remembered that so it seems odd.
[17:28:30] <Doobes> Wish I could help, but I'm still stuck in 1080p land. :D
[17:32:30] <CWalther> Chiso is looking fantastic in 1080p! :)
[17:47:03] <Hoikas> I haven't seen any issues on my 3840x1600 display, CWalther
[17:47:42] <CWalther> I almost suspected as much, otherwise you’d have fixed them :)
[17:48:47] <CWalther> I can try playing with the Windows settings.
[17:50:03] <Hoikas> I know we had a complaint recently from someone playing with scaling enabled, but I think that was a case of the UI being offscreen
[17:51:18] <CWalther> With 100% scale it looks good (but I don't want my Windows desktop that tiny).
[17:51:33] <Hoikas> :)
[17:52:27] <Hoikas> I don't touch scaling myself. I invariantly use programs that look terrible with it on. Instead, I "just" (a year ago) bought an ultrawide monitor
[17:54:44] <CWalther> With 200% and “fix scaling for apps” enabled (whatever that is), no difference, still blocky.
[17:57:06] <CWalther> With “custom scaling” 300%, pixel-tripled blocky.
[...]

PR #1269, opened 2022-12-05
Introducing backing scale

colincornaby:
This change introduces the idea of a backing scale into the engine. Backing scale refers to the difference between the resolution of the window that Uru is running in, and the resolution of the frame buffer backing it.

For example, on a 4k Retina display, the window may only have a resolution of 1920x1080, while the actual Metal/DX/OpenGL frame buffer has a resolution of 3840x2160. The backing scale would be 2.

Different UI elements can use the backing scale to resize accordingly. This change alters the behavior of progress bars and the intro movie to account for the backing scale. Cursor should also change with the backing scale - I'm still working on pulling that together for this pull request.

Text rendering should also take into account the backing scale (although this PR does not make all the required changes.) A backing scale of 2 would imply that a 12 point font should render at twice the DPI. I've made some initial progress for this on macOS.

Backing scale is also useful if Uru ever chose to use DLSS/MetalFX upscaling. If the engine is under-rendering, backing scale can be used to push the UI elements back into a size the user expects.


colincornaby:
I'm looking at cursor scale again. Cursor scale is its own function right now in the Mac client. In some ways this makes sense - plMouseDevice is not owned by plClient and can't trivially just inherit settings. There are accessibility reasons that the cursor may also want to be its own scale down the road.

I may reintroduce the cursor scale as is - even though it's parallel to the backing scale and not directly affected by it. There could be further iteration on cursor scale later.

colincornaby:
Something weird is going on in Windows font rendering. It's not behaving how I expect it to, and this pull request is causing font spacing to not behave correctly on Windows. Still checking it over.

colincornaby:
I'm kind of fumbling around in the Win32 client (never done serious Win32 before) and I've got backing and cursor scale hooked up in Windows as well.

I enabled HiDPI in Windows by tweaking the manifest. I have not committed that change in since it probably requires more testing, and there have been known issues around HiDPI.

```
<?xml version="1.0" encoding="UTF-8"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
	<asmv3:application>
		<asmv3:windowsSettings>
			<dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
		</asmv3:windowsSettings>
	</asmv3:application>
  <dependency>
    <dependentAssembly>
      <assemblyIdentity type="win32" name="Microsoft.Windows.Common-Controls" version="6.0.0.0" processorArchitecture="*" publicKeyToken="6595b64144ccf1df" language="*"/>
    </dependentAssembly>
  </dependency>
</assembly>
```

colincornaby:
Thanks! I was unsure how to opt in to functions from newer versions of Windows. There are better DPI check functions to use on newer Windows as well.

I'm also unsure if cursor scale and backing scale should be more closely related or stay separate options - but that's why this PR is a draft.

Hoikas:
I've built a set of [DPI awareness changes](https://github.com/H-uru/Plasma/compare/master...Hoikas:Plasma:backing-scale-changes) for Windows on top of this PR on my fork that might be useful. My branch seems to fix the problems in #636.

colincornaby:
Your changes look good to me. I'm going to pull them in. I don't have the Mac client code in that takes advantage of them, but I added the message signaling locally and it seemed fine.

I'll reset this to your branch's head and then graduate this to a full PR. Your comments indicate there might be a few things not done yet, but we can look at that on deeper review.

colincornaby:
I wrote that I'd bring in your branch's head. It's already here! Never mind.

colincornaby:
@Hoikas - I'm going to look at this again on macOS. Was there any more work you wanted to do on Windows? So far this has worked well on Windows for me, but I know there are some todos in the code comments.

Hoikas:
I don't think there's anything else I want to do right now.

colincornaby:
I'll do a final checkover. I'm fine dealing with any Mac specific issues directly on the Mac branch. But I do want to make sure there isn't anything critical with the core implementation.

colincornaby:
Something is subtlety wrong with Windows font rendering at HiDPI. The spacing is off. I think it's due to an issue with the font map that's created. I'll investigate more.

colincornaby:
Some details about the Windows font issue I'm looking at.

Here's what the current master branch looks like on the load screen:
1x Loading

And here's what this branch looks like on a HiDPI display:
2x Loading

The HiDPI issue has some subtle issues. The "Downloading" text is slightly too high and the detail text below the progress bar is spaced slightly too close.

Here's a graphic that better shows the issue:
Comparison

I spent some time going through the rendering code, and have started to believe the issue might be in the Win32 code that creates the font texture. On the left is the HiDPI font texture, and on the right is the normal font texture. I did a 2x nearest neighbor scale on the low rest font texture to make it comparable to the 1x texture.

Comparison

There seems to have been a padding change over the original 1x map. This could be working its way into the renderer causing the text to be pushed up. That would push the top text away from the progress car, and the bottom text closer to the progress bar.

This is my best lead so far. Examining the original Direct3D rendering with PIX and the metrics being created on the fly - everything in those components seems to be in order. The font map is the only thing I'm finding that looks wrong.

My strength is in Quartz2D text rendering but I'll do some Win32 digging to see what I can figure out. I might pull this code out into it's own app to do some quicker debugging.

Hoikas:
It's probable that there are some non-DPI aware APIs being called in the debug text renderer that I didn't catch. 😞

colincornaby:
> It's probable that there are some non-DPI aware APIs being called in the debug text renderer that I didn't catch. 😞

Possible. But this code is strange in that Cyan created it to be DPI aware - but it may have never really worked. AFAIK the DPI check would never have been tested at anything but the default 96 DPI.

dpogue:
Alternatively, maybe we write an implementation of this using freetype2 so that it's cross-platform? (Since I'm going to end up needing to do that for OpenGL anyways, if someone wants to do that work for me... 😛)

colincornaby:
Update on where I am with this:

There are issues with how the text rendering code was originally written. Plasma wasn't intended to be DPI aware when it shipped, so they're understandable.

I'm going to grab this graphic from Stackoverflow to demonstrate the issue:


TrueType fonts have layout metrics that define the layout of the glyphs. This spacing is _not_ guaranteed to be stable between font sizes or faces.

Plasma lays out text from the top of the character. But Courier New specifically does not have consistent top padding between font sizes for its glyphs. This is causing the text's offset to shift between font sizes.

A fix would be using the text baseline (the green line) for text layout. This line is stable in relation to the glyphs and would be more predictable between font sizes. Unfortunately, this requires passing through additional metrics about the font and re-writing some rendering code.

I have most the work done, but right now I'm just making sure I didn't break the standard DPI on Windows.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -136,23 +157,57 @@ uint16_t  *plTextFont::IInitFontTexture()
     // Loop through characters, drawing them one at a time
     RECT    r;
     r.left = r.top = 0;
-    r.right = r.bottom = 10;
+    r.right = 1;
+    r.bottom = 10;
     FillRect( hDC, &r, (HBRUSH)GetStockObject( GRAY_BRUSH ) );
 
     // (Make first character a black dot, for filling rectangles)
     SetPixel( hDC, 0, 0, RGB( 255, 255, 255 ) );
```
colincornaby, line 165: Does anyone know what lines 158 through 165 are about? The first character is going to be a space, which feels sufficient. I have no idea what this other code is doing. @dpogue @Hoikas

Hoikas: That looks a little bit like it's drawing a cursor, assuming you mean:
```C++
    RECT    r;
    r.left = r.top = 0;
    r.right = r.bottom = 10;
    FillRect( hDC, &r, (HBRUSH)GetStockObject( GRAY_BRUSH ) );

    // (Make first character a black dot, for filling rectangles)
    SetPixel( hDC, 0, 0, RGB( 255, 255, 255 ) );
```
Not sure about the `SetPixel()` call though.

colincornaby:
I noticed a few issues remaining:
- The banner image in the login window doesn't fill it's rect. I can change that to scale.
- The patcher isn't HiDPI.
- The default window size when running in debug is now physically smaller. Not sure if that is a feature or a bug.

I'm not moving macOS to FreeType yet. I'll take care of that on the Mac branch where I can test it better.

Sources/Plasma/Apps/plClient/win32/plWinDpi.cpp
```diff
@@ -0,0 +1,287 @@
...
+            if (SUCCEEDED(perMonitorV1Result.value())) {
+                LogGreen("Using Per Monitor V1");
+                break;
+            } else {
+                LogRed("Failed to set DPI Awareness Per Monitor v1: {}", hsCOMError(perMonitorV1Result.value()));
+            }
+        }
+
+        // System DPI Awareness.
+        if (SetProcessDPIAware() != FALSE) {
```
colincornaby, line 119: Hey @Hoikas - a few questions...

I noticed this function wasn't wrapped in an optional win call - could this cause trouble on Windows XP in since this function is only available on Vista?

I'm also looking at implementing the minimum functionality needed in the patcher to set the DPI of the app. Don't have a problem implementing it - just curious if there is a reason to query/go to the newer API and if that's relevant to the patcher.

dpogue: Our dependencies aren't compatible with Windows XP anymore, so I don't think that's much of a concern at this point.

colincornaby: Ah - is OpenUru also dropping Windows XP, or is that a larger conversation?

dpogue: I think it's currently a non-conversation because you know how talking to them goes... Realistically, we're already having to patch Python 3.10 to retain Windows 7 compatibility so the ship has to sail at some point. If we ever update to PhysX 5.1, they've dropped 32-bit support.

Hoikas: I think @dpogue has already answered everything here, but, if you want, you can change this to use the optional wincall thing to maintain compatibility, but that's not really something that we really care about in general.

I would really prefer to see the code shared as much as possible - do you think you could reorganize `plWinDpi` to not depend on `plMessage` instead of duplicating the code?

colincornaby: Let me think it through. I could probably move the core parts into their own module while making sure the stuff that is only needed for the engine lands in the engine. Maybe I could put a base class in it's own module and then extend the Plasma specific class with the rest.

patmauro:
Given the extensive text work here, it's possible this may directly address and resolve https://github.com/H-uru/Plasma/issues/1214

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -3337,6 +3337,7 @@ bool  plDXPipeline::CaptureScreen( plMipmap *dest, bool flipVertical, uint16_t d
     }
     else
     {
+        // FIXME: DPI awareness
```
dpogue, line 3340: `fDevice.fHwnd` is the HWND for the client window, if that's useful here

colincornaby:
I've committed changes that will scale the banner images for both the patcher and client login screens.

The way HiDPI is enabled for the patcher is temporary. I'm still working on a more permanent solution involving plWinDPI.

Sources/Plasma/Apps/plUruLauncher/winmain.cpp
```diff
@@ -188,6 +189,44 @@ static void PumpMessages()
     } while (s_launcher->PumpNetCore());
 }
 
+void SetDPIAwareness() const
```
Hoikas, line 192:
```suggestion
static void SetDPIAwareness()
```
I know this is temporary code, but this being `const` doesn't make sense because it is not part of a class.

Sources/Plasma/CoreLib/hsWindows.h
```diff
@@ -142,6 +143,55 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
             ST::format_type(format, output, hr.fResult);
         }
     }
+
+    template<class, class...>
+    class plOptionalWinCall;
```
dpogue, line 148: @Hoikas For what it's worth, I wanted to play around with something similar for some Linux stuff, and ended up slightly modifying this to work with `dlopen`/`dlsym` too. I also tried to add a map of open libraries so we can avoid re-opening the same library multiple times if we're pulling a bunch of functions from it.
My code (which is probably not perfect): https://gist.github.com/dpogue/6ac36a0fdd78397211441825b4ad5405

Sources/Plasma/Apps/plUruLauncher/winmain.cpp
dpogue: Yeah, the `const` modifier is breaking CI:
```
plUruLauncher\winmain.cpp(193,1): error C2270: 'SetDPIAwareness': modifiers not allowed on nonmember functions
```

colincornaby:
@Hoikas - Take a look at the latest changes. I pushed any message specific stuff in plWinDpi out into the client itself (and made plWinDpi stand alone.) Took my best guess at what you might want to see for the Windows client. It's a little messier but at least prevents the patcher from needing to touch the messaging system.

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -172,14 +173,21 @@ static void AuthFailedStrings (ENetError authError,
 
 void DebugMsgF(const char* format, ...);
 
+void HandleDpiChange(HWND hWnd, UINT dpi, float scale, const RECT& rect)
```
Hoikas, line 176:
```suggestion
static void HandleDpiChange(HWND hWnd, UINT dpi, float scale, const RECT& rect)
```

Sources/Plasma/PubUtilLib/plWinDpi/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+)
+
+target_link_libraries(
+    plWinDpi
+    PUBLIC
+        CoreLib
+        plStatusLog
+)
+
+endif()
```
Hoikas, line 22: Newline at EOF

Sources/Plasma/PubUtilLib/plWinDpi/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+if(WIN32)
```
Hoikas, line 1: This should probably be in the parent directory's CMakeLists

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.cpp
```diff
@@ -252,17 +247,7 @@ void plWinDpi::IEnableNCScaling(HWND hWnd) const
...
-    float scale = float(dpi) / 96.0f;
-    LogWhite("Window DPI changed to {} (scale factor: {.02f})", dpi, scale);
-
-    // Inform the engine about the new DPI.
-    auto* msg = new plDisplayScaleChangedMsg(scale, ConvertRect(rect));
-    msg->Send();
-}
-
-std::optional<LRESULT> plWinDpi::WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam) const
+std::optional<LRESULT> plWinDpi::WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam, const std::function <void (HWND, UINT, FLOAT, RECT&)>& dpiChangedCallback) const
```
Hoikas, line 250:
```suggestion
std::optional<LRESULT> plWinDpi::WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam, const std::function<void (HWND, UINT, FLOAT, RECT&)>& dpiChangedCallback) const
```

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
```diff
@@ -114,23 +113,12 @@ class plWinDpi
...
-    }
-
 private:
     BOOL ICalcWinSize(HWND hWnd, UINT dpi, SIZE& hWndSize) const;
     void IEnableNCScaling(HWND hWnd) const;
-    void IHandleDpiChange(HWND hWnd, UINT dpi, const RECT& rect) const;
 
 public:
-    std::optional<LRESULT> WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam) const;
+    std::optional<LRESULT> WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam, const std::function <void(HWND, UINT, FLOAT, RECT&)>& dpiChangedCallback) const;
```
Hoikas, line 121:
```suggestion
    std::optional<LRESULT> WndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lParam, const std::function<void(HWND, UINT, FLOAT, RECT&)>& dpiChangedCallback) const;
```

Hoikas (commented):
Looks like this is moving in the right direction.

Sources/Plasma/CoreLib/hsWindows.h
Hoikas: > I also tried to add a map of open libraries so we can avoid re-opening the same library multiple times if we're pulling a bunch of functions from it.

My thought process with `plOptionalWincall` was that there would be a fairly centralized place for the function pointers to only be bound once, like all of the DPI stuff only ever being inside of the `plWinDpi` singleton. I think it would make more sense to just bind any other function pointers we need in a single TU at application start up time.

dpogue: yeah, but even within plWinDpi you're pulling 9 functions from user32, which is going to cause `LoadLibraryW`/`FreeLibrary` to get called 9 times for the same library. That's what I was hoping to avoid by caching the loaded library.

Hoikas: That's true, but `USER32` is already loaded, so we're really just looking up the library by string name, incrementing the reference counter, then decrementing the reference counter. I'm not sure we really need to micro-optimize that.

colincornaby:
@Hoikas - I've applied your suggestions.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -104,183 +98,137 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+    CFRelease(fulfilledFontDescriptor);
+    CFRelease(fontURL);
+    CFRelease(fileSystemPath);
+    
+    FT_UInt freeTypeResolution = 192;
+#else
+    FT_Done_FreeType(library);
+    FT_UInt freeTypeResolution = 0;
+    return nullptr;
+#endif
```
colincornaby, line 161: I've put both Mac and Windows onto FreeType. The only thing platform specific is the TTF lookup. There isn't a huge amount of code - I think we can do this without a third party library. Hoping it's sufficient for @dpogue to get fonts working.

dpogue: I am looking into `fontconfig` on Linux as I type this 😉

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -57,13 +57,17 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#include FT_GLYPH_H
+
+#include <memory>
+
 #if defined(HS_BUILD_FOR_APPLE)
 #import <CoreGraphics/CoreGraphics.h>
 #import <CoreText/CoreText.h>
 #endif
 
-#define DisplayableChar(c) (c >= 0 && c <= 128)
```
colincornaby, line 65: This check was only being run on the char type so the compiler advised me it was redundant. Kind of wondering if that's true? This only goes to 128.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -104,183 +98,137 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+    CFStringGetCString(fileSystemPath, cPath, PATH_MAX, kCFStringEncodingUTF8);
+    
+    ftError = FT_New_Face(library, cPath, 0, &face);
     
-    bitmapBits = (uint32_t *)CGBitmapContextGetData(bitmapContext);
+    CFRelease(fulfilledFontDescriptor);
+    CFRelease(fontURL);
+    CFRelease(fileSystemPath);
+    
+    FT_UInt freeTypeResolution = 192;
```
colincornaby, line 156: This is a giant ugly hack. I need to come up with something to get the DPI through in since we don't have access to Cocoa in here.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -104,183 +98,137 @@ uint16_t  *plTextFont::IInitFontTexture()
...
-            tBits++;
-        }
-    }
+    SelectObject(hDC, hFont);
+    
+    DWORD fontDataSize = GetFontData( hDC, 0, 0, 0, 0 );
+    void* fontData = std::malloc( fontDataSize );
+    GetFontData(hDC, 0, 0, fontData, fontDataSize);
+    ftError = FT_New_Memory_Face(library, (FT_Byte *) fontData, fontDataSize, 0, &face);
+    FT_UInt freeTypeResolution = GetDeviceCaps(hDC, LOGPIXELSY);
```
colincornaby, line 131: I don't know a better way to load the font on Windows other than actually loading it as a Win32 font and pulling the TTF data. Didn't see anything better on their internet either for Win32.

However: There may be better options in things like DirectWrite. I had avoided other options because of the XP-as-baseline policy.

Hoikas: Windows 7 APIs are fine to use

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
Hoikas: `fontconfig` works on all three platforms, right? It would be nice to use one library instead of having to maintain all of this going forward.

dpogue: I don't think fontconfig is something we'd want to use outside of Linux (and if I can avoid using it there, I'd like to 😛)

colincornaby: I too would like to avoid using Fontconfig - the macOS code here should be pretty stable. Looking at Windows - the DirectWrite stuff might be even more trivial (mostly because it avoids the CFStringRef stuff going on with macOS.)

CMakeLists.txt
```diff
@@ -98,7 +98,7 @@ find_package(CURL REQUIRED)
 find_package(DirectX)
 find_package(epoxy)
 find_package(expat REQUIRED)
-find_package(freetype)
+find_package(freetype REQUIRED)
```
dpogue, line 101: Can you update `cmake/Dependencies.cmake` (around line 45) to mark this as required there too?

Sources/Plasma/PubUtilLib/plPipeline/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ target_link_libraries(plPipeline
         plSurface
         pfCamera # plCaptureRender, plDynamicEnvMap, plDXPipeline
         pfGameGUIMgr # plCaptureRender
+        freetype
```
dpogue, line 74: Out of curiosity, should this also have a (platform-specific) link for `CoreText.framework`?

colincornaby: CoreText is under the Cocoa/Cocoa.h umbrella, so it should get loaded under Cocoa. Cocoa is kind of a catch all for a bunch of stuff.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
colincornaby: After some more looking, I'm inclined to leave these GDI calls in place at least for the short term. Windows has the complication that TTFs are not guaranteed to be files, so they may not have a file path. I don't think that applies to Courier as shipped, but the copy-font-as-data path may not be as wrong as it feels on Windows. But I'm guessing this might be revisited in the future.

Hoikas:
There is a crashing bug on 32-bit Windows due to a calling convention problem in `plOptionalWinCall`. The following patch fixes it:
```diff
diff --git a/Sources/Plasma/CoreLib/hsWindows.h b/Sources/Plasma/CoreLib/hsWindows.h
index 89a00f560..afbb914b5 100644
--- a/Sources/Plasma/CoreLib/hsWindows.h
+++ b/Sources/Plasma/CoreLib/hsWindows.h
@@ -153,7 +153,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
     >
     class plOptionalWinCall<_ReturnT(_ArgsT...)>
     {
-        using _FuncPtrT = _ReturnT(*)(_ArgsT...);
+        using _FuncPtrT = _ReturnT (WINAPI*)(_ArgsT...);

         HMODULE fHModule;
         _FuncPtrT fProc;
 ```

**Comment by Hoikas** [comment]
freetype results seem a bit non-optimal in Windows
image

**Comment by colincornaby** [comment]
Yuck, that doesn't look right at all. Is that standard DPI?

**Comment by Hoikas** [comment]
Yes, that's at 100% scaling.

**Comment by colincornaby** [comment]
I'll debug and take a look. At HiDPI - I only found a few pixels different between Freetype's font rendering and Windows. So that is very unexpected.

**Comment by dpogue** [comment]
Just for reference, this is what I see on Linux at 1x DPI:
Screenshot from 2023-05-14 21-31-34

**Comment by colincornaby** [comment]
I'll take a look at the bitmap being generated. The bitmap should be the same across all systems. Possible that's not true for some reason, or the bitmap isn't playing nice with the Direct3D font renderer. I'd still consider that a FreeType issue in since it should generate the same-ish output as GDI.

**Comment by dpogue** [comment]
This is the texture I get for GL pipeline (upside down because GL and it's whole "0,0 is at the bottom left" thing 🙄)
courier-freetype

**Comment by colincornaby** [comment]
Windows Texture

Here's the Windows texture. I was going to grab what Windows is on the master branch - but this is pretty clearly broken all by itself.

**Comment by colincornaby** [comment]
FWIW - I'm not yet blaming FreeType. Without antialiasing - FreeType puts out 1 bit images. That's ok - Uru was actually flattening everything down to either 0 or int max in it's font buffers. But it's possible that maybe something is going wrong at the lower DPI with the 1 bit texture read?

Could still be FreeType - just saying don't rush to blame it yet. I'll do some debugging.

**Comment by colincornaby** [comment]
I think we're on the same page already - but I did check the FreeType output buffers on Windows, and they match exactly what we're seeing. I'll try tweaking some FreeType options to see if I can improve anything.

**Comment by colincornaby** [comment]
I had some flags set to prevent FreeType from using dramatically different metrics between font sizes that could affect layouts that were engineered at one DPI. I've reversed one of them and it seems to have fixed Windows font rendering. Give the last commit a try.

This comes with the disclaimer that it may allow FreeType to layout the high DPI font in ways that may not be helpful. But so far I haven't seen the issues I had when I tried the same thing with GDI.

**Comment by Hoikas** [comment]
Much better!

image

... however, I was able to trigger a crash, but I'm not sure what I did:

```
UruLive.1.918 - dad38ec - Internal.Release
OS: Windows 10 Professional (Build 19045)
CPU: Intel(R) Core(TM) i7-4770K CPU @ 3.50GHz
RAM: 32768 MiB

Stack Trace:
Level 00: plClient!plDebugTextManager::CalcStringWidth (File: plDebugText.cpp:269)
Level 01: plClient!plDebugTextManager::DrawToDevice (File: plDebugText.cpp:239)
Level 02: plClient!plDXPipeline::RenderScreenElements (File: plDXPipeline.cpp:3097)
Level 03: plClient!plClient::IDraw (File: plClient.cpp:1772)
Level 04: plClient!plClient::MainLoop (File: plClient.cpp:1527)
Level 05: plClient!WinMain (File: winmain.cpp:1290)
Level 06: plClient!__scrt_common_main_seh (File: exe_common.inl:288)
Level 07: KERNEL32!BaseThreadInitThunk
Level 08: ntdll!RtlGetAppContainerNamedObjectPath
```

**Comment by colincornaby** [comment]
That's odd. Looks like a bad font pointer. If you can figure out how to repro it, let me know. I'll do some digging. It's not directly related to the FreeType work I don't think - somehow a font pointer is bad.

**Comment by colincornaby** [comment]
@Hoikas - Were you ever able to repro that issue? Wondering if maybe it's an out of bound drawing issue. I can add some asserts in the draw code to make sure it never draws out of bounds. Otherwise I'm not sure there is a good explanation - there is a null check in that function call so fFont would just have to be corrupt.

File: Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp [hunk 17]
```diff
@@ -132,6 +129,10 @@ uint16_t  *plTextFont::IInitFontTexture()
 
     DeleteDC( hDC );
     DeleteObject( hFont );
+    
+    int iScale = int( ceil( GetDeviceCaps(hDC, LOGPIXELSY) / 96.0f ) );
```
Hoikas, line 133: Maybe use `plWinDpi` for this?

Hoikas:
I haven't been able to repro the crash. As best I can tell, a `nullptr` font is being dereferenced.

colincornaby:
> I haven't been able to repro the crash. As best I can tell, a `nullptr` font is being dereferenced.

That code is surrounded by a null check so it almost has to be a bad pointer. I upped the size of the textures being used as the texture size should probably increase as the DPI increases - and I added those asserts. It could be an out of bounds buffer write.

Address Sanitizer could help here but it has issues with Plasma. It finds some issues immediately that are probably ok - just funky stuff happening in Plasma.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
colincornaby: I decided to just use fFontHeight instead which should vary based on the DPI. I was worried my change would introduce non-power-of-two textures which is not a problem in Metal but might be on older Direct3D hardware. Not sure about the status of non power of two in Direct3D 9.

Hoikas:
```
Stack Trace:
Level 00: KERNELBASE!RaiseException
Level 01: VCRUNTIME140!CxxThrowException
Level 02: plClient!__scrt_throw_std_bad_array_new_length (File: throw_bad_alloc.cpp:40)
Level 03: 0x009AF0BC (process space)
Level 04: plClient!plTextFont::IInitFontTexture (File: plTextFont.cpp:159)
Level 05: plClient!plTextFont::IInitObjects (File: plTextFont.cpp:276)
Level 06: plClient!plDXTextFont::SaveStates (File: plDXTextFont.cpp:288)
Level 07: plClient!plDebugTextManager::DrawToDevice (File: plDebugText.cpp:220)
Level 08: plClient!plDXPipeline::RenderScreenElements (File: plDXPipeline.cpp:3097)
Level 09: plClient!plClient::IDrawProgress (File: plClient.cpp:1679)
Level 10: plClient!plClient::IDraw (File: plClient.cpp:1781)
Level 11: plClient!plClient::IProgressMgrCallbackProc (File: plClient.cpp:1273)
Level 12: plClient!plProgressMgr::IUpdateCallbackProc (File: plProgressMgr.cpp:202)
Level 13: plClient!plProgressMgr::IRegisterOperation (File: plProgressMgr.cpp:140)
Level 14: plClient!plClient::IOnAsyncInitComplete (File: plClient.cpp:2128)
Level 15: plClient!plClient::MsgReceive (File: plClient.cpp:845)
Level 16: plClient!plDispatch::IMsgDispatch (File: plDispatch.cpp:369)
Level 17: plClient!plDispatch::MsgSend (File: plDispatch.cpp:500)
Level 18: plClient!plDispatch::MsgQueueProcess (File: plDispatch.cpp:523)
Level 19: plClient!plClient::IUpdate (File: plClient.cpp:1576)
Level 20: plClient!plClient::MainLoop (File: plClient.cpp:1524)
Level 21: plClient!WinMain (File: winmain.cpp:1295)
Level 22: plClient!__scrt_common_main_seh (File: exe_common.inl:288)
Level 23: KERNEL32!BaseThreadInitThunk
Level 24: ntdll!RtlGetAppContainerNamedObjectPath
```

colincornaby:
Well thats embarrassing. The texture buffer was allocated before width and height were assigned. That means the width and height were probably trash values. I moved things around.

dpogue:
For what it's worth, this is what I added to get Linux support with fontconfig: https://github.com/dpogue/Plasma/commit/7a74bfcca3bf64e7252203d39840770f48e93f80

It's probably worth saving that for a follow-up PR though

colincornaby:
Yeah, the Mac backing scale is hardcoded right now so there will need to be further refinements to this code on the Mac side too.

Also wondering if there will need to be a "flush font texture" feature if the backing scale changes at runtime. I'm not sure if the engine will do that.

colincornaby:
@Hoikas - Could you re-review? I think you mention stylistic changes. There might be more things or maybe I lost track of a comment.

Sources/Plasma/CoreLib/hsWindows.h
dpogue: I would still sorta prefer to see this made not Windows-specific (so I can use it for X11 vs Wayland soft linking) but I'm also fine dealing with that in a future PR.

My proposal (which supports variables as well as functions, but is missing the WINAPI calling convention stuff): https://gist.github.com/dpogue/6ac36a0fdd78397211441825b4ad5405

colincornaby:
Looks like CI is still failing.

How do we want to balance the feedback with getting it into this PR vs future PRs? I'm holding off any Mac specific stuff here (especially since it's wrapped up in the Metal client.) I could also see that enhancements meant for the Linux client could be in future PRs.

Part of that is going to be that I'm probably not the right person to implement some of the features meant for Linux. So I'm not sure what of this feedback is actionable for me.

dpogue:
CI should work after rebasing.

I'm good with merging this and leaving any Linux/macOS follow-ups to future PRs.

Sources/Plasma/PubUtilLib/plPipeline/CMakeLists.txt
dpogue: We need to (conditionally on Windows, I guess) add `plWinDpi` as a private target link library here to fix build issues with the unit tests and 3DS Max plugin

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -146,6 +146,7 @@ target_link_libraries(
         plStatGather
         plStatusLog
         plUnifiedTime
+        $<$<BOOL:${WIN32}>:plWinDpi>
```
Hoikas, line 149:
```suggestion
```

This is pulled in transitively by plPipeline.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
Hoikas: This function is never used anywhere, so omitting should be fine for now, IMO.

Sources/Plasma/PubUtilLib/plPipeline/CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ target_link_libraries(plPipeline
...
```
Hoikas, line 74: Please separate 3rd party from the Plasma libraries. Also, please use the imported target `Freetype::Freetype` instead of `freetype`. When CMake sees the namespace delimiter `::`, it will try to resolve the dependency to a target that CMake knows about and will error if it is not found, providing a free configure-time check. Providing a library name (such as `freetype`) will silently pass the equivalent of `-lfreetype` to the linker if no matching CMake target is found, which can potentially result in surprising errors at link time.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -82,205 +90,153 @@ plTextFont::~plTextFont()
...
-        if( (uint32_t)( x + size.width + 1 ) > fTextureWidth )
+        if( (uint32_t)( x + size.cx + 1 ) > fTextureWidth )
         {
             x = 0;
-            y += size.height + 1;
+            y += size.cy + 1;
         }
+        int offset = int( fFontHeight - cBox.yMax );
+
+        //blit the bitmap into place
```
Hoikas, line 208:
```suggestion
        // blit the bitmap into place
```

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -82,205 +90,153 @@ plTextFont::~plTextFont()
...
+        for (int glyphY = bitmap.rows - 1; glyphY >= 0 ; glyphY--) {
+            for (int glyphX = 0; glyphX < bitmap.width; glyphX++) {
+                //1 bit Bitmap as source
```
Hoikas, line 211:
```suggestion
                // 1 bit Bitmap as source
```

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -82,205 +90,153 @@ plTextFont::~plTextFont()
...
-        CGPoint position = CGPointMake(x, -y - size.height - rect.origin.y);
-        CTFontDrawGlyphs(font, &glpyh, &position, 1, bitmapContext);
+                if ((cValue & (128 >> (glyphX & 7))) != 0) {
+                    src = UINT16_MAX;
+                }
+                int destY = y + glyphY + offset;
+                int destX = glyphX + x + face->glyph->bitmap_left;
+                int destIndex = (destY * fTextureWidth) + destX;
+                hsAssert( destX < fTextureWidth, "Destination X cannot be out of bounds");
+                hsAssert( destY < fTextureHeight, "Destination Y cannot be out of bounds");
```
Hoikas, line 225:
```suggestion
                hsAssert(destX < fTextureWidth, "Destination X cannot be out of bounds");
                hsAssert(destY < fTextureHeight, "Destination Y cannot be out of bounds");
```

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -82,205 +90,153 @@ plTextFont::~plTextFont()
...
     
     hDC = CreateCompatibleDC(nullptr);
-    hBitmap = CreateDIBSection(hDC, &bmi, DIB_RGB_COLORS, (void **)&bitmapBits, nullptr, 0);
     SetMapMode( hDC, MM_TEXT );
+    // Get the font data
 
-    nHeight = -MulDiv( fSize, GetDeviceCaps( hDC, LOGPIXELSY ), 72 );
-    fFontHeight = -nHeight;
-
+    int nHeight = -MulDiv( fSize, plWinDpi::Instance().GetDpi(), 72);
```
Hoikas, line 111: Is this ever re-done when the DPI changes while the game is running?

CMakeLists.txt
Hoikas: Bump

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
colincornaby: DPI change is treated as a device reset IIRC just like any other resolution change - which should flush textures like this one.

I don't know that this path has ever been fully tested however.

Sources/Plasma/PubUtilLib/plPipeline/CMakeLists.txt
colincornaby: The namespacing doesn't work. Findfreetype.cmake seems to muck with this.

dpogue: yeah, this is (unfortunately) consistent with what plFontConverter is doing: https://github.com/H-uru/Plasma/blob/fd591cbd88f538fc88ff2eb66a8ebb78f9085022/Sources/Tools/plFontConverter/CMakeLists.txt#L44

Hoikas: I had forgotten that upstream freetype exports only `freetype`. CMake's find modules expose the `Freetype::Freetype` target, but we will generally not get that due to loading the freetype cmake config.

Hoikas:
Still waiting on the update to cmake/Dependencies.cmake, then we should be good to go.

colincornaby:
> Still waiting on the update to cmake/Dependencies.cmake, then we should be good to go.

Not seeing that comment, guessing it's just buried in the others. But I'm guessing the change you'd like to see is FreeType moving to required from optional?

Hoikas:
[Yes](https://github.com/H-uru/Plasma/pull/1269/files#r1186971545), per @dpogue

Hoikas (approved):

dpogue (approved):

PR #1270, opened 2022-12-06
Additional Cyan Trailer Cameras

Hazado:
Files needed for Cyan Trailer camera pages to work


Hoikas (commented):
It looks like the cleft trailer camera page is set to not load by default (it has `,1` after its page number in the .age file). We should probably do the same thing for these.

Hazado:
only problem I see with that is most of these ages don't page load at all.

If the goal is to not have them loaded in normal play at all, ill adjust it.

Hazado:
Okay pushed changes to make all the trailer cam pages page loaded.
Also removed the page from any python that does page loading currently.
It shouldn't load in normal gameplay ever.
You now need to remove the `,1` from the age file, or add the page to the page loading script.

Hoikas (approved):

PR #1276, opened 2022-12-16
Run macOS builds with vcpkg again

dpogue:
This sets up the workflow to do arm64 builds in the future, but they're disabled now because of issues getting python3 to build with vcpkg for arm64.

We originally moved from vcpkg to homebrew for macOS dependencies because of issues with vcpkg caching and slow build times when macOS had to recompile several large dependencies (PhysX, OpenSSL, Python, etc.) on every single build. With the caching fixed, these builds should run much more quickly now (after the first run), comparable to the runtime with homebrew dependencies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,7 +245,8 @@ jobs:
     strategy:
       matrix:
         platform:
-          - { str: macos-x64 }
+          #- { str: macos-arm64, arch: arm64, triplet: arm64 }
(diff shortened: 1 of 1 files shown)


dpogue:
> I wonder if the Python problem on Mac on ARM was fixed by [microsoft/vcpkg#27830](https://github.com/microsoft/vcpkg/pull/27830)

Just tested and it still failed to build 😞

Hoikas (approved):

PR #1278, opened 2022-12-17
Fix OpenSSL 3 deprecation warnings from plBigNum

dpogue:
The "fast" BigNum prime check function is deprecated in OpenSSL 3 because it apparently doesn't do a sufficient job of testing prime-ness. This newer method does it properly, which probably means it's slower, but probably not to a degree that it noticeably impacts performance (if this method is called at all, which it doesn't seem to be?).

Also, the `Div` method was complaining about a ternary that tested whether `this` was null, which isn't possible in well-defined C++ code. I tried to look at where this was being used to see how that could happen and it turns out the function wasn't being used at all, so I just removed it entirely.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnEncryption/plBigNum.h
+++ b/Sources/Plasma/NucleusLib/pnEncryption/plBigNum.h
@@ -113,14 +113,6 @@ public:
         *remainder = (uint32_t)BN_div_word(m_number, b);
     }
 
-    void Div(const plBigNum& a, const plBigNum& b, plBigNum* remainder)
-    {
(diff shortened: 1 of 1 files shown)


zrax (approved):

zrax:
Hmm, I thought I had done this same change before...  But it looks like it was part of a work in progress to port to the EVP APIs that I never completed due to one of the algorithms not being usable from EVP.

dpogue:
Yeah, RC4 isn't supported by the EVP APIs unless you enable some legacy mode that apparently isn't enabled by default on  some builds of OpenSSL 😕

PR #1279, opened 2022-12-17
Try to automatically init submodules with CMake

dpogue:
This will hopefully allow people to just use Visual Studio to clone the main Plasma repo, and have it deal with git submodule initialization as part of the build process, rather than needing people to manually run a git command which clones submodules recursively.

### DIFF
--- a/Sources/Tests/CMakeLists.txt
+++ b/Sources/Tests/CMakeLists.txt
@@ -1,3 +1,9 @@
+include(SubmoduleUtils)
+plasma_init_git_submodule("Sources/Tests/gtest" RESULT has_gtest_module)
+if(NOT has_gtest_module)
+    return()
+endif()
+
 # Why is this OFF by default?  It doesn't work with CMake's defaults, and
 # should really only be off if the project has specifically been set up to
 # use the static MSVC CRT
--- a/cmake/SubmoduleUtils.cmake
+++ b/cmake/SubmoduleUtils.cmake
@@ -1,0 +1,34 @@
+# Some optional dependencies are included as git submodule, and there's a
+# chance the repo wasn't cloned with submodules recursively. In that case, we
+# want to try to automatically initialize them.
+function(plasma_init_git_submodule MODULE)
+    cmake_parse_arguments(PARSE_ARGV 1 _pigs "" "RESULT" "")
+
+    if(EXISTS "${CMAKE_SOURCE_DIR}/${MODULE}/.git")
+        set(return_value TRUE)
+    else()
+        # First, we need to see if we have git available...
+        find_package(Git QUIET)
+
+        if(Git_FOUND)
+            execute_process(
+                COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive -- ${MODULE}
+                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+                RESULT_VARIABLE RETURNCODE
+            )
+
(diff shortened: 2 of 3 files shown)


Sources/Tests/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+if(NOT EXISTS "${CMAKE_SOURCE_DIR}/Sources/Tests/gtest/.git")
+    # gtest is included as a git submodule, and there's a chance the repo
+    # wasn't cloned with submodules recursively. In that case, we want to try
+    # to automatically initialize it.
+    find_package(Git REQUIRED QUIET)
```
Hoikas, line 5: While this is in fact required, I think that the error message this will produce (eg something like `can't find package Git`) might be a little misleading. The problem isn't really that git is missing - rather that the submodule hasn't been initialized. I wonder, then, if we should raise a custom error if git isn't found when the submodule is not initialized.

Hoikas (commented):
This is a good idea. I feel like we might want to make these two snippets into a single CMake function to reduce code duplication, though.

dpogue:
I've turned this into a function that tries to more gracefully fail when git can't be found

cmake/SubmoduleUtils.cmake
```diff
@@ -0,0 +1,34 @@
...
+            else()
+                set(return_value TRUE)
+            endif()
+        else()
+            message("Could not automatically initialize git submodule at ${MODULE}")
+            set(return_value FALSE)
+        endif()
+    endif()
+
+    if (_pigs_RESULT)
```
Hoikas, line 31:
```suggestion
    if(_pigs_RESULT)
```

cmake/SubmoduleUtils.cmake
```diff
@@ -0,0 +1,34 @@
...
+        find_package(Git QUIET)
+
+        if(Git_FOUND)
+            execute_process(
+                COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive -- ${MODULE}
+                WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}"
+                RESULT_VARIABLE RETURNCODE
+            )
+
+            if (RETURNCODE)
```
Hoikas, line 20:
```suggestion
            if(RETURNCODE)
```

cmake/VcpkgToolchain.cmake
```diff
@@ -22,6 +22,12 @@ if(NOT USE_VCPKG)
     return()
 endif()
 
+include(SubmoduleUtils)
+plasma_init_git_submodule("vcpkg" RESULT has_vcpkg_module)
+if(NOT has_vcpkg_module)
+    return()
+endif()
```
Hoikas, line 29: To make errors a little bit clearer, we might want to move this above the "Welcome to vcpkg-land" bit, and change `option(USE_VCPKG` to use [`cmake_dependent_option()`](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) to clear out `USE_VCPKG` as a hint that vcpkg can't be enabled.

dpogue: I had been hoping to avoid cloning the submodule if vcpkg hadn't been turned on

Hoikas: Hmm... Maybe we can defer this for another addition to the submodule tools.

Hoikas (approved):

PR #1280, opened 2022-12-21
Move plClient Windows-specific code to win32 folder

dpogue:
Eventually (hopefully soon) we'll have `macos` and `linux` folders with those client implementations, so let's gather all the Windows-specific stuff in one place.


Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -56,8 +56,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include <curl/curl.h>
 
-#include "plClient.h"
-#include "plClientLoader.h"
+#include "../plClient.h"
+#include "../plClientLoader.h"
```
Hoikas, line 60: Maybe we should add the `plClient` directory as `target_include_directories()` of the plClient target.

dpogue: I think that might already work... I started that way and then thought it was weird for plClient to be referring to itself in the 3rd person, but I vaguely recall `..` is bad for performance in include paths?

Hoikas: Possibly. I'd just prefer to avoid `../` if at all possible because it makes things a bit confusing, IMO.

zrax: Adding extra `target_include_directories()` is actually vastly preferable to using `../`, since the latter will have to search the parent of ALL possible include paths, whereas the former will only add a single extra search path.

dpogue: Updated to use `plClient/` instead of `../`

Hoikas: What @zrax and I were suggesting was to do:
```cmake
target_include_directories(plClient PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
```
so that no matter what subfolder of `plClient` we're in, we can include things from the `plClient` directory.

dpogue: Ahhh okay.

On some level that probably risks confusions about it not being clear where headers are coming from, but this is a small and specific enough case that it seems fine.

dpogue:
Updated

Hoikas (approved):

PR #1285, opened 2022-12-30
Fix DX Pipeline externed global variables

dpogue:
There are two global variables defined in plDXPipeline that are used by the Console.

* One is the maximum blur scale for shadows. There are already several configurable values in plShadowMaster, so I've just added this as another value there so it has a proper home instead of being referenced via extern.

* The other is a debug flag that restricts the rendering capabilities of the DirectX pipeline to that of a GeForce 2 card. This seems pretty DirectX-specific, so I opted to just wrap the console commands in `#ifdef` so that they are only enabled when the DX pipeline is enabled (and thus the externed variable is defined).

These were showing up as linker when I tried to compile a Linux build with plGLPipeline, because they were referenced from the console commands.

/cc @colincornaby because I saw you had to declare these in plMetalPipeline too, even though they serve no purpose there.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
@@ -1089,7 +1089,7 @@ PF_CONSOLE_CMD( Graphics, BumpW, "", "Set bump mapping method to cheapest availa
     pfConsole::GetPipeline()->SetDebugFlag(plPipeDbg::kFlagBumpW, true);
 }
 
-
+#ifdef PLASMA_PIPELINE_DX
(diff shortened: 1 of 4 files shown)


Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -1240,10 +1241,12 @@ PF_CONSOLE_CMD( Graphics_Shadow,
                "...", 
                "Max shadowmap blur size." )
 {
-    extern float blurScale;
-    if( numParams > 0 )
-    {
-        blurScale = (float)atof( params[ 0 ] );
+    float blurScale;
+    if (numParams > 0) {
+        blurScale = (float)atof(params[0]);
```
Hoikas, line 1246: While we're here...

```suggestion
        blurScale = strtof(params[0]);
```

Hoikas (approved):

PR #1287, opened 2022-12-31
Fixing endian encoding on *nix systems for UUIDs

colincornaby:
Windows encodes the UUID fields in little endian, but *nix encodes as an array of unsigned chars (which is a big endian encode.) This was leading to unexpected failures against the Cyan patch server when the product UUID was sent over the wire.

This change makes it so that UUIDs are consistently encoded internally in Plasma. A little endian encoding will always be used in Plasma because UUIDs are encoded directly into network packets, and the server expects a little endian encoding.

I don't know what implications this could have for Windows/ARM. I think Windows/ARM is little endian as well - but a lot of this is baking in assumptions about how Windows encodes it's GUID structure and building an ad hoc standard off of that.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
+++ b/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
@@ -50,16 +50,41 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ 
 
 struct plUUIDHelper
 {
+    typedef struct _GUID {
+      uint32_t Data1;
+      uint16_t Data2;
+      uint16_t Data3;
+      uint8_t  Data4[8];
+    } GUID;
+    
     static inline void CopyToPlasma( plUUID * dst, const uuid_t & src )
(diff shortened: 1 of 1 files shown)


Hoikas:
As I understand the Microsoft uuids, we only want to perform the byteswap on little endian machines, so we'll want to use the hsToBE macro instead of unconditionally byteswapping.

colincornaby:
> As I understand the Microsoft uuids, we only want to perform the byteswap on little endian machines, so we'll want to use the hsToBE macro instead of unconditionally byteswapping.

uuid_t should always be big endian implicitly in since it encodes each 8 bit hex value into a sequential array. fData is always little endian because that's what the server expects. I think - even if the host were big endian - we'd still need to swap. The source uuid_t is always big endian and the server expects a little endian GUID structure.

So _I think_ we'd always want to swap. But I could be missing something.

dpogue (approved):
This makes sense to me.

You do always need to swap, regardless of machine architecture, otherwise it won't match what the server is expecting.

Hoikas (approved):

PR #1288, opened 2022-12-31
Code for age reset mechanic from books

Hazado:
PRP needed - https://github.com/H-uru/moul-assets/pull/200

Adds code so that you can use a bookmark from the linking book itself to reset the age back to default.

### DIFF
--- a/Scripts/Python/xLinkingBookGUIPopup.py
+++ b/Scripts/Python/xLinkingBookGUIPopup.py
@@ -296,7 +296,13 @@ class xLinkingBookGUIPopup(ptModifier):
                                 PtSetShareAgeInstanceGuid(pelletCaveGUID)
                             ClosedBookToShare = 1
                             self.HideBook()
-
+                        elif event[2] == xLinkingBookDefs.kShareBookDeleteID:
+                            ageName = self.IGetAgeFilename()
+                            PtDebugPrint("xLinkingBookGUIPopup:Book: hit delete bookmark for age '%s'" % (ageName))
+                            if ageName == "Neighborhood":
+                                PtYesNoDialog(self.key, PtGetLocalizedString("Personal.Bookshelf.DeleteNeighborhoodBook"))
+                            else:
+                                PtYesNoDialog(self.key, PtGetLocalizedString("Personal.Bookshelf.DeleteBook"))
                         elif event[2] >= xLinkingBookDefs.kFirstLinkPanelID or event[2] == xLinkingBookDefs.kBookMarkID:
                             PtDebugPrint("xLinkingBookGUIPopup:Book: hit linking panel %s" % (event[2]))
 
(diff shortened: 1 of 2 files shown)


Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -409,6 +415,48 @@ def OnNotify(self,state,id,events):
...
+                                    PtDebugPrint("psnlBookshelf.IGetLinkFromBook():\tSet Age to Volatile", ageName)
+                                    link.setVolatile(True)
+                                    link.save()
+                                
+                            else:
+                                PtDebugPrint("psnlBookshelf.IGetLinkFromBook():\tSet Age to Volatile", ageName)
+                                link.setVolatile(True)
+                                link.save()
+                        else:
+                            PtDebugPrint("xLinkingBookGUIPopup:Book:\tunable to find Owned link", ageName)
```
Hoikas, line 459: I realize this is based on some of Cyan's other code, but it's very non-optimal because it is very difficult to read, and the functionality it's trying to implement is already supported by the C++ side. So, I suggest:

```suggestion
                elif event[0] == kVariableEvent:
                    if event[1] == "YesNo" and event[3] == 1:
                        ageName = self.IGetAgeFilename()
                        vault = ptVault()
                        ageStruct = ptAgeInfoStruct()
                        ageStruct.setAgeFilename(ageName)

                        if link := vault.getOwnedAgeLink(ageStruct):
                            PtDebugPrint(f"xLinkingBookGUIPopup.OnNotify():\tfound Owned link for deletion {ageName=}", level=kDebugDumpLevel)

                            if ageName in {"Ahnonay", "AhnonayCathedral"}:
                                ahnonayAgeStruct = ptAgeInfoStruct()
                                ahnonayAgeStruct.setAgeFilename("Personal")

                                if ageLinkNode := vault.getOwnedAgeLink(ahnonayAgeStruct):
                                    if ageInfoNode := ageLinkNode.getAgeInfo():
                                        ageDataTemplate = ptVaultFolderNode()
                                        ageDataTemplate.folderSetName("AgeData")
                                        if ageDataFolder := ageInfoNode.findNode(ageDataTemplate):
                                            chronTemplate = ptVaultChronicleNode()
                                            chronTemplate.chronicleSetName("AhnonayVolatile")
                                            if (chron := ageDataFolder.findNode(chronTemplate)) and (chron := chron.upcastToChronicleNode()):
                                                chron.chronicleSetValue("1")
                                            else:
                                                PtDebugPrint(f"xLinkingBookGUIPopup.OnNotify():\tHmmm... The AhnonayVolatile chronicle is missing! This deletion may cause oddness.")
                                        else:
                                            PtDebugPrint(f"xLinkingBookGUIPopup.OnNotify():\tHmmm... The AgeData chronicle is missing! This deletion may cause oddness.")
                                else:
                                    PtDebugPrint("xLinkingBookGUIPopup.OnNotify():\tHmmm... The Relto AgeLink is missing!")
                            if ageName != "Ahnonay":
                                PtDebugPrint(f"xLinkingBookGUIPopup.OnNotify():\tMarking {ageName=} volatile.", level=kWarningLevel)
                                link.setVolatile(True)
                                link.save()
                        else:
                            PtDebugPrint(f"xLinkingBookGUIPopup.OnNotify():\tHmmm... The {ageName=} AgeLink is missing!")
```

I removed the check for `vault.inMyPersonalAge()` because it seems that we need to mark Ahnonay as volatile no matter where we delete it. I suspect that this was an attempt by Cyan to ensure that only the owner of the current Relto (assuming you copied and pasted this from the Relto bookshelf) would be able to delete Ahnonay. If so, that logic was incorrect in that not marking Ahnonay as volatile would leave things in a weird discombobulated state.

Hazado:
So im tempted to also off a method to clear any journey cloths the age has when you delete the age.

Shouldnt be hard to do but its a step  further than the current method on the relto shelf.

Hoikas (approved):
I think I would hold off on journey cloth resetting for now.

PR #1292, opened 2023-01-06
Replace last uses of MSVC-specific `_DEBUG` with `HS_DEBUGGING`

dgelessus:
This allows removing some compiler-specific CMake code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,10 @@ option(PRODUCT_EMBED_BUILD_INFO "Embed build revision information into plProduct
 cmake_dependent_option(PRODUCT_EMBED_BUILD_TIME "Embed build time into plProduct" ON [[PRODUCT_EMBED_BUILD_INFO]] OFF)
 
 # HeadSpin Configuration
+
+# Define HS_DEBUGGING for debug builds
+add_compile_definitions($<$<CONFIG:Debug>:HS_DEBUGGING>)
+
 if(WIN32 AND NOT CYGWIN)
     add_definitions(-DHS_BUILD_FOR_WIN32)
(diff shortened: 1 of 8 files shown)


dgelessus:
Clearly I didn't grep right the first time and missed a few more uses of `_DEBUG`, sorry about that.

But if I'm reading the Linux CI failures correctly, the PhysX headers expect `_DEBUG` to be defined in debug mode, even for non-MSVC compilers? In that case, the extra CMake code probably can't be removed after all.

Hoikas:
We should probably move the definition into the PhysX target in that case.

dgelessus:
Should both be fixed now. The `PhysX` library targets now define `_DEBUG` as needed. All uses of `_DEBUG` are gone, except for one in Sources/Tools/MaxPlasmaLights/plRTLights.rc, because `HS_DEBUGGING` isn't available there and that code is Windows-specific anyway.

dpogue (approved):

CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ option(PRODUCT_EMBED_BUILD_INFO "Embed build revision information into plProduct
 cmake_dependent_option(PRODUCT_EMBED_BUILD_TIME "Embed build time into plProduct" ON [[PRODUCT_EMBED_BUILD_INFO]] OFF)
 
 # HeadSpin Configuration
+
+# Define HS_DEBUGGING for debug builds
+add_compile_definitions($<$<CONFIG:Debug>:HS_DEBUGGING>)
```
Hoikas, line 39: Ideally we would be moving away from these scoped commands and toward target based stuff, but there's already a precedent for this below, so I don't think this is a blocker.

Hoikas (approved):

PR #1293, opened 2023-01-08
Guard against invalid data when reading vault nodes

dgelessus:
Previously the game crashed when reading truncated vault node data.

I implemented these checks in a similar way as in the other `IRead*` functions further up in pnNpCommon. Perhaps these two sets of functions should be merged... Or maybe all of these functions should be replaced with `hsReadOnlyStream`/`hsRAMStream`? (Most of the `hsStream::Read*` methods don't really handle truncated data either, but at least they don't crash right away.)

### DIFF
--- a/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
@@ -59,7 +59,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 # pragma message("Compiling pnNetCli with debugging on")
 # define NCCLI_LOG  LogMsg
 #else
-# define NCCLI_LOG  LogMsg
+# define NCCLI_LOG(...) ((void)0)
 #endif
 
 #if !defined(PLASMA_EXTERNAL_RELEASE) && defined(HS_BUILD_FOR_WIN32)
--- a/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
@@ -54,9 +54,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <mutex>
 #include <string>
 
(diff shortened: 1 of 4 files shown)


Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -54,14 +54,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <mutex>
 #include <string>
 
-//#define NCCLI_DEBUGGING
-#ifdef NCCLI_DEBUGGING
-# pragma message("Compiling pnNetCli with debugging on")
-# define NCCLI_LOG  LogMsg
-#else
-# define NCCLI_LOG  LogMsg
-#endif
```
dpogue, line 63: The fact that it was logging in both cases appears to have been my fault: https://github.com/H-uru/Plasma/commit/8e05891be0e16bb5e6a79956a5f054bed4a1d596

We probably do want conditional logging...

Hoikas: That explains why our OLD_ASYNC_LOG files are always so big 🤪

dgelessus: Ah okay, makes sense. I think I was also confused because most logging in pnNcCli calls `LogMsg` directly without going through the macro. But looking at it again, the logic seems to be that unconditional `LogMsg` is used for errors and the conditional `NCCLI_LOG` for non-error debug messages.

dpogue: My guess is that `__VA_ARGS__` would solve whatever issues MinGW had circa-2012...

```cpp
// Do we need a separate non-intuitive NCCLI define for this?
#ifdef HS_DEBUGGING
#   define NCCLI_LOG(...) LogMsg(__VA_ARGS__)
#else
#   define NCCLI_LOG(...)
#endif
```

Hoikas (approved):

PR #1295, opened 2023-01-08
Remove `HSMemory::*` in favor of standard memory manipulation functions

dgelessus:
All of these custom functions have very close equivalents in the C standard library, so use those instead for simplicity.

I don't have 3DS Max, so I can't test my changes to that part of the code, but the changes are pretty straightforward.


Sources/Plasma/PubUtilLib/plGImage/plMipmap.cpp
```diff
@@ -1448,8 +1448,8 @@ void plMipmap::Filter(float sig)
     {
         uint8_t *dst = (uint8_t *)(fImage);
 
-        uint8_t* src = (uint8_t*)HSMemory::New(fRowBytes * fHeight);
-        HSMemory::BlockMove(dst, src, fRowBytes * fHeight);
+        uint8_t* src = new uint8_t[fRowBytes * fHeight];
```
zrax, line 1451: Since this is just local to this function, an even better choice would be to use a `std::unique_ptr<uint8_t[]>` or even a `std::vector<uint8_t>`

zrax (approved):

PR #1297, opened 2023-01-11
Fix mis-encoded text when starting chat with a non-ASCII character

dgelessus:
To reproduce:

1. Use a keyboard layout with non-ASCII keys
2. Have "Start Chat" not mapped to anything in the Key Map
3. Be in-game without chat open
4. Start chatting by pressing a non-ASCII key

Without this fix, the first character was incorrectly re-encoded from UTF-8 to Latin 1 (e. g. "ä" became "Ã¤").

### DIFF
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -174,7 +174,7 @@ class xKIChat(object):
             if not KIMini.dialog.isEnabled():
                 self.ClearBBMini(0)
             if firstChar:
-                chatEdit.setString(firstChar)
+                chatEdit.setStringW(firstChar)
                 chatEdit.end()
             else:
                 chatEdit.clearString()


dgelessus:
I don't quite understand why this fix is necessary though. I thought that `setString`, etc. in the H'uru codebase were meant to support Unicode properly as well, and that you don't need to call `setStringW` explicitly? See e. g. this previous discussion: https://github.com/H-uru/Plasma/pull/856#discussion_r589747003

> With the update to Python 3, the difference between the getString, getStringW, setString, and setStringW methods has been erased, so ignore them

And indeed various parts of the KI code use `setString` and still seem to work properly with non-ASCII text. In this case, `setString` is definitely mishandling the Unicode string though...

Any reason not to make `setString`, etc. Unicode-native and just get rid of `setStringW`, etc.?

Hoikas:
The content of these methods is the same....

https://github.com/H-uru/Plasma/blob/938bc36fcbc92d8a8e432bc7c0c56c180ba97cbb/Sources/Plasma/FeatureLib/pfPython/pyGUIControlTextBoxGlue.cpp#L75-L97

dgelessus:
Ah, apparently only `ptGUIControlTextBox` was fully migrated to Unicode then. `ptGUIControlEditBox`, which the KI chat box uses, still has different implementations for `setString` and `setStringW`:

https://github.com/H-uru/Plasma/blob/3d352d274637841a31af61c5067f26b9667c23d7/Sources/Plasma/FeatureLib/pfPython/pyGUIControlEditBoxGlue.cpp#L100-L132

Same for `ptGUIControlMultiLineEdit` apparently.

So I suppose the long-term fix is to make those classes' non-`W` methods use the Unicode implementation as well?

Hoikas:
Whoops, I looked at the wrong class.

> So I suppose the long-term fix is to make those classes' non-W methods use the Unicode implementation as well?

Yeah, ideally these methods are just a pointer to the same code instead of being carbon copies... or, better yet, deleted entirely.

Hoikas (approved):

PR #1299, opened 2023-01-16
Show Share Book symbol when a book is shared with you

Hazado:
This makes the share book symbol show on a book that is shared with you.
The symbol is not interactible, this is to have the book match what you would see normally.

Before:
image
After:
image


Deledrius:
What's the goal with this change?  Consistency of appearance?  It seems to be at the cost of consistency of interface, because you're now presenting the user with an unclickable clickable.  I suspect this was an intentional choice that was made originally at the expense of a small amount of suspension of disbelief.

Hazado:
Goal is to have the in-game model match the GUI.
The player is presented another players linking book as it appears in world to both players.
Before you would be presented with a book that is not visible anywhere in the game usually.

Yes it now has the sharable symbol no longer have a clickable. I prefer that to showing me a book that does not exist as I see it. The suspension of disbelief is much smaller in this instance now than before in my opinion. 
Not being able to have the symbol be clickable vs it disappearing altogether.

Deledrius:
The original method showed you a slightly different book, which has the OOC effect of communicating through the UI that it is a single-purpose game element.  It is not realistic that its appearance is different from the in-world book, no, but it is simplifying the interface for the user in order to communicate something (Uru doesn't always get this right, but here I feel it does).  I acknowledge that this has always felt a bit odd.

While it would be preferable to find some way to accomplish both things, I don't think that showing an interactable UI element that's not usable is itself an improvement.  Of course, making it greyed-out, or changing the text of the "Share Book" verbiage to amend this would land back in the same problem of showing a different book than the one that's visible in-world.

I don't see any way to make it both visually consistent *and* present a clean UI that isn't lying to the player at the same time, so this isn't an improvement, it's simply an opposing design opinion on which to prioritize.

Hazado:
> I don't think that showing an interactable UI element that's not usable is itself an improvement. 
This change makes it so the shareable page, has no interactable element on mouseover anymore for the person its shared too.
Just wanted to clarify that point. 

> it's simply an opposing design opinion on which to prioritize.
I agree with you here. I felt removing the interactable element from the page while not changing the look is more inline with other design elements of the game. With additional coding we could include the shareable page working when the book is presented to the other player, but then you get in a weird fight between what book is being shared to other players.

I could remove everything from the other page if that would be a better experience. So that the shareable page is not visible, but a blank page instead and you only see the linking page.

Hoikas:
I'm in agreement with @Deledrius that the status quo is better. To be honest, I've always felt that the in-game GUIs are OOC. Book sharing and instancing a gameplay mechanics, IMO, so that stuff has no place in the actual game world itself. Unfortunately, Cyan disagrees on the latter. That being said, to my mind, it would make the most sense, to me, for share book symbols to not be visible on the in-world books. But, that's a very different PR. I'm thinking we may want to close this one?

PR #1300, opened 2023-01-19
Add Video Selector buttons for Video Options Menu GUI

Hazado:
PRP and other required files - https://github.com/H-uru/moul-assets/pull/206

Adds code for two new buttons in Trailer GUI
First video in AVI directory is now first video played
Ignores any videos smaller than 1000 bytes
Videos will loop if you keep clicking buttons
GUI will exit when video is over (retains previous behaviour)

### DIFF
--- a/Scripts/Python/xOptionsMenu.py
+++ b/Scripts/Python/xOptionsMenu.py
@@ -1002,7 +1002,7 @@ class xOptionsMenu(ptModifier):
                             gCurrentMovieName = gLiveMovieList[len(gLiveMovieList)-1]
                         else:
                             gCurrentMovieName = gLiveMovieList[tempIndex]
-                        gLiveMovie = ptMoviePlayer(f'{kLiveMovieDir}\{gCurrentMovieName}',self.key)
+                        gLiveMovie = ptMoviePlayer(os.path.join(kLiveMovieDir, gCurrentMovieName),self.key)
                         gLiveMovie.play()
                 elif tpID == kTrailerNextButton:
                     if gLiveMovie:
@@ -1012,7 +1012,7 @@ class xOptionsMenu(ptModifier):
                             gCurrentMovieName = gLiveMovieList[0]
                         else:
                             gCurrentMovieName = gLiveMovieList[tempIndex]
-                        gLiveMovie = ptMoviePlayer(f'{kLiveMovieDir}\{gCurrentMovieName}',self.key)
+                        gLiveMovie = ptMoviePlayer(os.path.join(kLiveMovieDir, gCurrentMovieName),self.key)
                         gLiveMovie.play()
                 elif tpID == kClickOnMeBtn:
                     if gLiveMovie:
@@ -1405,7 +1405,8 @@ class xOptionsMenu(ptModifier):
                 else:
                     gLiveMovieList = []
                     for file in os.listdir(kLiveMovieDir):
-                        if os.stat(f'{kLiveMovieDir}\{file}').st_size > 1000 and os.path.splitext(f'{kLiveMovieDir}\{file}')[1] == ".webm":
+                        tempfile = os.path.join(kLiveMovieDir, file)
+                        if os.stat(tempfile).st_size > 1000 and os.path.splitext(tempfile)[1] == ".webm":
                             gLiveMovieList.append(file)
                     else:
                         if gLiveMovieList == []:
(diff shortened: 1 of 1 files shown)


Scripts/Python/xOptionsMenu.py
```diff
@@ -1403,7 +1433,12 @@ def OnTimer(self,id):
             if PtIsDemoMode():
                 gLiveMovie = ptMoviePlayer(kDemoMovieName,self.key)
             else:
-                gLiveMovie = ptMoviePlayer(kLiveMovieName,self.key)
+                gLiveMovieList = []
+                for file in os.listdir(kLiveMovieDir):
+                    if os.stat(f'{kLiveMovieDir}\{file}').st_size > 1000:
```
Deledrius, line 1438: Is there a specific reason for this particular choice in filesize?

Scripts/Python/xOptionsMenu.py
```diff
@@ -1403,7 +1433,12 @@ def OnTimer(self,id):
...
+                        gLiveMovieList.append(file)
+                gCurrentMovieName = gLiveMovieList[0]
```
Deledrius, line 1440: This will fail if there are no video files present.

Deledrius (changes requested):

Scripts/Python/xOptionsMenu.py
Hazado: Line 1406 will cause it to error if no video files are present and prevent this code from being called

Scripts/Python/xOptionsMenu.py
Hazado: Moula has a file called "PreBranding.webm" that is 826 bytes, this was to prevent that from being shown and causing the Trailer GUI to exit immediately due to it closing when movies end.

Scripts/Python/xOptionsMenu.py
Deledrius: Ah, so it does.  The code folding in the diff helped me miss it.

Deledrius: Actually, I'm not so certain.  I believe the code at 1406 will fail to guard against this if there are files present but they are all below the size threshold.  It's also possible now for the movie list to contain files that aren't movies if they exist in this folder.

Scripts/Python/xOptionsMenu.py
Deledrius: It might be a good idea to specifically exclude that video if that's the purpose rather than implicitly by some heuristic.

Filtering out dot files is probably a good idea to do as well.  I think a whitelist of supported video extensions might be the better route for that.

Hazado: I don't want to specifically exclude it because there may be a time when the PreBranding video actually has legit video to play, and it will then be larger than 1000.
Right now the code will also play the cyan worlds intro.

Yeah, I agree that this code needs some more work to only allow supported video extensions

Hazado: Okay after much struggle, ive got code that only allows webm files greater than 1000 bytes.
Tested with No files, Folder ending in ".webm", and a text file in the avi folder

Hoikas:
These changes will error on non-Windows platforms due to the hard-coded use of Windows style paths (ie using blackslashes `\`).

Hazado:
> These changes will error on non-Windows platforms due to the hard-coded use of Windows style paths (ie using blackslashes `\`)

I used the wrong slash!!

Deledrius:
`os.path.join` is a great way to let Python handle it. :)

Hoikas:
Please use `pathlib` instead of `os.path`.

Deledrius:
Ah, yes.  Even better!

Hoikas:
After discussion, `pathlib` cannot be used in Plasma because it imports `urllib`, which imports `socket`, which is disallowed.

Hazado:
Okay, pushed a change I think will fix non-windows platforms breaking

dgelessus:
> After discussion, `pathlib` cannot be used in Plasma because it imports `urllib`, which imports `socket`, which is disallowed.

Does it? AFAICT, `pathlib` only imports `urllib.parse` and not e. g. `urllib.request`, so `socket` should never be imported.

Scripts/Python/xOptionsMenu.py
```diff
@@ -1373,7 +1403,14 @@ def OnTimer(self,id):
                 if PtIsDemoMode():
                     os.stat(kDemoMovieName)
                 else:
-                    os.stat(kLiveMovieName)
+                    gLiveMovieList = []
+                    for file in os.listdir(kLiveMovieDir):
+                        tempfile = os.path.join(kLiveMovieDir, file)
+                        if os.stat(tempfile).st_size > 1000 and os.path.splitext(tempfile)[1] == ".webm":
```
dgelessus, line 1409:
```suggestion
                        if os.stat(tempfile).st_size > 1000 and tempfile.endswith(".webm"):
```

Hoikas:
@Deledrius do the latest revisions resolve your concerns?

Deledrius (approved):

PR #1303, opened 2023-01-22
Fix potentially catastrophic Age init errors related to being alone.

Hoikas:
A common anti-pattern for determining whether or not we are the only person in an Age is to check the player list returned by `PtGetPlayerList()`. Generally, if this list is empty, some initialization to a default state is done or some restoration work is done if it is not empty. What the gameplay programmers seem to have not realized, though, is that the list returned by `PtGetPlayerList()` [omits any players that are in your ignore list](https://github.com/H-uru/Plasma/blob/2136c25ed7b8bc16183ece2af62b446007a07ad2/Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp#L1076). Therefore, the wrong initialization path may be executed if there are players in the Age, but all of them are in your ignore list. This could have consequences that range from odd but amusing to catastrophic, depending on the script in question.

Therefore, we fix this problem by introducing `PtIsSolo()`, which tells us whether or not we are the only player in the Age. This function is not vulnerable to ignore shennanigans, and we avoid creating lots of temporary `ptPlayer` objects that we don't actually care about.

### DIFF
--- a/Scripts/Python/psnlBugs.py
+++ b/Scripts/Python/psnlBugs.py
@@ -110,7 +110,7 @@ class psnlBugs(ptResponder):
         # check for all the cases where it would be raining, and if its not then turn on bugs
         sdl = PtGetAgeSDL()
         bugState = sdl["psnlBugsVis"]
-        if rainState == 1 or (rainState == 4 and len(PtGetPlayerList()) == 0) or (rainState == 3 and len(PtGetPlayerList()) > 0):
+        if rainState == 1 or (rainState == 4 and PtIsSolo() or (rainState == 3 and not PtIsSolo())):
             PtDebugPrint("turning off bugs")
             if bugState != 0:
                 sdl["psnlBugsVis"] = (0,)
--- a/Scripts/Python/psnlBugs.py
+++ b/Scripts/Python/psnlBugs.py
@@ -110,7 +110,7 @@ class psnlBugs(ptResponder):
         # check for all the cases where it would be raining, and if its not then turn on bugs
         sdl = PtGetAgeSDL()
         bugState = sdl["psnlBugsVis"]
-        if rainState == 1 or (rainState == 4 and PtIsSolo() or (rainState == 3 and not PtIsSolo())):
+        if rainState == 1 or (rainState == 4 and PtIsSolo()) or (rainState == 3 and not PtIsSolo()):
             PtDebugPrint("turning off bugs")
             if bugState != 0:
                 sdl["psnlBugsVis"] = (0,)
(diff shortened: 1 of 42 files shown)


Scripts/Python/psnlBugs.py
```diff
@@ -110,7 +110,7 @@ def OnServerInitComplete(self):
         # check for all the cases where it would be raining, and if its not then turn on bugs
         sdl = PtGetAgeSDL()
         bugState = sdl["psnlBugsVis"]
-        if rainState == 1 or (rainState == 4 and len(PtGetPlayerList()) == 0) or (rainState == 3 and len(PtGetPlayerList()) > 0):
+        if rainState == 1 or (rainState == 4 and PtIsSolo() or (rainState == 3 and not PtIsSolo())):
```
zrax, line 113:
```suggestion
        if rainState == 1 or (rainState == 4 and PtIsSolo()) or (rainState == 3 and not PtIsSolo()):
```

Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp
```diff
@@ -1017,6 +1017,21 @@ void cyMisc::PrintToScreen(const char* msg)
 #include "plPipeline.h"
 #include "plGImage/plMipmap.h"
 
+/////////////////////////////////////////////////////////////////////////////
+//
+//  Function   : IsSolo
+//  PARAMETERS : 
+//
+//  PURPOSE    : Return if we are the only player in the Age
```
zrax, line 1025:
```suggestion
//  PURPOSE    : Return whether we are the only player in the Age
```

Sources/Plasma/FeatureLib/pfPython/cyMisc.h
```diff
@@ -424,6 +424,15 @@ class cyMisc
     static PyObject* GetLocalAvatar(); // returns pySceneObject
     static PyObject* GetLocalPlayer(); // returns pyPlayer
 
+    /////////////////////////////////////////////////////////////////////////////
+    //
+    //  Function   : IsSolo
+    //  PARAMETERS : 
+    //
+    //  PURPOSE    : Return if we are the only player in the Age
```
zrax, line 432:
```suggestion
    //  PURPOSE    : Return whether we are the only player in the Age
```

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue.cpp
```diff
@@ -131,6 +133,11 @@ PYTHON_GLOBAL_METHOD_DEFINITION_NOARGS(PtGetLocalPlayer, "Returns a ptPlayer obj
     return cyMisc::GetLocalPlayer();
 }
 
+PYTHON_GLOBAL_METHOD_DEFINITION_NOARGS(PtIsSolo, "Returns if we are the only player in the Age")
```
zrax, line 136:
```suggestion
PYTHON_GLOBAL_METHOD_DEFINITION_NOARGS(PtIsSolo, "Returns whether we are the only player in the Age")
```

zrax (approved):

PR #1305, opened 2023-01-25
Wall game code adjustments

Hazado:
**grsnMainWallPython.py**

1. Blocker announcements will now only happen during a game.
You could before stay in the game area after it ended and wait for someone to sit at the North panel, and then it would allow announcements again.

**grsnWallImagerDisplay(N|S).py**
1. fix a code crash

**grsnWallPython.py**
Lot of code movement on this change. 
1. Added a check in the OnNotify section to only trigger code for the player that initiated the action @ line 466.
This change made it so I could move only the things that needed to be multiplayer above it that and cause less code to run for players that don't need it.
2. Reduced the for loops for the events. It will now only run through the events loop once for everything that needs that loop.
3. Was able to make it so setting the SDL was not needed to be done by the game master and could be done by the player that triggered the action instead. There are still a few spots that it needs the game master to be the only player setting the SDL, but that is greatly reduced.
4. The panel lights section (SetPanelMode @ line 820) got a few bug fixes and adjustments.
Now disable the panel completely once you lock in the blockers. Before you could click the blockers and cause OnNotify events to trigger even after you locked in the blockers.

### DIFF
--- a/Scripts/Python/grsnMainWallPython.py
+++ b/Scripts/Python/grsnMainWallPython.py
@@ -66,30 +66,35 @@ class grsnMainWallPython(ptResponder):
         ptResponder.__init__(self)
         self.id = 52394
         self.version = 2
+        self.oldNorthWall = None
+        self.oldSouthWall = None
 
     def OnServerInitComplete(self):
         PtDebugPrint("grsnMainWallPython::OnServerInitComplete")
         ageSDL = PtGetAgeSDL()
+        self.oldNorthWall = ageSDL["northWall"]
+        self.oldSouthWall = ageSDL["southWall"]
         
         ageSDL.setNotify(self.key, "nState", 0.0)
         ageSDL.setNotify(self.key, "sState", 0.0)
 
     def OnClimbingBlockerEvent(self,blocker):
         ageSDL = PtGetAgeSDL()
-        if(ageSDL['nState'][0] == kEnd):
+        if(ageSDL['nState'][0] != kGameInProgress):
             return
 
-        for i in range(0,171):
-            if(northBlockers.value[i] == blocker):
-                northWall.value[i].runAttachedResponder(kBlockerBlink)
-                if(eventHandler):
-                    eventHandler.HandleBlocker()
-                break
-            elif(southBlockers.value[i] == blocker):
-                southWall.value[i].runAttachedResponder(kBlockerBlink)
-                if(eventHandler):
-                    eventHandler.HandleBlocker()
-                break
+        try:
+            index = northBlockers.value.index(blocker)
+            northWall.value[index].runAttachedResponder(kBlockerBlink)
+        except:
+            try:
+                index = southBlockers.value.index(blocker)
+                southWall.value[index].runAttachedResponder(kBlockerBlink)
+            except:
+                PtDebugPrint("grsnMainWallPython::OnClimbingBlockerEvent: Blocker not found on either panel")
+                return
+        if(eventHandler and self.IAmMaster()):
+            eventHandler.HandleBlocker()
 
     def OnSDLNotify(self,VARname,SDLname,playerID,tag):
         ageSDL = PtGetAgeSDL()
@@ -97,6 +102,8 @@ class grsnMainWallPython(ptResponder):
         nState = ageSDL["nState"][0]
         sState = ageSDL["sState"][0]
         if(nState == sState == kEnd):
+            self.oldNorthWall = ageSDL["northWall"]
+            self.oldSouthWall = ageSDL["southWall"]
             for blocker in ageSDL["northWall"]:
                 if(blocker == -1):
                     break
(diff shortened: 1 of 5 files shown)


Scripts/Python/grsnMainWallPython.py
```diff
@@ -106,9 +113,16 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
+            if self.oldNorthWall:
+                for blocker in self.oldNorthWall:
+                    if(blocker == -1):
+                        break
+                    northWall.value[blocker].runAttachedResponder(kBlockerOff)
+            if self.oldSouthWall:
+                for blocker in self.oldSouthWall:
+                    if(blocker == -1):
+                        break
+                    southWall.value[blocker].runAttachedResponder(kBlockerOff)
```
Hoikas, line 125:
```suggestion
            for blocker in self.oldNorthWall:
                if blocker == -1:
                    break
                northWall.value[blocker].runAttachedResponder(kBlockerOff)
            for blocker in self.oldSouthWall:
                if blocker == -1:
                    break
                southWall.value[blocker].runAttachedResponder(kBlockerOff)
```

There shouldn't be any need to test the truthyness of the old wall sequences. An loop over an empty sequence does nothing. Also, we may want to consider changing the
```python
for blocker in foo:
    if blocker == -1:
       break
     # do work here
     pass
```
anti-patter to something more pythonic, such as [`itertools.takewhile`](https://docs.python.org/3/library/itertools.html#itertools.takewhile):
```python
for blocker in itertools.takewhile(lambda x: x!= -1, foo):
    # do work here
    pass
```

Scripts/Python/grsnMainWallPython.py
```diff
@@ -66,37 +66,44 @@ def __init__(self):
         ptResponder.__init__(self)
         self.id = 52394
         self.version = 2
+        self.oldNorthWall = None
+        self.oldSouthWall = None
```
Hoikas, line 70:
```suggestion
        self.oldNorthWall = []
        self.oldSouthWall = []
```

Scripts/Python/grsnWallImagerDisplayN.py
```diff
@@ -61,14 +61,16 @@ def __init__(self):
         ptResponder.__init__(self)
         self.id = 52398
         self.version = 2
+        self.oldNorthWall = None
```
Hoikas, line 64:
```suggestion
        self.oldNorthWall = []
```

Scripts/Python/grsnWallImagerDisplayS.py
```diff
@@ -61,14 +61,16 @@ def __init__(self):
         ptResponder.__init__(self)
         self.id = 52397
         self.version = 2
+        self.oldSouthWall = None
```
Hoikas, line 64:
```suggestion
        self.oldSouthWall = []
```

Scripts/Python/grsnWallPython.py
```diff
@@ -329,53 +331,161 @@ def OnServerInitComplete(self):
...
+                        if currentgender == kFemaleClothingGroup:
+                            clothing = FemaleSuit
+                        else:
+                            clothing = MaleSuit
+                        if not wornItem:
+                            wornItem = avatar.avatar.getAvatarClothingList()
+                            DefaultColor1 = {item[1]: avatar.avatar.getTintClothingItem(item[0],1) for item in wornItem}
+                            DefaultColor2 = {item[1]: avatar.avatar.getTintClothingItem(item[0],2) for item in wornItem}
+                        for index, item in enumerate(clothing):
+                            if item == None:
```
Hoikas, line 428:
```suggestion
                            if item is None:
```

Scripts/Python/grsnWallPython.py
```diff
@@ -329,53 +331,161 @@ def OnServerInitComplete(self):
...
+                    team = kNorth
+                    try:
+                        index = NorthPanel.value.index(pickedBlockerObj)
+                    except:
+                        try:
+                            index = SouthPanel.value.index(pickedBlockerObj)
+                            team = kSouth
+                        except:
+                            PtDebugPrint("grsnWallPython::OnNotify: Blocker not found on either panel")
+                            continue
```
Hoikas, line 467: This seems a bit sus to me. I feel like we should know for certain which panel we're working with, not trying one then trying the other for funsies. 😕

Hazado: I wish, it returns a scene object.
I could probably get the name of the object and parse north or south from that as every panel object starts with N or S

Hazado: could also seperate the check into north and south instead of combining them, would make the south panel one less index to check

Hoikas: Yeah, it would probably be a good idea to set the team by doing something like:
```python
if id == NorthPanelClick.id:
    team = kNorth
    blockerObjs = NorthPanel.value
else:
    team = kSouth
    blockerObjs = SouthPanel.value
index = blockerObjs.index(pickedBlockerObjs)
```

Hazado: Do you think it would be better to set the north and south panels to a constant on age initialization instead of setting up a dictionary everytime?

Hoikas (approved):

PR #1308, opened 2023-01-28
Fix `OnOwnershipChanged` callbacks.

Hoikas:
Although we are indeed using a simplified object ownership model, we do still want to send out broadcasts to game scripts about when the ownership was changed.

This is needed for the upcoming rewrite of xBlueSpiral to use swappable backends. See #1217.


dpogue:
Does this have any potential implications for the Gahreesen Wall? I think people said there were sometimes issues when the Age owner left, and @Hazado reworked a bunch of stuff recently to make it less dependent on the Age owner, but maybe this improves the situation even further?

Hoikas:
It could potentially be used to help. Right now, nothing is actually processing the notifications this sends out.

dpogue (approved):

PR #1315, opened 2023-02-07
Fan Relto video to replace Yeesha Intro

Hazado:
Data files - https://github.com/H-uru/moul-assets/pull/218

### DIFF
--- a/Scripts/Python/xOpeningSequence.py
+++ b/Scripts/Python/xOpeningSequence.py
@@ -223,36 +223,11 @@ class xOpeningSequence(ptModifier):
         elif id == OrientationDlg.id:
             if event == kDialogLoaded:
                 # see if the there actually is a movie to play
-                skipMovie = 1
-                try:
-                    ageSDL = PtGetAgeSDL()
-                    if IsTutorialPath():
-                        gMovieFilePath = kAtrusIntroMovie
-                    elif ageSDL["psnlIntroMovie"]:
-                        gMovieFilePath = ageSDL["psnlIntroMovie"][0]
-                    else:
-                        gMovieFilePath = kReltoIntroMovie
-                    os.stat(gMovieFilePath)
+                if self.CheckMovie():
                     # its there! show the background, which will start the movie
                     PtShowDialog("IntroBahroBgGUI")
-                    skipMovie = 0
-                except:
-                    skipMovie = 1
-                if skipMovie:
-                    # no movie... just show the help screen
-                    # start rendering the scene again
-                    PtEnableRenderScene()
-                    PtGUICursorOn()
-                    OrientationDlg.dialog.show()
-                    PtDebugPrint("xOpeningSequence - no intro movie!!!",level=kDebugDumpLevel)
-                    if IsTutorialPath():
-                        OrientationPBIcon01.value.draw.disable()
-                        OrientationPBIcon02.value.draw.disable()
-                        OrientationPBIcon01Zandi.value.draw.enable()
-                        ptGUIControlTextBox(OrientationDlg.dialog.getControlFromTag(kOrientPBText)).setString(PtGetLocalizedString("GUI.OrientationGUI.OrientPBTextZandi"))
-                    else:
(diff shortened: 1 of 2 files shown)


Scripts/Python/xOpeningSequence.py
```diff
@@ -75,7 +75,7 @@
 
 gIntroMovie = None
 kAtrusIntroMovie = "avi/AtrusIntro.webm"
-kYeeshaIntroMovie = "avi/URULiveIntro.webm"
+kIntroMovie = "avi/NewPlayerIntro.webm"
```
dpogue, line 78: Having variables named `gIntroMovie` and `kIntroMovie` (particularly when `kIntroMovie` is not entirely a constant) seems prone to confusion...

Maybe something like `gMovieFilePath` makes more sense as the variable, and have `kAtrusIntroMovie` and `kReltoIntroMovie` as string constants that get assigned to that filepath variable (along with the vault override)?

Hazado: Done

dpogue (approved):
LGTM

Hazado:
So I needed to delay the SDL check for setting the video for 3 seconds.
Without the delay it was getting the SDL but it would have the default values instead of what was set on the server.
With a delay in there it gets the correct server SDL

dpogue:
Rather than a hardcoded delay, can you use `OnServerInitComplete` or something? There should be a Python callback for when the server has sent the initial SDL state...

Scripts/Python/xOpeningSequence.py
```diff
@@ -223,10 +225,14 @@ def OnGUINotify(self,id,control,event):
                 # see if the there actually is a movie to play
                 skipMovie = 1
                 try:
+                    ageSDL = PtGetAgeSDL()
                     if IsTutorialPath():
-                        os.stat(kAtrusIntroMovie)
+                        gMovieFilePath = kAtrusIntroMovie
+                    elif ageSDL["psnlIntroMovie"]:
```
Hoikas, line 231:
```suggestion
                    elif ageSDL["psnlIntroMovie"][0]:
```
The current code works, but it's somewhat misleading. What we're really testing is the actual value of the variable being non-empty.

Scripts/Python/xOpeningSequence.py
```diff
@@ -223,10 +225,14 @@ def OnGUINotify(self,id,control,event):
...
+                        gMovieFilePath = ageSDL["psnlIntroMovie"][0]
                     else:
-                        os.stat(kYeeshaIntroMovie)
+                        gMovieFilePath = kReltoIntroMovie
+                    os.stat(gMovieFilePath)
```
Hoikas, line 235:
```suggestion
```

This doesn't do anything useful.

Scripts/Python/xOpeningSequence.py
```diff
@@ -328,10 +334,14 @@ def OnGUINotify(self,id,control,event):
         elif id == -1:
             if event == kShowHide:
                 if control.isEnabled():
+                    ageSDL = PtGetAgeSDL()
                     if StartInCleft():
-                        gIntroMovie = ptMoviePlayer(kAtrusIntroMovie, self.key)
+                        gMovieFilePath = kAtrusIntroMovie
+                    elif ageSDL["psnlIntroMovie"]:
```
Hoikas, line 340:
```suggestion
                    elif ageSDL["psnlIntroMovie"][0]:
```

Scripts/Python/xOpeningSequence.py
dpogue: Isn't the intention here to check if the file exists, and go into the `except` block if it's not found?

Hoikas: Hmm... The `except` block isn't visible in the context lines. Ok, carry on.

Hazado:
Consolidated a bit of the code for the movie check
It will now properly fail and skip the movie GUI if the movie is not found

Added your requested changes Adam

PR #1319, opened 2023-02-12
Return a real bool from `PYTHON_RETURN_BOOL`.

Hoikas:
Previously, an integer was returned. Returning an actual boolean enables improved Python code (eg `meow() is True`) and allows f-strings with these booleans to print the words `True` and `False` instead of the integers `1` and `0`.


zrax (approved):

dgelessus:
This could be replaced by CPython's own [`PyBool_FromLong`](https://docs.python.org/3/c-api/bool.html#c.PyBool_FromLong), right?

Hoikas:
It could, but I wanted to avoid a function call.

PR #1321, opened 2023-02-16
Fix the Cleft physical proxies crashing.

Hoikas:
The Cleft has a physical proxy that produces more than 32,000 indices. Therefore, we need to dice that up into multiple geometry spans to prevent bad things from happening.


Hazado:
Tested the client with the cleft and was able to see the physicals with the Graphics.Show.Physicals command

PR #1325, opened 2023-02-17
Initial stub of Linux client

dpogue:
This adds a preliminary X11 implementation of the game client window and a CLI-based login prompt for a Linux client. It does have basic keyboard and mouse input working, and uses the stub OpenGL pipeline (which currently doesn't render anything).

This isn't complete, but is functional enough to be useful as a test environment for ongoing work on the OpenGL pipeline, and there are a few proposed architectural things that I'd like some eyes on.

## plClientWindow
Right now, plClient has a few methods that need platform-specific implementations, and the actual creation of the game window happens as part of the main entry point. I'm proposing that we abstract the game client window logic into a plClientWindow subclass. This is partly because Linux will need different implementations for X11 and Wayland and need to pick between them at runtime.

In this PR, I introduce a plClientWindow base class, and provide the implementation for plX11ClientWindow. In the short term, I handle the client window as a global variable, but if this idea gets adopted across all the platforms then I would have plClient own the plClientWindow instance.

I know that this pattern can be adopted fairly easily for the existing Win32 client (and previously had a local branch with an implementation of that). I want to believe the same thing can be true on macOS, but it's more complicated with things being tied to the ApplicationDelegate, so that needs some more thinking.

## plLoginWindow (future)
If plClientWindow works out, I'd like the apply the same sort of pattern to the login window.

The end goal would be that we don't need a separate `main` entry point with a bunch of duplicated-but-slightly-different code for each platform, and as much of the platform-specific stuff could be wrapped in these implementation-specific subclasses.


dpogue:
I've pretty well completely rewritten this in a different approach, but it should be more scalable to support a Linux Wayland client as well (and with the hope of eventually moving Win32 and macOS over to this approach too).

@Hoikas I'd be interested to hear any feedback you have about the plClientWindow pattern I'm proposing.
@colincornaby Would also be interested to hear if/how you think this could work for Apple platforms. One thing I had not considered but came to mind while writing these notes is whether a plClientWindow subclass could be implemented in Swift for SwiftUI on visionOS or something like that.

Feel free to gloss over the (gross) implementation details of the X11 window stuff and just look at it from a high level.

Hoikas:
It seems like a good idea to me. It would be really nice if we could unify all of the clients under a pattern like that such that we only have one (win)main.cpp, but I know that macOS really wants to do its own bespoke thing, so it may not be possible. I do think it looks like the Windows and Linux clients could be unified under your proposal.

colincornaby:
Thanks. I had been thinking about a window class as a solution to the questions posed in #1860. It would be a good wrapper for creating a frame buffer. I hadn't decided if I should build out something similar to the static functions that are already there - or a class that could be passed in as the window argument. This would align us with a class based approach.

The problems I'm thinking about don't have to be solved here. But I was trying to come up with an elegant way to add requirements to the window class based on what renderers have been included in the build.

As examples:

- A window/client that supports OpenGL would be required to supply a context to render to. (Or not? An OpenGL client could just bind the context on render thread and never pass it through.)
- A window/client that supports Metal must supply Metal frame buffers to render to
- A window/client that supports Vulkan or DX12 must supply a swap chain

Obviously a window would not want to supply all those things at once - the pipeline demanding one of those things would cause the window to configure itself for one of those APIs on demand.

I don't have any solutions here. Last night I was thinking with concepts but we don't support C++20. Maybe just wrapping things in ifdefs is sufficient.

> It would be really nice if we could unify all of the clients under a pattern like that such that we only have one (win)main.cpp, but I know that macOS really wants to do its own bespoke thing

I'm not happy with the duplication or structure of the macOS main file. I think it could do with some ordering and deduplication. And because the iOS/visionOS prototype has to have a different main - I'm suffering from the duplication problem there too.

I think most of the what the macOS client does could be moved into a window subclass.

> One thing I had not considered but came to mind while writing these notes is whether a plClientWindow subclass could be implemented in Swift for SwiftUI on visionOS or something like that.

Swift can't subclass but it can integrate into a template sort of architecture. I hope subclassing is added in the future. But if we wanted to template plClientWindow on a platform specific type - that would allow Swift to participate.

Sources/Plasma/Apps/plClient/plClientWindow.h
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Updates the progress meter on the client window (if applicable) to
+     * represent the current status of an ongoing operation.
+     */
+    virtual void UpdateProgressIndicator(plOperationProgress* progress) {}
+
+    /**
+     * Displays a splash screen window while the client finishes initializing.
+     */
+    virtual void ShowLoadingSplashScreen() {}
```
colincornaby, line 117: Was thinking about this PR. There is a set of functions defined elsewhere here that are definitely a fit for a client window.

Some of these other functions though are part of the lifecycle before the client window appears. Looking at the pure abstract interface there are things that seem to be client concerns - but not necessarily window concerns.

I don't know if we should leave those things out of the interface for now. I think this is trying to drive towards cleaning up the platform mains. But maybe we need a client abstraction and a window abstraction.

dpogue: yeah, these are not strictly the client window, but we already have an instance of plClientWindow at the time that we're wanting to show these, and they will need to be implemented using the same windowing APIs as the client window itself (i.e., Win32, Cocoa, X11, etc.) so it seemed worth sticking them here instead of needing to do a bunch more conditional logic to pick a matching splashscreen window implementation.

They have (empty) default implementations, so they're technically optional to implement.

dpogue:
I think I'm going to mark this as ready to review. As far as a Linux client goes, it's still very much just a stub, but it's enough to be able to test the OpenGL pipeline work on X11 (or with XWayland), and merging this would allow the Windows and macOS clients to start adopting the plClientWindow API.

Future work to do:
* Linux Wayland window implementation
* API interface for a login window and implementations of that (currently this prompts on the terminal)
* Fullscreen handling, resolution changing, etc.
* Align Win32 client with plClientWindow and future login window API
* Try to align macOS client with plClientWindow and future login window API

Sources/Plasma/Apps/plClient/linux/plX11ClientWindow.h
```diff
@@ -0,0 +1,80 @@
...
+    void ShowClientWindow() override;
+    bool ProcessEvents() override;
+    void DeInit() override;
+
+    void ResizeClientWindow(uint16_t width, uint16_t height, bool windowed) override;
+
+    void ShowLoadingSplashScreen() override;
+    void HideLoadingSplashScreen() override;
+
+    hsWindowHndl GetWindowHandle() const override {
```
Hoikas:
```suggestion
    hsWindowHndl GetWindowHandle() const override
    {
```

Sources/Plasma/Apps/plClient/linux/plX11ClientWindow.h
```diff
@@ -0,0 +1,80 @@
...
+        return (hsWindowHndl)(uintptr_t)fXWindow;
+    }
+    hsDisplayHndl GetDisplayHandle() const override {
```
Hoikas:
```suggestion
    hsDisplayHndl GetDisplayHandle() const override
    {
```

Hoikas (commented):
Just a couple of nits from me.

PR #1327, opened 2023-02-18
Limit HQ DynamicCamMaps to reflections.

Hoikas:
It was pointed out in H-uru/moul-assets#209 that the resolution setting in DCMs can be extremely useful for non-reflections. Therefore, use the artist settings for things that aren't reflections.


dpogue (approved):
This is a more elegant fix than the flags hacking I assumed we were going to have to do :+1:

PR #1328, opened 2023-02-19
Miscellaneous simple pfPython cleanups

dgelessus:
Edit: I'm realizing now that the `Py_RETURN_NONE` change will cause merge conflicts for all open PRs that touch the Python glue...

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.h
+++ b/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.h
@@ -87,11 +87,6 @@ private:
     static plCyDebServer debugServer;
 #endif
 
-    static void initPlasmaModule();
-    static void initPlasmaConstantsModule();
-    static void initPlasmaNetConstantsModule();
-    static void initPlasmaVaultConstantsModule();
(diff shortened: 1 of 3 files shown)


Hoikas:
The only concern I have is with [b26d35f](https://github.com/H-uru/Plasma/pull/1328/commits/b26d35f84e63a61f24cf946979012439d1851b23) -- while it is annoying to have duplicated macros, it's nice, IMO, to have a consistent design language around our python bindings, which we get from our macros. Meaning, everything is cased the same. I wonder if it would be better to redefine our macros to point to the standard ones? :thinking:

dgelessus:
Yeah, it's nice to have `PYTHON_RETURN_NONE` for symmetry with `PYTHON_RETURN_ERROR`, instead of `Py_RETURN_NONE`. I undid the mass replace and did what you suggested (I had to redo that commit anyway because of conflicts).

Generally I would argue against keeping custom versions of standard/established APIs like this - but with all the other custom macros that the Python glue uses, these two don't make a big difference.

Hoikas (approved):

PR #1330, opened 2023-02-20
Add Chiso Preniv Book to Relto bookshelf

Hazado:
Assets - https://github.com/H-uru/moul-assets/pull/220

These changes add a hardcode link to the Relto Bookshelf for linking to the Chiso Preniv Age.

Changes the Chiso Preniv Age so its no longer inviteable in the KI or visible in the Nexus under Personal and Invited links.

Adds a script to click an object and add it to the Owned age links.
This is used to add Chiso Preniv Book to the owned ages. This will be added by @DoobesURU in a seperate PR.

### DIFF
--- a/Scripts/Python/xRegisterAge.py
+++ b/Scripts/Python/xRegisterAge.py
@@ -41,6 +41,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
  *==LICENSE==* """
 
+from __future__ import annotations
+
 from Plasma import *
 from PlasmaTypes import *
 from PlasmaKITypes import *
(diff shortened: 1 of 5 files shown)


Scripts/Python/xRegisterAge.py
```diff
@@ -0,0 +1,120 @@
...
+        for content in contents:
+            link = content.getChild().upcastToAgeLinkNode()
+            if link is not None:
+                info = link.getAgeInfo()
+            else:
+                link = content.getChild()
+                info = link.upcastToAgeInfoNode()
+
+            if info.getAgeFilename().lower() == ageName:
+                return link
```
Hoikas, line 86: Is there a known case where the Age gets stashed into the vault with the wrong case?

Scripts/Python/xRegisterAge.py
```diff
@@ -0,0 +1,120 @@
...
+        ageStruct = ptAgeInfoStruct()
+        ageStruct.setAgeFilename(ageFileName)
+        ageLinkNode = vault.getOwnedAgeLink(ageStruct)
+        if not ageLinkNode:
+            info = ptAgeInfoStruct()
+            info.setAgeFilename(ageFileName)
+            info.setAgeInstanceName(ageInstanceName)
+
+            playerName = PtGetClientName()
+            ageGuid = PtGuidGenerate()
```
Hoikas, line 101: I'm pretty sure that if you don't set the guid, then the server will generate one for you.

Scripts/Python/xRegisterAge.py
```diff
@@ -0,0 +1,120 @@
...
+            userDefName = ""
+            desc = ""
+
+            if playerName[-1] == "s" or playerName[-1] == "S":
+                userDefName = "%s'" % playerName
+                desc = "%s' %s" % (playerName, info.getAgeInstanceName())
+            else:
+                userDefName = "%s's" % playerName
+                desc = "%s's %s" % (playerName, info.getAgeInstanceName())
```
Hoikas, line 110: I think we have a localization tool to manage possessives now. Is that right, @ZarothYe?

Scripts/Python/xRegisterAge.py
```diff
@@ -0,0 +1,120 @@
...
+
+            info.setAgeInstanceGuid(ageGuid)
+            info.setAgeUserDefinedName(userDefName)
+            info.setAgeDescription(desc)
+
+            link = ptAgeLinkStruct()
+            link.setAgeInfo(info)
+
+            ptVault().registerOwnedAge(link)
+            PtDebugPrint("xRegisterAge.CheckAndRegisterAge(): Registered age - ", ageFileName)
```
Hoikas, line 120: Missing newline at EOF.

Scripts/Python/xRegisterAge.py
Hazado: This function was taken from [nxusBookMachine.py](https://github.com/H-uru/Plasma/blob/af1395d8b6da492817b78224948966238db0c2c3/Scripts/Python/nxusBookMachine.py#L534)
I don't know why this is done this way.

Scripts/Python/xRegisterAge.py
Hazado: I took this function from https://github.com/H-uru/Plasma/blob/af1395d8b6da492817b78224948966238db0c2c3/Scripts/Python/xLinkingBookGUIPopup.py#L1045

Scripts/Python/xRegisterAge.py
Hazado: The user defined name is also set here, that function will work for the the  age description
Should I just leave the userDefName part in for another PR to have it covered by xLocTools?

Scripts/Python/xRegisterAge.py
Hoikas, line 120: - The owned Age check is run three times in the OP. Once brute forced to figure out if the message should be shown (and proceeding), once before registering the Age in the Python, and again before registering the Age in the C++. Reduce to once for showing the message (and proceeding) and the C++ side check.
- No need to spec a guid. The server will generate one for us.
- Use f-strings.
- No need to use `xLocTools`. It looks like the assumption is that English gets stored in the vault and `xLocTools` massages that.

```suggestion
    def OnNotify(self, state, id, events):
        if PtWasLocallyNotified(self.ley) and state and id == act.id:
            if not self.IHaveAgeLink(ageFileName.value):
                if resp.value:
                    resp.run(self.key, events=events)
                self.IRegisterAge(ageFileName.value, ageInstanceName.value)
                PtSendKIMessage(kKILocalChatStatusMsg, PtGetLocalizedString("Global.Messages.AgeBookGet", [ageInstanceName.value]))

    def IHaveAgeLink(self, ageName: str) -> bool:
        vault = ptVault()
        info = ptAgeInfoStruct()
        ais.setAgeFilename(ageName)
        return vault.getOwnedAgeLink(ais) is not None

    def IRegisterAge(self, ageFileName: str, ageInstanceName: str):
        PtDebugPrint(f"xRegisterAge.CheckAndRegisterAge(): {ageFileName=} {ageInstanceName=}", level=kWarningLevel)
        if not ageLinkNode:
            playerName = PtGetClientName()
            userDefName = f"{playerName}'" if playerName[-1] in {"s", "S"} else f"{playerName}'s" 
            desc = f"{userDefName} {ageInstanceName}"

            info = ptAgeInfoStruct()
            info.setAgeFilename(ageFileName)
            info.setAgeInstanceName(ageInstanceName)
            info.setAgeUserDefinedName(userDefName)
            info.setAgeDescription(desc)

            link = ptAgeLinkStruct()
            link.setAgeInfo(info)

            ptVault().registerOwnedAge(link)
            PtDebugPrint(f"xRegisterAge.CheckAndRegisterAge(): Registered age {ageFileName=}", level=kWarningLevel)
 ```

File: Scripts/Python/xRegisterAge.py [hunk 5]
```diff
@@ -0,0 +1,120 @@
...
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+ *==LICENSE==* """
+
```
Hoikas, line 43:
```suggestion

from __future__ import annotations

```

Hazado:
Suggestions applied

Hoikas (approved):

PR #1336, opened 2023-02-26
A few more skinning improvements.

Hoikas:
- Remove manual input/output temporary buffer initialization because it slows down the loop significantly.
- Position and normal now share a single 32-byte aligned buffer for (future?) AVX skinning.
- Removed SSE 4.1 skinning because the SSE 3 codepath is significantly faster after these changes.
- Fixed a problem in the FPU variant whereby the Y component of the normal was skinned instead the Z component.


ZarothYe (approved):

Hoikas:
Are there any objections to this? It includes a fairly important bugfix to the FPU path.

dpogue (approved):

Deledrius (approved):

PR #1339, opened 2023-02-26
Fix Linking Book cover function

### LINKED ISSUE #1338: Linking Books and covers
It appears that `xLinkingBookGUIPopup.py` has the means to locate a cover in `xLinkingBookDefs.py` using the usual `<cover src="">` in the latter script, but for some reason, other parts of the GUI script are forcing the Books open.

I've tried fiddling with things, and while I'd been successful at getting the single Book cover to work, it ended up breaking the city Book as well as other Books with multiple pages.

The definition in question is `IsThereACover`, FYI.

DoobesURU:
Adjust script to actually use cover image defined in `xLinkingBookDefs.py`. Fixes #1338 

The script currently makes the Relto shelf and some other Books gain the "default" cover and/or extra pages, so it's a draft for now.

### DIFF
--- a/Scripts/Python/xLinkingBookGUIPopup.py
+++ b/Scripts/Python/xLinkingBookGUIPopup.py
@@ -529,6 +529,8 @@ class xLinkingBookGUIPopup(ptModifier):
                 # make sure there is a cover to show
                 if not showOpen and not self.IsThereACover(bookdef):
                     showOpen = 1
+                elif showOpen and not self.IsThereACover(bookdef):
+                    showOpen = 1
                 else:
                     showOpen = 0
                 gLinkingBook.setGUI(gui)
--- a/Scripts/Python/xLinkingBookGUIPopup.py
+++ b/Scripts/Python/xLinkingBookGUIPopup.py
@@ -527,7 +527,7 @@ class xLinkingBookGUIPopup(ptModifier):
                 gLinkingBook = ptBook(bookdef,self.key)
                 gLinkingBook.setSize( width, height )
                 # make sure there is a cover to show
-                if showOpen and self.IsThereACover(bookdef):
+                if showOpen and not self.IsThereACover(bookdef):
                     showOpen = 0
                 gLinkingBook.setGUI(gui)
                 gLinkingBook.show(showOpen)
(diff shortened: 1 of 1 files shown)


Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -527,9 +527,10 @@ def IShowBookNoTreasure(self):
                 gLinkingBook = ptBook(bookdef,self.key)
                 gLinkingBook.setSize( width, height )
                 # make sure there is a cover to show
-                if not showOpen:
-                    if not self.IsThereACover(bookdef):
-                        showOpen = 1
+                if not showOpen and not self.IsThereACover(bookdef):
+                    showOpen = 1
+                else:
+                    showOpen = 0
```
Hazado, line 533:
```suggestion
                if showOpen and self.IsThereACover(bookdef):
                    showOpen = 0
```

Hazado: Also Line 867 ` if idx > 0:` should be ` if idx >= 0:`
Anything other than -1 is true for string find

Hazado (changes requested):
``

Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -527,9 +527,8 @@ def IShowBookNoTreasure(self):
                 gLinkingBook = ptBook(bookdef,self.key)
                 gLinkingBook.setSize( width, height )
                 # make sure there is a cover to show
-                if not showOpen:
-                    if not self.IsThereACover(bookdef):
-                        showOpen = 1
+                if showOpen and self.IsThereACover(bookdef):
+                    showOpen = 0
```
Hoikas, line 531: This doesn't feel right to me.... Whether or not a book should start open is hardwired to the following "logic":

```python
        if len(actBookshelf.value) == 0:
            agePanel = TargetAge.value
            showOpen = 1  # start pedestal books open
        else:
            agePanel = stringAgeRequested
            showOpen = 0    # start bookself books closed
  ```

Cyan seems to think anything we just find on a pedestal out in the world is "open" by definiton, but anything on a shelf is closed. That means these lines I'm commenting on *should* just be a safety valve for not showing a cover that doesn't exist, which would make this:

```suggestion
                if showOpen and not self.IsThereACover(bookdef):
                    showOpen = 0
```

Hoikas (commented):
A very belated review.....

PR #1340, opened 2023-02-27
Port `ptVaultChronicleNode` setters to `ST::string`

dgelessus:
Includes a lot of dumb Python changes to remove the duplicate "legacy" glue methods. The actual `ST::string` changes are very small.

Chronicles are almost always pure ASCII, so this has little effect on overall Unicode support. French key bindings might be *less* broken now (they are still somewhat broken).


dgelessus:
The macOS CI is having a sad for some reason, but that shouldn't have anything to do with this PR (I didn't touch the CMake files).

dpogue:
> The macOS CI is having a sad for some reason, but that shouldn't have anything to do with this PR (I didn't touch the CMake files).

I re-ran and it seems to be okay now 🤷🏼‍♂️

Hoikas (approved):

PR #1343, opened 2023-02-28
Migrate remainder of pfGameGUIMgr to `ST::string`

### LINKED ISSUE #473: ST::stringify all the pfGUI classes

dgelessus:
This one is also just internal refactoring with (I think) no player-visible Unicode improvements. Some of this is code is probably dead anyway - the `Game.GUI` console commands seem to be unused...

Closes #473, except for a few small bits that I'll do later (together with other modules).

### DIFF
--- a/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMenuItem.h
+++ b/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMenuItem.h
@@ -50,6 +50,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #ifndef _pfGUIMenuItem_h
 #define _pfGUIMenuItem_h
 
+#include <string_theory/string>
+
 #include "pfGUIButtonMod.h"
 
 class pfGUISkin;
@@ -68,7 +70,7 @@ class pfGUIMenuItem : public pfGUIButtonMod
 
     protected:
 
-        wchar_t         *fName;
+        ST::string      fName;
         bool            fReportingHover;
 
         HowToSkin       fHowToSkin;
@@ -117,8 +119,8 @@ class pfGUIMenuItem : public pfGUIButtonMod
         void    PurgeDynaTextMapImage() override;
 
 
-        void        SetName( const wchar_t *name );
-        const wchar_t   *GetName() const { return fName; }
+        void        SetName(ST::string name);
+        ST::string GetName() const { return fName; }
     
         void    GetTextExtents( uint16_t &width, uint16_t &height );
 
--- a/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIPopUpMenu.cpp
+++ b/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIPopUpMenu.cpp
@@ -232,8 +232,9 @@ void    pfGUIPopUpMenu::Read( hsStream *s, hsResMgr *mgr )
     fMenuItems.resize(count);
     for (uint32_t i = 0; i < count; i++)
     {
-        char readTemp[ 256 ];
-        s->Read( sizeof( readTemp ), readTemp );
+        char readTemp[257];
+        s->Read(sizeof(readTemp) - 1, readTemp);
+        readTemp[sizeof(readTemp) - 1] = '\0';
         fMenuItems[ i ].fName = ST::string::from_latin_1(readTemp);
         
         fMenuItems[ i ].fHandler = pfGUICtrlProcWriteableObject::Read( s );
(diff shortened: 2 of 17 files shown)


Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIPopUpMenu.h
```diff
@@ -93,7 +92,7 @@ class pfGUIPopUpMenu : public pfGUIDialogMod
         {
             // Simple wrapper class that tells us how to build our menu
             public:
-                std::wstring        fName;
+                ST::string          fName;
```
Hoikas, line 95: Same as previous RE an `ST::string` member.

Sources/Plasma/FeatureLib/pfPython/pyGUIPopUpMenu.h
```diff
@@ -127,9 +126,9 @@ class pyGUIPopUpMenu
     virtual void        SetBackSelColor( float r, float g, float b, float a );
 
     // Menu item functions
-    virtual void        AddConsoleCmdItem( const std::wstring& name, const char *consoleCmd );
-    virtual void        AddNotifyItem( const std::wstring& name );
-    virtual void        AddSubMenuItem( const std::wstring& name, pyGUIPopUpMenu &subMenu );
+    virtual void        AddConsoleCmdItem(const ST::string& name, const ST::string& consoleCmd);
+    virtual void        AddNotifyItem(const ST::string& name);
+    virtual void        AddSubMenuItem(const ST::string& name, pyGUIPopUpMenu &subMenu);
```
Hoikas, line 131: While we're here, let's remove the redundant `virtual`...

```suggestion
    void        AddConsoleCmdItem(const ST::string& name, const ST::string& consoleCmd);
    void        AddNotifyItem(const ST::string& name);
    void        AddSubMenuItem(const ST::string& name, pyGUIPopUpMenu &subMenu);
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIPopUpMenu.cpp
```diff
@@ -263,9 +261,8 @@ void    pfGUIPopUpMenu::Write( hsStream *s, hsResMgr *mgr )
     for (const pfMenuItem& item : fMenuItems)
     {
         char writeTemp[ 256 ];
-        char *sName = hsWStringToString(item.fName.c_str());
-        strncpy( writeTemp, sName, sizeof( writeTemp ) );
-        delete [] sName;
+        ST::char_buffer sName = item.fName.to_latin_1();
+        strncpy(writeTemp, sName.c_str(), sizeof(writeTemp));
```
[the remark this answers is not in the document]
zrax: The buffer types always reserve (and initialize) one extra character for the nul terminator...  However, strncpy will NOT null terminate a string if the source string is longer than the specified size, so your point about null terminating the `writeTemp` buffer is still valid (and in fact was also true of the old non-ST code as well)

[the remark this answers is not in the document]
dgelessus: The string is written as a fixed-size field, so arguably it doesn't need a terminator in the data if the string is exactly as long as the field size. I've adjusted the reading code instead to handle unterminated inputs correctly.

Sources/Plasma/FeatureLib/pfPython/pyGUIPopUpMenu.h
dgelessus: pfPython seems to have a lot of unneeded `virtual`s - it might be better to remove them all at once in a separate PR...

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMenuItem.h
```diff
@@ -117,8 +119,8 @@ class pfGUIMenuItem : public pfGUIButtonMod
         void    PurgeDynaTextMapImage() override;
 
 
-        void        SetName( const wchar_t *name );
-        const wchar_t   *GetName() const { return fName; }
+        void        SetName(ST::string name);
+        const ST::string& GetName() const { return fName; }
```
Hoikas, line 123: Why not return `ST::string` (a non-const value)?

dgelessus: Good question! I think I didn't know the convention here and just blindly translated all `const wchar_t *` to `const ST::string&`. Apparently getters urually return `ST::string` by value - I'll adjust it.

Hoikas (approved):

PR #1344, opened 2023-03-02
Fix "Value out of range" warnings from openal-soft

dgelessus:
A lot of warnings like this appear on the console for debug builds:

```
[ALSOFT] (WW) Error generated on context 00000178631E5D10, code 0xa003, "Value out of range"
```

These are caused by out-of-range values for `plSound.fOuterVol` read from some PRPs. Because there are quite a few of these incorrect values (e. g. for many avatar sound effects), the easiest fix is to clamp the value  on read to the expected range.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAudio/plSound.cpp
+++ b/Sources/Plasma/PubUtilLib/plAudio/plSound.cpp
@@ -48,6 +48,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plProfile.h"
 #include "plgDispatch.h"
 
+#include <algorithm>
+
 #include "plAudioSystem.h"
 #include "plSound.h"
 #include "plWin32Sound.h"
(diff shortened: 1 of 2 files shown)


Hoikas:
I wonder if the clamping should be done closer to the point of use (ie closer to alsoft). I'm not sure if the -5000 is an openal-soft limitation or if it's something that would never be valid anywhere, but I think if it's the former, we'd prefer to let data that's legal in general but illegal in openal-soft persist.

dgelessus:
I don't know much about Plasma's sound system, but as far as I can tell, Plasma expects `plSound.fOuterVol` to be in the range [-5000, 0] (inclusive). You can see this in `plDSoundBuffer::SetConeOutsideVolume`, where it's translated to the range [0.0, 1.0] (inclusive) that OpenAL expects:

https://github.com/H-uru/Plasma/blob/0082e6a9b1f1d1b19fc378d56c24afb4be7d841a/Sources/Plasma/PubUtilLib/plAudio/plDSoundBuffer.cpp#L663-L670

The Max plugin also only produces volumes in the range [-5000, 0], except in one case where it produced -10000, probably by accident from how it does the calculation. (See `plSound3DEmitterComponent::ISetParameters` and the param block for `kSoundOConeVolumeSlider`.)

I did consider clamping the volume at the point where it's passed to OpenAL, but this would also mask legitimate warnings in case the engine actually calculates an incorrect volume somehow.

Hoikas:
That makes sense :+1:

Hoikas (approved):

PR #1349, opened 2023-03-04
Remove unused `Game.GUI` commands and `plTextGenerator`

dgelessus:
The `Game.GUI` commands are the last users of some almost unused code, namely `plTextGenerator` (which seems to do more or less the same thing as `plTextFont`) and most of `pfGUICtrlGenerator`.

As far as I can tell, these commands were only meant for testing/prototyping and aren't actively used anywhere. Removing them would allow dropping this otherwise dead code.

If somebody uses these commands in their testing setup or something, they could probably be ported to `plTextFont` instead, so that `plTextGenerator` can still be removed. But if nobody cares about them, it's easier to just remove it all.


Hoikas (approved):
I'm a little sad to see this go because I've always thought that GUIs in the game need to be more code-driven, and this is the only place such a mechanism is exposed. Unfortunately, it seems to be more for testing or mockup purposes than actual production GUIs.

PR #1354, opened 2023-03-06
Use DirectXMath in hsMatrix44.

Hoikas:
As a follow up to discussion around #1334, this moves the Windows CPU optimized paths of `hsMatrix44` to DirectXMath. When the SSE3 code was written, we saw a significant uplift over the FPU codepath. However, over the last 10 years, MSVC has gotten better at auto-vectorizing floating point code and now emits SSE2 instructions by default. Our SSE3 optimization offers only a moderate uplift at this point. Therefore, we drop the runtime CPU detection from `hsMatrix44` and farm out the matrix multiplication to `DirectXMath`, which, by default, uses hand-rolled SSE2. The in-game profiler indicates that in the case of `IMatrixMul34`, the time to perform all the DSpan transforms in Relto with one avatar on screen decreases from 0.04ms to 0.03ms on my system. I imagine there is a better uplift on the general purpose matrix multiplications due to the removal of the function pointer call.

In the case that someone wants extreme optimization, DirectXMath can use more recent instruction sets (eg SSE4, AVX, and AVX2). A future pull request might expose this to CMake.


dpogue:
> In the case that someone wants extreme optimization, DirectXMath can use more recent instruction sets (eg SSE4, AVX, and AVX2). A future pull request might expose this to CMake.

Just for the point of clarity, it would use these unconditionally, right? The theoretical advantage of the hsCPU function dispatch stuff was that it would pick from several options which implementation to run based on the actual CPU capabilities. If we built a client with DirectXMath using AVX2, that would fail to run on older machines, right?

Sources/Plasma/CoreLib/hsMatrixMath.h
```diff
@@ -51,13 +51,27 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    ret.fMap[3][0] = ret.fMap[3][1] = ret.fMap[3][2] = 0.f;
+    ret.fMap[3][3] = 1.f;
+#elif defined(HS_BUILD_FOR_WIN32)
+    // Trying to interpret an hsMatrix44's fMap as an XMFLOAT3X4 doesn't work, sadly.
+    // This is still a win for performance over the auto-vectorized FPU version below,
+    // especially once the manual initialization is removed.
+    DirectX::XMMATRIX m1 = DirectX::XMLoadFloat4x4A((DirectX::XMFLOAT4X4A*)lhs.fMap);
+    DirectX::XMMATRIX m2 = DirectX::XMLoadFloat4x4A((DirectX::XMFLOAT4X4A*)rhs.fMap);
+    DirectX::XMMATRIX result = m1 * m2;
+    DirectX::XMStoreFloat4x4A((DirectX::XMFLOAT4X4A*)ret.fMap, result);
```
dpogue, line 74: Isn't this missing the bit where it sets `fMap[3]` to `{0, 0, 0, 1}` at the end?

Or, if that is not in fact needed, is there a reason not to treat all of these like 4×4 multiplications?

Hoikas: The reason why these are not 4x4 matrices is because it's a special case optimization for when the entire matrix is rarely an identity matrix and only the first three rows are used in the result. So it is an optimization to ignore the 4th row in the math and avoid the branching in the main multiplication code path. Unfortunately, `DirectXMath` is using its own float array struct that isn't laid out quite like ours in the 3x4 case, so we must do a full-fat 4x4 multiplication here. The full fat multiplication gets the proper `0, 0, 0, 1` value from the input matrices, so the result's `fMap[3]` will be `0, 0, 0, 1` automatically. The other code paths must do this initialization manually because they do not perform the full 4x4 multiplication, and `hsMatrix44` does not have a default initializer (probably for performance reasons).

In my testing `DirectXMath`'s full 4x4 multiplication is faster than the FPU 3x4 multiplication + `fMap[3]` initialization.

Hoikas:
> Just for the point of clarity, it would use these unconditionally, right? The theoretical advantage of the hsCPU function dispatch stuff was that it would pick from several options which implementation to run based on the actual CPU capabilities. If we built a client with DirectXMath using AVX2, that would fail to run on older machines, right?

That is correct. We probably don't want to distribute anything using AVX2 (or even AVX) instructions by default.

Hoikas:
@colincornaby Does this look like it moves in the right direction considering your accelerate changes? Also, off the top of your head, does accelerate require aligned storage? You were mentioning in IRC removing the padding, and I think that's a good idea. Just want to verify where we are in terms of everyone needing alignment.

colincornaby:
> @colincornaby Does this look like it moves in the right direction considering your accelerate changes? Also, off the top of your head, does accelerate require aligned storage? You were mentioning in IRC removing the padding, and I think that's a good idea. Just want to verify where we are in terms of everyone needing alignment.

Accelerate doesn't require aligned storage. AFAIK - the Accelerate implementation is only multiplying using one matrix at a time. So it doesn't care about the packing so far.

However - the packing has been important for the matrix palettes in GPU compute based character skinning. I'm streaming the matrices to the GPU as is - including the padding for alignment. This is bloating the transfers to the GPU - and possibly causing performance issues in the shader.

If we ever remove the padding, I'll need to update the compute based skinning to match. Metal Shading Language can also import C++ headers and types - but hsMatrix44.h declares a lot of other functions that MLSL wouldn't be able to import.

Hoikas:
I think it would be nice to have an even lower-level type than `hsMatrix44` that is *just* the unaligned storage of the matrix *without* the flags field for this purpose. I'm going to run into a similar problem on the DX9 end for the vertex texture method.

Hoikas:
I'm going to close this for now. There are conflicts, and I don't feel like fixing them atm.

PR #1358, opened 2023-03-09
InputCore Unicode improvements (and slight cleanup)

dgelessus:
Migrates pnInputCore and plInputCore to `ST::string`, fixing some weird issues with non-ASCII characters in the code that translates `plKeyCombo`s from/to strings. As a result, you can now successfully change all key bindings in French without any hangs or silent errors.

This does *not* address some of the other internationalization issues in the input system, such as localized strings being passed around internally and stored in the vault. It only fixes Unicode support in the strings.

I also did other small modernization/refactoring where I think it's safe, and re-enabled some commented out Spanish/Italian translations. See commits for details (the single commit diffs may also be easier to read too).

PR #1367, opened 2023-04-01
Migrate plLocalization to `ST::string`

dgelessus:
And remove some unused and/or obsolete methods.

The changes touch a few parts of the Max plugin, which I cannot test properly, as usual.

This class could use some further cleanup too:

* The language numbers are typed weirdly - they're often cast back and forth between `plLocalization::Language` and plain `int`, sometimes offset by 1 for mysterious reasons.
* `kNumLanguages` is one higher than `GetNumLocales()`, presumably so that `kJapanese` (the last language in the enum) isn't shown to players.
* The `LocalToString`/`StringToLocal` mechanism seems to be obsolete - everything now uses pfLocalizationDataMgr instead of storing translations in the PRPs. `pfGUITextBoxMod` is the only place still using this old mechanism, but even there it's effectively dead. PRPs still have `$`-separated strings stored in text boxes, but have all translations blank except English, relying on pfLocalizationDataMgr for the actual translations.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
+++ b/Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
@@ -428,15 +428,7 @@ void    plDynamicTextMap::SetFont( const ST::string &face, uint16_t size, uint8_
 //  if( !IIsValid() )
 //      return;
 
-    if (plLocalization::UsingUnicode())
-    {
-        // unicode has a bunch of chars that most fonts don't have, so we override the font choice with one
-        // that will work with the desired language
-        hsStatusMessageF("We are using a unicode language, overriding font choice of %s", face.c_str("nil"));
-        fFontFace = "Unicode";
-    }
-    else
-        fFontFace = face;
+    fFontFace = face;
     fFontSize = size;
     fFontFlags = fontFlags;
     fFontAntiAliasRGB = antiAliasRGB;
--- a/Sources/Tools/MaxComponent/plGUIComponents.cpp
+++ b/Sources/Tools/MaxComponent/plGUIComponents.cpp
@@ -3056,9 +3056,9 @@ public:
 class plGUITextBoxProc : public ParamMap2UserDlgProc
 {
 private:
-    std::vector<M_STD_STRING> fTranslations;
+    std::vector<ST::string> fTranslations;
     int fCurLanguage;
-    void ISetTranslation(int lang, M_STD_STRING text)
+    void ISetTranslation(int lang, ST::string text)
     {
         while (lang >= fTranslations.size())
             fTranslations.emplace_back();
@@ -3080,14 +3080,14 @@ public:
                 if ( pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) )
                 {
                     fTranslations = plLocalization::StringToLocal(pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) );
-                    SetDlgItemText( hWnd, IDC_GUI_INITTEXT, fTranslations[0].c_str() );
+                    SetDlgItemText(hWnd, IDC_GUI_INITTEXT, ST2T(fTranslations[0]));
                 }
                 else
                     // if there is no text, then there is nothing to translate
                     SetDlgItemText( hWnd, IDC_GUI_INITTEXT, pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) );
                 SendMessage( GetDlgItem( hWnd, IDC_GUI_LANGUAGE ), CB_RESETCONTENT, 0, 0 );
                 for (i=0; i<plLocalization::kNumLanguages; i++)
-                    SendMessageA( GetDlgItem( hWnd, IDC_GUI_LANGUAGE ), CB_ADDSTRING, 0, (LPARAM)plLocalization::GetLanguageName((plLocalization::Language)i) );
+                    SendMessage(GetDlgItem(hWnd, IDC_GUI_LANGUAGE), CB_ADDSTRING, 0, (LPARAM)ST2T(plLocalization::GetLanguageName((plLocalization::Language)i)));
                 SendMessage( GetDlgItem( hWnd, IDC_GUI_LANGUAGE ), CB_SETCURSEL, 0, 0 );
                 fCurLanguage = 0;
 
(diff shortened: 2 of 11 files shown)


Sources/Tools/MaxComponent/plGUIComponents.cpp
```diff
@@ -3080,14 +3080,17 @@ class plGUITextBoxProc : public ParamMap2UserDlgProc
...
                 }
                 else
                     // if there is no text, then there is nothing to translate
                     SetDlgItemText( hWnd, IDC_GUI_INITTEXT, pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) );
                 SendMessage( GetDlgItem( hWnd, IDC_GUI_LANGUAGE ), CB_RESETCONTENT, 0, 0 );
-                for (i=0; i<plLocalization::kNumLanguages; i++)
-                    SendMessageA( GetDlgItem( hWnd, IDC_GUI_LANGUAGE ), CB_ADDSTRING, 0, (LPARAM)plLocalization::GetLanguageName((plLocalization::Language)i) );
+                for (i=0; i<plLocalization::kNumLanguages; i++) {
+                    ST::wchar_buffer languageNameBuf = plLocalization::GetLanguageName((plLocalization::Language)i).to_wchar();
+                    SendMessageW(GetDlgItem(hWnd, IDC_GUI_LANGUAGE), CB_ADDSTRING, 0, (LPARAM)languageNameBuf.c_str());
```
zrax, line 3092: I'm not entirely sure this is safe...  If the message is processed right away (as part of the SendMessage call), it should be fine, but if it's processed asynchronously, the buffer may be gone by the time it's used.  Unfortunately, I have no way to test it right now either :(

Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
```diff
@@ -428,15 +428,7 @@ void    plDynamicTextMap::SetFont( const ST::string &face, uint16_t size, uint8_
...
-    if (plLocalization::UsingUnicode())
-    {
-        // unicode has a bunch of chars that most fonts don't have, so we override the font choice with one
-        // that will work with the desired language
-        hsStatusMessageF("We are using a unicode language, overriding font choice of %s", face.c_str("nil"));
-        fFontFace = "Unicode";
-    }
-    else
-        fFontFace = face;
+    fFontFace = face;
```
zrax, line 431: This concerns me a little, since we already know that most of the fonts don't include any characters past `U+00FF`...  At some point, we'll probably need to expand the fonts if we want to actually render languages that require other unicode characters.

dpogue: Interestingly, this code appears to have been copied from Myst V (which does have giant Unicode P2F files), while Uru never shipped any Unicode font files

dgelessus: Yeah, I agree this isn't optimal, but the code that I removed here is dead and broken in its current state - it would need to be reworked anyway if we ever actually support languages not covered by the existing fonts. I explained my reasoning a bit more in the commit message for 7042d18163c9983f650da4056ec03cabf72266ed.

Sources/Tools/MaxComponent/plGUIComponents.cpp
dgelessus: Hrm... I don't have *that* much experience with Win32, so I'm not super familiar with the expected lifetimes for arguments passed via pointers. I would have expected `SendMessage` to make its own copy of the string if it needs to access it later - otherwise I'm not sure how we're supposed to know when the string is safe to deallocate...

dgelessus: Okay, according to [the docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendmessagew), `SendMessageW` "does not return until the window procedure has processed the message", so this *should* be safe.

Sources/Tools/MaxComponent/plGUIComponents.cpp
```diff
@@ -3080,14 +3080,17 @@ class plGUITextBoxProc : public ParamMap2UserDlgProc
                 if ( pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) )
                 {
                     fTranslations = plLocalization::StringToLocal(pmap->GetParamBlock()->GetStr( plGUITextBoxComponent::kRefInitText ) );
-                    SetDlgItemText( hWnd, IDC_GUI_INITTEXT, fTranslations[0].c_str() );
+                    ST::wchar_buffer buf = fTranslations[0].to_wchar();
+                    SetDlgItemTextW(hWnd, IDC_GUI_INITTEXT, buf.c_str());
```
Hoikas, line 3084:
```suggestion
                    SetDlgItemText(hWnd, IDC_GUI_INITTEXT, ST2T(fTranslations[0]));
```

Sources/Tools/MaxComponent/plGUIComponents.cpp
Hoikas, line 3092:
```suggestion
                    SendMessage(GetDlgItem(hWnd, IDC_GUI_LANGUAGE), CB_ADDSTRING, 0, ST2T(plLocalization::GetLanguageName((plLocalization::Language)i)));
```

Sources/Tools/MaxComponent/plGUIComponents.cpp
```diff
@@ -3176,8 +3175,10 @@ class plGUITextBoxProc : public ParamMap2UserDlgProc
                         int idx = (int)SendDlgItemMessage(hWnd, IDC_GUI_LANGUAGE, CB_GETCURSEL, 0, 0);
                         if (idx >= fTranslations.size())
                             SetDlgItemText( hWnd, IDC_GUI_INITTEXT, _T("") );
-                        else
-                            SetDlgItemText( hWnd, IDC_GUI_INITTEXT, fTranslations[idx].c_str() );
+                        else {
+                            ST::wchar_buffer buf = fTranslations[idx].to_wchar();
+                            SetDlgItemTextW(hWnd, IDC_GUI_INITTEXT, buf.c_str());
```
Hoikas, line 3180:
```suggestion
                            SetDlgItemText(hWnd, IDC_GUI_INITTEXT, ST2T(fTranslations[idx]));
```

Sources/Tools/plLocalizationEditor/plAddDlgs.cpp
```diff
@@ -161,10 +161,8 @@ std::vector<ST::string> IGetAllLanguageNames()
 
     for (int curLocale = 0; curLocale <= numLocales; curLocale++)
     {
-        const char *name = plLocalization::GetLanguageName((plLocalization::Language)curLocale);
-        wchar_t *wName = hsStringToWString(name);
-        retVal.push_back(ST::string::from_wchar(wName));
-        delete [] wName;
+        ST::string name = plLocalization::GetLanguageName((plLocalization::Language)curLocale);
+        retVal.push_back(name);
```
Hoikas, line 165:
```suggestion
        retVal.emplace_back(std::move(name));
```

zrax:
Looks like plInputMap.cpp is missing an `<algorithm>` include, exposed by header include changes here

dgelessus:
Ah, thanks for the heads-up. This wasn't noticeable on Windows, because hsWindows.h includes `<algorithm>`. Pushed a fix to #1358 - to avoid causing merge conflicts, because that PR already touches the plInputMap.cpp includes.

Hoikas:
I think this needs to be rebased to fix the two CI failures.

Sources/Tools/plLocalizationEditor/plAddDlgs.cpp
```diff
@@ -161,10 +161,7 @@ std::vector<ST::string> IGetAllLanguageNames()
 
     for (int curLocale = 0; curLocale <= numLocales; curLocale++)
     {
-        const char *name = plLocalization::GetLanguageName((plLocalization::Language)curLocale);
-        wchar_t *wName = hsStringToWString(name);
-        retVal.push_back(ST::string::from_wchar(wName));
-        delete [] wName;
+        retVal.push_back(plLocalization::GetLanguageName((plLocalization::Language)curLocale));
```
Hoikas, line 164:
```suggestion
        retVal.emplace_back(plLocalization::GetLanguageName((plLocalization::Language)curLocale));
```

dgelessus: I'm pretty sure that won't make a difference in this case. Now that the `GetLanguageName` call is inline instead of stored in a variable, `push_back` can do a move instead of a copy. Using `emplace_back` here would just call the move constructor of `ST::string`, which does the same thing.

Hoikas: Technically speaking, `push_back` means "copy." In practice, MSVC's `push_back` simply calls `emplace_back`, so there is no difference. However, I like to see `emplace_back` used to indicate "move/construct in place" and `push_back` for "copy".

dgelessus: It's not just an MSVC thing - [`push_back(T&&)`](https://en.cppreference.com/w/cpp/container/vector/push_back) is part of C++11. But if you're using `emplace_back` as a convention here I have no problem with that.

Hoikas (approved):

PR #1370, opened 2023-04-07
Reverts #730

Hazado:
Reverts https://github.com/H-uru/Plasma/pull/730
Did not actually fix the animations for others linking out

### DIFF
--- a/Scripts/Python/xLinkingBookGUIPopup.py
+++ b/Scripts/Python/xLinkingBookGUIPopup.py
@@ -338,7 +338,7 @@ class xLinkingBookGUIPopup(ptModifier):
                                         #PtDebugPrint("agePanel = ",agePanel)
                                         if agePanel in xLinkingBookDefs.CityBookLinks:
                                             self.IDoCityLinksChron(agePanel)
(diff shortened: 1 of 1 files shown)


Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -490,7 +490,7 @@ def IShowBookNoTreasure(self):
                     sdl["CleftVisited"] = (1,)
                     vault.addChronicleEntry("TomahnaLoad",1,"yes")
                     PtDebugPrint("Chronicle entry TomahnaLoad not present, adding entry and setting to yes")
-                respLinkResponder.run(self.key,avatar=PtGetLocalAvatar())
+                respLinkResponder.run(self.key,avatar=PtGetLocalAvatar(),netPropagate=0)
```
dgelessus, line 493: FWIW, this call didn't have `netPropagate=0` before, but adding it shouldn't hurt probably.

Hazado: Your right, I can remove if its a problem.

Hoikas (approved):

PR #1371, opened 2023-04-09
Various small fixes identified by cppcheck

dpogue:
Best to review this one commit at a time and I can drop any commits that people deem controversial.

This does not clear up all the issues that cppcheck found, but these were some low-hanging fruit.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -353,7 +353,7 @@ void DebugMsg(const char* fmt, ...);
 
 #if defined(_MSC_VER)
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); __assume(0); break;
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) || defined(__clang__)
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); __builtin_unreachable(); break;
 #else
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); break;
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
@@ -859,6 +859,16 @@ void plArmatureMod::SpawnAt(int spawnNum, double time)
 
 void plArmatureMod::SetFollowerParticleSystemSO(plSceneObject *follower)
 {
+    if (!follower) {
+        if (fFollowerParticleSystemSO) {
+            plAttachMsg* attMsg = new plAttachMsg(GetTarget(0)->GetKey(), fFollowerParticleSystemSO, plRefMsg::kOnRemove, GetKey());
+            plgDispatch::MsgSend(attMsg);
+        }
+
+        fFollowerParticleSystemSO = follower;
+        return;
+    }
+
     // TODO: Check for old one and clean up.
     hsPoint3 trans = GetTarget(0)->GetLocalToWorld().GetTranslate() - follower->GetLocalToWorld().GetTranslate();
     
--- a/Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCoop.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCoop.cpp
@@ -151,7 +151,7 @@ bool plAvBrainCoop::RelayNotifyMsg(plNotifyMsg *msg)
 
     proMultiStageEventData * mtevt = static_cast<proMultiStageEventData *>(msg->FindEventRecord(proEventData::kMultiStage));
     if(mtevt)
-        DebugMsg("COOP: Relaying multi-stage event to %d recipients (via plAvBrainCoop)", fRecipients.size());
+        DebugMsg("COOP: Relaying multi-stage event to %zu recipients (via plAvBrainCoop)", fRecipients.size());
 
     if(fRecipients.size() != 0)
     {
--- a/Sources/Plasma/PubUtilLib/plDrawable/plWaveSet7.cpp
+++ b/Sources/Plasma/PubUtilLib/plDrawable/plWaveSet7.cpp
@@ -4035,13 +4035,11 @@ void plWaveSet7::ICreateGraphEdgeLayer(hsGMaterial* mat, int iPass)
 
 void plWaveSet7::ICreateGraphShoreMaterials()
 {
-    int i;
-    for( i = 0; i < kGraphShorePasses; i++ )
-    {
+    for (size_t i = 0; i < kGraphShorePasses; i++) {
         // Create our material
         // and send ourselves a ref.
         hsGMaterial* mat = ICreateEmptyMaterial("GraphShoreMat", kRefGraphShoreMat, i);
-        
+
         // GraphShoreMat's are the materials used to generate the shore texture layers
         // which are then used on rendering the shore to the screen.
 
@@ -4065,17 +4063,16 @@ void plWaveSet7::ICreateGraphShoreMaterials()
 
 }
 
-void plWaveSet7::IAddGraphVShader(hsGMaterial* mat, int iPass)
+void plWaveSet7::IAddGraphVShader(hsGMaterial* mat, size_t iPass)
 {
-    if( !fGraphVShader[iPass] )
-    {
+    if (!fGraphVShader[iPass]) {
         plShader* vShader = new plShader;
         ST::string buff = ST::format("{}_GraphVS_{}", GetKey()->GetName(), iPass);
         hsgResMgr::ResMgr()->NewKey(buff, vShader, GetKey()->GetUoid().GetLocation());
         vShader->SetIsPixelShader(false);
 
         vShader->SetNumConsts(plGraphVS::kNumConsts);
-        
+
         vShader->SetVector(plGraphVS::kNumericConsts, 0, 0.5f, 1.f, 2.f);
         vShader->SetVector(plGraphVS::kPiConsts, 1.f / hsConstants::two_pi<float>,
                                                  hsConstants::half_pi<float>,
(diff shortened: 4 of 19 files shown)


Sources/Plasma/PubUtilLib/plDrawable/plWaveSet7.cpp
```diff
@@ -4325,7 +4325,7 @@ void plWaveSet7::IInitGraph(int iPass)
 
 void plWaveSet7::IShuffleDownGraphs(int iPass)
 {
-    int i;
+    unsigned int i;
     for( i = iPass+1; i < kGraphShorePasses; i++ )
```
dgelessus, line 4329: `iPass` is a signed `int`, so theoretically there could still be overflow weirdness here. The proper fix would be to update the entire call chain to use `unsigned int` or `size_t` for this index.

Sources/Plasma/PubUtilLib/plGLight/plShadowMaster.cpp
```diff
@@ -91,7 +91,7 @@ void plShadowMaster::SetGlobalMaxSize(uint32_t s)
             if( (1 << i) & s )
                 break;
         }
-        s = 1 << i;
+        s = 1 << std::max(i, 0);
```
dgelessus, line 94: > In practice, I'm pretty sure this won't happen and maybe an hsAssert is a better option here?

Yeah - if I'm reading the code correctly, `i` could only get down to -1 if `s` is 0, but in that case `((s-1) & ~s) != (s-1)` would be false, so we would never get here in the first place.

```suggestion
        ASSERT(i >= 0);
        s = 1 << i;
```

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -411,7 +411,7 @@ void DebugMsg(const char* fmt, ...);
 
 #if defined(_MSC_VER)
 #define  DEFAULT_FATAL(var)  default: FATAL("No valid case for switch variable '" #var "'"); __assume(0); break;
-#elif defined(__GNUC__)
+#elif defined(__GNUC__) || defined(__clang__)
```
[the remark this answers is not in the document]
zrax: Hmm, is that accurate?  I would have expected this to be more of a `clang` vs `clang-cl` distinction, not a Unix vs Windows distinction...

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
```diff
@@ -1706,7 +1706,7 @@ class plBDFCharsParser : public plBDFSectParser
             {
                 // If we're doing data, all lines are hex values until we hit "ENDCHAR"
                 if( fRowsLeft == 0 )
-                    throw;
+                    throw false;
```
zrax, line 1709: Hmm, this is suspicious...  `throw` without any parameter will re-throw the current exception or call `std::terminate` if there isn't one...  But what does throwing a bool accomplish?  Who's catching from this call?  I feel like either changing this to throw a proper exception (like `std::runtime_error`) or an explicit crash (like `hsAssert(false)`) would be better.

Sources/Plasma/CoreLib/HeadSpin.h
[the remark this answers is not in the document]
dgelessus: Yeah, that's what I would have expected too, but it seems to make no difference. I tested using the Clang 12 from VS 2019, and both `clang` and `clang-cl` define only `__clang__` and not `__GNUC__`.

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
dpogue: There are several other `throw` statements in this function that were throwing bools, so this was just making it consistent, but I'll try to investigate whether anything is actually catching these errors. Maybe `hsAssert` is a better option.

Sources/Plasma/PubUtilLib/plGLight/plShadowMaster.cpp
```diff
@@ -88,10 +88,11 @@ void plShadowMaster::SetGlobalMaxSize(uint32_t s)
         int i;
         for( i = 31; i >= 0; i-- )
         {
-            if( (1 << i) & s )
+            if ((1u << i) & s)
                 break;
         }
-        s = 1 << i;
+        hsAssert(i >= 0, "Invalid non-POT shadow size");
+        s = 1u << i;
```
dpogue, line 95: cppcheck is still warning about `i` potentially being `uint32_t(-1)`, but I think that's because `hsAssert` doesn't guarantee end of execution when `HS_DEBUGGING` is set.

Sources/Plasma/CoreLib/HeadSpin.h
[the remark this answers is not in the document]
zrax: Ok, it might be more of an SDK thing then -- Clang from MSYS2, for example, does define `__GNUC__` even on Windows, but that version would be using the MinGW SDK rather than the MSVC SDK.  Either way, the fix is appropriate, since Clang doesn't always define `__GNUC__`...  It just piqued my curiosity :)

Sources/Plasma/PubUtilLib/plDrawable/plWaveSet7.cpp
```diff
@@ -4380,12 +4370,8 @@ void plWaveSet7::IUpdateGraphShader(float dt, int iPass)
 
 void plWaveSet7::IUpdateGraphShaders(plPipeline* pipe, float dt)
 {
-    if( fGraphShoreDraw[0] )
-    {
-        int i;
-        for( i = kGraphShorePasses-1; i >= 0; i-- )
-        {
+    if (fGraphShoreDraw[0]) {
+        for (size_t i = kGraphShorePasses-1; i >= 0; i--)
```
zrax, line 4374: `size_t >= 0` is always true, so this loop will never terminate.

Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCoop.cpp
```diff
@@ -151,7 +151,7 @@ bool plAvBrainCoop::RelayNotifyMsg(plNotifyMsg *msg)
 
     proMultiStageEventData * mtevt = static_cast<proMultiStageEventData *>(msg->FindEventRecord(proEventData::kMultiStage));
     if(mtevt)
-        DebugMsg("COOP: Relaying multi-stage event to %d recipients (via plAvBrainCoop)", fRecipients.size());
+        DebugMsg("COOP: Relaying multi-stage event to %lu recipients (via plAvBrainCoop)", fRecipients.size());
```
dgelessus, line 154:
```suggestion
        DebugMsg("COOP: Relaying multi-stage event to %zu recipients (via plAvBrainCoop)", fRecipients.size());
```

(`z` is the length modifier for `size_t`)

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
dpogue: So it appears that calls to parse the BDF contents are wrapped in a `try`/`catch` block, and catch both `std::exception` and non-exceptions: https://github.com/H-uru/Plasma/blob/5b47a7592770d2d1c96975266ac3d1e0b16e9bb3/Sources/Plasma/PubUtilLib/plGImage/plFont.cpp#L2100-L2112

It seems the only place this BDF code is actually used is in plFontConverter, and freetype2 has built-in BDF support so it's unclear to me if this code is still serving any purpose or if it should be removed.

dgelessus (approved):

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
```diff
@@ -859,6 +859,11 @@ void plArmatureMod::SpawnAt(int spawnNum, double time)
 
 void plArmatureMod::SetFollowerParticleSystemSO(plSceneObject *follower)
 {
+    if (!follower) {
+        fFollowerParticleSystemSO = follower;
```
Hoikas, line 863: This doesn't seem to set up the key reference for the new follower (see line 873), which could cause the follower to be thrown away from under us.

dpogue: In this case, `follower` is null

Hoikas:
What is the status here?

dpogue:
I think this was just waiting for re-review

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
Hoikas: I'm still concerned about this. While it is true that `follower` is null, this seems like it's leaking the reference established on line 873. I do think it would be better to make it more obvious that `fFollowerParticleSystemSO` is being set to `nullptr`.

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
Hoikas: It looks like this should probably throw a real `std::exception` for logging purposes.

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
dpogue: It looks like there is a case (albeit perhaps unused in normal gameplay) where this method gets called with `nullptr`, and in theory the current code will crash.

If your preference to use the existing fFollowerParticleSystemSO to remove the reference (and maybe change `SetAttachedToAvatar(false)` on the particle system itself?) before setting it to null?

Hoikas: I believe that's correct. We'll basically want to send another `plAttachMsg` with `plRefMsg::kOnRemove`, I think.

Deledrius (approved):

Sources/Plasma/PubUtilLib/plAvatar/plArmatureMod.cpp
dpogue: I've updated this to send a detach message

Hoikas: For future reference, I went through and verified that sending a `plAttachMsg` with the context `kOnRemove` will release the key reference. This happens in `plCoordinateInterface::IDetachChild()`.

Hoikas (approved):

PR #1377, opened 2023-04-11
Replace `hsError` (`hsOK`, `hsFail`) with better alternatives

dgelessus:
Most uses of `hsOK` and `hsFail` typed them as generic integers and sometimes mixed them with `bool`s or various enums. This PR tries to clean that up a bit by declaring things as `hsError`, or by replacing them with other enums where it's clearly more appropriate.

~I wonder if we want to keep `hsError` around at all... For the simple cases that just need a success/failure status code, a `bool` would work just as well (and various places already use `bool` returns to indicate success/failure). For the complex cases that use more than just the two values `hsOK` and `hsFail`, it may be better to add a dedicated enum for each purpose, so that it's clear which values are actually possible/valid and what they mean.~

Edit: This PR now removes `hsError` entirely, replacing it mainly with `ENetError` and `bool`.


Sources/Plasma/PubUtilLib/plAvatar/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ target_link_libraries(
         plMessage
         plModifier
         plPhysical
+        pfCCR # FIXME
```
dgelessus, line 90: This is a terrible solution obviously - I left it in here so I can ask about it. The `plAvatarMgr::WarpPlayer...` methods use the `plCCRError` enum from pfCCR as a status return value, causing this backwards dependency from PubUtilLib on FeatureLib.

The obvious solution would be to move the `plCCRError` enum into one of the PubUtilLib modules, but I don't know which one would be appropriate. Alternatively, we could simply stop using this enum and just return success/failure from the `WarpPlayer...` methods. Especially considering that nothing actually checks the return values currently, and apparently these methods are the only real user of pfCCR in the entire open-source codebase...

Sources/Plasma/PubUtilLib/plAvatar/plAvatarMgr.h
```diff
@@ -160,9 +162,9 @@ class plAvatarMgr : public hsKeyedObject
     int NumSpawnPoints() { return fSpawnPoints.size(); }
     int FindSpawnPoint( const char *name ) const;
     // \}
-    static int WarpPlayerToAnother(bool iMove, uint32_t remoteID);
-    static int WarpPlayerToXYZ(float x, float y, float z);
-    static int WarpPlayerToXYZ(int pid, float x, float y, float z);
+    static plCCRError WarpPlayerToAnother(bool iMove, uint32_t remoteID);
+    static plCCRError WarpPlayerToXYZ(float x, float y, float z);
+    static plCCRError WarpPlayerToXYZ(int pid, float x, float y, float z);
```
dgelessus, line 167: (these are the `plAvatarMgr::WarpPlayer...` methods in question)

Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.h
```diff
@@ -151,9 +151,9 @@ void NetCommActivateMsgDispatchers();
 
 // Return this value from your registered msg handler
 // to stop further dispatching of incoming msg.
-const unsigned kOK_MsgConsumed  = hsOK + 1;
+static constexpr hsError kOK_MsgConsumed = hsOK + 1;
```
dgelessus, line 154: As far as I can tell, nothing ever returns this value, so this message handling code can probably be simplified a bit as well.

Hoikas:
I think it would make more sense to move toward a boolean return instead of `hsError`, as you suggest. Of course, the engine doesn't exactly agree right now on whether or not true or false is success... Probably a holdover from `BOOL` where `FALSE` means "OK" :(

Sources/Plasma/PubUtilLib/plAvatar/CMakeLists.txt
dgelessus: Fixed now by making the `WarpPlayer` methods return just `hsOK`/`hsFail` and no specific error codes. I now also removed `plCCRError` (and the remaining near-empty CCR code) because nothing else used it.

Sources/Plasma/PubUtilLib/plMessage/plVaultNotifyMsg.h
```diff
@@ -75,8 +76,8 @@ class plVaultNotifyMsg : public plMessage
     uint16_t  GetType() const { return fType; }
     void      SetType( uint16_t v ) { fType=v; }
 
-    int8_t  GetResultCode() const { return fResultCode; }
-    void    SetResultCode( int8_t v ) { fResultCode=v; }
+    ENetError GetResultCode() const { return fResultCode; }
+    void SetResultCode(ENetError resultCode) { fResultCode = resultCode; }
```
Hoikas, line 80: This message probably predates `ENetError`, but I don't think it's ever serialized, so as long as we internally agree on what the error codes mean, everything will be fine. This PR moves us toward more consistency, so this is probably for the best.

Hoikas (commented):
This is looking good to me. I'm holding back on approval while it's a draft, though.

dgelessus:
Yeah, I made it a draft again because I wanted to see if I could also get rid of the last remaining `hsError` uses. The answer is yes 🙂

Everything now uses `ENetError` or `bool` instead of `hsError`. In a few places I removed the status returns completely, where the code always returned success and the callers never checked the status. I think this makes the code easier to follow, but let me know if any of this goes too far.

Sources/Plasma/PubUtilLib/plMessage/plVaultNotifyMsg.h
dgelessus: Yeah, I didn't see any place where `plVaultNotifyMsg` gets sent over the network (and I assume it shouldn't appear in any PRPs...). At least one place already stored an `ENetError` value in this field anyway, so it seemed like a logical choice.

Sources/Plasma/PubUtilLib/plNetTransport/plNetTransport.cpp
```diff
@@ -241,14 +241,14 @@ int plNetTransport::SendMsg(int chan, plNetMessage* netMsg) const
...
-            ret=0;  // sent
+            ret = 0; // sent
         }
     }
     else
     {
         hsStatusMessage("EMPTY TRANSPORT GROUP\n");
     }
-    return ret;
+    //return ret;
```
Hoikas, line 251: It looks like there's still some unused return value stuff in this function.

dgelessus: This entire code is dead actually, because the second line of the method is an unconditional `return`. I wasn't sure if the code was left in for a reason, so I just commented out the return for now, and wanted to take a better look through the network message code later.

Sources/Plasma/FeatureLib/pfPython/pyVaultNode.cpp
```diff
@@ -119,7 +119,7 @@ void pyVaultNode::pyVaultNodeOperationCallback::VaultOperationComplete( uint32_t
                 pyObjectRef tup = PyTuple_New(2);
                 PyTuple_SET_ITEM(tup.Get(), 0, pyVaultNode::New(fNode));
                 PyTuple_SET_ITEM(tup.Get(), 1, fPyNodeRef.Release());
-                plPython::CallObject(func, context, std::move(tup), resultCode);
+                plPython::CallObject(func, context, std::move(tup), static_cast<int32_t>(result));
```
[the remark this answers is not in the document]
dgelessus: I tried, yes. :P I think I ran into problems with references when defining the overload outside of plPythonConvert.h - perhaps it will work better if I add it directly in the header.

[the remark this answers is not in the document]
dgelessus: Nevermind, I just tried again and it works either way now... 🤷‍♀️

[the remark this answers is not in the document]
Hoikas: Those overloads can be very fragile for reasons that I don't understand completely 😞

Sources/Plasma/PubUtilLib/plNetTransport/plNetTransport.cpp
Hoikas: I'm holding off on approval until we see where this goes.

dgelessus: Ah, by "later" I meant "probably much later as its own PR". Didn't want to change the network code much, so that significant logic changes don't get overlooked in between all the simple type renamings. But I can take out this dead code if you prefer - that should be safe enough.

Hoikas: Yes, I think we want to see dead code deleted instead of commented out.

Hoikas (approved):

Sources/Plasma/PubUtilLib/plNetCommon/plNetMsgHandler.h
```diff
@@ -58,14 +56,22 @@ class plNetMsgHandler
 protected:
     plNetApp* fNetApp;
 public:
+    enum class Status
```
dpogue, line 59: FYI on Linux, building the client, I'm getting an error from this: `error: unnamed scoped enum is not allowed`

I think if this is being declared as a `enum class` it needs to "inherit" from a type like `int32_t`?

dpogue: Nevermind, this is because of X11 doing stupid stuff... 🤦🏼‍♂️ 

```
/usr/include/X11/Xlib.h:83:16: note: expanded from macro 'Status'
#define Status int
               ^
```

PR #1379, opened 2023-04-16
Improve Unicode support in the console

### LINKED ISSUE #479: ST::stringify the console
pfConsoleEngine and pfConsole could both use some ST::string/ST::format love.

dgelessus:
This implements full Unicode support in the console engine internals (fixes #479). The in-game console itself only has limited Unicode support for now:

* A few string fields are still limited to 8-bit Latin-1 characters, namely the input line and display buffer.
* `plDebugText`, which the console uses for text rendering, still only supports ASCII characters (not even Latin-1). Non-ASCII characters are not visible in the input and output.

Non-ASCII characters *can* be typed into the input, and even though they are invisible, they will be passed through correctly to Python and/or the console engine (unlike before). Cut/copy/paste also work correctly with non-ASCII characters.

So for example you can type or paste this command to bind the Ü key (semicolon on US keyboards):

```
Keyboard.BindConsoleCmd Ü Graphics.Show.Physicals
```

And you can run this in the Python console and get the correct output:

```python
>>> print(ascii("Ü"))
'\xdc'
```

The previous implementation used very "C-style" string management (not in a good way). I've reimplemented most of it using `ST::string`, which should make the code easier to follow. The diff is quite messy though - apologies in advance to whoever decides to review this. You might want to skip over pfConsoleCommands.cpp and the other command implementations at first - they mostly contain boring `char *` -> `ST::string` changes.

Obviously I didn't test every single console command - only a few common ones and those where Unicode is obviously relevant.


Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -230,7 +230,7 @@ bool hsStream::AtEnd()
 
 bool hsStream::IsTokenSeparator(char c)
 {
-    return (isspace(c) || c==',' || c=='=');
+    return (isspace(static_cast<unsigned char>(c)) || c==',' || c=='=');
```
zrax, line 233: ??  ctype functions (`isspace`, `isalnum`, etc) take an `int`, which both a signed and unsigned implementation of `char` should automatically and safely promote to...  Why all the casts?

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -278,19 +278,17 @@ PF_CONSOLE_FILE_DUMMY(Main)
...
 //
 plKey FindObjectByName(const ST::string& name, int type, const ST::string& ageName,
-                       const char** statusStr, bool subString = false)
+                       ST::string& statusStr, bool subString = false)
 {
     if (name.empty())
     {
-        if (statusStr)
-            *statusStr = "Object name is nil";
+        statusStr = "Object name is nil";
```
zrax, line 285: These would be a good place to use `ST_LITERAL` or the `_st` literal suffix

Sources/Plasma/CoreLib/hsStream.cpp
dgelessus: That's the problem. `char` is signed and gets sign-extended to `int`, but the ctype functions are [only defined for `unsigned char` values and `EOF`](https://en.cppreference.com/w/cpp/string/byte/isspace). See also e. g. https://nullprogram.com/blog/2023/02/11/#character-classification-and-mapping

MSVCRT has a debug assertion ensuring that you don't pass any other negative values. That started failing once I typed non-ASCII characters into the console, so I went ahead and fixed the issue for all ctype calls (there weren't all that many).

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
dgelessus: Ah yep. I used it more consistently elsewhere, but got lazy when I saw the 6000-line file 🙃

Sources/Plasma/CoreLib/hsStream.cpp
zrax: Ah, that make sense...  Although perhaps it makes more sense to test for that specific case, since characters outside the 0-127 range will always be UTF-8 multibyte sequence bytes when coming from ST::strings.

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
dgelessus: Okay, added `ST_LITERAL` in the relevant places. I didn't adjust every single `PrintString` call though - that would be quite a bit of work and not really worth it. (This is the dev console - I don't think we need to micro-optimize performance and memory usage that much.)

Sources/Plasma/CoreLib/hsStream.cpp
dgelessus: Not sure what you mean exactly... something like `c < 0x80 && isspace(c)`? I don't think that's necessary if `c` is already `char`-sized. We don't use locales, so all the `is` functions will return `false` for all bytes outside the ASCII range.

(I *did* use `c < 0x80` checks where `c` is a larger type like `wchar_t` or `uint32_t`, because just a cast isn't sufficient there.)

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -798,14 +778,12 @@ class BriefDocGenIterator : public pfConsoleCmdIterator
...
-                strcpy(fGrpName, g->GetName());
+                fGrpName = g->GetName();
             else 
             {
                 pfConsoleCmdGroup *parentGrp;
                 parentGrp = g->GetParent();	
-                strcpy(fGrpName, parentGrp->GetName());
-                strcat(fGrpName,".");
-                strcat(fGrpName,g->GetName());
+                fGrpName = parentGrp->GetName() + '.' + g->GetName();
```
Hoikas, line 786:
```suggestion
                fGrpName = ST::format("{}.{}", parentGrp->GetName(), g->GetName());
```

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -2418,12 +2389,11 @@ PF_CONSOLE_CMD(App,
                "Verify the given age is really unloaded into logfile logs/<age>.log")
 {
     hsAssert(0, "Fixme");
-    char* age = params[0];
-    char str[256];
-    sprintf(str, "%s.log", age);
+    const ST::string& age = params[0];
+    ST::string str = age + ".log";
```
Hoikas, line 2393:
```suggestion
    ST::string str = ST::format("{}.log", age);
```

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -2840,9 +2817,9 @@ PF_CONSOLE_CMD( Camera,     // groupName
 
 PF_CONSOLE_CMD( Camera, SwitchTo, "string cameraName", "Switch to the named camera")
 {
-    const char *status = "";
-    ST::string foo = ST::format("{}_", (char*)params[0]);
-    plKey key = FindObjectByNameAndType(foo, "plCameraModifier1", "", &status, true);
+    ST::string status;
+    const ST::string& cameraName = params[0];
+    plKey key = FindObjectByNameAndType(cameraName + "_", "plCameraModifier1", {}, status, true);
```
Hoikas, line 2822:
```suggestion
    plKey key = FindObjectByNameAndType(ST::format("{}_", cameraName), "plCameraModifier1", {}, status, true);
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -367,118 +328,72 @@ int  pfConsoleCmdGroup::IterateCommands(pfConsoleCmdIterator* t, int depth)
...
 char    pfConsoleCmd::fSigTypes[ kNumTypes ][ 8 ] = { "int", "float", "bool", "string", "char", "void", "..." };
 
-pfConsoleCmd::pfConsoleCmd(const char *group, const char *name,
-                            const char *paramList, const char *help, 
-                            pfConsoleCmdPtr func, bool localOnly )
+pfConsoleCmd::pfConsoleCmd(const ST::string& group, ST::string name,
+                            const ST::string& paramList, ST::string help, 
+                            pfConsoleCmdPtr func)
+    : fName(std::move(name)), fHelpString(std::move(help)), fFunction(func),
+        fNext(nullptr), fPrevPtr(nullptr), fParentGroup(nullptr)
```
Hoikas, line 335:
```suggestion
    : fName(std::move(name)), fHelpString(std::move(help)), fFunction(func),
      fNext(), fPrevPtr(), fParentGroup()
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -367,118 +328,72 @@ int  pfConsoleCmdGroup::IterateCommands(pfConsoleCmdIterator* t, int depth)
...
-            fSigLabels.Push( hsStrcpy( tok2 ) );            
+            fSigLabels.emplace_back(std::move(parts[1]));
+        } else {
+            fSigLabels.emplace_back();
         }
-        else
-            fSigLabels.Push(nullptr);
 
         // Find type
+        ST::string type = std::move(parts[0]);
```
Hoikas, line 372:
```suggestion
        ST::string& type = parts[0];
```

This seems to just be for self documentation, so no need to actually move the contents of the string.

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -555,32 +470,26 @@ uint8_t   pfConsoleCmd::GetSigEntry( uint8_t i )
...
 
-const char  *pfConsoleCmd::GetSignature()
+ST::string pfConsoleCmd::GetSignature()
 {
-    static char string[256];
-    
-    int     i;
-    char    pStr[ 128 ];
+    ST::string_stream string;
+    string << fName << " ";
```
Hoikas, line 477:
```suggestion
    string << fName << ' ';
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -555,32 +470,26 @@ uint8_t   pfConsoleCmd::GetSigEntry( uint8_t i )
...
-            sprintf( pStr, "[%s %s]", fSigTypes[ fSignature[ i ] ], fSigLabels[ i ] );
-
-        hsAssert( strlen( string ) + strlen( pStr ) + 2 < sizeof( string ), "Not enough room for signature string" );
-        strcat( string, ( i > 0 ) ? ", " : " " );
+        if (i > 0) {
+            string << ", ";
+        }
 
-        strcat( string, pStr );
+        string << "[" << fSigTypes[fSignature[i]];
```
Hoikas, line 485:
```suggestion
        string << '[' << fSigTypes[fSignature[i]];
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -555,32 +470,26 @@ uint8_t   pfConsoleCmd::GetSigEntry( uint8_t i )
...
+        if (!fSigLabels[i].empty()) {
+            string << " " << fSigLabels[i];
```
Hoikas, line 487:
```suggestion
            string << ' ' << fSigLabels[i];
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleCmd.cpp
```diff
@@ -555,32 +470,26 @@ uint8_t   pfConsoleCmd::GetSigEntry( uint8_t i )
...
+        }
+        string << "]";
```
Hoikas, line 489:
```suggestion
        string << ']';
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleEngine.cpp
```diff
@@ -540,40 +507,25 @@ bool    pfConsoleEngine::FindPartialCmd( char *line, bool findAgain, bool preser
...
-    line[ 0 ] = 0;
-    IBuildCmdNameRecurse( cmd->GetParent(), line );
-    strcat( line, cmd->GetName() );
-    strcat( line, " " );
+    ST::string name = cmd->GetName() + ' ';
+    pfConsoleCmdGroup* group = cmd->GetParent();
+    while (group != nullptr && group != pfConsoleCmdGroup::GetBaseGroup()) {
+        name = group->GetName() + '.' + name;
+        group = group->GetParent();
+    }
```
Hoikas, line 523: This should probably use `ST::string_stream` to avoid extra allocations.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -221,17 +221,46 @@ pfConsole * pfConsole::GetInstance () {
...
     fDisplayBuffer = new char[ fNumDisplayLines * kMaxCharsWide ];
     memset( fDisplayBuffer, 0, fNumDisplayLines * kMaxCharsWide );
 
-    memset( fWorkingLine, 0, sizeof( fWorkingLine ) );
-    fWorkingCursor = 0;
+    IClearWorkingLine();
 
-    memset( fHistory, 0, sizeof( fHistory ) );
+    for (size_t i = 0; i < kNumHistoryTypes; i++) {
+        fHistory[i] = {};
```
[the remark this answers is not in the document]
dgelessus: Yeah, Plasma already uses `.clear()` elsewhere, despite the CMakeLists.txt only saying string_theory 2.0 - that should probably be updated...

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
dgelessus: Out of curiosity, is this a style preference, or is `ST::format` better than plain concatenation for some other reason?

Hoikas: Using `ST::format` should result in at least one less allocation here, I think. String concats are very rarely the right choice, IME, although they certainly can be the easiest.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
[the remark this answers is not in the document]
dgelessus: Oh wait, I remember now - `fHistory[i]` is actually a struct. This line initializes/clears all fields in the struct, not just the `ST::string`, so `clear()` doesn't work here.

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleEngine.cpp
dgelessus: This isn't easily implemented with `ST::string_stream`, because that only allows appending to the end, but the string here is built from back to front. I suppose the strings could be collected into a `std::vector` and then joined at the end, to avoid at least some intermediate allocations.

Hoikas (approved):
This LGTM, but I'd like to ensure that all of @zrax's items are cleared up before proceeding.

dgelessus:
Rebased to fix a conflict.

I think I had addressed/fixed all the things from the reviews - let me know if I missed anything.

Hoikas:
Barring any objections, I intend to merge this tonight.

Hoikas:
A couple of days late, but here we go...

PR #1392, opened 2023-05-05
Conditional compiling as workaround for _PyObject_CallOneArg when using Python 3.8

dpogue:
Patch from Prad to fix building with Python 3.8.

@Hoikas mentioned that this API did exist as experimental/non-standard functionality in Python 3.8, so we might be able to get away with including an internal header instead.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
+++ b/Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
@@ -104,7 +104,11 @@ namespace plPython
             // Use Python's built-in vectorcall optimization for one argument.
             pyObjectRef arg = ConvertFrom(std::forward<Args>(args)...);
             plProfile_BeginTiming(PythonUpdate);
+#if PY_VERSION_HEX >= 0x03090000
             pyObjectRef result = _PyObject_CallOneArg(callable, arg.Get());
+#else
+            pyObjectRef result = PyObject_CallFunctionObjArgs(callable, arg.Get(), NULL);
+#endif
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfPython/plPythonCallable.h
```diff
@@ -104,7 +104,11 @@ namespace plPython
             // Use Python's built-in vectorcall optimization for one argument.
             pyObjectRef arg = ConvertFrom(std::forward<Args>(args)...);
             plProfile_BeginTiming(PythonUpdate);
+#if PY_VERSION_HEX >= 0x03090000
             pyObjectRef result = _PyObject_CallOneArg(callable, arg.Get());
+#else
+            pyObjectRef result = PyObject_CallFunctionObjArgs(callable, arg.Get(), NULL);
```
Hoikas, line 110: I haven't been able to find `_PyObject_CallOneArg` in the Python 3.8 headers like the commentary from the 3.9+ header indicate. It might be an API that is never actually publicly exposed at all in 3.8. Regardless, this fix is not correct because `PyObject_CallFunctionObjArgs` will allocate heap memory for the arguments. The whole point of this code is to use more optimized calling conventions present in modern Python versions that avoid unnecessary conversions and allocations. The correct fix is probably to remove this branch of the `if` statement and let single arguments be handed by the `else if constexpr (sizeof...(args) < 64)` block. Under the hood, `_PyObject_CallOneArg` does the same thing that this block does, so handling the two cases separately is probably not necessary.

Hoikas (changes requested):

Issue #1381, opened 2023-04-19
Corrupted / unclickable Relto page in Cleft


mgiuca:
I just built H'uru from source, connecting to my own Dirtsand using the Dockersand script, using the `/LocalData` arg.

Versions (everything fresh as of 2023-04-19):

* H'uru: 13b0b9cc16b
* Dirtsand: `569ee14`
* Assets: From `UruLive.Live.1.918.OpenUru#211`

Create a new character -> Choose Returning Character (Relto) and run the Cleft.

Basically everything is working.

When I get to the Totem Relto page, I see this corrupted image which appears to be possibly **all** of the Relto pages drawn on top of each other:

![image](https://user-images.githubusercontent.com/228433/233090631-ce5908cd-08ec-4be5-a295-032fe31ceebf.png)

Clicking it brightens the mountain image but has no effect. There is no way I can proceed with the game (or at least, get to Gahreesen, Teledahn, Eder Gira or Kadish Tolesa).

I have not tried a Cleft start.

I have not tried any other Relto pages (I'm actually not sure where to find any since I can't get to the main ages).

****

I'm novice at Uru's Python code but looking in `xYeeshaPages.py`, the `IDrawLinkPanel` function appears to try to hide all 26 buttons, then show just the one needed. It looks correct, but hypothetically, if this code wasn't doing _anything_, that would explain what I'm seeing, right? Therefore, is it possible that those `kYeeshaPage` codes became wrong at some point, and these are having no effect?

I tried deleting all the "`show()`" calls except for the one for `kYeeshaPage25` to see if it would hackily fix it, but it didn't have any effect.

Hoikas:
You might be missing some of the new Relto page GUI items due to having out of date copies of the game files. This would cause execution of the script to end with an exception. Are you using the latest assets from H-uru/moul-assets?

mgiuca:
Perfect, thanks for the advice, @Hoikas ! I got it working.

So this doesn't appear to be in the README at all. (There's no mention of moul-assets other than a "related project".) Could that be added? I can send a PR if you like.

Let me get this straight -- let me know if there are any unnecessary steps here:

1. You _still_ need to copy the `avi`, `dat` and `sfx` folders from the official client, right? (I first tried deleting those to just use the `moul-assets` ones, but I got a freeze on startup loading age "StartUp".)
[...]

PR #1393, opened 2023-05-10
Adds being able to disable cluster groups in Relto age

Hazado:
Adds back the cluster group disabling from URU CC

This is not currently used on the Huru branches, but Blue Flowers Page requires this code to work.
Otherwise the flowers will stay enabled all the time while using the Huru client on any shard that has that page included with it.

### DIFF
--- a/Scripts/Python/psnlYeeshaPageChanges.py
+++ b/Scripts/Python/psnlYeeshaPageChanges.py
@@ -62,7 +62,7 @@ respAudioStop = ptAttribResponder(4,"Audio stop responder")
 
 respEnable = ptAttribResponder(5, "Enabled resp (if necessary)")
 respDisable = ptAttribResponder(6, "Disabled resp (if necessary)")
-bushDistrib = ptAttribClusterList(7, "Bush distributor")
+clusterList = ptAttribClusterList(7, "Cluster Group object list")
 
 #globals
 TotalPossibleYeeshaPages = len(xLinkingBookDefs.xYeeshaPages)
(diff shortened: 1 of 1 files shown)


Hazado:
https://github.com/H-uru/Plasma/issues/1381#issuecomment-1516218449
This should resolve the comment about the blue flowers being on by default in the above issue.

Scripts/Python/psnlYeeshaPageChanges.py
```diff
@@ -61,9 +62,9 @@
 
 respEnable = ptAttribResponder(5, "Enabled resp (if necessary)")
 respDisable = ptAttribResponder(6, "Disabled resp (if necessary)")
+bushDistrib = ptAttribClusterList(7, "Bush distributor")
```
Hoikas, line 65: I thought this was the blue flowers, not the bushes?

Scripts/Python/psnlYeeshaPageChanges.py
```diff
@@ -237,6 +238,9 @@ def EnableDisable(self, val):
 
             respAudioStop.run(self.key,avatar=None,fastforward=0)
             respDisable.run(self.key,avatar=None,fastforward=1)
+            if len(bushDistrib.value) > 0:
+                for x in bushDistrib.value:
+                    x.setVisible(0)
```
Hoikas, line 243:
```suggestion
            for i in bushDistrib.value:
                i.setVisible(False)
```

Scripts/Python/psnlYeeshaPageChanges.py
Hazado: thats what Cyan named it in UruCC, probably better to rename it cluster components

Hazado: Pushed a change with better naming

Hoikas (approved):

PR #1394, opened 2023-05-21
Expose just the hsGMatState enums as a header

dpogue:
For shader-based pipelines with Metal and OpenGL, we need to make use of these hsGMatState flag values in the shader code, but the shaders can't include the existing hsGMatState.h file with C++ class definitions.

This extracts just the enum values into their own header that can be imported by shaders, with one major caveat: Rather than plain enum declarations, the enums are generated from macro statements.

You must define `GMAT_STATE_ENUM_START`, `GMAT_STATE_ENUM_VALUE`, and `GMAT_STATE_ENUM_END` before importing this header.

This is because OpenGL shaders don't support enums except as an extension, so it would be better for these to be defined as constants in the shader in that situation.

/cc @colincornaby because I know you had to do some hacking of hsGMatState.h to include it in your Metal shaders

### DIFF
--- a/Sources/Plasma/NucleusLib/inc/CMakeLists.txt
+++ b/Sources/Plasma/NucleusLib/inc/CMakeLists.txt
@@ -2,6 +2,7 @@ set(pnNucleusInc_HEADERS
     hsGDeviceRef.h
     hsGMatState.h
     hsGMatState.inl
+    hsGMatStateEnums.h
     hsResMgr.h
     hsTimer.h
     plAudible.h
--- a/Sources/Plasma/NucleusLib/inc/hsGMatState.h
+++ b/Sources/Plasma/NucleusLib/inc/hsGMatState.h
@@ -48,127 +48,24 @@ class hsStream;
 
 class hsGMatState {
 public:
-enum hsGMatBlendFlags {
-    kBlendTest  = 0x1,                          // dev
-    // Rest of blends are mutually exclusive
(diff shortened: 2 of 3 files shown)


Hoikas (commented):
This seems like a lot for a situation where we just need to change the way that the constants are exposed in a certain compile mode. I think it would make more sense to pull the enumerations out into the global namespace (or another namespace such as `hsGMatSateConst`). To maintain compatibility with the rest of the engine, using statements, like those found [here](https://godbolt.org/z/7re4YT14o) can be used. Note that these using statements will require bumping to C++20.

dpogue:
Keeping them as enums doesn't work for use in GLSL though. I basically need to end up with `const int kBlendAlpha = 0x2;` in the GLSL shader code 😞

dgelessus:
How about simple `#define` macro constants? Not an optimal solution obviously, because you lose namespacing and types, but would be overall simpler to understand and work with.

dpogue:
`#define` constants would work in GLSL fine, but we do need the namespacing for hsGMatState or it's going to involve a tonne of C++ changes. I'm trying to keep the interface identical on the C++ side.

Hoikas (approved):

colincornaby:
I have been watching this conversation - this will be useful for the Metal shaders. I wish GLSL was more up to speed on C++ but it is what it is. 😛

PR #1396, opened 2023-05-27
Simplify a few bits of `plUnifiedTime`

dgelessus:
Simplifies some code to make it easier to follow, reduces use of platform-dependent APIs, and removes bits that were entirely unused (notably 500 lines of date/time string parsing code). This *should* make no functional difference.


Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
```diff
@@ -171,7 +104,20 @@ void plUnifiedTime::SetSecsDouble(double secs)
...
+
+#if defined(HS_BUILD_FOR_APPLE)
+    // timespec_get is only supported since macOS 10.15,
+    // but clock_gettime exists since macOS 10.12.
+    int res = clock_gettime(CLOCK_REALTIME, &ts);
+    hsAssert(res == 0, "clock_gettime failed");
+#else
+    int res = timespec_get(&ts, TIME_UTC);
+    hsAssert(res != 0, "timespec_get failed");
+#endif
```
dpogue, line 117: In theory we can use `__builtin_available(macOS 10.15, *)` to do runtime detection here, but the ifdefs are going to get ugly...

```cpp
#ifdef HS_BUILD_FOR_APPLE
#   include <AvailabilityMacros.h> // For MAC_OS_X_VERSION_MIN_REQUIRED
#endif

#ifdef HS_BUILD_FOR_APPLE
    #ifdef HAVE_BUILTIN_AVAILABLE
        #if MAC_OS_X_VERSION_MIN_REQUIRED >= 101500
        // timespec_get only available in macOS 10.15 and newer
        if (__builtin_available(macOS 10.15, iOS 13.0, *) {
            int res = timespec_get(&ts, TIME_UTC);
            hsAssert(res != 0, "timespec_get failed");
        } else
        #endif
        #if MAC_OS_X_VERSION_MIN_REQUIRED >= 101200
        // fallback to clock_gettime on macOS 10.12 and newer
        if (__builtin_available(macOS 10.12, iOS 10.0, *) {
            int res = clock_gettime(CLOCK_REALTIME, &ts);
            hsAssert(res == 0, "clock_gettime failed");
        } else
        #endif
    #endif
    {
        // Fall back to gettimeofday :(
        // In which case, we actually need to provide our own timespec struct too
        struct timeval tv;
        int res = gettimeofday(&tv, nullptr);
        hsAssert(res == 0, "gettimeofday failed");

        ts.tv_sec = tv.tv_sec;
        ts.tv_nsec = tv.tv_usec * 1000;
    }
#else
    int res = timespec_get(&ts, TIME_UTC);
    hsAssert(res != 0, "timespec_get failed");
#endif
```

Just flagging that timespec is only available on macOS 10.9 and newer

dgelessus: I don't think we need runtime detection here - as far as I can tell, `timespec_get` and `clock_gettime` do the same thing, just with a slightly different function signature.

If you're interested in very old OS X compatibility (because of PowerPC or something), I would simply make the Mac branch use `gettimeofday` all the time. `plUnifiedTime` only stores microseconds, so we don't care about the extra precision from `struct timespec`. I switched to that just because it's natively supported by all relatively modern platforms.

> Just flagging that timespec is only available on macOS 10.9 and newer

Where do you see that? `struct timespec` itself exists as far back as Mac OS X **10.2** (I don't have any older headers easily available). The functions for getting the current time with that precision were only added much later though.

dpogue: > Where do you see that? `struct timespec` itself exists as far back as Mac OS X **10.2** (I don't have any older headers easily available). The functions for getting the current time with that precision were only added much later though.

Interesting, I was going based on https://developer.apple.com/documentation/kernel/timespec which says 10.9

dgelessus: That's the kernel API documentation - that usually matches the userspace, but I wouldn't trust it. For some reason, Apple doesn't have a proper documentation page for `struct timespec` (or many other Unix APIs).

Either that or Apple is giving inaccurate version numbers for old APIs, like how current Microsoft docs claim that very old Windows APIs only exist since Windows 2000.

Hoikas:
We will probably merge this once the conflicts are corrected.

dgelessus:
Done, thanks for the reminder 🙂

Hoikas (approved):

PR #1397, opened 2023-06-03
Prepare supporting additional languages

dgelessus:
Technical improvements to allow introducing new languages gradually and without breaking anything. Specifically:

* Fixed various issues in code that iterates over all defined languages.
* Added a list of "usable" languages that work well enough for normal play. For now, these are all the languages that were already selectable on the login screen (English, French, German, Spanish, Italian). All other languages (e. g. Japanese) are not considered "usable" and can only be used with an internal client.
* Added language definitions for Dutch, Russian, Polish, and Czech. This matches Myst V's internal language list and the definitions should be fully compatible. These languages are not marked as "usable", because they have no translations (though some Russian translations have already been submitted in H-uru/moul-assets#235).

Standard disclaimer: I cannot test changes to the Max plugin beyond "it compiles".


dgelessus:
@Hoikas When you have the time, could you pull this PR into TrollLand? It would help with testing H-uru/moul-assets#235. I think this PR has a small conflict with my other PR #1379 though - not sure if that will cause issues for you...

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -2431,27 +2431,20 @@ PF_CONSOLE_CMD(App,
...
-        plLocalization::SetLanguage(plLocalization::kGerman);
-    else if (stricmp(params[0], "spanish") == 0)
-        plLocalization::SetLanguage(plLocalization::kSpanish);
-    else if (stricmp(params[0], "italian") == 0)
-        plLocalization::SetLanguage(plLocalization::kItalian);
-    else if (stricmp(params[0], "japanese") == 0)
-        plLocalization::SetLanguage(plLocalization::kJapanese);
-
+    for (const auto &lang : plLocalization::GetAllLanguages()) {
+        if (plLocalization::GetLanguageName(lang).compare_i(params[0]) == 0) {
```
dpogue, line 2443: How would we feel about ISO language codes (i.e., `en`, `fr`, `de`, `it`, `es`, etc.) rather that names? That would open us to support region-specific localizations (such as spelling "Neighbourhood" properly in `en-GB`)

Although I guess that complicates the .loc files too, eh?

Hoikas:
[3f17917](https://github.com/H-uru/Plasma/pull/1397/commits/3f1791757182cc9a3912598e30b5de2f62d00b52) is now on TL.

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
dgelessus: Ideally we would use standard language codes everywhere instead of custom names, yeah. Migrating the existing .loc files would be quite painful though - it would be a big incompatible change to all existing content, which would break open PRs and require changes to other tools like Korman.

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -2431,27 +2431,20 @@ PF_CONSOLE_CMD(App,
...
-    else if (stricmp(params[0], "german") == 0)
-        plLocalization::SetLanguage(plLocalization::kGerman);
-    else if (stricmp(params[0], "spanish") == 0)
-        plLocalization::SetLanguage(plLocalization::kSpanish);
-    else if (stricmp(params[0], "italian") == 0)
-        plLocalization::SetLanguage(plLocalization::kItalian);
-    else if (stricmp(params[0], "japanese") == 0)
-        plLocalization::SetLanguage(plLocalization::kJapanese);
-
+    for (const auto &lang : plLocalization::GetAllLanguages()) {
```
Hoikas, line 2442:
```suggestion
    for (const auto& lang : plLocalization::GetAllLanguages()) {
```

Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
```diff
@@ -547,12 +538,13 @@ void LocalizationDatabase::IVerifyElement(const ST::string &ageName, const ST::s
...
 
-    for (const auto& language : languageNames)
-    {
-        if (theElement.find(language) == theElement.end())
-        {
+    for (auto lang : plLocalization::GetAllLanguages()) {
+        if (
+            plLocalization::IsLanguageUsable(lang)
+            && theElement.find(plLocalization::GetLanguageName(lang)) == theElement.end()
+        ) {
```
Hoikas, line 545: This is difficult to parse, IMO.

Sources/Plasma/PubUtilLib/plResMgr/plLocalization.cpp
```diff
@@ -45,41 +45,51 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
-
-const ST::string plLocalization::fLangNames[] =
-{
-    ST_LITERAL("English"), // kEnglish
-    ST_LITERAL("French"),  // kFrench
-    ST_LITERAL("German"),  // kGerman
-    ST_LITERAL("Spanish"), // kSpanish
-    ST_LITERAL("Italian"), // kItalian
-    ST_LITERAL("Japanese") // kJapanese
+const std::array<std::set<ST::string>, plLocalization::kNumLanguages> plLocalization::fLangCodes = {{
```
Hoikas, line 69: Why not an array of `std::unordered_set`?

Sources/Plasma/PubUtilLib/plResMgr/plLocalization.cpp
```diff
@@ -112,70 +122,69 @@ plFileName plLocalization::GetLocalized(plFileName const &name)
...
-    {
-        for (int j=i; j<kNumLanguages; j++)
-        {
-            if (tagLocs[sortedTagLocs[i]] > tagLocs[sortedTagLocs[j]])
-                sortedTagLocs[i]^=sortedTagLocs[j]^=sortedTagLocs[i]^=sortedTagLocs[j]; // swap the contents (yes, it works)
-        }
-    }
-    // now sortedTagLocs has the indexes of tagLocs sorted from smallest loc to highest loc
+
+    std::sort(sortedLangs.begin(), sortedLangs.end(), [&](auto a, auto b) {
```
Hoikas, line 167:
```suggestion
    std::sort(sortedLangs.begin(), sortedLangs.end(), [&tagLocs](auto a, auto b) {
```

We prefer capture by name.

Sources/Plasma/PubUtilLib/plResMgr/plLocalization.cpp
dgelessus: It was already an ordered `std::set` before, so I kept it that way. Looks like nothing relies on the order though, so it should be safe to change to `std::unordered_set`.

Sources/Plasma/FeatureLib/pfLocalizationMgr/pfLocalizationDataMgr.cpp
dgelessus: I can pull `plLocalization::GetLanguage(lang)` out into a variable I suppose?

Hoikas (approved):
This LTGM, but I'll hold off if you're still waiting on some Russian strings.

dgelessus:
Russian translations of strings added now.

As mentioned in https://github.com/H-uru/moul-assets/pull/237#pullrequestreview-1489103395, there are still problems with Cyrillic text in journals, but I can also fix those separately later (if they're even caused by a code bug at all).

dgelessus:
> problems with Cyrillic text in journals

For reference, these are tracked in #1403 and fixed in #1405. The underlying problem was unrelated to this PR or the Russian/Cyrillic fonts.

PR #1401, opened 2023-06-18
Some NucleusLib Cleanups

dpogue:
This fixes a longstanding cross-project linking issue from plLogicModBase (by extracting that particular bit of behaviour into plLogicModifier).

### DIFF
--- a/Sources/Plasma/NucleusLib/inc/hsResMgr.h
+++ b/Sources/Plasma/NucleusLib/inc/hsResMgr.h
@@ -145,6 +145,9 @@ private:
 public:
     static hsResMgr* ResMgr() { return (hsResMgr*)fResMgr; }
 
+    // External modifier use only
+    static void SetTheResMgr(hsResMgr* mgr) { fResMgr = mgr; }
+
     static plDispatchBase* Dispatch() { hsAssert(fResMgr, "No resmgr"); return fResMgr->Dispatch(); }
 
     static bool Init(hsResMgr* m);
--- a/Sources/Plasma/NucleusLib/pnModifier/CMakeLists.txt
+++ b/Sources/Plasma/NucleusLib/pnModifier/CMakeLists.txt
@@ -28,12 +28,9 @@ target_link_libraries(
         pnNetCommon
     PRIVATE
         pnMessage
-        plNetMessage
         pnNucleusInc
         pnSceneObject
 )
-
-target_include_directories(pnModifier PRIVATE "${PLASMA_SOURCE_ROOT}/PubUtilLib")
 
 source_group("Header Files" FILES ${pnModifier_HEADERS})
 source_group("Source Files" FILES ${pnModifier_SOURCES})
--- a/Sources/Plasma/NucleusLib/pnModifier/plLogicModBase.h
+++ b/Sources/Plasma/NucleusLib/pnModifier/plLogicModBase.h
@@ -76,13 +76,13 @@ protected:
     bool                    fDisabled;
 
     bool IEval(double secs, float del, uint32_t dirty) override { return false; }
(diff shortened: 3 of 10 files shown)


Sources/Plasma/NucleusLib/pnModifier/CMakeLists.txt
```diff
@@ -28,7 +28,6 @@ target_link_libraries(
         pnNetCommon
     PRIVATE
         pnMessage
-        plNetMessage
```
dgelessus, line 31: GitHub doesn't let me comment in the right place, but: you can now remove the extra PubUtilLib include directory a few lines below:

```cmake
target_include_directories(pnModifier PRIVATE "${PLASMA_SOURCE_ROOT}/PubUtilLib")
```

Sources/Plasma/NucleusLib/inc/hsResMgr.h
```diff
@@ -145,6 +145,9 @@ class hsgResMgr
 public:
     static hsResMgr* ResMgr() { return (hsResMgr*)fResMgr; }
 
+    // External modifier use only
+    static void SetTheResMgr(hsResMgr* mgr) { fResMgr = mgr; }
```
dgelessus, line 149: This is for ModDLL stuff I assume? Does this actually work in any useful way? I would have expected that the singleton ResMgr object gets passed around and stored in some places that won't be updated by this call.

Sources/Plasma/NucleusLib/pnModifier/CMakeLists.txt
dpogue: ooh, good catch! Thanks

Sources/Plasma/NucleusLib/inc/hsResMgr.h
dpogue: This is used by ModDLL stuff, but for setting the ResMgr _in_ the DLL to the existing instance in the host application. It matches how the plFactory and plTimerCallbackMgr are handled.

The ModDLL stuff does work, and after these changes you can actually define external creatable classes and use them... whether that ends up being useful or not is up for debate 🤷🏼‍♂️

dgelessus: Oh, I see, the DLL has its own copy of the variable that stores the singleton ResMgr. I assumed that it would share the variable with the main executable and that this would *replace* the existing ResMgr - but the variable is static, so of course it isn't shared that way. Makes sense now, thanks.

Sources/Plasma/NucleusLib/pnModifier/plLogicModBase.h
```diff
@@ -76,13 +76,13 @@ class plLogicModBase : public plSingleModifier
...
 
     bool IEval(double secs, float del, uint32_t dirty) override { return false; }
-    void IUpdateSharedState(bool triggered) const;
     void IHandleArbitration(class plServerReplyMsg* msg);
     bool IEvalCounter();
     virtual void PreTrigger(bool netRequest);
     virtual void Trigger(bool netRequest);
     virtual void UnTrigger();
-    
+    virtual void UpdateSharedState(bool triggered) const { }
```
Hoikas, line 84: This may be a bit of a personal opinion, but I think this should probably be a pure virtual function. Generally, logic modifiers are waiting for the server response here, so, having a blank implementation in some theoretical non-`plLogicModifier` subclass would leave those `plLogicModBase` instances dead in the water.

dpogue: Updated to be pure virtual

Hoikas (approved):

PR #1405, opened 2023-06-24
Fix journals not rendering Unicode text correctly (fixes #1403)

### LINKED ISSUE #1403: Journal text rendering broken for non-English languages
Originally discovered here: https://github.com/H-uru/moul-assets/pull/237#pullrequestreview-1489103395

Several journals render badly in non-English languages. At some point after a non-ASCII character appears, the text stops rendering until the end of the current double page. The following pages sometimes work, sometimes they're partially broken too, and sometimes also completely blank. Sometimes a page renders partially at first, but then turns blank after flipping back and forth a few times.

This affects at least the credits journal, Yeesha's book for the pillar ages on the Relto bookshelf, and the KI/Nexus notebook in the hood classroom. I've reproduced this with French, German, and Russian.

How it's supposed to look in German - with Gehn 1.28, based on 2f1c126ee583dc82768259ebf5d3520d4535ddaf from 2021-04-09:

German credits journal, working

How it looks right now - with 730d2a3d66f230f7c5000d6f8cf55f0cb1c9cb00 from 2023-06-21 and H-uru/moul-assets@08f4fdfbd9421867256ae8c7599b0ea657382f95 from 2023-05-17:

German credits journal, broken

I have no idea yet what causes this bug or when it was introduced (other than "sometime in the last two years").

dgelessus:
For the `ST::string` overloads of the text rendering methods, the `firstClippedChar` output parameter still counted in `wchar_t` units, not UTF-8 bytes as needed for indexing into `ST::string`.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
+++ b/Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
@@ -592,7 +592,16 @@ void    plDynamicTextMap::SetFirstLineIndent( int16_t indent )
 void    plDynamicTextMap::CalcWrappedStringSize( const ST::string &text, uint16_t *width, uint16_t *height, uint32_t *firstClippedChar, uint16_t *maxAscent, uint16_t *lastX, uint16_t *lastY )
 {
     // TEMP
-    CalcWrappedStringSize(text.to_wchar().data(), width, height, firstClippedChar, maxAscent, lastX, lastY);
+    ST::wchar_buffer wcharBuf = text.to_wchar();
+    uint32_t firstClippedWchar;
+    CalcWrappedStringSize(wcharBuf.data(), width, height, &firstClippedWchar, maxAscent, lastX, lastY);
+
+    if (firstClippedChar != nullptr) {
+        // Convert the firstClippedChar offset from wchar_t units to UTF-8 byte units.
+        // This is a bit inefficient, because it creates an actual UTF-8 string
+        // even though we just need the count, but string_theory has no better alternative.
+        *firstClippedChar = ST::string::from_wchar(wcharBuf.data(), firstClippedWchar).size();
+    }
 }
 
 void    plDynamicTextMap::CalcWrappedStringSize( const wchar_t *text, uint16_t *width, uint16_t *height, uint32_t *firstClippedChar, uint16_t *maxAscent, uint16_t *lastX, uint16_t *lastY )
--- a/Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
+++ b/Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
@@ -2364,22 +2364,19 @@ void    pfJournalBook::IRenderPage( uint32_t page, uint32_t whichDTMap, bool sup
                             idx--;
                         break;
                     }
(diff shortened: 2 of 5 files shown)


Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
```diff
@@ -592,7 +592,16 @@ void    plDynamicTextMap::SetFirstLineIndent( int16_t indent )
...
-    CalcWrappedStringSize(text.to_wchar().data(), width, height, firstClippedChar, maxAscent, lastX, lastY);
+    ST::wchar_buffer wcharBuf = text.to_wchar();
+    uint32_t firstClippedWchar;
+    CalcWrappedStringSize(wcharBuf.data(), width, height, &firstClippedWchar, maxAscent, lastX, lastY);
+
+    if (firstClippedChar != nullptr) {
+        // Convert the firstClippedChar offset from wchar_t units to UTF-8 byte units.
+        // This is a bit inefficient, because it creates an actual UTF-8 string
+        // even though we just need the count, but string_theory has no better alternative.
+        *firstClippedChar = ST::string::from_wchar(wcharBuf.data(), firstClippedWchar).size();
```
Hoikas, line 603: Any thoughts about this, @zrax?

dgelessus: By the way, this conversion will probably go away again in the long term. I assume at some point we'll convert the rendering APIs to use `ST::string` all the way through, which makes it trivial to calculate `firstClippedChar` as an UTF-8 count.

zrax: string_theory does have a private API (`_ST_PRIVATE::utf8_measure_from_utf16/32`) that could do this, but I don't think we really want to rely on that.  Another option, since the font rendering code currently (unfortunately) assumes that all renderable characters fit into a single `wchar_t`, is to write a simplified length calculation that just ignores the possibility of UTF-16 surrogates.  It could look something like (completely untested)
```c++
static uint32_t WcharToUtf8Position(const wchar_t *text, uint32_t position)
{
    const wchar_t *end = text + position;
    uint32_t u8pos = 0;
    while (text < end) {
        if (*text < 0x80)
            u8pos += 1;
        else if (*text < 0x800)
            u8pos += 2;
        else /* Assumes all input characters are 0x0000 - 0xffff */
            u8pos += 3;
        ++text;
    }
    return u8pos;
}
```

Then you could just call `*firstClippedChar = WcharToUtf8Position(wcharBuf.data(), firstClippedWchar)`

But yeah, at some point we'll need to make the whole font rendering code work better with ST types, and break some of its faulty assumptions around `wchar_t`

dgelessus: Yeah, I wanted to avoid writing a manual UTF-8 length calculation here - seems easy to get wrong in some way :/ I don't think it's a big enough performance/memory issue to be worth the effort, especially if we're going to throw it out again soon. 

(Actually, the previous line wrapping code did an identical conversion elsewhere, which I was able to remove now, so this change shouldn't even make a difference at all.)

Hoikas (approved):

dpogue (approved):

PR #1407, opened 2023-06-26
fontconfig text font loading for *nix

dpogue:
This uses fontconfig to find the closest match font for the requested name (which I think is always "Trebuchet MS Bold" or "Courier New" right now) on Linux for plTextFont


vcpkg.json
```diff
@@ -6,6 +6,10 @@
       "asio",
       "curl",
       "expat",
+      {
+        "name": "fontconfig",
+        "platform": "linux"
```
Hoikas, line 11:
```suggestion
        "platform": "!(osx | windows)"
```

CMake and the code seem to want fontconfig for non-Apple unixy systems, so we should match that with our platform expression.

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -150,9 +152,37 @@ uint16_t  *plTextFont::IInitFontTexture()
...
+
+    if (result) {
+        FcPatternDestroy(match);
+
+        FT_Done_FreeType(library);
+        return nullptr;
+    }
+
+    char filename[PATH_MAX];
+    strncpy(filename, (char*)(value.u.s), PATH_MAX - 1);
```
Hoikas, line 180:
```suggestion
    strncpy(filename, (char*)(value.u.s), std::size(filename) - 1);
    filename[std::size(filename) - 1] = 0;
```

vcpkg.json
dpogue: This matched what we had for libuuid, but I've updated it as requested.

However, I think the `$<$<PLATFORM_ID:Linux>:Fontconfig::Fontconfig>` check might still only pick this up on Linux specifically?

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
dgelessus: Could all the copying logic be skipped if you pass `value.u.s` directly into `FT_New_Face`?

dpogue: I feel like I tried that initially and it blew up 😕

Hoikas (approved):
#YOLO

PR #1408, opened 2023-06-26
Fix Win32 ifdef check in DisplayScaleChangedMsg

dpogue:
This is more consistent with how we do compile-time checking in the rest of the codebase, and avoid issues on compilers producing Windows binaries that somehow don't define `WIN32` by default (aka mingw-w64)


zrax (approved):
For reference, mingw does define `_WIN32`, which is generally preferred anyway since `WIN32` is defined by the SDK, not the compiler.  However, using `HS_BUILD_FOR_WIN32` is appropriate for consistency with the rest of the Plasma codebase.

PR #1411, opened 2023-06-26
Update Windows Icon

patmauro:
Replaces the ICO files used to compile plClient and plUruLauncher, effectively updating the app icon for Windows. This icon is based on the icon used in the 2007 OSX client, but has been re-centered and updated to support up to 256px (the max size supported by Win11).

To ensure this is future-proof, there is [ongoing work](https://github.com/H-uru/Plasma/discussions/1391) to make sure we have an HQ source asset available in case this needs to be updated or modified for **Mac or Linux** clients, but this is being handled as a separate action item. For now, we have everything we need for **Windows** for the foreseeable future.

Note that this change is NOT being submitted to the OpenUru repo (yet) - instead, I would rather this be "absorbed" into OpenUru organically once it (eventually) catches up with H'uru changes. The reasoning here is twofold: 
(a) While the OU client is still out of sync with the H'uru client, **this new icon will serve as a convenient way to visually distinguish the H'uru client from OU/legacy clients**.
(b) Once OU client eventually reaches full parity with H'uru, we can sync in this icon change as well - This will be a great way of visually communicating to end users that the client has just undergone a significant overhaul, even though the improvements may not be immediately apparent.

Before: 
MOULa

After:
UruIcon256_Centered fw

[For more info, see the thread about this effort in Discussions.](https://github.com/H-uru/Plasma/discussions/1391)


Deledrius (approved):

Hoikas (commented):
Also needs to be done for plCrashHandler.

patmauro:
> Also needs to be done for plCrashHandler.

Good catch - done; phCrashHandler ICO replaced.

Hoikas (approved):

PR #1413, opened 2023-06-28
Fix pfMoviePlayer when libwebm is not available

dpogue:
`fLastImg` is only declared in the header if `USE_WEBM` is true, and we can't call `vpx_img_free` if `USE_WEBM` is not true


Deledrius (approved):
👍🏼

Hoikas (approved):

PR #1414, opened 2023-06-28
Fix Python Parameter display on pre-2022 Max Plugins.

Deledrius:
The change introduced by 107af8272 breaks the plugin as far back as Max 7.  This applies the fix only to versions that were confirmed as requiring it.

### DIFF
--- a/Sources/Tools/MaxComponent/plAutoUIBase.cpp
+++ b/Sources/Tools/MaxComponent/plAutoUIBase.cpp
@@ -501,6 +501,7 @@ void plAutoUIBase::ICreateControls()
     RECT rect;
     GetWindowRect(fhDlg, &rect);
 
+#if MAX_VERSION_MAJOR >= 24 // Max 2022
     // This used to use MoveWindow() to resize the rollup, but in Max 2022,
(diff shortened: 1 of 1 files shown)


Hoikas:
@Deledrius Did you want to test `MoveWindow(..., TRUE);` as well?

Deledrius:
> @Deledrius Did you want to test `MoveWindow(..., TRUE);` as well?

I think it's a better fix, if it works.  No `#ifdef` required.  I tested it with Max 7 and it doesn't introduce breakage like https://github.com/H-uru/Plasma/commit/107af8272d8bfa9200c27017bd4d5439da9b8097 did, so if it works for later versions as well, I think it's the correct way to go.

Hoikas:
If you provide a build or commit for me, I will test it ;)

Hoikas (approved):
We'll probably have to adjust the ifdef some as we find more versions affected by the problem, sadly. The `MoveWindow(...,, TRUE);` proposal doesn't seem to work, unfortunately.

Deledrius:
Ah well, it was worth a try.

PR #1415, opened 2023-06-30
Drop SafeString "backward compat hack" from 20 years ago

dgelessus:
Happy 20th birthday to this code that was only meant to exist for a week or so! I think it's safe to remove by now. This "old" format is never produced by `WriteSafeString` and doesn't seem to be used in any data files.

### DIFF
--- a/Sources/Plasma/CoreLib/hsStream.cpp
+++ b/Sources/Plasma/CoreLib/hsStream.cpp
@@ -155,12 +155,7 @@ ST::string hsStream::ReadSafeString()
     ST::char_buffer name;
     uint16_t numChars = ReadLE16();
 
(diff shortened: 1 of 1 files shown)


Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -155,12 +155,7 @@ ST::string hsStream::ReadSafeString()
     ST::char_buffer name;
     uint16_t numChars = ReadLE16();
 
-#ifndef REMOVE_ME_SOON
-    // Backward compat hack - remove in a week or so (from 6/30/03)
```
zrax, line 159: 🎂

zrax (approved):

Deledrius (approved):

PR #1418, opened 2023-07-08
Specify older Pillow version to fix cairosvg build

dpogue:
Update to latest cairosvg too, since that seems to work fine.

On July 1st, a new major version of Pillow was released, and it requires zlib to compile. For whatever reason, that's not working when pip tries to install it. However cairosvg is fine using the older version so it's easiest to rollback for the moment to fix the builds.

### DIFF
--- a/Scripts/Ports/python3-cairosvg/vcpkg.json
+++ b/Scripts/Ports/python3-cairosvg/vcpkg.json
@@ -1,7 +1,6 @@
 {
   "name": "python3-cairosvg",
-  "version-semver": "2.5.2",
-  "port-version": 1,
+  "version-semver": "2.7.0",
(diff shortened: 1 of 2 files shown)


Hoikas (commented):
You also need to update the port version in `vcpkg.json` to `2.7.0`.

dpogue:
Good catch, thanks

zrax (approved):

Scripts/Ports/python3-cairosvg/vcpkg.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "python3-cairosvg",
-  "version-semver": "2.5.2",
+  "version-semver": "2.7.0",
   "port-version": 1,
```
Hoikas, line 4:
```suggestion
```

Hoikas (changes requested):

Hoikas (approved):

PR #1419, opened 2023-07-09
Prerelease publishing

### LINKED ISSUE #1417: Publish build artifacts to a "last successful build" prerelease after each merge
Currently, we have a CI system that runs on every pull request and on every merge to master. It builds Windows 32-bit clients, Windows 64-bit clients, a whole bunch of 3DS Max plugins, and a subset of tools for macOS and Linux. These are attached to the GitHub Actions build results as build artifacts and can be downloaded.

However, these are only available to download to people who are logged in to GitHub, and it's a bit difficult to know where to go to find them (click on Actions, find the right run of the CI workflow, click into it, scroll down in a page with multiple scrollable areas, etc.)

A more user-friendly option would be to publish these build artifacts to a release on GitHub that would be featured on the main repository page and would allow downloads from anyone without requiring a GitHub login. We probably don't want to create a new release for every merge though, so a continuously-updated "last successful build" pre-release is likely the best option.

Hoikas has set up publishing of artifacts to releases as part of the korman workflows, and I've done some similar stuff for projects at work.

dpogue:
Closes #1417.

### DIFF
--- a/.github/PRERELEASE_NOTES.md
+++ b/.github/PRERELEASE_NOTES.md
@@ -1,0 +1,36 @@
+Here you will find the build artifacts for the last successful build of the master branch of the Plasma repository. These builds are automatically generated and are not tested. They may include bugs, errors, and partially-completed work.  
+**These are made available for you to use at your own risk.**
+
+For Age building and development activities, use an **Internal** client.  
+
+The clients for macOS and Linux are not functional at this time, but other tools should run as expected.
+
+Windows Clients & Tools
+-----------------------
+* 32-bit Internal Client: `plasma-windows-x86-internal-release.zip`
+* 32-bit External Client: `plasma-windows-x86-external-release.zip`
+* 64-bit Internal Client: `plasma-windows-x64-internal-release.zip`
+* 64-bit External Client: `plasma-windows-x64-external-release.zip`
+
+macOS Clients & Tools
+---------------------
+Currently the automated build system does not produce Apple Silicon (M1) builds.
+
+* 64-bit Intel Internal Client: `plasma-macos-x64-internal-release.zip`
+* 64-bit Intel External Client: `plasma-macos-x64-external-release.zip`
+
+Linux Clients & Tools
+---------------------
+* 64-bit Internal Client: `plasma-linux-x64-internal-release.zip`
+* 64-bit External Client: `plasma-linux-x64-external-release.zip`
+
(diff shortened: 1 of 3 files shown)


.github/PRERELEASE_NOTES.md
```diff
@@ -0,0 +1,38 @@
+Here you will find the build artifacts for the last successful build of the master branch of the Plasma repository. This builds are automatically generated and are not tested. They may include bugs, errors, and partially-completed work.  
+**These are made available for you to use at your own risk.**
+
+For Age building and development activities, use an **Internal** client.  
+For playing Uru on public servers, use an **External** client.
```
dgelessus, line 5: I would remove this line, because AFAIK there are currently no public shards where you can safely play with a mainline H'uru client. Even if there was one, you would still need an internal client to use `/LocalData` and prevent the client from "updating" itself to the standard client provided by the server.

```suggestion
```

dgelessus (commented):
This has the slightly annoying side effect that every pre-release upload triggers a GitHub notification that a new release was made. You can suppress this by changing your "Watch" settings for this repo to "Custom" and checking everything except releases. (I hope that doesn't silently disable any other notifications that don't fall into any category...)

.github/PRERELEASE_NOTES.md
Hoikas: Agreed. External clients are currently only useful for shard operators.

dpogue: Do we want to archive them here then? I guess maybe it's still useful as a place for shard operators to grab them

Hoikas: Yeah. I think we still want to have them archived for anyone who wants to set up their own complete shard.

.github/PRERELEASE_NOTES.md
```diff
@@ -0,0 +1,36 @@
+Here you will find the build artifacts for the last successful build of the master branch of the Plasma repository. This builds are automatically generated and are not tested. They may include bugs, errors, and partially-completed work.
```
Hoikas, line 1:
```suggestion
Here you will find the build artifacts for the last successful build of the master branch of the Plasma repository. These builds are automatically generated and are not tested. They may include bugs, errors, and partially-completed work.  
```

Hoikas (approved):

PR #1421, opened 2023-07-13
Improve compatibility with some unix systems

dpogue:
Not to set expectations too high, but with these changes I was able to compile the majority of the engine on an OpenIndiana (Solaris) virtual machine.

Tested and working:
* `test_CoreLib`
* `test_pnEncryption`
* `test_plUnifiedTime`
* `test_plLocalization`
* `test_pfPython`
* `plSystemInfo`
* `plFilePatcher`
* `plPageInfo`

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -209,8 +209,8 @@ inline double hsSwapEndianDouble(double dvalue)
 // Generic versions for use in templates
 template <typename T> inline T hsToLE(T value) = delete;
 template <> inline char hsToLE(char value) { return value; }
-template <> inline int8_t hsToLE(int8_t value) { return value; }
-template <> inline uint8_t hsToLE(uint8_t value) { return value; }
+template <> inline signed char hsToLE(signed char value) { return value; }
+template <> inline unsigned char hsToLE(unsigned char value) { return value; }
 template <> inline int16_t hsToLE(int16_t value) { return (int16_t)hsToLE16((uint16_t)value); }
 template <> inline uint16_t hsToLE(uint16_t value) { return hsToLE16(value); }
 template <> inline int32_t hsToLE(int32_t value) { return (int32_t)hsToLE32((uint32_t)value); }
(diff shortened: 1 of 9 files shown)


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -209,7 +213,9 @@ inline double hsSwapEndianDouble(double dvalue)
 // Generic versions for use in templates
 template <typename T> inline T hsToLE(T value) = delete;
 template <> inline char hsToLE(char value) { return value; }
+#ifndef _CHAR_IS_SIGNED
 template <> inline int8_t hsToLE(int8_t value) { return value; }
+#endif
```
zrax, line 218: This probably isn't actually the best fix for this (after all, the difference is that `int8_t` is defined as `char` rather than `signed char`, not that `char` is signed).  Instead, we should probably rewrite these to use `signed char` and `unsigned char` instead of `int8_t` and `uint8_t`, which also matches the fixes we made to libhsplasma.

Sources/Plasma/CoreLib/hsThread_Unix.cpp
```diff
@@ -81,8 +81,14 @@ hsGlobalSemaphore::hsGlobalSemaphore(int initialValue, const ST::string& name)
     fPSema = nullptr;
     fNamed = !name.empty();
     if (fNamed) {
+        const char* semName = name.c_str();
+        if (semName[0] != '/') {
+            /* UNIX named semaphores need to start with a slash */
+            semName = ST::format("/{}", name).c_str();
+        }
```
zrax, line 88: To avoid using a dangling pointer (the string produced by `ST::format` goes out of scope)
```suggestion
        ST::string semName = name;
        if (semName.front() != '/') {
            /* UNIX named semaphores need to start with a slash */
            semName = ST::format("/{}", name);
        }
```

Sources/Plasma/CoreLib/hsThread_Unix.cpp
```diff
@@ -81,8 +81,14 @@ hsGlobalSemaphore::hsGlobalSemaphore(int initialValue, const ST::string& name)
...
+
         /* Named semaphore shared between processes */
-        fPSema = sem_open(name.c_str(), O_CREAT, 0666, initialValue);
+        fPSema = sem_open(semName, O_CREAT, 0666, initialValue);
```
zrax, line 91:
```suggestion
        fPSema = sem_open(semName.c_str(), O_CREAT, 0666, initialValue);
```

zrax (approved):

PR #1425, opened 2023-07-15
Assign sequential object IDs at Max export time

dpogue:


Sources/Plasma/PubUtilLib/plResMgr/plRegistryKeyList.cpp
```diff
@@ -201,6 +201,18 @@ bool plRegistryKeyList::SetKeyUnused(plKeyImp* key, LoadStatus& loadStatusChange
     return foundKey != nullptr;
 }
 
+void plRegistryKeyList::PrepForWrite()
+{
+    uint32_t objectID = 1;
+    for (auto it = fKeys.begin(); it != fKeys.end(); ++it) {
+        plKeyImp* key = *it;
```
Hoikas, line 208:
```suggestion
    for (plKey& key : fKeys) {
```

dpogue: This would have to be `for (plKeyImp* key : fKeys) {` instead (since we can't call `SetObjectID` on `plKeyData`), but I'm happy to make that change

Hoikas: Whoops! fixed

Hoikas (approved):

Sources/Plasma/PubUtilLib/plResMgr/plRegistryKeyList.h
```diff
@@ -97,6 +97,10 @@ class plRegistryKeyList
     void SetKeyUsed(plKeyImp* key) { ++fReffedKeys; }
     bool SetKeyUnused(plKeyImp* key, LoadStatus& loadStatusChange);
 
+    // Export time only.  Before we write to disk, assign all the static keys
+    // object ID's that they can use to do fast lookups at load time.
```
Deledrius, line 101:
```suggestion
    // object IDs that they can use to do fast lookups at load time.
```

Deledrius (changes requested):

Deledrius (approved):

Hoikas:
So, it occurs to me that this has the same problem that libhsplasma had previously. What's happening here is that a PRP's ObjIDs are being sorted, then the Prp is written. Then, the next PRP's ObjIDs are sorted and the PRP is written. So, any cross-page references between PRPs may be shattered due to wrong ObjIDs being written if a reference is written out to another PRP before the other PRP is sorted. Therefore, the PrepForWrite function needs to be called on *all* RegistryPageNodes before any PRP is written.

dpogue:
@Hoikas I had the same thought... except that this appears to be consistent with the behaviour from before https://github.com/H-uru/Plasma/pull/264 was merged 🤷🏼‍♂️

Hoikas:
Let's go ahead and preemptively fix that problem so we don't have any land mines waiting on us. Cyan tended to be pretty good about limiting themselves one or two PRPs per max file, but we need to be more resilient due to not being on the other side of the hallway from our users 😉

dpogue:
I exported the Nexus max files with this, and it appears that all the Object IDs are sequential.

Sources/Plasma/PubUtilLib/plResMgr/plRegistryNode.cpp
```diff
@@ -204,6 +204,18 @@ void plRegistryPageNode::UnloadKeys()
...
+void plRegistryPageNode::PrepForWrite()
+{
+    if (!fIsNewPage)
+        return;
+
+    KeyMap::const_iterator it;
+    for (it = fKeyLists.begin(); it != fKeyLists.end(); it++) {
+        plRegistryKeyList* keyList = it->second;
+        keyList->PrepForWrite();
+    }
```
Hoikas, line 216:
```suggestion
    for (auto [idx, keyList] : fKeyLists)
        keyList->PrepForWrite();
```

Hoikas (approved):

PR #1426, opened 2023-07-15
Minor fixes for GameMgr-based marker games

dgelessus:
Fixes a few small errors in #1217 that I encountered during normal testing. (Haven't done any in-depth testing of marker or GameMgr stuff, so there may be other problems lurking still.)

One of the tracebacks was longer than the `LogPythonTraceback` network message formally allows (1023 characters), so I added a workaround for that as well.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
@@ -58,8 +58,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pyMatrix44.h"
 #include "pyObjectRef.h"
 
-#include "pnNetBase/pnNetBase.h"
-
 // CPython specific init stuff
 #include <cpython/initconfig.h>
 #include <pylifecycle.h>
@@ -112,8 +110,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plPythonSDLModifier.h"
 
 // For printing to the log
+#include "plNetClientComm/plNetClientComm.h"
 #include "plStatusLog/plStatusLog.h"
-#include "plNetGameLib/plNetGameLib.h"
 
 // vault
 #include "pyVaultNode.h"
@@ -687,11 +685,6 @@ PYTHON_CLASS_CONVERT_FROM_IMPL(ptOutputRedirector, pyOutputRedirector)
 class pyErrorRedirector
 {
 private:
-    static constexpr char16_t kTracebackTruncatedMarker[5] = {u'[', u'.', u'.', u'.', u']'};
-    // Maximum length of a traceback chunk to still leave room for the truncation marker.
-    // Like kMaxTracebackLength, includes one char16_t extra for the zero terminator.
-    static constexpr unsigned int kTracebackChunkLength = kMaxTracebackLength - std::size(kTracebackTruncatedMarker);
-
     static bool fTypeCreated;
 
     ST::string_stream fData;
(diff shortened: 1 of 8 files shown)


Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
```diff
@@ -81,6 +81,7 @@ const unsigned kMaxPublisherAuthKeyLength       = 64;
 const unsigned kMaxGTOSIdLength                 = 8;
 const unsigned kMaxGameScoreNameLength          = 64;
 const unsigned kMaxEmailAddressLength           = 64;
+const unsigned kMaxTracebackLength              = 1024;
```
dgelessus, line 84: It would be nice if we could just increase this value. It's an artificial limit - at the protocol level, the length is sent as a `uint16_t`. DIRTSAND and MOSS already accept longer lengths, but Cyan's server probably enforces the 1024-char limit like the client does :(

Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -685,6 +687,11 @@ PYTHON_CLASS_CONVERT_FROM_IMPL(ptOutputRedirector, pyOutputRedirector)
 class pyErrorRedirector
 {
 private:
+    static constexpr char16_t kTracebackTruncatedMarker[5] = {u'[', u'.', u'.', u'.', u']'};
```
Hoikas, line 690: It might be more readable to do something like

```suggestion
    static constexpr std::u16string_view kTracebackTruncatedMarker{ U"[...]" };
```

Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -725,7 +732,20 @@ class pyErrorRedirector
...
+            // so split it up into multiple small messages
+            // and add a truncation marker at the end of every message but the last.
+            auto pos = wdata.begin();
+            for (; wdata.end() - pos >= kMaxTracebackLength - 1; pos += kTracebackChunkLength - 1) {
+                ST::utf16_buffer chunk = ST::utf16_buffer(pos, kMaxTracebackLength - 1);
+                memcpy(&chunk[kTracebackChunkLength - 1], kTracebackTruncatedMarker, sizeof(kTracebackTruncatedMarker));
+                NetCliAuthLogPythonTraceback(chunk.data());
+            }
+            NetCliAuthLogPythonTraceback(pos);
+        }
```
Hoikas, line 748: I wonder if it would be better to move this into `NetCliAuthLogPythonTraceback()`? I feel like users of this API should probably not be burdened with needing to chunk the traceback, even though this is currently the only user of the API.

Hoikas (commented):
Thanks for sniffing these out!

Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
dgelessus: It would definitely be more convenient in case anything else wants to send Python errors. Right now we have this pattern that every `NetCli` function is a thin wrapper around a network message, so I didn't want to make one of them send multiple messages. But perhaps that doesn't matter much.

dgelessus: There's now a new `NetComm` function that handles the chunk splitting before calling the `NetCli` function.

Sources/Plasma/PubUtilLib/plNetClientComm/plNetClientComm.cpp
```diff
@@ -1255,6 +1259,42 @@ void NetCommSendFriendInvite (
...
 }
 
+static constexpr std::u16string_view kTracebackTruncatedMarker = u"[...]"sv;
+// Maximum length of a traceback chunk to still leave room for the truncation marker.
+// Like kMaxTracebackLength, includes one char16_t extra for the zero terminator.
+static constexpr unsigned int kTracebackChunkLength = kMaxTracebackLength - std::size(kTracebackTruncatedMarker);
+
+typedef void LogErrorFunc(const char16_t* errorText);
+
+template<LogErrorFunc logErrorFunc> static void NetCommLogChunkedError(const ST::string& errorText)
```
Hoikas, line 1269:
```suggestion
template<LogErrorFunc logErrorFunc>
static void NetCommLogChunkedError(const ST::string& errorText)
```

Hoikas (commented):
Minor style nitpick

Hoikas (approved):

PR #1427, opened 2023-07-16
Let zlib handle the gzip header instead of parsing it manually

dgelessus:
For some reason, the code in `IValidateGzHeader` was copied from zlib, when they could have just used it through the API... (Perhaps this is *really* old code from before the main zlib API had gzip support?)


Hoikas (commented):
The diff is a bit hard to read, but I think this looks good. I think I want to deploy to TL before I approve, though. (No blocker on someone else approving, though.)

dgelessus:
Good idea - `plZlibStream` is only used by the patcher and that doesn't get tested much outside of TrollLand. (I already did a quick test with this change and the patcher *seems* to work.)

Deledrius (approved):

Hoikas (approved):
The TL patcher works great 👍

PR #1428, opened 2023-07-16
`ptGUIControl` construction improvement.

Hoikas:
This allows the scripter to avoid indirections through `ptKey` in their python code such that instead of writing something like:
```python
ctrl = ptGUIControlButton(MyGUIAttr.dialog.getControlFromTag(kButtonTag))
ctrl.show()
```

The following can be used:
```python
ctrl = MyGUIAttr.dialog.getControlModFromTag(kButtonTag))
ctrl.show()
```

This allows for more concise Python code and avoids constructing temporary `pyKey` objects in the engine. This API change is backwards incompatible, so I opted to make this a new method. I elected to not make `ptGUIDialog` a mapping type because controls can be fetched by both ID and tag, which are distinct values. It's better to be explicit in this case, IMO.

Further, I have applied the new methods to all Python code except for the KI (ki module and xOptionsMenu) because there are roughly 500 instances of the old junk in those files, and I'm getting tired.

### DIFF
--- a/Scripts/Python/xAvatarCustomization.py
+++ b/Scripts/Python/xAvatarCustomization.py
@@ -640,7 +640,7 @@ class xAvatarCustomization(ptModifier):
         "Activated... we just landed in the AvaCusta Age"
         if state and id == InRoomActivator.id:
             ZoomResponder.run(self.key, state="ZoomOut")
-            ptGUIControlButton(AvCustGUI.dialog.getControlFromTag(kAvatarCameraID)).disable()
+            AvCustGUI.dialog.getControlModFromTag(kAvatarCameraID).disable()
             TestMap.textmap.clearToColor(ptColor(0.92,0.82,0.63,1))
             TestMap.textmap.flush()
             # disable the KI until further notice
@@ -652,7 +652,7 @@ class xAvatarCustomization(ptModifier):
             # callback from the reset confirmation dialog box
             if state:
                 self.dirty = 0 # we are resetting the avatar, so hide the reset button
-                ptGUIControlButton(AvCustGUI.dialog.getControlFromTag(kAvatarResetID)).hide()
+                AvCustGUI.dialog.getControlModFromTag(kAvatarResetID).hide()
                 self.IResetAvatar()
 
     def OnGUINotify(self,id,control,event):
@@ -678,7 +678,7 @@ class xAvatarCustomization(ptModifier):
                 # is it one of the knobbies?
                 if isinstance(control,ptGUIControlValue):
                     self.dirty = 1 # we changed something! so show the "reset" button
-                    ptGUIControlButton(AvCustGUI.dialog.getControlFromTag(kAvatarResetID)).show()
+                    AvCustGUI.dialog.getControlModFromTag(kAvatarResetID).show()
                     if tagID >= kMorphSliderOffset and tagID < kMorphSliderOffset+kNumberOfMorphs:
                         # it is a morph slider
                         self.IMorphItem(tagID)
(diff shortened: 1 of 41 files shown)


Scripts/Python/xAvatarCustomization.py
```diff
@@ -1373,11 +1373,7 @@ def IInitAvaCusta(self):
         SetDefaultSettings() # grab the stuff we came in with
         PtDisableAvatarCursorFade()
 
-        if InAvatarCloset:
-            ptGUIControlButton(AvCustGUI.dialog.getControlFromTag(kAvatarSwitchID)).hide()
-        else:
-            ptGUIControlButton(AvCustGUI.dialog.getControlFromTag(kAvatarSwitchID)).hide()
-
+        AvCustGUI.dialog.getControlModFromTag(kAvatarSwitchID).hide()
```
Hoikas, line 1376: :trollface:

Scripts/Python/xAvatarCustomization.py
```diff
@@ -1547,18 +1543,17 @@ def IColorShowingItem(self,controlID):
                 # if the saturation is zero then don't allow tiniting
                 if controlID == kColor1ClickMap or controlID == kColor2ClickMap:
                     #Make sure that we're not zoomed in (changing eye color)
-                    panelRG = ptGUIControlRadioGroup(AvCustGUI.dialog.getControlFromTag(kPanelsRGID))
```
Hoikas, line 1550: This is already fetched on line 1534. No need to do it again.

Sources/Plasma/FeatureLib/pfPython/pyGUIControlButton.cpp
```diff
@@ -53,9 +53,9 @@ pyGUIControlButton::pyGUIControlButton(plKey objkey) : pyGUIControl(std::move(ob
 {
 }
 
-bool pyGUIControlButton::IsGUIControlButton(pyKey& gckey)
+bool pyGUIControlButton::IsGUIControlButton(const plKey& key)
 {
-    if ( gckey.getKey() && pfGUIButtonMod::ConvertNoRef(gckey.getKey()->ObjectIsLoaded()) )
+    if ( key && pfGUIButtonMod::ConvertNoRef(key->ObjectIsLoaded()) )
```
dpogue, line 58: `key` is a reference, so this can never be null, right?

Sources/Plasma/FeatureLib/pfPython/pyGUIDialog.cpp
```diff
@@ -246,6 +279,26 @@ PyObject* pyGUIDialog::GetControl( uint32_t idx )
     PYTHON_RETURN_ERROR;
 }
 
+PyObject* pyGUIDialog::GetControlMod(uint32_t tagID) const
```
dpogue, line 282: This is control index, not tag ID, right?

Sources/Plasma/FeatureLib/pfPython/pyGUIControlButton.cpp
Hoikas: Right, but the underlying `plKeyData*` can be `nullptr`, so we must test before accessing.

Sources/Plasma/FeatureLib/pfPython/pyGUIDialog.cpp
Hoikas: Good point, I'll fix this.

Sources/Plasma/FeatureLib/pfPython/pyGUIControlButton.cpp
dpogue: I guess behind the scenes this is actually doing `key != (plKeyData*)nullptr` and calling this operator? https://github.com/H-uru/Plasma/blob/ad3863057e51ffb65a43d30c6856d7ddb769b139/Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h#L81

dgelessus: It's implicitly using `operator plKeyImp*() const` to get at the underlying `fKeyData` pointer, and then does a normal bool conversion (`!= nullptr`) on that pointer.

(Coincidentally, I'm working on a change that will make this bit of `plKey` less convoluted.)

Hoikas: Yes. This is a fairly common pattern in Plasma to make sure we don't crash dereferencing null keys. The only thing I did in these functions is change the argument from `plKey`'s Python wrapper class `pyKey` to `plKey` itself. We were previously and still are checking that the underlying `plKeyData` is not null.

dpogue (approved):
LGTM

Deledrius (approved):

PR #1429, opened 2023-07-17
Avoid exposing `plKey` implementation details

dgelessus:
`plKey`'s contents are separated into a public interface `plKeyData` and an internal implementation subclass `plKeyImp`. However, the main `plKey` API had an implicit conversion to `plKeyImp`, making it relatively easy to access the implementation details anyway.

This changes that to require an explicit method call, which only becomes available if you include the implementation header. This will make it easier to find all the places that rely on this conversion.

Confusingly, the implicit conversion was used quite often whenever a `plKey` was used as a boolean (to check if its underlying pointer is null). Those uses don't actually need the implementation details and are now handled by a simple `operator bool`.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h
+++ b/Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h
@@ -82,7 +82,7 @@ public:
     plKeyData* operator->() const;
     plKeyData& operator*() const;
 
-    operator bool() const { return fKeyData; }
+    operator bool() const { return fKeyData != nullptr; }
 
     static plKey Make(plKeyData* data) { return plKey(data); }
 
--- a/Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h
+++ b/Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h
@@ -92,7 +92,7 @@ protected:
     void IIncRef();
     void IDecRef();
 
-    // Internal constructor, extra param is to distinguish it from the void* constructor
+    // Internal constructor
     plKey(plKeyData* data);
 };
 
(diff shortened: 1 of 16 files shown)


Sources/Plasma/NucleusLib/pnKeyedObject/plKey.h
```diff
@@ -83,7 +82,7 @@ class plKey
     plKeyData* operator->() const;
     plKeyData& operator*() const;
 
-    operator plKeyImp*() const { return (plKeyImp*)fKeyData; }
+    operator bool() const { return fKeyData; }
```
zrax, line 85: Suggestion for clarity:
```suggestion
    operator bool() const { return fKeyData != nullptr; }
```

Sources/Plasma/NucleusLib/pnKeyedObject/plKeyImp.h
```diff
@@ -172,6 +177,9 @@ class plKeyImp : public plKeyData
     mutable int16_t             fPendingRefs;   // Outstanding requests I have out.
     mutable std::vector<plKeyImp*>  fClones;    // clones of me
     mutable plKey               fCloneOwner;    // pointer for clones back to the owning key
+
+    friend class pfConsoleActiveRefPeeker;
+    friend class plPageOptimizer; // needs to update fStartPos
```
zrax, line 182: I don't love these, but they're infinitely better than the UB we were relying on before...  :trollface:

zrax (approved):

zrax (approved):

Hoikas:
The CI failure is due to nasm.us being down again, unfortunately.

PR #1430, opened 2023-07-17
Vendor the PhysX 4.1.2 vcpkg port for now

dpogue:
It was [replaced upstream](https://github.com/microsoft/vcpkg/pull/32568) with the PhysX 5.1 port, but that drops support for several of our target platforms.


Hoikas:
😞

Hoikas (dismissed):

Hoikas (commented):
On second thought, #1420 will cause the vcpkg submodule to be deep cloned, so we could pin the last PhysX 4.1.2 version from vcpkg in our vcpkg.json file. We may eventually want to vendor the port if we fix any additional bugs in PhysX.

dpogue:
Superseded by #1444

PR #1431, opened 2023-07-18
Update website download and build instructions

dpogue:
* Encourage using VS2022 and the built-in git functionality for a more seamless process that doesn't involve dealing with git bash and submodules
* Keep the list of dependencies up to date
* Link to our new pre-release tag with artifacts for download instead of sending people to mine for zip files in the Actions tab

### DIFF
--- a/building.html
+++ b/building.html
@@ -45,7 +45,8 @@
           <li><a href="https://www.openssl.org/">OpenSSL</a></li>
           <li><a href="https://openal-soft.org/">OpenAL Soft</a></li>
           <li><a href="https://libexpat.github.io/">Expat</a></li>
-          <li><a href="https://libjpeg-turbo.org/">libjpeg</a></li>
+          <li><a href="https://www.freetype.org/">FreeType</a></li>
+          <li><a href="https://libjpeg-turbo.org/">libjpeg-turbo</a></li>
           <li><a href="http://www.libpng.org/">libpng</a></li>
           <li><a href="https://zlib.net/">zlib</a></li>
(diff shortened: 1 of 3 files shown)


building.html
```diff
@@ -69,24 +71,20 @@ <h3 id="dependencies">Library Dependencies</h3>
...
+      <p>Compiling on Windows requires Visual Studio. We recommend <a href="https://visualstudio.microsoft.com/downloads/">Visual Studio 2022 Community Edition</a>, which is available as a free download. Install the “Desktop development with C++” workload.</p>
 
       <ol>
         <li>
-          <p>Clone the repository from GitHub, including all submodules, using a git client. On the command line, this can be done by executing the following:<br>
-            <code>git clone --recurse-submodules https://github.com/H-uru/Plasma.git</code></p>
-
-          <p class="infonote" data-infotype="warning"><b>Note:</b> We recommend using <a href="https://git-scm.com/download/win">Git for Windows</a>. Using the git program provided by Cygwin will not set permissions properly on some files and result in build failures.</p>
-
-          <p class="infonote" data-infotype="warning"><b>Note:</b> Some dependencies may fail to build if the path to the code contains spaces.</p>
```
dgelessus, line 81: Was this removed on purpose? Last I checked we still have at least one autoconf-based dependency, which won't work properly with spaces in the path.

dpogue: oh, that is a good catch, I'll add that back to the step about cloning to a folder

dgelessus (approved):

Hoikas (approved):

PR #1432, opened 2023-07-18
Update README to point to website instructions

dpogue:
It's better to maintain these in a single place rather than multiple places where they can get out of date and out of sync (although yes it would be nice for the instructions to be available after cloning the repo).

Also update the list of libraries to keep it accurate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,6 @@ enhanced features.
 
 For more information on Myst Online, see http://mystonline.com/developers/
 
-For a project roadmap, see https://github.com/H-uru/Plasma/wiki/Roadmap
-
 
 Related Projects
 ----------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,8 +8,6 @@ enhanced features.
 
 For more information on Myst Online, see http://mystonline.com/developers/
 
-For a project roadmap, see https://github.com/H-uru/Plasma/wiki/Roadmap
```
dgelessus, line 11: Would be good to add this link on the website's "Get Involved" page instead. I don't think this wiki page (or any part of the wiki...) is linked anywhere else.

dpogue: There's not much left on that page, and we've been trying to move more towards GitHub Discussions for things like roadmap ideas.

There is a lack of pointing to GitHub Discussions on the website "Get Involved" page though, which should probably be fixed...

dgelessus:
> although yes it would be nice for the instructions to be available after cloning the repo

A possible solution: currently the website is on its own `gh-pages` branch, but GitHub Pages also supports putting the website files in a "docs" folder on the default branch, which would be easily accessible in a local checkout. I can think of some minor downsides to doing this, but overall it would be more convenient than the current situation.

A small difficulty is that we already have a "Docs" folder on the master branch, so you can't also have a "docs" folder on a case-insensitive file system. Though perhaps it would be a good opportunity to make those docs available on the website somewhere?

Hoikas (approved):

PR #1434, opened 2023-07-21
Minor tweaks to hsMatrix44 Accelerate.framework support

dpogue:
1. This is a C++ file, and Accelerate is a C library, so we should use `#include` instead of `#import`.

2. Somewhere around Mac OS X 10.10, some parameters to `vDSP_mmul` were changed from `float*` to `const float*`. But earlier versions complain if you try to pass them `const float*` values, so just cast to `float*`.

Tested and successfully compiled on Mac OS X 10.5 PowerPC 😛

### DIFF
--- a/Sources/Plasma/CoreLib/hsMatrix44.cpp
+++ b/Sources/Plasma/CoreLib/hsMatrix44.cpp
@@ -52,7 +52,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <cmath>
 
 #ifdef HS_BUILD_FOR_APPLE
-#import <Accelerate/Accelerate.h>
+#include <Accelerate/Accelerate.h>
 #endif
 
 static hsMatrix44 myIdent = hsMatrix44().Reset();
@@ -115,9 +115,9 @@ hsMatrix44 hsMatrix44::mult_accelerate(const hsMatrix44 &a, const hsMatrix44 &b)
         return b;
     if( b.fFlags & hsMatrix44::kIsIdent )
         return a;
-    
-    vDSP_mmul((const float*)a.fMap, 1, (const float*)b.fMap, 1, (float*)&(c.fMap), 1, 4, 4, 4);
-    
+
+    vDSP_mmul(const_cast<float*>(a.fMap[0]), 1, const_cast<float*>(b.fMap[0]), 1, c.fMap[0], 1, 4, 4, 4);
+
     return c;
 }
 #endif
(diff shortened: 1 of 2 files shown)


colincornaby:
> Tested and successfully compiled on Mac OS X 10.5 PowerPC 😛

You're a monster! 😆

colincornaby (approved):

Sources/Plasma/CoreLib/hsMatrix44.cpp
```diff
@@ -115,9 +115,9 @@ hsMatrix44 hsMatrix44::mult_accelerate(const hsMatrix44 &a, const hsMatrix44 &b)
         return b;
     if( b.fFlags & hsMatrix44::kIsIdent )
         return a;
-    
-    vDSP_mmul((const float*)a.fMap, 1, (const float*)b.fMap, 1, (float*)&(c.fMap), 1, 4, 4, 4);
-    
+
+    vDSP_mmul((float*)a.fMap, 1, (float*)b.fMap, 1, (float*)&(c.fMap), 1, 4, 4, 4);
```
dgelessus, line 119: Would this work to make it clear that we mean to cast away the `const` here? (Replacing `fMap` with `fMap[0]` should make the other part of the cast unnecessary, because `fMap[0]` will decay to a `float*`.)

```suggestion
    vDSP_mmul(const_cast<float*>(a.fMap[0]), 1, const_cast<float*>(b.fMap[0]), 1, const_cast<float*>(c.fMap[0]), 1, 4, 4, 4);
```

dpogue: The `fMap` property in hsMatrix44 is not `const`, so I don't think there's any issue here. It's fine to pass a `float*` to a function that's expecting `const float*` (the `const` just means that the function guarantees it won't modify the parameter).

`fMap[0]` might be better than casting though... let me try that

dgelessus: `a` and `b` are `const` references though, so their `fMap`s will be `const` I think - it's an array field, not a pointer, so the `const`-ness of the entire object will "propagate" to it. `c.fMap` won't need a cast though, as you say.

dpogue: Ahh, good catch there, I've updated the PR

dgelessus (approved):

Hoikas (approved):

PR #1436, opened 2023-07-21
Fix footstep sounds being in Chinese.

### LINKED ISSUE #1435: PlasmaMax footstep sounds are garbled unicode
plasmamax-footsteps

Hoikas:
Fixes #1435.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.h
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.h
@@ -109,7 +109,7 @@ public:
         kMaxSurface,
         kFootNoSurface = kMaxSurface,
     };  
-    static const char *SurfaceStrings[];
+    static ST::string SurfaceStrings[];
 };
 
 class plArmatureEffect : public hsKeyedObject
--- a/Sources/Tools/MaxComponent/plResponderLink.cpp
+++ b/Sources/Tools/MaxComponent/plResponderLink.cpp
@@ -1560,7 +1560,7 @@ public:
         {
         case WM_INITDIALOG:
             for (i = 0; i < plArmatureEffectsMgr::kMaxSurface; i++)
-                ComboBox_AddString(hCB, plArmatureEffectsMgr::SurfaceStrings[i]);
+                ComboBox_AddString(hCB, ST2T(plArmatureEffectsMgr::SurfaceStrings[i]));
 
             ComboBox_SetCurSel(hCB, pb->GetInt(ParamID(kSurface)));
 
@@ -1604,13 +1604,10 @@ const TCHAR* plResponderCmdFootSurface::GetInstanceName(IParamBlock2 *pb)
 {
     static TCHAR name[256];
 
-#ifdef UNICODE
-    _snwprintf(name, std::size(name), L"Foot Surface (%S)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
-#else
-    _snprintf(name, std::size(name), "Foot Surface (%s)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
-#endif
+    _sntprintf(
+        name, std::size(name), _T("Foot Surface (%s)"),
+        ST2T(plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))])
+    );
     return name;
 }
 
--- a/Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.cpp
@@ -58,28 +58,30 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pfAudio/plRandomSoundMod.h"
 #include "pfMessage/plArmatureEffectMsg.h"
 
-const char *plArmatureEffectsMgr::SurfaceStrings[] = 
+using namespace ST::literals;
+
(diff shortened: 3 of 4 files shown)


dpogue (approved):

Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.h
```diff
@@ -109,7 +109,7 @@ class plArmatureEffectsMgr : public hsKeyedObject
         kMaxSurface,
         kFootNoSurface = kMaxSurface,
     };  
-    static const char *SurfaceStrings[];
+    static ST::string SurfaceStrings[];
```
dgelessus, line 112: Technically, `ST::string` hasn't been forward declared yet in this header.

Sources/Tools/MaxComponent/plResponderLink.cpp
```diff
@@ -1606,10 +1606,10 @@ const TCHAR* plResponderCmdFootSurface::GetInstanceName(IParamBlock2 *pb)
 
 #ifdef UNICODE
     _snwprintf(name, std::size(name), L"Foot Surface (%S)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
+        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))].c_str());
 #else
     _snprintf(name, std::size(name), "Foot Surface (%s)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
+        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))].c_str());
 #endif
```
dgelessus, line 1613: `.c_str()` is iffy here - technically this passes UTF-8 into a format expecting ANSI (in practice it doesn't matter, because the strings are pure ASCII). But this entire block can be simplified to:

```suggestion
    _sntprintf(name, std::size(name), _T("Foot Surface (%s)"),
        ST2T(plArmatureEffectsMgr::GetSurfaceName(pb->GetInt(ParamID(kSurface)))));
```

dgelessus (changes requested):

Sources/Tools/MaxComponent/plResponderLink.cpp
Hoikas: This suggestion won't work correctly - it passes a non-trival type (`ST::string`) into printf, and the wrong formatter is used when `UNICODE` is not defined (on Max 2012 and lower). Needs to be `%S` in that situation.

Sources/Plasma/PubUtilLib/plAvatar/plArmatureEffects.h
Hoikas: True, but I we know it's available due to this being an `hsKeyedObject` subclass.

Sources/Tools/MaxComponent/plResponderLink.cpp
dgelessus: `_sntprintf` maps to the narrow/wide variant depending on whether `UNICODE` is enabled, and `ST2T` does the conversion from `ST::string` to `TCHAR*`, no? (This follows the same pattern as most of the other `GetInstanceName` implementations.)

Hoikas: It maps to either `_snprintf` (which requires char* to be `%s`) when `UNICODE` is not defined or to `_snwprintf` (which requires char* to be `%S`) when `UNICODE` is defined. Further, non-trival types, such as `ST::string` cannot be passed as arguments to `printf` in general. We had lots of fun fixing those kinds of crashing bugs when we first changed many  C strings to `plString`.

dgelessus: I don't understand... Unless I'm missing something, it should match up in both cases:

* If `UNICODE` is defined, then `ST2T(s)` becomes `s.to_wchar().data()`, which is a `wchar_t*`, which is appropriate for `%s` in `_snwprintf`
* If `UNICODE` is *not* defined, then `ST2T(s)` becomes `s.to_latin_1().data()`, which is a `char*`, which is appropriate for `%s` in `_snprintf`

It shouldn't make any difference that it was an `ST::string` originally, because it's explicitly converted to a plain C string before it reaches the printf. The C string *is* based on a temporary buffer, but that shouldn't be a problem - IIRC, C++ guarantees that temporary intermediate values stay alive until the end of the statement (i. e. during the entire `_sntprintf` call).

Hoikas: I think the problem here is that I was forgetting ehat exactly `ST2T` returns. Your suggestion is fine.

Sources/Tools/MaxComponent/plResponderLink.cpp
```diff
@@ -1604,13 +1604,10 @@ const TCHAR* plResponderCmdFootSurface::GetInstanceName(IParamBlock2 *pb)
...
 
-#ifdef UNICODE
-    _snwprintf(name, std::size(name), L"Foot Surface (%S)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
-#else
-    _snprintf(name, std::size(name), "Foot Surface (%s)",
-        plArmatureEffectsMgr::SurfaceStrings[pb->GetInt(ParamID(kSurface))]);
-#endif
+    _sntprintf(
+        name, std::size(name), L"Foot Surface (%s)",
```
dgelessus, line 1608:
```suggestion
        name, std::size(name), _T("Foot Surface (%s)"),
```

dgelessus (approved):

dpogue (approved):

PR #1437, opened 2023-07-22
Better finding of the macOS CPU Vendor name

dpogue:
macOS doesn't support CPUID, so previously all macOS CPUs were being reported as "Unknown".

On newer versions, we can get the CPU vendor string from `sysctl`, and on very old versions we can look it up from the mach host info.

---

Results of plSystemInfo on Ventura:
```
OS: macOS 13.4.1
CPU: Apple M1 Pro
RAM: 16384 MiB

User Data Path: /Users/dpogue/Library/Application Support/Uru Live
```

Results of plSystemInfo on Catalina:
```
OS: Mac OS X 10.15.7
CPU: Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
RAM: 8192 MiB

User Data Path: /Users/dpogue/Library/Application Support/Uru Live
```

Results of plSystemInfo on Leopard:
```
OS: Mac OS X 10.5.8
CPU: PowerPC 7400
RAM: 1024 MiB

User Data Path: /Users/dpogue/Library/Application Support/Uru Live
```

### DIFF
--- a/Sources/Plasma/CoreLib/hsSystemInfo.cpp
+++ b/Sources/Plasma/CoreLib/hsSystemInfo.cpp
@@ -67,6 +67,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #endif
 
 #ifdef HS_BUILD_FOR_APPLE
+#    include <mach/mach.h>
 #    include <CoreFoundation/CoreFoundation.h>
 
     extern "C" {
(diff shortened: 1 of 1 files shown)


colincornaby:
How far back does sysctl work? Wondering if this should be a runtime check instead of a compile time check. But if the modern Mac client doesn't fall along a sysctl compatibility boundary I'd be fine with compile time.

dpogue:
`sysctl` itself goes way back (probably to 10.0, since it exists in FreeBSD), but the `machdep` values don't go as far back as 10.5 (or maybe they do but only on Intel)

Hoikas:
What benefit do we get by supporting both the new `sysctl` API and the older mach one instead of just the latter?

dpogue:
On newer machines, the old mach call just returns "ARM64E" for the CPU name instead of "Apple M1 Pro" 😒 

Also, using the mach stuff is... not deprecated, but let's say discouraged... sysctl is the supported way to get this information

colincornaby:
Yeah, anything Mach tends to be in a discouraged state these days.

I'm fine with the ARM64E return right now. It would be nice to get the CPU name... but if we log GPU details the GPU name would let us know the exact SoC.

Does the sysctl approach default to on? In since a hypothetical PPC client would likely be the only place the fallback is needed I'd want to make sure the right path for most builds is the default.

Sources/Plasma/CoreLib/hsSystemInfo.cpp
```diff
@@ -114,6 +115,39 @@ static inline void ICPUID(cpuid_t* info, int function_id)
...
+#ifdef HS_BUILD_FOR_APPLE
+static inline ST::string IGetAppleCPUVendor()
+{
+    ST::string result;
+
+#ifdef HAVE_SYSCTL
+    size_t bufsize = 100;
+    char buffer[bufsize];
+
+    if (sysctlbyname("machdep.cpu.brand_string", &buffer, &bufsize, nullptr, 0) == 0) {
```
Hoikas, line 127: Do we know that `sysctlbyname` is always available when `sysctl` is? `HAVE_SYSCTL` is only testing the latter and not the former, IIRC.

dpogue: `sysctlbyname` exists in macOS 10.0, so I think that's probably safe to depend on

dpogue:
> Does the sysctl approach default to on? In since a hypothetical PPC client would likely be the only place the fallback is needed I'd want to make sure the right path for most builds is the default.

`HAVE_SYSCTL` should default to on for macOS, since that's what we're using to get the RAM values on all macOS versions

Hoikas:
@colincornaby does this look good to you?

Hoikas:
Pinging @colincornaby for reaction

colincornaby:
This looks good to me.

PR #1438, opened 2023-07-22
Use standard versions of `snprintf`/`swprintf` on Windows

dgelessus:
I thought there was some reason why the non-standard versions were still used, but apparently not. After re-reading the Microsoft docs ([`snprintf`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170), [`swprintf`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/sprintf-sprintf-l-swprintf-swprintf-l-swprintf-l?view=msvc-170)), I'm pretty sure we can safely switch to the standard functions on Windows too.

`snprintf` is supported since VS 2015 and the standard-compliant version of `swprintf` since at least VS 2010, which is definitely enough for our purposes.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -319,8 +319,6 @@ int hsMessageBoxWithOwner(hsWindowHndl owner, const wchar_t* message, const wcha
 
 #if HS_BUILD_FOR_WIN32
      // This is for Windows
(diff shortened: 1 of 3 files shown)


zrax (approved):
I believe the only reason those were still being used is because much of the code predates those MSVC versions.

PR #1439, opened 2023-07-22
Clean up and consistently implement the `hsStream` API

dgelessus:
Some `hsStream` methods weren't implemented in all subclasses even though they should have been. This fixes those cases - either implementing the missing methods, or removing them from the base `hsStream` if they actually only make sense for some subclasses.

A few things in the stream classes were implemented twice under different names for no apparent reason, so I removed those duplicates.

Notably, this merges `hsRAMStream` and `hsVectorStream`. Both provided a memory-based read/write stream, but with two different implementations for the dynamic memory buffer. I kept just the `std::vector` version from `hsVectorStream`, because I don't see any advantage to the custom `hsAppender` that `hsRAMStream` used before.


Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -660,88 +624,96 @@ uint32_t  plReadOnlySubStream::GetEOF()
...
+
+void hsRAMStream::FastFwd()
+{
+    fPosition = fVector.size();
 }
 
 void hsRAMStream::Truncate()
 {
-    Reset();
+    fVector.resize(fPosition);
```
zrax, line 679: These two things are not equal...  The old `hsRAMStream::Truncate()` would always reset the stream to 0 size.  Did you check everywhere Truncate was called to see which behavior was expected?

dgelessus: Oh, good point, I didn't even notice that. Both `hsUNIXStream` and the former `hsVectorStream` implement `Truncate` as "delete all data after current seek position", so I would say the old `hsRAMStream::Truncate` was simply incorrect.

I'll have to check if anything relied on the old `hsRAMStream` behavior...

dgelessus: Okay, thankfully `hsStream::Truncate` is basically never used. It's only called three times in all of Plasma, and two of those calls are inside other `Truncate` implementations. The one "real" call was on a `hsRAMStream`, *of course*, so I changed it to `Reset`.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -660,88 +624,96 @@ uint32_t  plReadOnlySubStream::GetEOF()
...
-    return (fBytesRead >= fAppender.Count() * fAppender.ElemSize());
+    return (fPosition >= fVector.size());
 }
 
 uint32_t hsRAMStream::Read(uint32_t byteCount, void * buffer)
 {
-    if (fBytesRead + byteCount > fAppender.Count() * fAppender.ElemSize())
-        byteCount = (fAppender.Count() * fAppender.ElemSize()) - fBytesRead;
+    if (fPosition + byteCount > fVector.size()) {
+//      hsStatusMessageF("Reading past end of hsRAMStream (read %u of %u requested bytes)", fVector.size() - fPosition, byteCount);
```
Hoikas, line 635:
```suggestion
```

Hoikas (approved):

PR #1441, opened 2023-07-25
Remove dead code in SDL parser.

zrax:
Note: The `SetInternal()` and `SetAlwaysNew()` methods are still used, but are flagged from `DISPLAYOPTION=hidden` and `DEFAULTOPTION=vault` respectively.


Sources/Plasma/PubUtilLib/plSDL/plSDLParser.cpp
```diff
@@ -278,22 +278,6 @@ bool plSDLParser::IParseVarDesc(const plFileName& fileName, hsStream* stream, ch
             }
         }
```
dgelessus, line 280:
```suggestion
```

dgelessus (approved):

Hoikas (approved):

PR #1442, opened 2023-07-26
Document SDL read/write options in more detail

dgelessus:
The code that handles the SDL blob read/write options is quite tangled, so it's difficult to see which options do what exactly and where they even have any effect at all.

This PR adds detailed comments for all the options' meanings (for both reading and writing) and explicitly lists all the options understood by each method. Also some minor cleanup to the option passing, but not much - this is 90% a documentation change.


Deledrius:
More code documentation is very appreciated!

Hoikas (approved):
Good find in c6c90a1. :+1:

dgelessus:
That wasn't a logic change btw. `kSkipNotificationInfo` should behave the same way before and after my change. It just makes it clearer (IMO) what the flag does.

PR #1443, opened 2023-07-27
Remove dead hsThread_Mac.cpp

dgelessus:
This code uses the Multiprocessing Services API from Carbon (i. e. inherited from Classic Mac OS), which is deprecated since OS X 10.8 and has never been available for 64-bit programs. Its official replacement is the standard POSIX threads (pthreads) API.

This file was never used since the open-source release. The CMake build already correctly uses the pthread-based hsThread_Unix.cpp on macOS.


dpogue (approved):
Even I'm not unhinged enough to try porting Uru to Mac OS 9 😝

zrax (approved):

PR #1446, opened 2023-08-02
Add popup quit when sitting at heek table

Hazado:
Stops the mouse movement keys when sitting at the heek table to prevent misclicking and leaving the game unintentionally.

Keyboard movement still works and allows you to leave the game still.
Closing the KI will also re-enable mouse movement keys.
Mouse movement key is re-enabled after standing back up.

### DIFF
--- a/Scripts/Python/nb01RPSGame.py
+++ b/Scripts/Python/nb01RPSGame.py
@@ -383,7 +383,7 @@ class nb01RPSGame(ptResponder, object):
 
     def OnControlKeyEvent(self,controlKey,activeFlag):
         if controlKey in [PlasmaControlKeys.kKeyMoveBackward, PlasmaControlKeys.kKeyRotateLeft, PlasmaControlKeys.kKeyRotateRight, PlasmaControlKeys.kKeyExitMode] and activeFlag:
-            PtYesNoDialog(self.key, "Do you want to leave the Ahyoheek game?")
+            PtYesNoDialog(self.key, PtGetLocalizedString("Heek.Messages.Quit"))
 
     def OnNotify(self, state, id, events):
         """Handle Plasma Notification Messages"""
@@ -951,7 +951,6 @@ class nb01RPSGame(ptResponder, object):
             args[event[1]] = event[3]
 
         if "YesNo" in args:
-            PtDebugPrint("YesNo is",args["YesNo"])
             if args["YesNo"]:
                 PtDisableControlKeyEvents(self.key)
                 PtEnableMovementKeys()
@@ -959,7 +958,7 @@ class nb01RPSGame(ptResponder, object):
                 return True
             else:
                 return False
-        elif "type" in args:
+        else:
             # Now, let's fire it off!
             type = args["type"]
             del args["type"]
--- a/Scripts/Python/nb01RPSGame.py
+++ b/Scripts/Python/nb01RPSGame.py
@@ -384,7 +384,7 @@ class nb01RPSGame(ptResponder, object):
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
     def OnControlKeyEvent(self, controlKey, activeFlag):
-        """Captures movement keys and prompt a yes/no dialog during a game or standup animation otherwise"""
+        #Captures movement keys and prompt a yes/no dialog during a game or standup animation otherwise
         if controlKey in [PlasmaControlKeys.kKeyMoveBackward, PlasmaControlKeys.kKeyRotateLeft, PlasmaControlKeys.kKeyRotateRight, PlasmaControlKeys.kKeyExitMode] and activeFlag:
             if self._round_played:
                 PtLocalizedYesNoDialog(self.key, "Heek.Messages.Quit")
@@ -456,10 +456,10 @@ class nb01RPSGame(ptResponder, object):
         # Manage game state if standing up
         if state:
             if PtWasLocallyNotified(self.key):
-                """Disable Yeesha Book to prevent linking out
-                   Enable control keys to capture button presses or mouse movements
-                   Disable all movement keys so sit modifier wont trigger before we want it to
-                   Enable the mouse movement so we can move the camera and use the sides of the screen to trigger a movement"""
+                #Disable Yeesha Book to prevent linking out
+                #Enable control keys to capture button presses or mouse movements
+                #Disable all movement keys so sit modifier wont trigger before we want it to
+                #Enable the mouse movement so we can move the camera and use the sides of the screen to trigger a movement
                 PtSendKIMessage(kDisableEntireYeeshaBook, 0)
                 PtEnableControlKeyEvents(self.key)
                 PtDisableMovementKeys()
@@ -733,10 +733,10 @@ class nb01RPSGame(ptResponder, object):
         PtSendKIMessage(kKILocalChatStatusMsg, msg)
 
     def _QuitGame(self, YesNo):
-        """Performs the standup animation when sitting down
-           Disable control key events
-           Enables the disabled movement keys
-           PtAvatarExitAFK() is actually just performing a GoToStage 2 which for the sit brain is the standup animation - https://github.com/H-uru/Plasma/blob/9956967363f383d91f43162d116ad3477148adb2/Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp#L1970"""
+        #Performs the standup animation when sitting down
+        #Disable control key events
+        #Enables the disabled movement keys
+        #PtAvatarExitAFK() is just doing IExitTopmostGenericMode()
         if YesNo:
             PtDisableControlKeyEvents(self.key)
             PtEnableMovementKeys()
--- a/Scripts/Python/nb01RPSGame.py
+++ b/Scripts/Python/nb01RPSGame.py
@@ -45,6 +45,7 @@ from Plasma import *
 from PlasmaConstants import *
 from PlasmaKITypes import *
 from PlasmaTypes import *
+import PlasmaControlKeys
 
 # detectors
 detButtonRock = ptAttribActivatorList(1, "Rock ButtonClick det", netForce=True)
@@ -380,6 +381,13 @@ class nb01RPSGame(ptResponder, object):
         else:
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
+    def OnControlKeyEvent(self,controlKey,activeFlag):
+        PtDebugPrint(controlKey, activeFlag)
+        if controlKey == PlasmaControlKeys.kKeyExitMode:
+            PtYesNoDialog(self.key, "Do you want to leave the Ahyoheek game?")
+        elif controlKey == PlasmaControlKeys.kKeyMoveBackward or controlKey == PlasmaControlKeys.kKeyRotateLeft or controlKey == PlasmaControlKeys.kKeyRotateRight:
+            PtYesNoDialog(self.key, "Do you want to leave the Ahyoheek game?")
+
     def OnNotify(self, state, id, events):
         """Handle Plasma Notification Messages"""
 
@@ -389,10 +397,10 @@ class nb01RPSGame(ptResponder, object):
 
         # Finished sitting down.
         if self._HandleNotify(state, id, events, self._sitting, self._OnSitDown):
-            if self.playing:
-                PtEnableMouseMovement()
-            else:
-                PtDisableMouseMovement()
+            if not self.playing:
+                PtEnableControlKeyEvents(self.key)
+                #PtDisableMouseMovement()
+                PtDisableMovementKeys()
             return
 
         # A rock/paper/sics button was mashed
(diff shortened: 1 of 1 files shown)


dpogue:
From Scharminius in Discord:
> What if mouse movement [keys] were disabled for all forms of sitting? I find it likewise annoying when clicking through chatlines or the playerlist, and missing the button by a hair and thus doing a stand-up routine in front of everyone.

Deledrius:
Maybe this makes sense as a universal preference "Disable Click-to-Stand mouse controls when sitting."?

DamnBriggsy:
> Maybe this makes sense as a universal preference "Disable Click-to-Stand mouse controls when sitting."?

This would be ideal, in my opinion.

Hoikas:
@DamnBriggsy raises a good point. I've always thought it would be better, honestly, for *any* attempt to leave the heek game, whether via movement keys or mouse clicks, pop up a yes/no confirmation dialog asking something to the effect of "Are you sure you want to leave and forfeit this game?" This will better defeat accidental leaves, which would still be possible with the options menu checkbox and will prevent accidental keypress quits as well.

Hazado:
> @DamnBriggsy raises a good point. I've always thought it would be better, honestly, for _any_ attempt to leave the heek game, whether via movement keys or mouse clicks, pop up a yes/no confirmation dialog asking something to the effect of "Are you sure you want to leave and forfeit this game?" This will better defeat accidental leaves, which would still be possible with the options menu checkbox and will prevent accidental keypress quits as well.

I also wanted this, but I dont see any movement keys being sent to the script
Right now the heek script only gets messages for the sit/stand animation completion

Hoikas:
We can start off with the incremental improvement for mouse clicking, the, and figure out keypresses in time.

Deledrius:
> @DamnBriggsy raises a good point. I've always thought it would be better, honestly, for _any_ attempt to leave the heek game, whether via movement keys or mouse clicks, pop up a yes/no confirmation dialog asking something to the effect of "Are you sure you want to leave and forfeit this game?" This will better defeat accidental leaves, which would still be possible with the options menu checkbox and will prevent accidental keypress quits as well.

I was thinking about this, and the main issue here is that it would break one of the deliberate core UX elements of Uru when it comes to keeping nearly everything IC.  Of course, there's always a balance of pushing that too far one way or the other.  What would the "Uru way" be in this case?  All options seem strange and tedious, and hardly more realistic.  It might be an instance where a simple confirmation dialog, despite breaking "the rules", would be relatively invisible compared to the more IC alternatives.

DamnBriggsy:
> I was thinking about this, and the main issue here is that it would break one of the deliberate core UX elements of Uru when it comes to keeping nearly everything IC. Of course, there's always a balance of pushing that too far one way or the other. What would the "Uru way" be in this case? All options seem strange and tedious, and hardly more realistic. It might be an instance where a simple confirmation dialog, despite breaking "the rules", would be relatively invisible compared to the more IC alternatives.

One spot I can think of where this confirmation dialog happens is for deleting a book from the Relto bookshelf. While I agree that the gamey-ness should be reduced as often as possible, there's some precedent for this already. I think accidental quits in Heek are pretty common, so this would be a good place to break from realism and enter video game territory.

Hazado:
Im going to set this to draft for the moment.
I want to do the popup when you try to stand during an active game.
It currently requires converting all the seats to multibeh instead of the standard sitting modifier.

Hazado:
Okay, so multibeh plan was causing me headaches.

You get a popup now when you attempt to leave, asking if you want to leave the game.
placeholder text there for the moment.
You can use the mouse to leave as well as the keyboard keys.
Test it a let me know if this is close to what everyone is looking for.

Scripts/Python/nb01RPSGame.py
```diff
@@ -941,17 +950,27 @@ def _HandleVariableNotify(self, events):
...
+        if "YesNo" in args:
+            PtDebugPrint("YesNo is",args["YesNo"])
+            if args["YesNo"]:
+                PtDisableControlKeyEvents(self.key)
+                PtEnableMovementKeys()
+                PtAvatarExitAFK()
+                return True
+            else:
+                return False
+        elif "type" in args:
```
Hoikas, line 962:
```suggestion
        else:
```

Previously, the code would raise a `KeyError` if `type` is missing. We would still like for that to happen so that we can sniff out incorrect notifications.

Scripts/Python/nb01RPSGame.py
```diff
@@ -380,6 +381,10 @@ def OnSDLNotify(self, VARname, SDLname, playerID, tag):
         else:
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
+    def OnControlKeyEvent(self,controlKey,activeFlag):
+        if controlKey in [PlasmaControlKeys.kKeyMoveBackward, PlasmaControlKeys.kKeyRotateLeft, PlasmaControlKeys.kKeyRotateRight, PlasmaControlKeys.kKeyExitMode] and activeFlag:
+            PtYesNoDialog(self.key, "Do you want to leave the Ahyoheek game?")
```
Hoikas, line 386: This will work for now, but we probably want to move toward a localized entry before merging this.

Hazado:
Quit message with localized strings now.
https://github.com/H-uru/moul-assets/pull/240

Scripts/Python/nb01RPSGame.py
```diff
@@ -389,6 +394,10 @@ def OnNotify(self, state, id, events):
 
         # Finished sitting down.
         if self._HandleNotify(state, id, events, self._sitting, self._OnSitDown):
+            if not self.playing:
+                PtEnableControlKeyEvents(self.key)
+                PtDisableMovementKeys()
+                PtEnableMouseMovement()
```
Hoikas, line 400: This should probably be done in `nb01RPSGame._OnSitDown()`. I wanted the `OnNotify()` method to be as slim as possible with effectively no gameplay or interface logic.

Scripts/Python/nb01RPSGame.py
```diff
@@ -941,17 +950,26 @@ def _HandleVariableNotify(self, events):
...
-            PtDebugPrint("nb01RPSGame._HandleVariableNotify():\tPyEvent '{}' has bad kwargs".format(type))
-        return True
+        if "YesNo" in args:
+            if args["YesNo"]:
+                PtDisableControlKeyEvents(self.key)
+                PtEnableMovementKeys()
+                PtAvatarExitAFK()
+                return True
+            else:
+                return False
```
Hoikas, line 960: Let's pull this out into an event handler method. This function is intended for teasing apart what a variable notification is, not actually doing the details of processing that notification.

Scripts/Python/nb01RPSGame.py
```diff
@@ -380,6 +381,10 @@ def OnSDLNotify(self, VARname, SDLname, playerID, tag):
         else:
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
+    def OnControlKeyEvent(self,controlKey,activeFlag):
```
Hoikas, line 384:
```suggestion
    def OnControlKeyEvent(self, controlKey, activeFlag):
```

Hazado:
@Hoikas Hoping these changes do what you requested

Set it up so the quit message popup only happens once your playing an active game using the ``self._round_played`` variable at line 388
Easy enough to remove that and have the popup quit message happen anytime you sit down instead.

Scripts/Python/nb01RPSGame.py
```diff
@@ -442,6 +456,9 @@ def _OnSitDown(self, state, seat, events):
         if state:
             if PtWasLocallyNotified(self.key):
                 PtSendKIMessage(kDisableEntireYeeshaBook, 0)
+                PtEnableControlKeyEvents(self.key)
+                PtDisableMovementKeys()
+                PtEnableMouseMovement()
```
Hoikas, line 461: A comment explaining what these are for would be good.

Scripts/Python/nb01RPSGame.py
```diff
@@ -710,6 +727,11 @@ def _SendLocalStatusMsg(self, msg):
         """Sends a status chat message (purple text) to the local player's KI."""
         PtSendKIMessage(kKILocalChatStatusMsg, msg)
 
+    def _QuitGame(self, YesNo):
+        if YesNo:
+            PtDisableControlKeyEvents(self.key)
+            PtEnableMovementKeys()
+            PtAvatarExitAFK()
```
Hoikas, line 734: A comment explaining what this does would be nice (we're not AFK, after all).

Scripts/Python/nb01RPSGame.py
```diff
@@ -710,6 +727,11 @@ def _SendLocalStatusMsg(self, msg):
         """Sends a status chat message (purple text) to the local player's KI."""
         PtSendKIMessage(kKILocalChatStatusMsg, msg)
 
+    def _QuitGame(self, YesNo):
+        if YesNo:
+            PtDisableControlKeyEvents(self.key)
+            PtEnableMovementKeys()
```
Hoikas, line 733: A comment explaining what's going on here would be good.

Scripts/Python/nb01RPSGame.py
```diff
@@ -380,6 +383,17 @@ def OnSDLNotify(self, VARname, SDLname, playerID, tag):
         else:
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
+    def OnControlKeyEvent(self, controlKey, activeFlag):
+        if controlKey in [PlasmaControlKeys.kKeyMoveBackward, PlasmaControlKeys.kKeyRotateLeft, PlasmaControlKeys.kKeyRotateRight, PlasmaControlKeys.kKeyExitMode] and activeFlag:
+            if self._round_played:
+                PtYesNoDialog(self.key, PtGetLocalizedString("Heek.Messages.Quit"))
```
Hoikas, line 389:
```suggestion
                PtLocalizedYesNoDialog(self.key, "Heek.Messages.Quit")
```

Hoikas (commented):
This is looking really good. I think the last thing that I would like to see is a set of comments explaining what exactly is going on. Even though you explained to me how this works, the disabling/enabling of certain controls is a bit mind-bending.

Scripts/Python/nb01RPSGame.py
Hazado: PTAvatarExitAFK() actually just does a gotostage 2 which works for the sit brain in this case

https://github.com/H-uru/Plasma/blob/9956967363f383d91f43162d116ad3477148adb2/Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp#L1970

Hoikas: Put that in a comment in the code 😉

Hazado:
Added comments and suggested change.
Let me know if this needs further explanation

Scripts/Python/nb01RPSGame.py
```diff
@@ -441,7 +456,14 @@ def _OnSitDown(self, state, seat, events):
         # Manage game state if standing up
         if state:
             if PtWasLocallyNotified(self.key):
+                """Disable Yeesha Book to prevent linking out
+                   Enable control keys to capture button presses or mouse movements
+                   Disable all movement keys so sit modifier wont trigger before we want it to
+                   Enable the mouse movement so we can move the camera and use the sides of the screen to trigger a movement"""
```
Hoikas, line 462: These should be comments (lines beginning with `#`) instead of multiline strings, which are Python objects that the compyler has to deal with.

Scripts/Python/nb01RPSGame.py
```diff
@@ -710,6 +732,15 @@ def _SendLocalStatusMsg(self, msg):
         """Sends a status chat message (purple text) to the local player's KI."""
         PtSendKIMessage(kKILocalChatStatusMsg, msg)
 
+    def _QuitGame(self, YesNo):
+        """Performs the standup animation when sitting down
+           Disable control key events
+           Enables the disabled movement keys
+           PtAvatarExitAFK() is actually just performing a GoToStage 2 which for the sit brain is the standup animation - https://github.com/H-uru/Plasma/blob/9956967363f383d91f43162d116ad3477148adb2/Sources/Plasma/FeatureLib/pfPython/cyAvatar.cpp#L1970"""
```
Hoikas, line 739: This should probably just list the C++ function that performs the GoToStage. We already have the source code in front of us, so a URL is a bit overkill.

Scripts/Python/nb01RPSGame.py
```diff
@@ -380,6 +383,18 @@ def OnSDLNotify(self, VARname, SDLname, playerID, tag):
         else:
             raise RuntimeError("Got an SDL notify for {}, but no CB".format(VARname))
 
+    def OnControlKeyEvent(self, controlKey, activeFlag):
+        #Captures movement keys and prompt a yes/no dialog during a game or standup animation otherwise
```
Hoikas, line 387:
```suggestion
        # Captures movement keys and prompt a yes/no dialog during a game or standup animation otherwise
```

Scripts/Python/nb01RPSGame.py
```diff
@@ -441,7 +456,14 @@ def _OnSitDown(self, state, seat, events):
         # Manage game state if standing up
         if state:
             if PtWasLocallyNotified(self.key):
+                #Disable Yeesha Book to prevent linking out
+                #Enable control keys to capture button presses or mouse movements
+                #Disable all movement keys so sit modifier wont trigger before we want it to
+                #Enable the mouse movement so we can move the camera and use the sides of the screen to trigger a movement
```
Hoikas, line 462:
```suggestion
                # Disable Yeesha Book to prevent linking out
                # Enable control keys to capture button presses or mouse movements
                # Disable all movement keys so sit modifier wont trigger before we want it to
                # Enable the mouse movement so we can move the camera and use the sides of the screen to trigger a movement
```

Scripts/Python/nb01RPSGame.py
```diff
@@ -710,6 +732,15 @@ def _SendLocalStatusMsg(self, msg):
         """Sends a status chat message (purple text) to the local player's KI."""
         PtSendKIMessage(kKILocalChatStatusMsg, msg)
 
+    def _QuitGame(self, YesNo):
+        #Performs the standup animation when sitting down
+        #Disable control key events
+        #Enables the disabled movement keys
+        #PtAvatarExitAFK() is just doing IExitTopmostGenericMode()
```
Hoikas, line 739:
```suggestion
        # Performs the standup animation when sitting down
        # Disable control key events
        # Enables the disabled movement keys
        # PtAvatarExitAFK() is just doing IExitTopmostGenericMode()
```

Hoikas (commented):
Minor nitpick. This should be all that's left.

Deledrius (approved):

Hoikas (approved):

PR #1448, opened 2023-08-05
Close `hsStream`s automatically in destructors

dgelessus:
A step towards RAII for `hsStream`. This makes streams much easier to manage correctly, especially when using `std::unique_ptr`, collections and such.

In fact, I removed `hsStream::Close` entirely, because almost nothing closed a stream *without* destroying it soon afterwards. Sometimes this was done to reuse a single stream object for multiple files, but it's not hard to recreate the stream for each file instead.

This isn't "proper full RAII", because *opening* a stream is still its own method call and not part of the constructor. I wasn't sure how to handle error reporting and virtual `Open` calls if the constructor does the opening. This isn't a big problem though I think - the automatic closing is the important part.

The diff is long, but it's mostly boring deleting of all the manual `Close` calls. The actually interesting changes are in `hsStream` and its implementations. The updated logic in plRegistryNode, plStreamSource, and pfPatcher could use some extra eyes perhaps.

### DIFF
--- a/Sources/Plasma/CoreLib/hsStream.cpp
+++ b/Sources/Plasma/CoreLib/hsStream.cpp
@@ -541,12 +541,9 @@ void hsUNIXStream::Flush()
 
 //////////////////////////////////////////////////////////////////////////////////////
 
-void    plReadOnlySubStream::Open( hsStream *base, uint32_t offset, uint32_t length )
+plReadOnlySubStream::plReadOnlySubStream(hsStream* base, uint32_t offset, uint32_t length)
+    : fBase(base), fOffset(offset), fLength(length)
 {
-    fBase = base;
-    fOffset = offset;
-    fLength = length;
-
     fBase->SetPosition( fOffset );
     IFixPosition();
 }
--- a/Sources/Plasma/CoreLib/hsStream.cpp
+++ b/Sources/Plasma/CoreLib/hsStream.cpp
@@ -427,7 +427,9 @@ void hsStream::WriteLEFloat(size_t count, const float values[])
 
 hsUNIXStream::~hsUNIXStream()
 {
-    // Don't Close here, because Sub classes Don't always want that behaviour!
+    if (fRef) {
+        fclose(fRef);
+    }
 }
 
 bool hsUNIXStream::Open(const plFileName &name, const char *mode)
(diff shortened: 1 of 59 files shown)


Sources/Plasma/PubUtilLib/plFile/plSecureStream.cpp
```diff
@@ -95,6 +95,18 @@ fDeleteOnExit(false)
...
 {
+    if (fOpenMode == kOpenWrite) {
+        fRAMStream->Rewind();
+        IWriteEncrypted(fRAMStream.get(), fWriteFileName);
+    }
+
+    if (fRef != INVALID_HANDLE_VALUE) {
+#if HS_BUILD_FOR_WIN32
+        CloseHandle(fRef);
+#elif HS_BUILD_FOR_UNIX
```
Hoikas, line 106:
```suggestion
#if defined(HS_BUILD_FOR_WIN32)
        CloseHandle(fRef);
#elif defined(HS_BUILD_FOR_UNIX)
```

Yeah, I see that it was wrong before 😞

Hoikas (commented):
This looks like it will be a good improvement.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -819,6 +810,39 @@ uint32_t hsWriteOnlyStream::Write(uint32_t byteCount, const void* buffer)
...
+
+void hsWriteOnlyStream::FastFwd()
+{
+    fPosition = GetEOF();
+    fData = fStop;
+}
+
+void hsWriteOnlyStream::Truncate()
+{
+    hsThrow("can't write to a readonly stream");
```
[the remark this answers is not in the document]
dgelessus: Oops. This was copied from `hsReadOnlyStream`, which was the superclass before. `hsWriteOnlyStream` really cannot be truncated though, because it's based on a fixed-size memory buffer. Will update the message.

Sources/Plasma/PubUtilLib/plFile/plSecureStream.cpp
dgelessus: Will do. The codebase is quite inconsistent here though. `#if HS_BUILD_FOR_WIN32`, `#ifdef HS_BUILD_FOR_WIN32`, and `#if defined(HS_BUILD_FOR_WIN32)` are all relatively common.

Hoikas: Yeah... Using `defined` here is more correct, though, because `hsConfig.h` is just doing `#define HS_BUILD_FOR_WIN32`, not `#define HS_BUILD_FOR_WIN32 1`, so we really need to check if the macro is defined, not its value. Unfortunately, `#elifdef` is only available in C++23 and later, but we're still on C++17.

Hoikas (commented):
This LGTM, but I want to toss it onto TL to test the patching implications.

Hoikas (approved):

PR #1449, opened 2023-08-06
Avoid runtime indirection when initializing `plCreatable` class indices

dgelessus:
The initialization went through a couple of getter/setter methods (one of them virtual), but the field can easily be made a constant initialized at compile time with no dynamic lookup.

I'm actually fixing a couple of IDE warnings here: one about virtual calls in constructors (`ClassIndex()`) and one about "unnecessary" semicolons after the plCreator.h `REGISTER_*` macros.


Hoikas (commented):
This LGTM, but I'd like for @dpogue to glance it over. He's particularly interested in external creatables, so I think he has a better picture of how all this fits together than I do.

dpogue (approved):
This looks to me like it should be fine

PR #1452, opened 2023-08-07
Fix ODR violation for `plFactory` and avoid including it everywhere

dgelessus:
Touching the core plCreatable headers, part two.

The private members of `plFactory` were only conditionally defined in plFactory.h, which is an ODR violation and gives warnings with GCC for example (although apparently it has never caused any issues in practice). I'm not sure why it was done that way (perhaps to reduce header pollution very slightly?), but it seems safe to define everything unconditionally.

Regardless, to avoid pulling in `plFactory` in many places where it's not needed, I also adjusted plCreatable.h to not automatically include plFactory.h anymore. This required removing a couple of `plCreatable` methods, but they weren't used anywhere and are trivial to replace.

### DIFF
--- a/Sources/Plasma/Apps/plPageInfo/plPageInfo.cpp
+++ b/Sources/Plasma/Apps/plPageInfo/plPageInfo.cpp
@@ -46,6 +46,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include <string_theory/format>
 
+#include "pnFactory/plFactory.h"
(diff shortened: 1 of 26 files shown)


dpogue (approved):
This seems good to me.

`plFactory::Create(plWhatever::Index())` works just as well as `plWhatever::Create()` for the esoteric cases that no normal person ever runs into.

Hoikas (approved):

PR #1453, opened 2023-08-08
Don't trust the local machine's system time.

Hoikas:
Previously, when asking for the server's time, the calculation was done by basically offsetting the local system time by some previously calculated delta. This was basically useless because the user can easily change the system time out from under us and cause huge deltas. The attempt to fix things in #173 by using `WM_TIMECHANGE` was a good idea but ultimately incorrect because `hsTimer` is a monotonic timer. At this point, it uses `std::chrono::steady_clock` as its backend, so attempting to use `hsTimer` to recalculate the server/client time offset is useless.

Therefore, this changes us to simply save the last time value that we received from the game server and the monotonic time that we received that update. This is definitely not perfect because we don't consider things like round trip time for the message and time spent in the update loop. However, it *is* a clear improvement in that now the KI will always display what the server thinks Mountain time is, and it should be impossible to perform system clock based exploits (such as instantly baking pellets)... without any platform specific code!

Future work: currently, we are implicitly depending on the server to send us periodic `plNetMessage`s with the time sent value to synchronize the time. If that's never done, then we simply return the client's time (yikes!). This is a decent assumption in all current content because we will receive `plNetMsgGameMessage` with `plServerReply` for region enters on spawn and `plNetSDLMsg` for the Age state (when an Age has any non-excluded SDL states), however, there is little guarantee that the server will continue to send us these messages if we are in an Age by ourselves doing nothing. One option would be to send an innocuous message such as `plNetMsgGetSharedState` periodically. I have not implemented that here because it is not immediately apparent whether or not the various server types implement this message.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetClient/plNetCliAgeJoiner.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetClient/plNetCliAgeJoiner.cpp
@@ -212,7 +212,7 @@ void plNCAgeJoiner::Start () {
 
     plNetClientMgr * nc = plNetClientMgr::GetInstance();
     nc->SetFlagsBit(plNetClientMgr::kPlayingGame, false);
-    nc->fServerTimeOffset               = 0;    // reset since we're connecting to a new server
+    nc->ResetServerTimeOffset();
     nc->fRequiredNumInitialSDLStates    = 0;
     nc->fNumInitialSDLStates            = 0;
     nc->SetFlagsBit(plNetClientApp::kNeedInitialAgeStateCount);
--- a/Sources/Plasma/PubUtilLib/plNetClient/plNetClientMgr.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetClient/plNetClientMgr.cpp
@@ -102,10 +102,10 @@ plNetClientMgr::PendingLoad::~PendingLoad()
 //
 plNetClientMgr::plNetClientMgr()
     : fLocalPlayerKey(), fMsgHandler(this), fJoinOrder(), fTaskProgBar(),
-      fMsgRecorder(), fServerTimeOffset(), fTimeSamples(), fLastTimeUpdate(),
-      fListenListMode(kListenList_Distance), fAgeSDLObjectKey(), fExperimentalLevel(),
-      fOverrideAgeTimeOfDayPercent(-1.f), fNumInitialSDLStates(), fRequiredNumInitialSDLStates(),
-      fDisableMsg(), fIsOwner(true), fIniPlayerID(), fPingServerType()
+      fMsgRecorder(), fLastLocalTime(), fListenListMode(kListenList_Distance),
+      fAgeSDLObjectKey(), fExperimentalLevel(), fOverrideAgeTimeOfDayPercent(-1.f),
+      fNumInitialSDLStates(), fRequiredNumInitialSDLStates(), fDisableMsg(), fIsOwner(true),
+      fIniPlayerID(), fPingServerType()
 {   
 #ifndef HS_DEBUGGING
     // release code will timeout inactive players on servers by default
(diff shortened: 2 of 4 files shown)


Sources/Plasma/PubUtilLib/plNetClient/plNetCliAgeJoiner.cpp
```diff
@@ -212,7 +212,7 @@ void plNCAgeJoiner::Start () {
 
     plNetClientMgr * nc = plNetClientMgr::GetInstance();
     nc->SetFlagsBit(plNetClientMgr::kPlayingGame, false);
-    nc->fServerTimeOffset               = 0;    // reset since we're connecting to a new server
+    nc->ResetServerTimeOffset();                // reset since we're connecting to a new server
```
dgelessus, line 215: Since this doesn't align with anything...

```suggestion
    nc->ResetServerTimeOffset(); // reset since we're connecting to a new server
```

Sources/Plasma/PubUtilLib/plNetClient/plNetClientMgr.cpp
```diff
@@ -403,50 +403,36 @@ void plNetClientMgr::IUnloadNPCs()
...
+        return;
 
-            DebugMsg("Setting server time offset to {f}", fServerTimeOffset);
-        }
-    }
+    fLastServerTime = msg->GetTimeSent();
+    fLastLocalTime = localTime;
 }
 
 void plNetClientMgr::ResetServerTimeOffset(bool delayed)
```
dgelessus, line 419: The `delayed` parameter should be deleted since the new code doesn't use it anymore.

dgelessus (commented):
This looks like a good solution to me.

dgelessus:
> If that's never done, then we simply return the client's time (yikes!).

This can never happen I think. When joining an age instance, the client always sends a `plNetMsgGameStateRequest` and will stay on the loading screen until it has received the corresponding `plNetMsgInitialAgeStateSent` reply from the server (and the corresponding initial SDL states, if any). Assuming that that message always has the `TimeSent` field filled out, this guarantees that the server time will be known by the time the loading screen ends.

> One option would be to send an innocuous message such as `plNetMsgGetSharedState` periodically. I have not implemented that here because it is not immediately apparent whether or not the various server types implement this message.

According to [my notes](https://nagus.readthedocs.io/en/latest/uru/game_server.html#plnetmessages), none of the fan servers support `plNetMsgGetSharedState`. I haven't checked if Cyan's server responds to it, but I feel like it's best not to rely on it if the client doesn't use it otherwise.

Perhaps `plNetMsgMembersListReq` would be safe to send as a "ping"? Though right now that message is only sent exactly once when joining the age instance, so we should check that all servers are fine with that message being sent multiple times...

dpogue (approved):
This looks good to me

Sources/Plasma/PubUtilLib/plNetClient/plNetClientMgr.cpp
```diff
@@ -403,50 +403,36 @@ void plNetClientMgr::IUnloadNPCs()
...
 
 //
 // return the gameservers time
 //
 plUnifiedTime plNetClientMgr::GetServerTime() const 
-{ 
-    if ( fServerTimeOffset==0 )     // offline mode or before connecting/calibrating to a server
+{
+    if (fLastServerTime.AtEpoch()) {
+        WarningMsg("WARNING: Someone asked for the server time, but we don't know it yet!");
```
dgelessus, line 431: `plLoggable::WarningMsg` already adds a `WRN: ` prefix

```suggestion
        WarningMsg("Someone asked for the server time, but we don't know it yet!");
```

dgelessus (approved):
One really minor thing, otherwise LGTM.

PR #1458, opened 2023-08-10
Adds SDL based upload check for imagers

Hazado:
Assets - https://github.com/H-uru/moul-assets/pull/241
Sets up the an SDL variable so you can set individual permissions for upload/pellet drops to everyone/member only/off

Requested by Cyan Shard operator to prevent pellet uploads in individual hoods

Adds new SDL variable to Neighborhood.sdl
nb01PelletUploadState

0 <= Everyone can upload (Default)
1 = Members only upload
2 >= No one can upload

### DIFF
--- a/Scripts/Python/xSimpleImager.py
+++ b/Scripts/Python/xSimpleImager.py
@@ -77,7 +77,7 @@ ImagerPelletUpload = ptAttribBoolean(10, "Pellet Score Imager?", 0)
 ImagerClueObject = ptAttribSceneobject(11, "Imager Object (for puzzle clue)")
 ImagerClueTime = ptAttribInt(12, "Number of seconds until clue image shows",default=870)
 ImagerRandomTime = ptAttribInt(13, "Random number added to make timer more variable",default=0)
-ImagerPermissionCheck = ptAttribString(14,"Imager upload SDL variable (optional)")
+ImagerPermissionCheck = ptAttribString(14, "Imager upload SDL variable (optional)")
 #----------
 # globals
 #----------
@@ -531,7 +531,7 @@ class xSimpleImager(ptModifier):
                 ageVault.setDeviceInbox(ImagerName.value, ageSDL[ImagerInboxVariable.value][0], self, kSettingDeviceInbox)
 
     def PermissionCheck(self):
-        #Age SDL Check to prevent unwanted access to imager
+        # Age SDL Check to prevent unwanted access to imager
         ageSDL = PtGetAgeSDL()
         permissionCheck = kPermissionEveryone
         if ImagerPermissionCheck.value:
--- a/Scripts/Python/xSimpleImager.py
+++ b/Scripts/Python/xSimpleImager.py
@@ -77,6 +77,7 @@ ImagerPelletUpload = ptAttribBoolean(10, "Pellet Score Imager?", 0)
 ImagerClueObject = ptAttribSceneobject(11, "Imager Object (for puzzle clue)")
 ImagerClueTime = ptAttribInt(12, "Number of seconds until clue image shows",default=870)
 ImagerRandomTime = ptAttribInt(13, "Random number added to make timer more variable",default=0)
+ImagerPelletCheckVariable = ptAttribString(14,"Pellet upload SDL variable (optional)")
 #----------
 # globals
 #----------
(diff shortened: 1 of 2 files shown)


dgelessus:
Relatively minor thing: the code looks like it would work for all imagers that pop up in the KI, not just the pellet imager. Perhaps the things in the script should be called `ImagerPermissionVariable`, `CheckPermission`, etc. instead, in case we later want to add permission checks to other imagers as well?

Scripts/Python/xSimpleImager.py
```diff
@@ -522,6 +523,19 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
+        ageSDL = PtGetAgeSDL()
+        permissionCheck = 0
+        if ImagerPermissionCheck.value:
+            permissionCheck = ageSDL[ImagerPermissionCheck.value][0]
+        if permissionCheck >= 2: # Locked for everyone
+            return False
+        elif permissionCheck == 1 and ptVault().amOwnerOfCurrentAge(): #Hood members only
+            return True
+        elif permissionCheck <= 0: # Anyone
+            return True
```
Hoikas, line 537: This feels like a good opportunity for named constants instead of magic numbers, then you can delete the inline comments, which violate PEP-8.

Scripts/SDL/Neighborhood.sdl
```diff
@@ -1966,3 +1966,133 @@ STATEDESC Neighborhood
...
+    VAR BOOL            nb01DarkShapeSwimsRun[1]                DEFAULT=0   DEFAULTOPTION=VAULT
+
+# State variables
+    VAR BYTE            nb01Ayhoheek5Man1State[1]               DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01CommunityAreaState[1]               DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01CityLightsArchState[1]              DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01LakeLightState[1]                   DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01PrivateRoomsState[1]                DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01PuzzleWallState[1]                  DEFAULT=0   DEFAULTOPTION=VAULT
+    VAR BYTE            nb01PelletUploadState[1]                DEFAULT=0
```
Hoikas, line 2041:
```suggestion
    VAR BYTE            nb01PelletUploadState[1]                DEFAULT=0   DEFAULTOPTION=VAULT
```

Hazado: I didnt make this a vault option because I intended it to be age specific and not a global

Hoikas: `DEFAULTOPTION=VAULT` doesn't mean the setting is global. It means that the setting in the vault SDL node (whether that be the global SDL node or the age instance SDL node) will override whatever is set in the AgeSDLHook game server state. Shard operators set these kinds of settings in vault SDL nodes, so this is the appropriate flag to use.

Hoikas (approved):

Hoikas:
As of #1450, we could theoretically merge this immediately without breaking MOULa. However, I think there's probably a discussion to be had about what our repository's SDL should look like compared to the rest of the ecosystem, so we should probably wait until either after that discussion or the SDL is deployed to MOULa.

Scripts/Python/xSimpleImager.py
```diff
@@ -77,6 +77,7 @@
 ImagerClueObject = ptAttribSceneobject(11, "Imager Object (for puzzle clue)")
 ImagerClueTime = ptAttribInt(12, "Number of seconds until clue image shows",default=870)
 ImagerRandomTime = ptAttribInt(13, "Random number added to make timer more variable",default=0)
+ImagerPermissionCheck = ptAttribString(14,"Imager upload SDL variable (optional)")
```
Hoikas, line 80:
```suggestion
ImagerPermissionCheck = ptAttribString(14, "Imager upload SDL variable (optional)")
```

Scripts/Python/xSimpleImager.py
```diff
@@ -522,6 +530,19 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
             else:
                 ageVault.setDeviceInbox(ImagerName.value, ageSDL[ImagerInboxVariable.value][0], self, kSettingDeviceInbox)
 
+    def PermissionCheck(self):
+        #Age SDL Check to prevent unwanted access to imager
```
Hoikas, line 534:
```suggestion
        # Age SDL Check to prevent unwanted access to imager
```

Hoikas (approved):

PR #1459, opened 2023-08-10
Refactor and add tests for console command parsing

dgelessus:
Converts the very "old C-like" console command parsing code into its own class that has a bit more structure and uses modern C++ data types. Also, unit tests!

Aside from cleaning up the code, my motivation here is to allow parsing console commands *without* actually executing them. This would be especially useful for parsing .ini files - thinking of server.ini or cases like in #1244.

Suggestions for improving the parsing API are welcome. It has to be a bit complex unfortunately, to support situations where the input should only be partially parsed (e. g. tab completion), and because of the wacky command name syntax where parsing depends on which commands are defined...

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleParser.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleParser.cpp
@@ -142,3 +142,16 @@ pfConsoleCmd* pfConsoleParser::ParseCommand()
     }
     return group->FindCommandNoCase(*token);
 }
+
+std::optional<std::vector<ST::string>> pfConsoleParser::ParseArguments()
+{
+    std::vector<ST::string> args;
+    while (auto token = fTokenizer.NextArgument()) {
+        args.emplace_back(std::move(*token));
+    }
+    if (!fTokenizer.fErrorMsg.empty()) {
+        // Parse error in argument
+        return {};
+    }
+    return args;
+}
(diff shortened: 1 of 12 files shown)


dpogue:
There's one (fairly trivial) conflict here, but this seems good to me from a quick skim over things

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleParser.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    if (begin == fPos) {
+        fErrorMsg.clear();
+        return {};
+    }
+
+    return begin;
+}
+
+std::pair<pfConsoleCmdGroup*, std::optional<ST::string>> pfConsoleParser::ParseGroupAndName()
```
Hoikas, line 120: Slight preference for this to be an `std::tuple` because of `std::pair` implying the usage of a map. But that's just a preference because a lot of implementations just have `pair` use `tuple`.

Hoikas (commented):
Looking forward to seeing this rebased so it can be merged

Sources/Plasma/FeatureLib/pfConsoleCore/pfConsoleParser.cpp
dgelessus: Ah hm, is there a convention that `std::pair` is specifically for map-related things? I assumed it was just a specialized alternative to a 2-element `std::tuple`, but I don't know the stdlib too well...

dgelessus:
(I'll get the rebase done this weekend probably)

Hoikas (approved):

PR #1464, opened 2023-08-14
Fix starting video failing to play

Hazado:
The starting video will fail to play with the default SDL due to "IndexError: tuple index out of range"
Adding this check catches that before we get an error.

### DIFF
--- a/Scripts/Python/xOpeningSequence.py
+++ b/Scripts/Python/xOpeningSequence.py
@@ -446,7 +446,7 @@ class xOpeningSequence(ptModifier):
             ageSDL = PtGetAgeSDL()
             if StartInCleft():
                 gMovieFilePath = kAtrusIntroMovie
-            elif ageSDL["psnlIntroMovie"][0]:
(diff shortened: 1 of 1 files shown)


Scripts/Python/xOpeningSequence.py
```diff
@@ -446,7 +446,7 @@ def CheckMovie(self):
             ageSDL = PtGetAgeSDL()
             if StartInCleft():
                 gMovieFilePath = kAtrusIntroMovie
-            elif ageSDL["psnlIntroMovie"][0]:
+            elif ageSDL["psnlIntroMovie"] and ageSDL["psnlIntroMovie"][0]:
```
Hoikas, line 449:
```suggestion
            elif any(ageSDL["psnlIntroMovie"]):
```

```
PS C:\Users\Adam> python
>>> any(tuple())
False
>>> any(tuple(""))
False
>>> any(tuple("meow.webm"))
True
>>>
```

Hoikas (approved):

PR #1465, opened 2023-08-15
Some code deduplication for plPythonSDLModifier.

Hoikas:


Sources/Plasma/FeatureLib/pfPython/plPythonSDLModifier.cpp
```diff
@@ -322,37 +350,17 @@ bool plPythonSDLModifier::IPythonVarIdxToSDL(plSimpleStateVariable* var, int var
     case plVarDescriptor::kByte:
     case plVarDescriptor::kBool:
     case plVarDescriptor::kInt:
-        if (PyLong_Check(pyVar)) {
-            int v = PyLong_AsLong(pyVar);
-            var->Set(v, varIdx);
-            if (!hintstring.empty())
-                var->GetNotificationInfo().SetHintString(hintstring);
-            return true;
-        } else if (PyLong_Check(pyVar)) {
```
Hoikas, line 331: :trollface:

Sources/Plasma/FeatureLib/pfPython/plPythonSDLModifier.cpp
```diff
@@ -314,6 +316,32 @@ void plPythonSDLModifier::IDirtySynchState(const ST::string& name, bool sendImme
...
+            return static_cast<T>(PyFloat_AS_DOUBLE(o));
+        if (PyNumber_Check(o)) {
+            pyObjectRef pyFloat = PyNumber_Float(o);
+            // pyFloat might have come from some strange land where they return
+            // unexpected things, so don't use the unsafe macro here.
+            return static_cast<T>(PyFloat_AsDouble(pyFloat.Get()));
+        }
+        return std::nullopt;
+    } else {
+        static_assert(false, "Unhandled C++ type conversion for Python number");
```
zrax, line 341: This is not a dependent expression, and therefore is ill-formed even if the function is never called with the `else` condition ("no diagnostic required" means that MSVC is allowed to ignore it even though it's invalid, but GCC and Clang do emit a diagnostic)...  Some options:
1) Leave the base template unimplemented and implement the two (well, three) cases as specializations instead
2) Redefine the static_assert to use a dependent condition (something like `static_assert(sizeof(T) != sizeof(T), ...)`). Although this is technically still ill-formed, GCC and clang might be happier.
3) Move the static_assert outside of the `if constexpr` block, testing all possible types, and then leave something like a `std::unreachable()` in the else condition (less DRY).

Sources/Plasma/FeatureLib/pfPython/plPythonSDLModifier.cpp
```diff
@@ -314,6 +316,32 @@ void plPythonSDLModifier::IDirtySynchState(const ST::string& name, bool sendImme
...
+{
+    if constexpr (std::is_same_v<T, int>) {
+        if (PyLong_Check(o))
+            return PyLong_AsLong(o);
+        if (PyNumber_Check(o)) {
+            pyObjectRef pyLong = PyNumber_Long(o);
+            return PyLong_AsLong(pyLong.Get());
+        }
+        return std::nullopt;
+    } else if constexpr (std::disjunction_v<std::is_same<T, float>, std::is_same<T, double>>) {
```
zrax, line 330: I'm fairly certain you can just
```suggestion
    } else if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double>) {
```

Sources/Plasma/FeatureLib/pfPython/plPythonSDLModifier.cpp
zrax: Presumably one of those used to be a `PyInt_*`...  I hope >.>

Sources/Plasma/FeatureLib/pfPython/plPythonSDLModifier.cpp
Hoikas: Yeah, that worked with MSVC... I just wasn't sure why gcc and clang were exploding, but you've clarified that nicely :)

zrax (approved):

Hoikas:
The most recent push removes an unneeded `type_traits` include.

zrax (approved):

PR #1466, opened 2023-08-15
hsMessageBox Cleanups

dpogue:
* Moves hsMessageBox implementation out of CoreLib and into PubUtilLib
* Splits the Win32-specific implementation into its own file (matching how macOS is handled)
* Adds overloads for `hsMessageBox` that take ST::string
* Silences (on gcc & clang) deprecation warnings from OpenSSL about RC4

I'm unsure on whether it makes sense to change the hsMessageBox stuff to always convert to ST::string and then have a single implementation, or if it's more performant having it split up as it currently is...


Sources/Plasma/CoreLib/HeadSpin_Win.cpp
```diff
@@ -0,0 +1,219 @@
...
+    case IDIGNORE:      return hsMBoxIgnore;
+    case IDYES:         return hsMBoxYes;
+    case IDNO:          return hsMBoxNo;
+    default:            return hsMBoxCancel;
+    }
+
+    return hsMBoxCancel;
+}
+
+int hsMessageBoxWithOwner(hsWindowHndl owner, const char* message, const char* caption, int kind, int icon)
```
zrax, line 119: Why not just pass these variants onto the `ST::string` variant, and then eventually deprecate/remove the non-`ST::string` overloads?  Otherwise, this is a lot of duplication...

dpogue: That's what I was thinking, and performance of string conversions probably doesn't matter much in the case of message boxes

Sources/Plasma/CoreLib/HeadSpin_Mac.mm
```diff
@@ -101,39 +104,19 @@ int hsMessageBoxWithOwner(hsWindowHndl owner, NSString* message, NSString* capti
...
-                                     [[[NSString alloc] initWithBytesNoCopy:(void *)caption
-                                                                     length:wcslen(caption)*sizeof(*caption)
-                                                                   encoding:NSUTF32LittleEndianStringEncoding freeWhenDone:NO
-                                      ] autorelease],
-                                     [[[NSString alloc] initWithBytesNoCopy:(void *)message
-                                                                     length:wcslen(caption)*sizeof(*caption)
-                                                                   encoding:NSUTF32LittleEndianStringEncoding freeWhenDone:NO
-                                      ] autorelease],
+                                     [NSString initWithUTF8String:message.c_str()],
+                                     [NSString initWithUTF8String:caption.c_str()],
```
dgelessus, line 119:
```suggestion
                                     [NSString stringWithSTString:message],
                                     [NSString stringWithSTString:caption],
```

dgelessus (changes requested):

dpogue:
> I think it would make sense to pull `hsMessageBox` out into its own header. Possibly even into another library, somewhere under PubUtilLib.

I like this idea

Sources/Plasma/CoreLib/HeadSpin_Mac.mm
dpogue: We can't use that here (reliably) because NSString+STString category is in plClient, and this might (hypothetically) get called from somewhere that's not part of plClient.

Sure would be convenient if we could do something like `message.toNSString()` as part of the base library :trollface:

dgelessus (changes requested):
Otherwise this looks good I think.

dgelessus (approved):

Sources/Plasma/NucleusLib/pnUtils/pnUtCrypt.cpp
```diff
@@ -95,7 +95,11 @@ static void Rc4Codec (
 ) {
     // RC4 uses the same algorithm to both encrypt and decrypt
     uint8_t * temp = (uint8_t *)malloc(bytes);
+
+    IGNORE_WARNINGS_BEGIN("deprecated-declarations")
```
Hoikas, line 99: The changes in this file probably belong in another PR

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox.h
```diff
@@ -0,0 +1,79 @@
...
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#ifndef _hsMessageBox_h_
+#define _hsMessageBox_h_
+
+#include "HeadSpin.h"
+#include <string_theory/string>
```
Hoikas, line 47: A forward declaration would be sufficient and therefore avoid pulling in heavy headers, such as `type_traits` via `string_theory/string`.

Sources/Plasma/PubUtilLib/plResMgr/plResManager.cpp
```diff
@@ -1366,46 +1367,44 @@ bool plResManager::VerifyPages()
...
         ST::string pageFile = pages[i]->GetPagePath().GetFileName();
-        if (strlen(buf) + pageFile.size() > bufSize - 5)
-        {
-            strcat(buf, "...\n");
-            break;
-        }
-
-        strcat(buf, pageFile.c_str());
-        strcat(buf, "\n");
+        ss << pageFile << ST_LITERAL("\n");
```
Hoikas, line 1384:
```suggestion
        ss << pageFile << '\n';
```

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox.h
dpogue: Ahh, now that we got rid of the non-ST::String overloads, I can do that!

Sources/Plasma/NucleusLib/pnUtils/pnUtCrypt.cpp
dpogue: Split to #1476

Hoikas (approved):

PR #1468, opened 2023-08-16
Port plVaultClientApi to standard C++ containers

dgelessus:
Here I go, touching more important and complex code - what could go wrong?

More seriously, this makes the vault logic *a bit* more readable and less complex, because it eliminates some manual memory management and wrapper structs. Some of the code is still quite verbose though, because of how the C++ map API is designed.

This eliminates the last use of the weird custom intrusive hash table pnUtHash (aka `HASHTABLEDECL`/`HASHLINK`).


Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -367,57 +324,53 @@ static void BuildNodeTree (
...
+            parentIt = s_nodes.emplace(refs[i].parentId, parentNode).first;
+        } else {
+            parentNode = parentIt->second;
             existingNodeIds->emplace_back(refs[i].parentId);
         }
-        RelVaultNodeLink * childLink = s_nodes.Find(refs[i].childId);
-        if (!childLink) {
+
+        auto childIt = s_nodes.find(refs[i].childId);
+        hsRef<RelVaultNode> childNode;
```
Hoikas, line 340: I wonder if this should be `hsWeakRef` so we don't have to explicitly steal the new reference in 343, and we don't have to do an atomic operation in the else block.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -547,16 +503,19 @@ static void VaultNodeFetched (
     }
 
     // Add to global node table
-    RelVaultNodeLink * link = s_nodes.Find(node->GetNodeId());
-    if (!link) {
-        link = new RelVaultNodeLink(false, 0, node->GetNodeId());
-        link->node->SetNodeId_NoDirty(node->GetNodeId());
-        s_nodes.Add(link);
+    auto it = s_nodes.find(node->GetNodeId());
+    hsRef<RelVaultNode> globalNode;
```
Hoikas, line 507: Another `hsWeakRef`?

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -728,13 +696,13 @@ static void SaveDirtyNodes () {
     if (!s_nextSaveMs || signed(s_nextSaveMs - currTimeMs) <= 0) {
         s_nextSaveMs = (currTimeMs + kSaveUpdateIntervalMs) | 1;
         unsigned bytesWritten = 0;
-        for (RelVaultNodeLink * link = s_nodes.Head(); link; link = s_nodes.Next(link)) {
+        for (const auto& [nodeId, node] : s_nodes) {
```
Hoikas, line 699:
```suggestion
        for (auto [nodeId, node] : s_nodes) {
```

Correct me if I'm wrong, but I think that C++17 structured bindings for tuples and pairs will return references.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1124,12 +1100,13 @@ bool RelVaultNode::IsParentOf (unsigned childId, unsigned maxDepth) {
         return false;
     if (maxDepth == 0)
         return false;
-    if (state->children.Find(childId))
+    if (state->children.find(childId) != state->children.end())
         return true;
-    RelVaultNodeLink * link = state->children.Head();
-    for (; link; link = state->children.Next(link))
-        if (link->node->IsParentOf(childId, maxDepth - 1))
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1105:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1139,24 +1116,24 @@ bool RelVaultNode::IsChildOf (unsigned parentId, unsigned maxDepth) {
...
-    RelVaultNodeLink * link = state->parents.Head();
-    if (!link) {
+    if (state->parents.empty()) {
         nodeIds->emplace_back(GetNodeId());
-    }
-    else {
-        for (; link; link = state->parents.Next(link))
-            link->node->GetRootIds(nodeIds);
+    } else {
+        for (const auto& [nodeId, node] : state->parents) {
```
Hoikas, line 1134:
```suggestion
        for (auto [nodeId, node] : state->parents) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1173,10 +1150,9 @@ void RelVaultNode::GetChildNodeIds (
 ) {
     if (!maxDepth)
         return;
-    RelVaultNodeLink * link = state->children.Head();
-    for (; link; link = state->children.Next(link)) {
-        nodeIds->emplace_back(link->node->GetNodeId());
-        link->node->GetChildNodeIds(nodeIds, maxDepth-1);
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1153:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1187,10 +1163,9 @@ void RelVaultNode::GetParentNodeIds (
 ) {
     if (!maxDepth)
         return;
-    RelVaultNodeLink * link = state->parents.Head();
-    for (; link; link = state->parents.Next(link)) {
-        nodeIds->emplace_back(link->node->GetNodeId());
-        link->node->GetParentNodeIds(nodeIds, maxDepth-1);
+    for (const auto& [nodeId, node] : state->parents) {
```
Hoikas, line 1166:
```suggestion
    for (auto [nodeId, node] : state->parents) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1203,17 +1178,16 @@ hsRef<RelVaultNode> RelVaultNode::GetParentNode (
...
-    link = state->parents.Head();
-    for (; link; link = state->parents.Next(link)) {
-        if (link->node->Matches(templateNode.Get()))
-            return link->node;
-    }
-    
-    link = state->parents.Head();
-    for (; link; link = state->parents.Next(link)) {
-        if (hsRef<RelVaultNode> node = link->node->GetParentNode(templateNode, maxDepth - 1))
+    for (const auto& [nodeId, node] : state->parents) {
```
Hoikas, line 1181:
```suggestion
    for (auto [nodeId, node] : state->parents) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1203,17 +1178,16 @@ hsRef<RelVaultNode> RelVaultNode::GetParentNode (
...
+        if (node->Matches(templateNode.Get())) {
             return node;
+        }
+    }
+
+    for (const auto& [nodeId, node] : state->parents) {
```
Hoikas, line 1187:
```suggestion
    for (auto [nodeId, node] : state->parents) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1227,17 +1201,16 @@ hsRef<RelVaultNode> RelVaultNode::GetChildNode (
     if (maxDepth == 0)
         return nullptr;
 
-    RelVaultNodeLink * link;
-    link = state->children.Head();
-    for (; link; link = state->children.Next(link)) {
-        if (link->node->Matches(templateNode.Get()))
-            return link->node;
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1204:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1227,17 +1201,16 @@ hsRef<RelVaultNode> RelVaultNode::GetChildNode (
...
+        if (link.node->Matches(templateNode.Get())) {
+            return link.node;
+        }
     }
-    
-    link = state->children.Head();
-    for (; link; link = state->children.Next(link)) {
-        if (hsRef<RelVaultNode> node = link->node->GetChildNode(templateNode, maxDepth-1))
+
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1210:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1314,13 +1285,12 @@ void RelVaultNode::GetChildNodes (
     unsigned                maxDepth,
     RelVaultNode::RefList * nodes
 ) {
-    RelVaultNodeLink * link;
-    link = state->children.Head();
-    for (; link; link = state->children.Next(link)) {
-        if (link->node->Matches(templateNode.Get()))
-            nodes->push_back(link->node);
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1288:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1465,8 +1451,9 @@ void RelVaultNode::Print (const ST::string& tag, unsigned level) {
 //============================================================================
 void RelVaultNode::PrintTree (unsigned level) {
     Print("", level);
-    for (RelVaultNodeLink * link = state->children.Head(); link; link = state->children.Next(link))
-        link->node->PrintTree(level + 1);
+    for (const auto& [nodeId, link] : state->children) {
```
Hoikas, line 1454:
```suggestion
    for (auto [nodeId, link] : state->children) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1573,11 +1551,10 @@ hsRef<RelVaultNode> VaultGetNode (
     hsWeakRef<NetVaultNode> templateNode
 ) {
     ASSERT(templateNode);
-    RelVaultNodeLink * link = s_nodes.Head();
-    while (link) {
-        if (link->node->Matches(templateNode.Get()))
-            return link->node;
-        link = s_nodes.Next(link);
+    for (const auto& [nodeId, node] : s_nodes) {
```
Hoikas, line 1554:
```suggestion
    for (auto [nodeId, node] : s_nodes) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -1610,15 +1586,17 @@ void VaultAddChildNode (
...
-            childLink->node->SetNodeId_NoDirty(childId);
-            s_nodes.Add(childLink);
-        }
-        else if (ownerId) {
-            childLink->ownerId = ownerId;
+    auto parentIt = s_nodes.find(parentId);
+    if (parentIt != s_nodes.end()) {
+        const hsRef<RelVaultNode>& parentNode = parentIt->second;
+        auto childIt = s_nodes.find(childId);
+        hsRef<RelVaultNode> childNode;
```
Hoikas, line 1593: Another `hsWeakRef`?

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -2050,9 +2033,10 @@ void VaultLocalFindNodes (
     hsWeakRef<NetVaultNode> templateNode,
     std::vector<unsigned> * nodeIds
 ) {
-    for (RelVaultNodeLink * link = s_nodes.Head(); link != nullptr; link = s_nodes.Next(link)) {
-        if (link->node->Matches(templateNode.Get()))
-            nodeIds->emplace_back(link->node->GetNodeId());
+    for (const auto& [nodeId, node] : s_nodes) {
```
Hoikas, line 2036:
```suggestion
    for (auto [nodeId, node] : s_nodes) {
```

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
dgelessus: Hmm, will that work out without potentially leaking references? In the `if` branch, something needs to take ownership of the initial reference from the `new RelVaultNode()` call. If I put that directly into a `hsWeakRef`, it will just leak, and if I `UnRef` it first, it deallocates itself before it reaches a strong `hsRef`...

Best solution I can think of would be a temporary strong `hsRef` that holds the `new RelVaultNode()` long enough to be reffed elsewhere, *and* a `hsWeakRef` that holds either the `new RelVaultNode()` or an already existing one. That *should* avoid temporarily increasing the refcount.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
dgelessus: That's a good question... When you dereference an iterator, that definitely results in a reference, but I'm not sure that the reference-ness is "transferred" to a variable declared with `auto`. (The structured binding might be irrelevant here - I think it would be the same thing if you store the entire pair in an `auto` variable, but I'm not sure.)

To be honest, I'm a bit out of my depth with this. I'm mostly going by the [examples on cppreference.com](https://en.cppreference.com/w/cpp/container/unordered_map#Example), which consistently use `const auto& ...` and not just `auto ...` when doing a range `for` over a map.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
dgelessus: Oh wait, the `RelVaultNode` gets stored into a strong `hsRef` by the `emplace` call a few lines later, so I can let that steal it and store it as a plain pointer locally.

dgelessus:
I wonder... considering that all `RelVaultNode`s are stored in a big global map, perhaps it would be possible to have that be the owner of all `RelVaultNode`s and skip the dynamic refcounting. No idea how well that would interact with the rest of the codebase though. In any case, that's not something I'll try in this PR 😅

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
Hoikas: Maybe @zrax can clarify this for us... I noticed that cppreference was using `const auto&` for maps *after* I submitted the review, but I'm not sure why exactly.

Hoikas:
I think that would be ideal (in another PR), but we'd want to make sure that `pyVaultNode` would be able to cope with that. I think that, strictly speaking, it is possible for Python scripts to persist after an Age's vault nodes have been culled, so we would probably be forced into maintaining a reference count for vault nodes.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -504,14 +502,12 @@ static void VaultNodeFetched (
...
     // Add to global node table
     auto it = s_nodes.find(node->GetNodeId());
-    hsRef<RelVaultNode> globalNode;
     if (it == s_nodes.end()) {
-        globalNode.Steal(new RelVaultNode());
-        globalNode->SetNodeId_NoDirty(node->GetNodeId());
-        it = s_nodes.emplace(node->GetNodeId(), globalNode).first;
-    } else {
-        globalNode = it->second;
+        hsRef<RelVaultNode> newGlobalNode(new RelVaultNode());
```
Hoikas, line 506: This will cause the node to leak - a new instance of `NetVaultNode` has a ref count of 1, and the `hsRef` constructor increments the reference counter.

dgelessus: Forgot a `hsStealRef` here, oops

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
zrax: From what I understand, the structured binding just extracts the declarations as if you had declared them individually in separate declarations.  Therefore, `auto [k, v] = tuple` will copy (or move) the values from `tuple` into non-reference types for `k` and `v`.  I think where it gets fuzzy/interesting is that the `std::unordered_map` iterator returns ***references to `std::pair<K, V>` objects***, and not a pair of references (!!).  So although a classic iterator would still give you references to the data, extracting them with a structured binding will give you copies/moves unless you use a reference in your binding.

Playground: https://godbolt.org/z/4xsPYG6vs

TLDR:  `const auto&` appears to be correct here.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -367,57 +324,51 @@ static void BuildNodeTree (
...
+        if (parentIt == s_nodes.end()) {
             newNodeIds->emplace_back(refs[i].parentId);
-            parentLink = new RelVaultNodeLink(false, 0, refs[i].parentId);
-            parentLink->node->SetNodeId_NoDirty(refs[i].parentId);
-            s_nodes.Add(parentLink);
-        }
-        else {
+            hsRef<RelVaultNode> newParentNode(new RelVaultNode(), hsStealRef);
+            newParentNode->SetNodeId_NoDirty(refs[i].parentId);
+            parentIt = s_nodes.emplace(refs[i].parentId, std::move(newParentNode)).first;
```
Hoikas, line 332: Shouldn't `parentIt` be set to the iterator returned by `emplace()`? It looks like `parentIt` is currently an iterator for `s_nodes` :/

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -367,57 +324,51 @@ static void BuildNodeTree (
...
+        if (childIt == s_nodes.end()) {
             newNodeIds->emplace_back(refs[i].childId);
-            childLink = new RelVaultNodeLink(refs[i].seen, refs[i].ownerId, refs[i].childId);
-            childLink->node->SetNodeId_NoDirty(refs[i].childId);
-            s_nodes.Add(childLink);
-        }
-        else {
+            hsRef<RelVaultNode> newChildNode(new RelVaultNode(), hsStealRef);
+            newChildNode->SetNodeId_NoDirty(refs[i].childId);
+            childIt = s_nodes.emplace(refs[i].childId, std::move(newChildNode)).first;
```
Hoikas, line 343: Same question as above.

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
```diff
@@ -547,16 +501,17 @@ static void VaultNodeFetched (
...
-    RelVaultNodeLink * link = s_nodes.Find(node->GetNodeId());
-    if (!link) {
-        link = new RelVaultNodeLink(false, 0, node->GetNodeId());
-        link->node->SetNodeId_NoDirty(node->GetNodeId());
-        s_nodes.Add(link);
+    auto it = s_nodes.find(node->GetNodeId());
+    if (it == s_nodes.end()) {
+        hsRef<RelVaultNode> newGlobalNode(new RelVaultNode(), hsStealRef);
+        newGlobalNode->SetNodeId_NoDirty(node->GetNodeId());
+        it = s_nodes.emplace(node->GetNodeId(), std::move(newGlobalNode)).first;
```
Hoikas, line 508: Same question as above

Sources/Plasma/PubUtilLib/plVault/plVaultClientApi.cpp
dgelessus: That's what the code does. Not sure I understand the question :/

The `.first` at the end is because [`std::unordered_map::emplace`](https://en.cppreference.com/w/cpp/container/unordered_map/emplace) actually returns a `std::pair<iterator, bool>`.

Hoikas: Ah, right, I was thinking `s_nodes` was an `std::list` for some reason.

Hoikas (approved):

Issue #1208, opened 2022-07-04
Linking to Gahreesen from Relto pillar chooses spawn point in Maintainer's Nexus


ZarothYe:
This behavior is happening currently on TrollLand. @Hazado any ideas why this would happen? Using the book on the bookshelf afterwards picks the correct spawn point in the well.

https://user-images.githubusercontent.com/94947971/177062470-327ceefc-618c-4c77-8054-7e86e7e2053f.mp4

dpogue:
I ran into the same issue. It doesn't seem like the actual _spawn point_ is the problem (the avatar isn't actually loading in the Maintainer Nexus), but it's picking up the wrong camera

ZarothYe:
In my first case, I couldn't see the avvie, but the camera followed... *something* down down down to the depths. There was no panic link. I had to manually link out.

In my second case (in the video shown), I still couldn't see the avvie, but the panic link did trigger after a few seconds. So I'm not convinced that the avatar wasn't there, otherwise why would he panic link? Could just be that they fell through the elevator. It was a notoriously tetchy elevator in testing on Minkata for a long time.

Hoikas:
What seems to be happening is that the player is triggering regions along the path from the position of the pole in Relto to the default link in point in Gahreesen. This trips one of both of the Gahreesen nexus/nexii entry regions, which executes this python code:

https://github.com/H-uru/Plasma/blob/f29f7711d307068165bc6bacaaa254e7b9572758/Scripts/Python/grsnNexusBookMachine.py#L202-L210

Thereby pushing the Nexus cutscene camera and warping the player somewhere off into the abyss.

Hazado:
Ive done some testing and its definitely triggering the ``entryTrigger`` region in the ``OnNotify()`` before arriving at the correct link-in point.
This is probably going to need a client fix that prevents any of those from running before the avatar is actually at the starting link-in point

PR #1469, opened 2023-08-16
Hack to prevent issue #1208 in Gahreesen age

Hazado:
Issue  https://github.com/H-uru/Plasma/issues/1208
Prevents the OnNotify from running before OnServerInitComplete is done.

Arriving in Gahreesen from the Bahro Pole in Relto seems to run through the entryTrigger region and triggers the subworld entry.
This causes the avatar to fall out of the world because they are not anywhere with solid ground.

### DIFF
--- a/Scripts/Python/grsnNexusBookMachine.py
+++ b/Scripts/Python/grsnNexusBookMachine.py
@@ -81,7 +81,6 @@ kElevDown = 0
 kElevMoving = 1
 kElevUp = 2
 elevatorStatus = kElevDown
-serverInitDone = False
 
 class grsnNexusBookMachine(ptResponder):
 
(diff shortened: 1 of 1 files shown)


Scripts/Python/grsnNexusBookMachine.py
```diff
@@ -81,6 +81,7 @@
 kElevMoving = 1
 kElevUp = 2
 elevatorStatus = kElevDown
+serverInitDone = False
```
Hoikas, line 84: A class member would probably be better than another global.

Hoikas (commented):
This is a bit of a hack that leaves the underlying problem unsolved. I would be curious to see the result of [this experiment](https://discord.com/channels/282045216221822978/282046743338811392/1141488858056372254) before accepting this.

Hazado:
> This is a bit of a hack that leaves the underlying problem unsolved. I would be curious to see the result of [this experiment](https://discord.com/channels/282045216221822978/282046743338811392/1141488858056372254) before accepting this.

I changed the spawnpoint names for the nexus elevators and the issue still happened.
I dont think the linked experiment will change whats happening here.
The avatar does arrive at the right spawn point, but ends up in a subworld.

Hoikas:
A real fix hasn't exactly been forthcoming, so we might as well merge this hack for now.

PR #1470, opened 2023-08-16
make xHighLevelStarTrekDoor.py to be less bad

Hazado:
Not saying this is better but its less bad

Removes age owner from controlling the triggering of the doors
Removes the SendNote() stuff it was doing
Removes the doorstack tracking
Makes the door opening and closing rely on the SDL triggering
Makes the SDL trigger only happen by the person who set off the door sensor

I think ive fixed some of the race conditions but this is a multiplayer script and those are hard to troubleshoot

### DIFF
--- a/Scripts/Python/xHighLevelStarTrekDoor.py
+++ b/Scripts/Python/xHighLevelStarTrekDoor.py
@@ -80,7 +80,6 @@ class xHighLevelStarTrekDoor(ptModifier):
         PtDebugPrint("DEBUG: xHighLevelStarTrekDoor.__init__: v. %d" % self.version)
 
         self.DoorEnabled = 1
-        rgnSensor.volumeSensorNoArbitration(True)
 
     ##########################################
     def OnServerInitComplete(self):
@@ -122,6 +121,8 @@ class xHighLevelStarTrekDoor(ptModifier):
             # the door is really shut, someone left it open
             self.SDL['DoorState'] = (doorSDLstates.closed,)
             ageSDL[strDoorClosedVar.value] = (1,)
+
+        rgnSensor.volumeSensorNoArbitration(True)
 
     ##########################################
     def OnSDLNotify(self,VARname,SDLname,playerID,tag):
--- a/Scripts/Python/xHighLevelStarTrekDoor.py
+++ b/Scripts/Python/xHighLevelStarTrekDoor.py
@@ -68,12 +68,12 @@ class xHighLevelStarTrekDoor(ptModifier):
         self.version = 3
 
         PtDebugPrint("DEBUG: xHighLevelStarTrekDoor.__init__: v. %d" % self.version)
-        
+
         self.PrevDoorState = 0
         self.DoorState = 0
-        self.DoorStack = []
         self.DoorEnabled = 1
         self.respondertime = 30
+        self.lastTriggered = -1
 
     ##########################################
     def OnServerInitComplete(self):
@@ -94,46 +94,31 @@ class xHighLevelStarTrekDoor(ptModifier):
                 PtDebugPrint("ERROR: xHighLevelStarTrekDoor.OnServerInitComplete():\tERROR: age sdl read failed, defaulting door enabled value")
         else:
             self.DoorEnabled = 1
-        
+
         try:
             self.DoorState = self.SDL['DoorState'][0]
         except:
             self.SDL['DoorState'] = (0,)
             self.DoorState = self.SDL['DoorState'][0]
-         
+
         PtDebugPrint("xHighLevelStarTrekDoor: self.SDL = %d" % self.DoorState)
 
         if not PtIsSolo():
-            
+
             PtDebugPrint("xHighLevelStarTrekDoor: Somebody is already in the age. Attempting to sync states.")
 
             if self.DoorState == doorSDLstates['opening'] or self.DoorState == doorSDLstates['movingopen'] or self.DoorState == doorSDLstates['opentoclose']:
                 respOpenDoor.run(self.key,netPropagate=0)
-                self.DoorStack.append("respOpenDoor.run(self.key,netPropagate=0)")
-
-                '''
-                respOpenDoor.run(self.key,fastforward=1)
-                self.DoorState = doorSDLstates['open']
-                PtDebugPrint("xHighLevelStarTrekDoor: Door is open.")
-                '''
-            
             elif self.DoorState == doorSDLstates['closing'] or self.DoorState == doorSDLstates['movingclosed'] or self.DoorState == doorSDLstates['closetoopen']:
                 respCloseDoor.run(self.key,netPropagate=0)
-                self.DoorStack.append("respCloseDoor.run(self.key,netPropagate=0)")
-                '''
-                respCloseDoor.run(self.key,fastforward=1)
-                self.DoorState = doorSDLstates['closed']
-                PtDebugPrint("xHighLevelStarTrekDoor: Door is closed.")
-                '''
-            
             elif self.DoorState == doorSDLstates['open']:
                 respOpenDoor.run(self.key,fastforward=1)
-                
 
         else:
             # the door is really shut, someone left it open
             self.SDL['DoorState'] = (doorSDLstates['closed'],)
             self.DoorState = self.SDL['DoorState'][0]
+            ageSDL[strDoorClosedVar.value] = (1,)
 
     ##########################################
     def OnSDLNotify(self,VARname,SDLname,playerID,tag):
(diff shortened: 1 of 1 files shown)


Hazado:
I did have one bug happen during my local testing (two clients on the same computer connecting to my local dockersand)
If the avatar that triggered the door sensor quit without leaving the sensor region, the sensor wouldn't trigger anymore
Think its a multi-client on the same computer bug and not one with the script

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -144,219 +129,85 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
-            elif self.DoorState == doorSDLstates['movingopen'] or self.DoorState == doorSDLstates['opening']:
-                self.UpdateDoorState(doorSDLstates['open'])
-
-            elif self.DoorState == doorSDLstates['closetoopen']:
-                self.UpdateDoorState(doorSDLstates['opening'])
-
-            elif self.DoorState == doorSDLstates['movingclosed'] or self.DoorState == doorSDLstates['closing']:
-                self.UpdateDoorState(doorSDLstates['closed'])
+                return
+            if int(playerID) == int(localPlayerID):
```
Hoikas, line 208:
```suggestion
            if playerID == localPlayerID:
```

These should already be integers.

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -144,219 +129,85 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
-                else:
-                    respCloseDoor.run(self.key,netPropagate=0)
-            else:
-                PtDebugPrint("xHighLevelStarTrekDoor.ExecCode(): ERROR! Invalid tag '%s'." % (tag))
-                self.DoorStack.pop(0)
-        except:
-            PtDebugPrint("xStandardDoor.ExecCode(): ERROR! Invalid code '%s'." % (code))
-            self.DoorStack.pop(0)
+                elif self.DoorState == doorSDLstates['closing']:
+                    ageSDL[strDoorClosedVar.value] = (1,)
```
Hoikas, line 213: You need to remove the `sendToClients` bit on lines 82-83 or there will be a storm of SDL updates from every client in the Age.

Hoikas (commented):
This is a quick review, but I did notice that there is a duplicated variable -- `self.SDL["DoorState"]` and `self.DoorState`. It would be nice if only one of these were used.

Hazado:
> This is a quick review, but I did notice that there is a duplicated variable -- `self.SDL["DoorState"]` and `self.DoorState`. It would be nice if only one of these were used.

Yeah, one is a script SDL and the other a class variable. I should probably just use the script SDL

Hazado:
Added the requested changes

Hazado:
Tested this change on minkata alpha server.
Doors always worked with one bug
If the person who opened the doors links out or quits while in the sensor region, the doors will stay open until someone else enters the sensor region.

Hoikas:
> Tested this change on minkata alpha server.

Is "minkata alpha server" now on H'uru? 😮

Hazado:
> > Tested this change on minkata alpha server.
> 
> 
> 
> Is "minkata alpha server" now on H'uru? 😮 

Im ready to test this on trolland :)

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -94,46 +90,24 @@ def OnServerInitComplete(self):
...
-        
-        try:
-            self.DoorState = self.SDL['DoorState'][0]
-        except:
-            self.SDL['DoorState'] = (0,)
-            self.DoorState = self.SDL['DoorState'][0]
-         
-        PtDebugPrint("xHighLevelStarTrekDoor: self.SDL = %d" % self.DoorState)
+
+        PtDebugPrint("xHighLevelStarTrekDoor: self.SDL = %d" % self.SDL['DoorState'][0])
```
Hoikas, line 94:
```suggestion
        PtDebugPrint(f"xHighLevelStarTrekDoor: {self.SDL['DoorState'][0]=}")
```

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -144,219 +118,76 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
-                if doorClosed == 0:
-                    self.SendNote("respOpenDoor;1")
-                    self.UpdateDoorState(doorSDLstates['open'])
-                elif doorClosed == 1:
-                    self.SendNote("respCloseDoor;1")
-                    self.UpdateDoorState(doorSDLstates['closed'])
+            if doorClosed:
+                respCloseDoor.run(self.key,netPropagate=0)
+            else:
+                respOpenDoor.run(self.key,netPropagate=0)
```
Hoikas, line 124: Since we're now explicitly doing everything locally, we should probably mark all of the `ptAttribResponder`s as `netPropagate=False` explicitly at the top of the file, then we can remove all of these manual `netPropagate=0`s.

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -144,219 +118,76 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
-            elif self.DoorState == doorSDLstates['open']:
-                PtDebugPrint("xHighLevelStarTrekDoor: Updating Age SDL to Open")
+                return
+            if self.SDL['DoorState'][0] == doorSDLstates['opening']:
                 ageSDL[strDoorClosedVar.value] = (0,)
-
-            elif self.DoorState == doorSDLstates['closed']:
-                PtDebugPrint("xHighLevelStarTrekDoor: Updating Age SDL to Closed")
+            elif self.SDL['DoorState'][0] == doorSDLstates['closing']:
                 ageSDL[strDoorClosedVar.value] = (1,)
```
Hoikas, line 193: This could be improved greatly by making the door state into a [property](https://www.freecodecamp.org/news/python-property-decorator/).

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -68,19 +68,15 @@ def __init__(self):
         self.version = 3
 
         PtDebugPrint("DEBUG: xHighLevelStarTrekDoor.__init__: v. %d" % self.version)
-        
+
         self.PrevDoorState = 0
-        self.DoorState = 0
-        self.DoorStack = []
         self.DoorEnabled = 1
         self.respondertime = 30
```
Hoikas, line 74:
```suggestion
```

This is unused

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -144,219 +118,76 @@ def OnSDLNotify(self,VARname,SDLname,playerID,tag):
...
                 return
             #Region Triggered
             for event in events:
                 #true when you enter the region
-                
-                if event[0] == 1 and event[1] == 1 and PtFindAvatar(events) == PtGetLocalAvatar():
-                    playerID = PtGetLocalPlayer().getPlayerID()
-                    triggerstr = "rgnTriggerEnter%d" % playerID
-                    self.SendNote(triggerstr)
+                if event[0] == 1 and event[1] == 1:
```
Hoikas, line 137: I wonder if there is a less tedious way to do this?

Hoikas (commented):
Because we've now changed the SDL tracking and door animations to be done locally, we no longer need to acquire the exclusive server lock when someone enters the door region sensor. Each player is responsible for doing their own book-keeping, so there is no message storm or race condition concerns. I haven't looked at the PRPs yet, so I don't know if these are `plVolumeSensorConditionalObject` (requests exclusive server lock) or `plVolumeSensorConditionalObjectNoArbitration` (no server lock) in your fixed PRPs. We can fix the issue by, in the `OnServerInitComplete()` or `OnInit()` routines doing
```python
rgnSensor.volumeSensorNoArbitration(True)
```
to force the issue. This should improve the reliability when many players are crossing the region boundary.

Further, since we are  making the script "less bad", I think we should convert the `doorSDLstates` dict to something more Pythonic, like an [`IntEnum`](https://docs.python.org/3.8/library/enum.html#enum.IntEnum).

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -94,46 +99,29 @@ def OnServerInitComplete(self):
...
-                '''
-                respCloseDoor.run(self.key,fastforward=1)
-                self.DoorState = doorSDLstates['closed']
-                PtDebugPrint("xHighLevelStarTrekDoor: Door is closed.")
-                '''
-            
-            elif self.DoorState == doorSDLstates['open']:
-                respOpenDoor.run(self.key,fastforward=1)
-                
+            if self.SDL['DoorState'][0] in (doorSDLstates['opening'], doorSDLstates['movingopen'], doorSDLstates['opentoclose']):
```
Hoikas, line 114:
```suggestion
            if self.SDL['DoorState'][0] in (doorSDLstates.opening, doorSDLstates.movingopen, doorSDLstates.opentoclose):
```

`doorSDLstates` should now be using members thusly. You'll need to apply these changes throughout the file.

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -68,19 +78,14 @@ def __init__(self):
...
 
         PtDebugPrint("DEBUG: xHighLevelStarTrekDoor.__init__: v. %d" % self.version)
-        
-        self.PrevDoorState = 0
-        self.DoorState = 0
-        self.DoorStack = []
+
         self.DoorEnabled = 1
-        self.respondertime = 30
+        rgnSensor.volumeSensorNoArbitration(True)
```
Hoikas, line 83: `rgnSensor` does not actually have the key of the `plVolumeSensorConditionalObject` yet, so this has no effect. You'd want to put this in a `def OnInit(self):` method.

Scripts/Python/xHighLevelStarTrekDoor.py
```diff
@@ -49,16 +49,26 @@
...
+#doorSDLstates = {'closed':0,'opening':1,'open':2,'closing':3,'opentoclose':4,'closetoopen':5,'movingopen':6,'movingclosed':7}
+class doorSDLstates(IntEnum):
+    closed = 0
+    opening = 1
+    open = 2
+    closing = 3
+    opentoclose = 4
+    closetoopen = 5
+    movingopen = 6
+    movingclosed = 7
```
Hoikas, line 71: I think I want us to work on these state names so that what is going on is a little more clear. I know that I still need to open this file in VS Code (instead of viewing the diff, which is basically useless) to really understand what all is going on here.

Hoikas (approved):
There's still room for improvement here, but I think we should go ahead and bring this in.

PR #1472, opened 2023-08-18
Include Winsock header only where it's actually used

dgelessus:
pnNetCommon is also the only library that actually links with Winsock, so including the header elsewhere doesn't make much sense.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnNetCommon/plNetAddress.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCommon/plNetAddress.cpp
@@ -91,7 +91,7 @@ void plNetAddress::Read(hsStream * s)
     fHost = s->ReadLE32();
     fPort = s->ReadLE16();
 
-    // Family is always AF_INET
+    // Family is always kInet
     (void) s->ReadLE16();
 }
 
@@ -100,6 +100,5 @@ void plNetAddress::Write(hsStream * s)
     s->WriteLE32(fHost);
     s->WriteLE16(fPort);
 
-    // Family is always AF_INET
-    s->WriteLE16(static_cast<uint16_t>(2));
+    s->WriteLE16(static_cast<uint16_t>(Family::kInet));
 }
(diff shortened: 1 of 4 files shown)


Sources/Plasma/NucleusLib/pnNetCommon/plNetAddress.cpp
```diff
@@ -105,5 +101,5 @@ void plNetAddress::Write(hsStream * s)
     s->WriteLE16(fPort);
 
     // Family is always AF_INET
-    s->WriteLE16((uint16_t)AF_INET);
+    s->WriteLE16(static_cast<uint16_t>(2));
```
zrax, line 104: I would rather see this left as the constant `AF_INET`.  We already need the winsock/BSD socket headers for pnNetCommon, so why not include them here too?

dgelessus: The previous code effectively did this already with its own `#define AF_INET 2` - this change just makes that explicit. But the main reason is explained in the commit message for afc2dc63fae98073d5c4bd5ca274d1c4a9ea42b9:

> Although AF_INET is always 2 on every relevant system, it's better not
> to assume this. Here, we use AF_INET only as a marker value in a
> serialization format and never pass it to any system functions, so it's
> more important that the value is consistent between systems than that it
> matches the system constant.

For `AF_INET` it doesn't really make a difference, but e. g. `AF_INET6` has different values between OSes, so in general you can't send raw `AF_*` values over the network. (That said, `plNetAddress::Write` is never used anywhere, so this is all theoretical anyway.)

zrax: The point about serializing it over the network is fair, but it's still preferable to have a meaning associated with the value being sent, rather than just an opaque "2" with no clear indication of where that "2" came from (the comment helps, but not much).  Perhaps a better choice would be to define a new constant or enum, which would also give us better flexibility if we need to do expand it to support IPv6 in the future.

zrax (approved):

PR #1473, opened 2023-08-26
pnNcCli should consistently convert to little endian

dpogue:
There are 3 special cases that are not handled here:

- UUIDs are handled as "data", but those should already be stored in the weird mixed endian mode that Windows uses (we had to fix this for macOS a while back)

- SHA digests are handled as "raw data", and I'm not sure of the endianness on these. I think this might have been fixed in my first batch of PowerPC fixes? This might need further tweaking.

- Plasma Net Messages are "variable pointers" but those should have all their data stored in little endian by the `Read()`/`Write()` methods on the classes, so we should be okay here.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
@@ -613,6 +613,8 @@ static bool DispatchData (NetCli * cli, void * param) {
                     cli->recvFieldBytes = 0;
                 }
                 break;
+
+                default: break;
             }
         }
 
--- a/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
@@ -300,54 +300,64 @@ static void BufferedSendData (
             case kNetMsgFieldInteger: {
                 const unsigned count = cmd->count ? cmd->count : 1;
                 const unsigned bytes = cmd->size * count;
-                void * temp = malloc(bytes);
-                
-                if (count == 1)
-                {
+                auto temp = std::make_unique<uint8_t[]>(bytes);
+
+                if (count == 1) {
                     // Single values are passed by value
                     if (cmd->size == sizeof(uint8_t)) {
-                        *(uint8_t*)temp = *(uint8_t*)msg;
+                        *(uint8_t*)&temp[0] = *(uint8_t*)msg;
                     } else if (cmd->size == sizeof(uint16_t)) {
-                        *(uint16_t*)temp = hsToLE16(*(uint16_t*)msg);
+                        *(uint16_t*)&temp[0] = hsToLE16(*(uint16_t*)msg);
                     } else if (cmd->size == sizeof(uint32_t)) {
-                        *(uint32_t*)temp = hsToLE32(*(uint32_t*)msg);
+                        *(uint32_t*)&temp[0] = hsToLE32(*(uint32_t*)msg);
                     } else if (cmd->size == sizeof(uint64_t)) {
-                        *(uint64_t*)temp = hsToLE64(*(uint64_t*)msg);
+                        *(uint64_t*)&temp[0] = hsToLE64(*(uint64_t*)msg);
                     }
-                }
-                else
-                {
+                } else {
                     // Value arrays are passed in by ptr
                     for (size_t i = 0; i < count; i++) {
                         if (cmd->size == sizeof(uint8_t)) {
-                            ((uint8_t*)temp)[i] = ((uint8_t*)*msg)[i];
+                            *(uint8_t*)&temp[i] = ((uint8_t*)*msg)[i];
                         } else if (cmd->size == sizeof(uint16_t)) {
-                            ((uint16_t*)temp)[i] = hsToLE16(((uint16_t*)*msg)[i]);
+                            *(uint16_t*)&temp[i] = hsToLE16(((uint16_t*)*msg)[i]);
                         } else if (cmd->size == sizeof(uint32_t)) {
-                            ((uint32_t*)temp)[i] = hsToLE32(((uint32_t*)*msg)[i]);
+                            *(uint32_t*)&temp[i] = hsToLE32(((uint32_t*)*msg)[i]);
                         } else if (cmd->size == sizeof(uint64_t)) {
-                            ((uint64_t*)temp)[i] = hsToLE64(((uint64_t*)*msg)[i]);
+                            *(uint64_t*)&temp[i] = hsToLE64(((uint64_t*)*msg)[i]);
                         }
                     }
                 }
-                
-                // Write values to send buffer
-                AddToSendBuffer(cli, bytes, temp);
 
-                free(temp);
+                // Write values to send buffer
+                AddToSendBuffer(cli, bytes, temp.get());
             }
             break;
 
             case kNetMsgFieldReal: {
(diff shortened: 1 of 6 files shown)


Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -341,13 +338,30 @@ static void BufferedSendData (
             case kNetMsgFieldReal: {
                 const unsigned count = cmd->count ? cmd->count : 1;
                 const unsigned bytes = cmd->size * count;
-                
-                if (count == 1)
+                void* temp = malloc(bytes);
```
zrax, line 341:
```suggestion
                auto temp = std::make_unique<uint8_t[]>(bytes);
```

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -356,11 +370,20 @@ static void BufferedSendData (
                 // we reserve one space for the NULL terminator
                 const uint16_t length = (uint16_t) std::char_traits<char16_t>::length((const char16_t *) *msg);
                 ASSERT_MSG_VALID(length < cmd->count);
+
                 // Write actual string length
                 uint16_t size = hsToLE16(length);
                 AddToSendBuffer(cli, sizeof(uint16_t), (const void*)&size);
+
+                char16_t* temp = new char16_t[length];
```
zrax, line 378:
```suggestion
                std::vector<char16_t> temp(length);
```

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -356,11 +370,20 @@ static void BufferedSendData (
...
+                for (size_t i = 0; i < length; i++) {
+                    temp[i] = hsToLE16(((char16_t*)*msg)[i]);
+                }
+
                 // Write string data
-                AddToSendBuffer(cli, length * sizeof(char16_t), (const void *) *msg);
+                AddToSendBuffer(cli, length * sizeof(char16_t), (const void*)temp);
+
+                delete[] temp;
```
zrax, line 386:
```suggestion
                AddToSendBuffer(cli, length * sizeof(char16_t), (const void*)temp.data());
```

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -341,13 +338,30 @@ static void BufferedSendData (
...
+                        } else if (cmd->size == sizeof(double)) {
+                            ((double*)temp)[i] = hsToLEFloat(((double*)*msg)[i]);
+                        }
+                    }
+                }
+
+                // Write values to send buffer
+                AddToSendBuffer(cli, bytes, temp);
+
+                free(temp);
```
zrax, line 364:
```suggestion
                AddToSendBuffer(cli, bytes, temp.get());
```

dpogue:
This is now sufficient for the `plFilePatcher` command-line to work on Mac OS X 10.5 Leopard PowerPC (big endian) and download files.

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -356,11 +366,18 @@ static void BufferedSendData (
...
                 AddToSendBuffer(cli, sizeof(uint16_t), (const void*)&size);
+
+                std::vector<char16_t> temp(length);
+                for (size_t i = 0; i < length; i++) {
+                    temp[i] = hsToLE16(((char16_t*)*msg)[i]);
+                }
+
                 // Write string data
-                AddToSendBuffer(cli, length * sizeof(char16_t), (const void *) *msg);
+                AddToSendBuffer(cli, length * sizeof(char16_t), (const void*)temp.data());
```
Hoikas, line 380: Do we need to explicitly cast this to `const void*`?

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
```diff
@@ -356,11 +366,18 @@ static void BufferedSendData (
                 // we reserve one space for the NULL terminator
                 const uint16_t length = (uint16_t) std::char_traits<char16_t>::length((const char16_t *) *msg);
                 ASSERT_MSG_VALID(length < cmd->count);
+
                 // Write actual string length
                 uint16_t size = hsToLE16(length);
                 AddToSendBuffer(cli, sizeof(uint16_t), (const void*)&size);
+
+                std::vector<char16_t> temp(length);
```
Hoikas, line 374: Why not `std::unique_ptr<char16_t[]>` for consistency with the rest of the pnNcCli buffers?

Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglFile.cpp
```diff
@@ -107,12 +116,12 @@ struct CliFileConn : hsRefCnt {
...
-    bool Recv_BuildIdReply (const File2Cli_BuildIdReply * msg);
-    bool Recv_BuildIdUpdate (const File2Cli_BuildIdUpdate * msg);
-    bool Recv_ManifestReply (const File2Cli_ManifestReply * msg);
-    bool Recv_FileDownloadReply (const File2Cli_FileDownloadReply * msg);
+    void Dispatch(const Cli2File_MsgHeader * msg);
+    bool Recv_PingReply(const File2Cli_PingReply* recv);
+    bool Recv_BuildIdReply(const File2Cli_BuildIdReply* recv);
+    bool Recv_BuildIdUpdate(const File2Cli_BuildIdUpdate* recv);
+    bool Recv_ManifestReply(const File2Cli_ManifestReply* recv);
+    bool Recv_FileDownloadReply(const File2Cli_FileDownloadReply* recv);
```
Hoikas, line 124: It would be nice to defer the style changes to another PR.

Hoikas (commented):
I think I am coming around to your sense of malaise around all of the manual endianness work in the file messages. I'm especially unhappy with the temporary copy of certain messages from the file server being used solely for byte swapping, which could probably be done in-place (assuming the buffer is not `const`). It's basically a wasted copy on LE machines.

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
dpogue: This is what Zrax suggested here: https://github.com/H-uru/Plasma/pull/1473#discussion_r1306498052

Sources/Plasma/NucleusLib/pnNetCli/pnNcCli.cpp
dpogue: No idea, I just kept the cast that it had previously 😛

dpogue:
> I think I am coming around to your sense of malaise around all of the manual endianness work in the file messages. I'm especially unhappy with the temporary copy of certain messages from the file server being used solely for byte swapping, which could probably be done in-place (assuming the buffer is not `const`). It's basically a wasted copy on LE machines.

yeah, I don't like the extra copy, particularly since it's not needed for almost all cases... I thought about putting it behind some sort of `HS_BIG_ENDIAN` ifdef but that also seemed ugly.

Hypothetically there should be no issue making the messages non-const and modifying the data directly...

dpogue:
Updated to do the byte swapping directly on the buffer data, and confirmed that it's still working on both LE and BE architectures.

Hoikas (approved):

PR #1477, opened 2023-08-27
Birds aren't real...

Hoikas:
In evolutionary biology, we say that dinosaurs didn't die off... they simply flew away. In this case, we don't give a rat's about birds or dinosaurs. All these state descriptors that predate MOULa are of no practical use. They are merely historical curiosities. Therefore, we toss them. Culling all this old crap will be great for H-uru/dirtsand#184.


dgelessus (approved):
The diff on GitHub looks weird in places because there are formatting/comment changes mixed in. I checked locally and everything looks good to me - there are no dangerous changes to the `STATEDESC`s that are *not* being deleted.

dgelessus:
I remembered again that #1254 still exists, so the second copies of animation.sdl and avatar.sdl will get further out of sync with this. I can fix that after this PR is merged though (#1254 could use a rebase anyway as it's almost a year old).

dpogue (approved):

PR #1480, opened 2023-08-30
Add hsDarwin.h with helpers for string conversion

dpogue:
The `NSString` category in the macOS plClient folder is useful, but there are several other spots where we're wanting to convert between `CFString`/`NSString` and `ST::string`, so it make sense to have helpers available throughout the engine. These helper functions are written so that they can be used from C++ code without needing to bring in Objective-C.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.h
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.h
@@ -47,7 +47,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface NSString (StringTheory)
 
-- (id)initWithSTString:(const ST::string&)string;
 + (id)stringWithSTString:(const ST::string&)string;
 
 - (const ST::string)STString;
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
@@ -40,24 +40,19 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
-#include "StringTheory_NSString.h"
+#include "NSString+StringTheory.h"
+#include "hsDarwin.h"
 
 @implementation NSString (StringTheory)
 
-- (id)initWithSTString:(const ST::string&)string
-{
-    self = [self initWithUTF8String:string.c_str()];
-    return self;
-}
-
 + (id)stringWithSTString:(const ST::string&)string
 {
-    return [[NSString alloc] initWithSTString:string];
+    return NSStringCreateWithSTString(string);
 }
 
 - (const ST::string)STString
 {
-    return ST::string([self UTF8String]);
+    return STStringFromNSString(self);
 }
 
 @end
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
@@ -43,7 +43,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #import "PLSLoginWindowController.h"
 #include <regex>
 #import "PLSServerStatus.h"
-#include "StringTheory_NSString.h"
+#import "NSString+StringTheory.h"
 #include "pfPasswordStore/pfPasswordStore.h"
 #include "plNetGameLib/plNetGameLib.h"
 #include "plProduct.h"
@@ -128,8 +128,8 @@ static void* StatusTextDidChangeContext = &StatusTextDidChangeContext;
     [[NSUserDefaults standardUserDefaults] synchronize];
 
     if (self.password && ![self.password isEqualToString:FAKE_PASS_STRING]) {
-        ST::string username = ST::string([self.username cStringUsingEncoding:NSUTF8StringEncoding]);
-        ST::string password = ST::string([self.password cStringUsingEncoding:NSUTF8StringEncoding]);
+        ST::string username = [self.username STString];
+        ST::string password = [self.password STString];
 
         pfPasswordStore* store = pfPasswordStore::Instance();
         if (self.rememberPassword)
@@ -160,8 +160,8 @@ static void* StatusTextDidChangeContext = &StatusTextDidChangeContext;
     //  Hash username and password before sending over the 'net.
     //  -- Legacy compatibility: @gametap (and other usernames with domains in them) need
     //     to be hashed differently.
-    ST::string username = ST::string([self.username cStringUsingEncoding:NSUTF8StringEncoding]);
-    ST::string password = ST::string([self.password cStringUsingEncoding:NSUTF8StringEncoding]);
+    ST::string username = [self.username STString];
+    ST::string password = [self.password STString];
     static const std::regex re_domain("[^@]+@([^.]+\\.)*([^.]+)\\.[^.]+");
     std::cmatch match;
     std::regex_search(username.c_str(), match, re_domain);
@@ -185,7 +185,7 @@ static void* StatusTextDidChangeContext = &StatusTextDidChangeContext;
     ShaDigest hash;
     [self storeHash:hash];
 
-    ST::string username = ST::string([self.username cStringUsingEncoding:NSUTF8StringEncoding]);
+    ST::string username = [self.username STString];
     NetCommSetAccountUsernamePassword(username, hash);
     NetCommSetAuthTokenAndOS(nullptr, u"mac");
 }
@@ -214,7 +214,7 @@ static void* StatusTextDidChangeContext = &StatusTextDidChangeContext;
 
     [self.window center];
     [self.productTextField
-        setStringValue:[NSString stringWithSTString:plProduct::ProductString().c_str()]];
+        setStringValue:[NSString stringWithSTString:plProduct::ProductString()]];
 }
 
 - (NSNibName)windowNibName
--- a/Sources/Plasma/CoreLib/hsDarwin.h
+++ b/Sources/Plasma/CoreLib/hsDarwin.h
@@ -1,0 +1,114 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
+PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+licensors of this Program grant you additional
+permission to convey the resulting work. Corresponding Source for a
+non-source form of such a combination shall include the source code for
+the parts of OpenSSL and IJG JPEG Library used as well as that of the covered
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#ifndef _hsDarwin_inc_
+#define _hsDarwin_inc_
+
+#include <string_theory/string>
+#include <string_theory/format>
+
+#ifdef HS_BUILD_FOR_APPLE
+#include <CoreFoundation/CoreFoundation.h>
+
+[[nodiscard]]
+#if __has_feature(attribute_cf_returns_retained)
+__attribute__((cf_returns_retained))
+#endif
+inline CFStringRef CFStringCreateWithSTString(const ST::string& str)
+{
+    return CFStringCreateWithBytes(kCFAllocatorDefault, (const UInt8*)str.data(), str.size(), kCFStringEncodingUTF8, false);
+}
+
+inline ST::string STStringFromCFString(CFStringRef str, ST::utf_validation_t validation = ST_DEFAULT_VALIDATION)
+{
+    CFRange range = CFRangeMake(0, CFStringGetLength(str));
+    CFIndex strBufSz = 0;
+    CFStringGetBytes(str, range, kCFStringEncodingUTF8, 0, false, nullptr, 0, &strBufSz);
+    ST::char_buffer buffer;
+    buffer.allocate(strBufSz);
+    CFStringGetBytes(str, range, kCFStringEncodingUTF8, 0, false, (UInt8*)buffer.data(), strBufSz, nullptr);
+
+    return ST::string(buffer, validation);
+}
+
+inline void format_type(const ST::format_spec &format, ST::format_writer &output, CFStringRef str)
+{
+    ST::char_buffer utf8 = STStringFromCFString(str).to_utf8();
+    ST::format_string(format, output, utf8.data(), utf8.size());
+}
+
+
+#ifdef __OBJC__
+@class NSString;
+
+[[nodiscard]]
+#if __has_feature(attribute_ns_returns_retained)
+__attribute__((ns_returns_retained))
+#endif
+inline NSString* NSStringCreateWithSTString(const ST::string& str)
+{
+#if __has_feature(objc_arc)
+    return (NSString*)CFBridgingRelease(CFStringCreateWithSTString(str));
+#else
+    return (NSString*)CFStringCreateWithSTString(str);
+#endif
+}
+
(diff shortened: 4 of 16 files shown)


Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -0,0 +1,105 @@
...
+#include <string_theory/format>
+
+#ifdef HS_BUILD_FOR_APPLE
+#include <CoreFoundation/CoreFoundation.h>
+
+#ifdef __OBJC__
+    @class NSString;
+#else
+    class NSString;
+#endif
```
dgelessus, line 56: Is this still needed? I think all the `NSString`-specific code could be moved behind `#ifdef __OBJC__` - then non-ObjC code doesn't need to know about `NSString`.

We should have no ODR issues here, because all the `NSString` code is macros and standalone functions, not methods (as in zrax/string_theory#26).

dgelessus (commented):
This is definitely an improvement for working with `CFStringRef`s. But for `NSString`, I don't really see the benefit, to be honest... If you're working with `NSString`s, you're already in Objective-C code, so you can make full use of the language and define a category, like we did previously. (If we want to reduce code duplication, you can still implement the category on top of hsDarwin.h, and have the bridging code in the category's ObjC source file.)

colincornaby:
My general thoughts on this...

I don't really see a problem - but like @dgelessus mentioned it would be nice to leave the Obj-C category intact. Part of my goals has been to provide a sane frontend for any future Swift code through Obj-C. This new code doesn't include nullability annotations - which is a step back. The existing Obj-C code could be refactored to sit on this new code however.

I think my perspective is also that Obj-C (or Swift) is always going to be implicitly part of writing an iOS or Mac app. hsMessageBox_Mac.mm is a good example of this. Maybe not directly related to this PR - but in general I'll be a little skeptical of code sections where C++ is used to avoid Obj-C for no clear reason. Objective-C++ can also be used to patch Obj-C into C++ classes (again with hsMessageBox_Mac.mm being an example.) CFStringRef should only be used where the integration directly requires it (like in keychain support.)

Also going to note that CoreFoundation is manually memory managed - while Obj-C is automatically memory managed. I'll check through this commit to see if that's properly handled. But needing memory management review is a reason I'd discourage use of CoreFoundation unless appropriate.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -116,7 +116,7 @@ - (void)save
 {
     // windows segments by product name here. in since user defaults belong to this product, we
     // don't need to do that.
-    NSString* serverName = [NSString stringWithSTString:GetServerDisplayName()];
+    NSString* serverName = ST2NSString(GetServerDisplayName());
```
colincornaby, line 119: I think this is a memory leak because ST2NSString is no longer ARC managed and returns a +1 CFString. Would prefer if the Obj-C interface was left operational to solve this issue.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -141,17 +141,17 @@ - (void)save
...
     self.username = [settingsDictionary objectForKey:@"LastAccountName"];
     self.rememberPassword = [[settingsDictionary objectForKey:@"RememberPassword"] boolValue];
 
     if (self.rememberPassword) {
         pfPasswordStore* store = pfPasswordStore::Instance();
-        ST::string username = [self.username STString];
+        ST::string username = NS2STString(self.username);
         ST::string password = store->GetPassword(username);
-        self.password = [NSString stringWithSTString:password];
+        self.password = ST2NSString(password);
```
colincornaby, line 154: Same issue in the above sections. I'm pretty sure these strings now leak.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -214,7 +214,7 @@ - (void)windowDidLoad
 
     [self.window center];
     [self.productTextField
-        setStringValue:[NSString stringWithSTString:plProduct::ProductString().c_str()]];
+        setStringValue:ST2NSString(plProduct::ProductString().c_str())];
```
colincornaby, line 217: Another possible leak. stringWithSTString returned an autoreleased string. ST2NSString does not.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -263,7 +263,7 @@ - (void)loginAttemptEndedWithResult:(ENetError)result
 
 - (IBAction)needAccountButtonHit:(id)sender
 {
-    NSString* urlString = [NSString stringWithSTString:GetServerSignupUrl()];
+    NSString* urlString = ST2NSString(GetServerSignupUrl());
```
colincornaby, line 266: Another leak - I'm going to stop calling out the leaks individually because there are a few now.

colincornaby:
Reading through - it would be better if the Obj-C interface was left working and returned an autoreleased string. There are a few other possible ways (changing the __bridge to __bridge_transfer for example) - but leaving the Obj-C interface intact would be the most unambiguous on what the memory rules are.

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -0,0 +1,105 @@
...
+    CFStringGetBytes(str, range, kCFStringEncodingUTF8, 0, false, nullptr, 0, &strBufSz);
+    ST::char_buffer buffer;
+    buffer.allocate(strBufSz);
+    CFStringGetBytes(str, range, kCFStringEncodingUTF8, 0, false, (UInt8*)buffer.data(), strBufSz, nullptr);
+
+    return ST::string(buffer, validation);
+}
+
+#define ST2CFString(str) CFStringCreateWithSTString(str)
+#define CF2STString(str) STStringFromCFString((CFStringRef)str)
```
colincornaby, line 80: It might be good to remove these shorthands. They mask the ownership rules. (The word create is pretty critical in the above function to understanding the memory ownership relationship.)

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -0,0 +1,105 @@
...
+
+*==LICENSE==*/
+
+#ifndef _hsDarwin_inc_
+#define _hsDarwin_inc_
+
+#include <string_theory/string>
+#include <string_theory/format>
+
+#ifdef HS_BUILD_FOR_APPLE
```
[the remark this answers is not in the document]
dpogue: This was copied from how hsWindows.h is implemented... only benefit I can really see is that we can include it in the Pch file

dpogue:
> A __bridge_transfer might still be necessary in the Obj-C wrapper - but it's better to do that changeover at the exact point that a CFString crosses from C to Obj-C. I worry that doing a __bridge_transfer inside of a C function is ambiguous.

It looks like `CFBridgingRetain` is the correct way to do this from a C function

Sources/Plasma/CoreLib/hsDarwin.h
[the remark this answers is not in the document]
colincornaby: Putting on my "technically correct" hat...

NSString and CFString are both cross platform types. They're part of Apple's open source versions of Foundation and CoreFoundation, and Swift for Windows and Linux would have these types.

That said: I doubt anyone is going to be building Plasma stuff using CoreFoundation or Foundation on Windows or Linux.

colincornaby:
I'm still not super excited about the removal of the Obj-C category - it could still be a front end for the new code. And it's a nice affordance to have when working in Obj-C. But I think the changes are an improvement for classes like the keychain handling. There would also be less churn in the Obj-C code if a category was maintained.

But I think in principal these are good changes - and the latest edits fix the memory issues.

dpogue:
Like I said, I don't mind putting the ObjC category back, but that still raises issues around using it in spots outside of plClient (such as the caption/message of hsMessageBox).

I've (minimally) tested that this code works in 10.5 PPC (which predates all the ARC stuff), but I'd still like to add some unit tests for it.

colincornaby:
I think stuff like hsMessage could use the CF bridged versions.

Trying the run the existing client and code on a pre-ARC system is a whole... thing. I did the move to ARC because there just aren't many non-ARC developers out there - and it kept the code accessible. 10.5 era AppKit is also very different.

Are you attempting to use the existing Mac client on 10.5? Or are you writing a new shell?

Obj-C code that sits in the engine itself could be moved back to manual memory management. That would really be hsMessageBox, the keychain stuff doesn't care about ARC. I don't know if the code would work with Obj-C garbage collection. GC was rarely used but also promised to work without retains/releases.

The actual API used could also end up being a problem for 10.5. And I think the xib encoding has changed since the PPC era tools. It might be better if 10.5 PPC had it's own client - if you were thinking of using the Mac client.

dpogue:
I'm pretty sure the current macOS client won't work with 10.5, so I haven't even tried it yet. I'm more interested in making sure the core engine stuff is able to work on 10.5 than the client bits right now.

IIRC the updates to pfPasswordStore for updating passwords cause it to use APIs that didn't exist in 10.5, and hsMessageBox doesn't currently work because it's using block syntax and `@autoreleasepool` that aren't supported by standard GCC.

dpogue:
I've updated this to restore the Objective C category for plClient code, but implemented using the new hsDarwin methods under the hood.

colincornaby:
Fun side quest - I let Xcode name the category, but I wanted to find out if there was convention.

The + format for categories IS actually defined convention:
https://developer.apple.com/library/archive/documentation/General/Conceptual/DevPedia-CocoaCore/Category.html

colincornaby:
One more thing I'll add - changes to the Mac code will probably be necessary, and I don't want to railroad any existing conventions. However - I'll also probably be a little conservative on changes for the time being because changes here may roll down to the Metal client and cause conflicts.

colincornaby (approved):

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -0,0 +1,114 @@
...
+@class NSString;
+
+[[nodiscard]]
+#if __has_feature(attribute_ns_returns_retained)
+__attribute__((ns_returns_retained))
+#endif
+inline NSString* NSStringCreateWithSTString(const ST::string& str)
+{
+#if __has_feature(objc_arc)
+    return (NSString*)CFBridgingRelease(CFStringCreateWithSTString(str));
```
dgelessus, line 90: I think this is incorrect, but I'm not super sure about all of this. The way I understand it, `CFBridgingRelease` takes in an "owned reference" and returns an "unowned reference". This conflicts with the "Create" function name and "returns retained" attribute, which indicate that the return value is an "owned" reference. I *think* a plain `__bridge` cast would be correct here - that way the "owned" reference stays "owned" as it's returned.

```suggestion
    return (__bridge NSString*)CFStringCreateWithSTString(str);
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.h
dgelessus, line 1: Could we move this into CoreLib so that all code can use the `NSString (StringTheory)` category, not just plClient? This would allow deleting the Objective-C-only `NSString` functions from hsDarwin.h, which *might* help with untangling the ARC bridging casts...

If we do that, CoreLib needs to link against the Foundation framework and not just Core Foundation, but I don't think that's a huge issue.

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
```diff
@@ -40,13 +40,14 @@
...
-#include "StringTheory_NSString.h"
+#include "NSString+StringTheory.h"
+#include "hsDarwin.h"
 
 @implementation NSString (StringTheory)
 
 - (id)initWithSTString:(const ST::string&)string
 {
-    self = [self initWithUTF8String:string.c_str()];
+    self = NSStringCreateWithSTString(string);
```
dgelessus, line 50: We're assuming ARC here, right? In non-ARC code, this would have to `release` the old `self`.

colincornaby: All Obj-C code right now is ARC - so this should assume ARC.

release isn't typically called here because init is only allowed to be called as a constructor. It also might release the alloc'd memory. But I don't think I've ever considered it strongly because init is a constructor and should only be called once.

Sources/Plasma/CoreLib/hsDarwin.h
colincornaby: I believe this code is correct. It's releasing CoreFoundation's handle to the code while transferring it to ARC.

If the string was CFRelease'd later, a normal bridge would be convention.

Apple's example shows a similar scenario:
https://developer.apple.com/documentation/foundation/1587932-cfbridgingrelease

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.h
colincornaby: CoreLib would also need to inherit some build settings like ARC. (Or maybe the code is un-ARC'd?)

FWIW - I'm not totally convinced Obj-C should go into CoreLib, but that might just be my own personal hangup. I'd rather have Obj-C than any other sort of Obj-C runtime hacking shenanigans if these sorts of features need to go into CoreLib.

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
dpogue: would `self = [self initWithString:NSStringCreateWithSTString(string)];` be better? It feels like it's potentially duplicating the string, but maybe it's smart enough not to do that?

colincornaby: Obj-C convention would be that the string should be duplicated - unless the function specifically calls that out in it's name. (EG the noCopy family of functions.)

I'd prefer duplication unless there's serious concerns about cost as it's usually the Obj-C default.

colincornaby: Oh, I think I see what you're saying a bit clearer. Yeah, there might be two duplications. It's not super efficient. The initializing from the cstr would be preferable, but not super essential.

dpogue: Actually, is there a compelling reason to have the initializer function, and not just return `NSStringCreateWithSTString` from the static `stringWithSTString:` method?

colincornaby: Self has already been allocated and has a pointer. Returning a different pointer will leak self I believe.

dpogue: I'm proposing this:

```diff
-- (id)initWithSTString:(const ST::string&)string
-{
-    self = NSStringCreateWithSTString(string);
-    return self;
-}
-
 + (id)stringWithSTString:(const ST::string&)string
 {
-    return [[NSString alloc] initWithSTString:string];
+    return NSStringCreateWithSTString(string);
 }
```

colincornaby: Ah, I missed you said static. I think that would be ok.

Sources/Plasma/CoreLib/hsDarwin.h
dgelessus: Hmm, I think I'm getting confused by the "release" terminology and the extra inline function here... Am I reading correctly that `CFBridgingRelease` in ARC mode does the same thing as a `__bridge_transfer` cast? I see now that we used `__bridge_transfer` before, but it was then changed to `CFBridgingRelease` and I don't understand why :/

You are right that the ownership needs to be transferred to ARC, so a plain `__bridge` would be incorrect/leaky, yeah. My issue is that we don't want to actually `release` the object at this point, because `NSStringCreateWithSTString` is declared to return a retained object. But perhaps I'm overthinking this - ARC *should* understand what needs to be done based on the "returns retained" attribute combined with a `__bridge_transfer` cast.

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.h
dgelessus: We could pull the Objective-C support into a separate pnObjC library or something - but as much as I like reducing unneeded dependencies, I think it wouldn't be worth the extra complexity. CoreLib is already not very minimal - it also does e. g. COM on Windows, even though most Plasma libraries don't depend on that.

Sources/Plasma/CoreLib/hsDarwin.h
colincornaby: > Hmm, I think I'm getting confused by the "release" terminology and the extra inline function here... Am I reading correctly that `CFBridgingRelease` in ARC mode does the same thing as a `__bridge_transfer` cast? I see now that we used `__bridge_transfer` before, but it was then changed to `CFBridgingRelease` and I don't understand why :/

I'd have to go back through the history but that might be my fault. I've been doing Swift too long where CoreFoundation is also ARC managed...

I might have thought __bridge_transfer was a bridge that didn't move ownership.

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
dgelessus: > Self has already been allocated and has a pointer. Returning a different pointer will leak self I believe.

That was my original worry - without ARC, the original `self` would need to be `release`d, otherwise it would leak. With ARC, everything should be fine. In any case, @dpogue's suggestion would be more readable, yeah. (`initWithSTString:` should never be necessary if we have `stringWithSTString:`.)

Sources/Plasma/CoreLib/hsDarwin.h
dgelessus: > moving it to the ObjC category might result in leaks in the plMessageBox stuff?

Yeah, that would only be possible if we move the category out of plClient and use it everywhere (as suggested below).

Hoikas:
Is this good to go?

dpogue:
> Is this good to go?

IMO, yes

colincornaby:
I don't have any more feedback. (Already approved.)

dgelessus:
Should be okay from my side too. The one thing that I thought was a bug is actually fine. The other discussions I opened are about API organization, so they're not essential and/or can be addressed later if needed.

PR #1485, opened 2023-09-16
Fix 3 problems with `plSubWorldMsg`.

Hoikas:
The impetus for this change is that Korman's new `Change Subworld` message node can cause all players in the Age to change to that subworld. The rationale here is that changing subworlds modifies a player's avatarPhysical SDL. So, really, we probably only want to be sending `plSubWorldMsg` to the local player from a Responder. Responders are primarily a zero-code artist logic tool, so they shouldn't have to care about that level of SDL detail. Further, there is precedent for major avatar/player related actions to be rejected from non-local sources. Namely, all camera messages are restricted in the same way in Responders.

As a bonus, I fixed a memory leak that wouldn't be detected by asan and fixed a logic error with the subworld SDL synch.


dpogue (approved):

Hoikas:
Update: I checked all of Cyan's responders, and none of them send subworld messages, so there is no danger of this breaking Cyan content.

PR #1487, opened 2023-09-17
Port NetCli/NetGameLib to standard C++ containers

dgelessus:
This eliminates the last uses of the weird custom intrusive linked list pnUtList (aka `LISTDECL`/`LIST`/`LINK`).

Most of this is straightforward. The only interesting change is in NetGameLib, where I'm replacing each `s_conns` list with a single `s_conn` variable, because the lists could never contain more than one element.


Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglTrans.cpp
```diff
@@ -252,31 +259,36 @@ void NetTransCancelByConnId (unsigned connId, ENetError error) {
...
-                        if (trans->m_hasSubTrans)
-                            parentCompleted.Link(trans);
-                        else
-                            completed.Link(trans);
+                        s_transactions.erase(it);
+                        if (trans->m_hasSubTrans) {
+                            parentCompleted.push_back(trans);
+                        } else {
+                            completed.push_back(trans);
+                        }
```
Hoikas, line 291: I think it would be more efficient to use `std::list<T>::splice()` to avoid deleting and reallocating list nodes.

Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
```diff
@@ -1249,7 +1248,7 @@ enum {
 
 static bool                         s_running;
 static std::recursive_mutex         s_critsect;
-static LISTDECL(CliAuConn, link)    s_conns;
+static CliAuConn* s_conn = nullptr;
 static CliAuConn *                  s_active;
```
Hoikas, line 1252: Are there ever any times that `s_conn` and `s_active` will not be the same? I'm wondering if we could just use `s_active` instead...

dgelessus: That's what I wanted to do initially, but unfortunately the two aren't exactly the same. `s_conns`/`s_conn` is always set before `AsyncSocketConnect`, whereas `s_active` is only set later once the connection is fully established (e. g. for the auth server only after a `Auth2Cli_ClientRegisterReply` is received).

This logic can probably be simplified somehow so that it only uses one variable, but I didn't want to do anything too risky as part of the legacy data structure cleanup.

Hoikas: That seems like a good approach. 👍

Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
```diff
@@ -1327,8 +1326,7 @@ static CliAuConn * GetConnIncRef (const char tag[]) {
 }
 
 //============================================================================
-static void UnlinkAndAbandonConn_CS (CliAuConn * conn) {
-    s_conns.Unlink(conn);
+static void AbandonConn(CliAuConn* conn) {
```
Hoikas, line 1329: I think this is my last question, but, just to verify... Does the caller not need to have the critical section locked anymore? I'm assuming *no* because this is no longer accessing any static variables, and you dropped `_CS` from the function name. If *yes*, we probably want to retain the `_CS` suffix on the function name.

dgelessus: Exactly. As far as I can tell, the `s_critsect` lock was only needed to safely update `s_conns`, so now that this function doesn't do that anymore, it should be safe to call without the `s_critsect` locked (although currently all calls still have the lock).

That said, now that I'm re-reading this... maybe `AbandonConn` should lock on `conn->critsect` instead, because it manipulates a few fields directly instead of just calling methods? Honestly, I'm not quite sure which parts need which lock exactly (and I think the existing code is a bit inconsistent too).

Hoikas: It's all a little muddy, for sure, so I took a look at the code. It looks like `CliAuConn`'s ping timer reconnect stuff what requires the lock on `CliAuConn::critsect`, so I don't think we need to lock that here. OTOH, the code *does* seem to lock `s_critsect` when accessing or mutating any `CliAuConn` instance's `cancelId` field (see `NotifyConnSocketConnectFailed` and the other notification callbacks), so I think we still need to lock `s_critsect` before calling this function and rename it to `AbandonConn_CS`. I'm assuming the other CliConn variants have the same ownership model :/

dgelessus: > Alternatively, you could just lock `s_critsect` inside `AbandonConn`

That seems like the best solution, yeah.

Hoikas (approved):

PR #1489, opened 2023-09-25
Support Unicode in `plTextFont` (avatar tooltips, console, etc.)

dgelessus:
Fixes the last remaining problems with rendering non-ASCII characters. The main effect is that avatar tooltips work for non-ASCII names now:

Screenshot of avatar tooltip for "Æthelred the Unready" with working "Æ" character

Aside from that, it fixes non-ASCII text rendering in the console, log drawer, other debug views, and the loading screen. With this, I *think* Plasma now supports Unicode (or at least Latin-1) in all places where it matters.

Internally, this expands the supported character range of `plTextFont` from ASCII to Latin-1. Characters beyond the supported range don't cause rendering glitches anymore and instead show a fallback character (currently `¿`, because the real replacement character `�` is not in Latin-1).

Because `plTextFont` uses the OS native fonts instead of the game's .p2f files, it's easy to extend the character range further, theoretically up to all of Unicode. The practical limit is currently much lower than that though, because `plTextFont` pre-renders the entire character set into a texture with hardcoded sizes. I stuck with Latin-1 for now, because it fits into the current texture sizes, and anything else isn't usable by players yet anyway (though that would change with H-uru/moul-assets#237 adding Russian fonts).

If we want to support more than Latin-1, we need to adjust the hardcoded texture sizes accordingly, or calculate the required texture size dynamically as the font is loaded. Might also be good to support non-contiguous ranges of characters - that's an open issue for `plFont` as well I think.

The diff is quite big, but most of it are boring `ST::string`ification changes for all the debug views. The actually interesting changes are are in `plTextFont` and `pfConsole`.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfAnimation/plAnimDebugList.cpp
+++ b/Sources/Plasma/FeatureLib/pfAnimation/plAnimDebugList.cpp
@@ -114,7 +114,7 @@ void plAnimDebugList::ShowReport()
 
     x = startX;
     y = startY;
-    txt.DrawString(x, y, "Material Animations:", 255, 255, 255, 255, plDebugText::kStyleBold);
+    txt.DrawString(x, y, ST_LITERAL("Material Animations:"), 255, 255, 255, 255, plDebugText::kStyleBold);
     y += yOff;
     for (const plKey& matKey : fMaterialKeys)
     {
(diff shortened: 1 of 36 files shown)


ZarothYe:
I have no ability to review this on a technical level, but I'm so freaking hyped for and immensely impressed by this! Well done!

Hoikas (approved):

PR #1494, opened 2023-10-01
More keychain fixes

colincornaby:
Minor fixes

- Keychain changes apparently need to happen from a secondary thread - and not the main thread. Xcode is now warning about it.
- Had some issues with Keychain not updating reliably on one of my machines. Some refinements that seem to make Keychain happier with updating password records by simplifying the queries.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
+++ b/Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
@@ -93,8 +93,8 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
     CFAutorelease(serviceName);
     CFAutorelease(passwordData);
 
-    const void* keys[] = { kSecClass, kSecAttrService, kSecReturnData, kSecValueData };
-    const void* values[] = { kSecClassGenericPassword, serviceName, kCFBooleanTrue, passwordData };
+    const void* keys[] = { kSecClass, kSecAttrService, kSecAttrAccount, kSecValueData };
+    const void* values[] = { kSecClassGenericPassword, serviceName, accountName, passwordData };
 
     CFDictionaryRef query = CFDictionaryCreate(nullptr, keys, values, 4, nullptr, nullptr);
     CFAutorelease(query);
@@ -104,14 +104,14 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
     if (err == errSecDuplicateItem) {
         // the keychain item already exists, update it
 
-        const void* queryKeys[2] = { kSecClass, kSecAttrService };
-        const void* queryValues[2] = { kSecClassGenericPassword, serviceName };
-        CFDictionaryRef updateQuery = CFDictionaryCreate(nullptr, queryKeys, queryValues, 2, nullptr, nullptr);
+        const void* queryKeys[] = { kSecClass, kSecAttrService, kSecAttrAccount };
+        const void* queryValues[] = { kSecClassGenericPassword, serviceName, accountName };
+        CFDictionaryRef updateQuery = CFDictionaryCreate(nullptr, queryKeys, queryValues, 3, nullptr, nullptr);
         CFAutorelease(updateQuery);
 
-        const void* attributeKeys[2] = { kSecAttrAccount, kSecValueData };
-        const void* attributeValues[2] = { accountName, passwordData };
-        CFDictionaryRef attributes = CFDictionaryCreate(nullptr, attributeKeys, attributeValues, 2, nullptr, nullptr);
+        const void* attributeKeys[1] = { kSecValueData };
+        const void* attributeValues[1] = { passwordData };
+        CFDictionaryRef attributes = CFDictionaryCreate(nullptr, attributeKeys, attributeValues, 1, nullptr, nullptr);
 
         err = SecItemUpdate(updateQuery, attributes);
 


dpogue (approved):

Deledrius (approved):

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
```diff
@@ -92,27 +92,27 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
...
     CFAutorelease(serviceName);
     CFAutorelease(passwordData);
-
-    const void* keys[] = { kSecClass, kSecAttrService, kSecReturnData, kSecValueData };
-    const void* values[] = { kSecClassGenericPassword, serviceName, kCFBooleanTrue, passwordData };
-
+    
+    const void* keys[] = { kSecClass, kSecAttrService, kSecAttrAccount, kSecValueData };
+    const void* values[] = { kSecClassGenericPassword, serviceName, accountName, passwordData };
+
```
Hoikas, line 98: Whitespace

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
```diff
@@ -92,27 +92,27 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
...
 
     OSStatus err = SecItemAdd(query, nullptr);
 
     if (err == errSecDuplicateItem) {
         // the keychain item already exists, update it
-
-        const void* queryKeys[2] = { kSecClass, kSecAttrService };
-        const void* queryValues[2] = { kSecClassGenericPassword, serviceName };
-        CFDictionaryRef updateQuery = CFDictionaryCreate(nullptr, queryKeys, queryValues, 2, nullptr, nullptr);
+
```
Hoikas, line 106: Whitespace

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
```diff
@@ -92,27 +92,27 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
...
+        const void* queryKeys[] = { kSecClass, kSecAttrService, kSecAttrAccount };
+        const void* queryValues[] = { kSecClassGenericPassword, serviceName, accountName };
+        CFDictionaryRef updateQuery = CFDictionaryCreate(nullptr, queryKeys, queryValues, 3, nullptr, nullptr);
         CFAutorelease(updateQuery);
-
-        const void* attributeKeys[2] = { kSecAttrAccount, kSecValueData };
-        const void* attributeValues[2] = { accountName, passwordData };
-        CFDictionaryRef attributes = CFDictionaryCreate(nullptr, attributeKeys, attributeValues, 2, nullptr, nullptr);
-
+
```
Hoikas, line 111: Whitespace

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
```diff
@@ -92,27 +92,27 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
...
-
-        const void* attributeKeys[2] = { kSecAttrAccount, kSecValueData };
-        const void* attributeValues[2] = { accountName, passwordData };
-        CFDictionaryRef attributes = CFDictionaryCreate(nullptr, attributeKeys, attributeValues, 2, nullptr, nullptr);
-
+        
+        const void* attributeKeys[1] = { kSecValueData };
+        const void* attributeValues[1] = { passwordData };
+        CFDictionaryRef attributes = CFDictionaryCreate(nullptr, attributeKeys, attributeValues, 1, nullptr, nullptr);
+
```
Hoikas, line 115: Whitespace

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -131,11 +132,13 @@ - (void)save
...
-            store->SetPassword(username, password);
-        else
-            store->SetPassword(username, ST::string());
+        dispatch_async(_keychainQueue, ^{
+            pfPasswordStore* store = pfPasswordStore::Instance();
+            if (self.rememberPassword)
+                store->SetPassword(username, password);
+            else
+                store->SetPassword(username, ST::string());
+        });
```
Hoikas, line 141: If I understand correctly, this is enqueuing the `pfPasswordStore` function call to happen on a secondary thread. The goal of `pfPasswordStore::Instance()` is to return a platform agnostic password store. The expectation is that all of the password stores have the same interface. The interface of the Win32 and Linux password stores is thread-agnostic. Therefore, it is important that the Apple password store be thread-agnostic as well. Thus, the worker thread fiddling needs to be done in `pfPasswordStore_APPLE` so that the interfaces of all password stores match exactly.

Hoikas (changes requested):

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
dpogue: That might be complicated in the case of retrieving the password though (you could probably do something with `std::promise` and changing the interface to return a `std::future<ST::string>` but that's a bit complicated...) 😕

colincornaby: For extra fun - reading the password seems fine? It's the writing that is the issue. The dispatch threading could be moved to the password store class itself (or be replaced by C++ threading but that's a longer effort.)

FWIW - I don't know why the debugger throws this warning at runtime. They warn it could stall the main thread but don't say why. We trigger the keychain unlock dialog (if needed) at launch. It could be that if the user waits long enough, the keychain will lock, setting the password will need a dialog, and the main thread will be held.

colincornaby:
I think this is stuck on the "should asyncronisity be baked in at an API level" question.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
colincornaby: Checking back in again - should re-arranging the underlying API to handle threading with promises be part of this PR? I can make the change - but I just want to make sure that's a direction we're all in agreement on.

Hoikas: I think it would be nice if we could get away with not using a defferred (read `std::promise`) for reading the password. IMO, it's ok if the main thread stalls there - I'm assuming that macOS has stolen the thread to prompt the user for some interaction. I would prefer to see the password saving dispatched inside `pfPasswordStore`, though.

colincornaby: Without knowing more about Apple's definition of blocking - I agree that this probably is fine on the main thread.

However - and this is what caused this PR - Xcode flags a runtime error/warning when we make this call off the main thread. So this issue is currently leading to us not getting clean runs out of Xcode.

colincornaby:
I'm circling back to this PR as I was running into some of these issues when testing deploying a server the other day.

I think where we got stuck is the main thread issues. Apple does not want us using the main thread - to the point where this gets raised as a runtime warning within Xcode. I'm not entirely clear on why not calling on the main thread is preferred. I can continue to investigate.

As stated in the previous comment chain, I think we'd maybe be fine using the main thread to store passwords. But it would be nice to know why Apple thinks we shouldn't be safe. And it would be nice to not have the runtime warning.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
colincornaby: I did a re-review of these changes - including looking through all the possible solutions again. I'm going to re-nudge this PR against the existing set of changes.

I think the client code should be responsible for the threading because there might be UI implications to these functions being asynchronous. For example - the UI should do something sane while getting the password on a thread. Unless we want to adopt a promise model in C++ that communicates this state - I'd be more comfortable with the UI managing threading. Because all other password stores should remain thread agnostic - this also wouldn't prevent macOS from adopting other password stores.

This is also only a runtime warning so if another client adopted the keychain store it wouldn't be fatal if they access the keychain password store from the main thread. And they could make similar changes.

Speaking of the runtime warning - I am not seeing it anymore on macOS Sequoia and Xcode 16. I haven't seen any formal change notes around this - so maybe it's just me.

colincornaby:
FYI - I've removed the changes that kept this PR from moving forward. I'll be opening another PR with those specific changes.

colincornaby:
I'll look at why the build failed. This branch might just be behind.

dpogue (approved):

PR #1497, opened 2023-10-01
Combine the two string tokenizer variants into a template

dgelessus:
`hsStringTokenizer` and `hsWStringTokenizer` were nearly identical, except for the character type. The Max plugin code also expects the two to have compatible APIs.

I would rather convert this to `ST::string` like all the other old string code, but one part of the Max plugin (UserPropMgr) depends heavily on how this implementation allocates and reuses strings, so it's difficult to move away from manually managed raw strings. I'm sure the Max code could be rewritten to use fewer weird tricks, but that's not something I want to do, because I have no way to test that code.


Sources/Plasma/CoreLib/hsStringTokenizer.h
```diff
@@ -45,66 +45,164 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
-    bool IsSep(wchar_t c);
+    static size_t StrLen(const CharT* str)
+    {
+        size_t i = 0;
+        while (str[i] != 0) {
+            i++;
+        }
+        return i;
+    }
+
```
Hoikas, line 185:
```suggestion
```

Sources/Plasma/CoreLib/hsStringTokenizer.h
```diff
@@ -45,66 +45,164 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+            *fLastTerminator = fLastRep;
+            fLastTerminator = nullptr;
+        }
+    }
+
+    void Reset(const CharT* string, const CharT* seps)
+    {
+        delete[] fString;
+        if (string) {
+            size_t count = StrLen(string);
```
Hoikas, line 135:
```suggestion
            size_t count = std::char_traits<CharT>::length(string);
```

Sources/Plasma/CoreLib/hsStringTokenizer.h
```diff
@@ -45,66 +45,164 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+            fString = new CharT[count + 1];
+            memcpy(fString, string, sizeof(CharT) * (count + 1));
+        } else {
+            fString = nullptr;
+        }
+
+        delete[] fSeps;
+        if (seps) {
+            fNumSeps = StrLen(seps);
```
Hoikas, line 144:
```suggestion
            fNumSeps = std::char_traits<CharT>::length(seps);
```

Sources/Plasma/CoreLib/hsStringTokenizer.h
```diff
@@ -45,66 +45,164 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        }
+    }
+
+    void Reset(const CharT* string, const CharT* seps)
+    {
+        delete[] fString;
+        if (string) {
+            size_t count = StrLen(string);
+            fString = new CharT[count + 1];
+            memcpy(fString, string, sizeof(CharT) * (count + 1));
```
Hoikas, line 137:
```suggestion
            std::char_traits<CharT>::copy(fString, string, count);
            fString[count] = 0;
```

Sources/Plasma/CoreLib/hsStringTokenizer.h
```diff
@@ -45,66 +45,164 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        } else {
+            fString = nullptr;
+        }
+
+        delete[] fSeps;
+        if (seps) {
+            fNumSeps = StrLen(seps);
+            fSeps = new CharT[fNumSeps + 1];
+            memcpy(fSeps, seps, sizeof(CharT) * (fNumSeps + 1));
```
Hoikas, line 146:
```suggestion
            std::char_traits<CharT>::copy(fSeps, seps, fNumSeps);
            fString[fNumSeps] = 0;
```

Hoikas (commented):
Just some minor suggestions for using standard features where available. Also, I'd prefer to see us continue to explicitly null the end of the string buffer.

Hoikas (approved):

PR #1498, opened 2023-10-04
Tools reorg

dpogue:
This is open for discussion, but seemed worthwhile to get it working first so people could take a look.

* Moves all the non-client tools out of `Sources/Plasma/Apps` and into `Sources/Tools`:
    * plFileEncrypt
    * plFileSecure
    * plPageInfo
    * plPageOptimizer
    * plPythonPack

* The exception is SoundDecompress which moved from `Sources/Tools` into `Sources/Plasma/Apps` (because it is sorta intended to be run by end users and distributed with the client)

* Move all the Max Plugin stuff out of `Sources/Tools` into its own `Sources/MaxPlugin` folder so that it is more self-contained:
    * MaxComponent
    * MaxConvert
    * MaxExport
    * MaxMain
    * MaxPlasmaLights
    * MaxPlasmaMtls

* Only build plLocalizationBenchmark when `PLASMA_BUILD_TOOLS` is on (arguably it's not a tool, but it was previously being built unconditionally)
* Consolidate building of plClient for multiple platforms


CMakeLists.txt
```diff
@@ -192,11 +192,16 @@ endif(MSVC)
 
 set(PLASMA_SOURCE_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/Sources/Plasma")
 set(PLASMA_TOOLS_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/Sources/Tools")
+set(PLASMA_MAX_PLUGIN_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/Sources/MaxPlugin")
 
 add_subdirectory(Scripts)
 add_subdirectory(Sources/Plasma)
 add_subdirectory(Sources/Tools)
```
dgelessus, line 199: The entire Tools directory can be made conditional now, like the MaxPlugin one:

```suggestion

if(PLASMA_BUILD_TOOLS)
    add_subdirectory(Sources/Tools)
endif()
```

(and then remove the `if(PLASMA_BUILD_TOOLS)` condition in Sources/Tools/CMakeLists.txt)

Sources/Plasma/Apps/CMakeLists.txt
```diff
@@ -3,27 +3,18 @@ include_directories("${PLASMA_SOURCE_ROOT}/FeatureLib")
...
+    add_subdirectory(plUruLauncher)
 endif()
 
 if(PLASMA_BUILD_TOOLS)
-    add_subdirectory(plPythonPack)
-    add_subdirectory(plFileSecure)
-    add_subdirectory(plFileEncrypt)
-    add_subdirectory(plPageInfo)
-    add_subdirectory(plPageOptimizer)
+    add_subdirectory(SoundDecompress)
```
dgelessus, line 19: This is now the only thing outside Sources/Tools that's conditional on `PLASMA_BUILD_TOOLS` - maybe it should use a different option? I don't really have any better ideas though. We probably don't want to build it unconditionally, but does it need its own `PLASMA_BUILD_SOUNDDECOMPRESS` option? I guess we could group it under `PLASMA_BUILD_LAUNCHER`..?

dgelessus (commented):
This is a good reorganization IMO - the reasoning all makes sense to me. The previous split of the tools between Plasma/Apps and Tools always seemed pretty random. The new directory structure also aligns better with the `PLASMA_BUILD_*` CMake options, which I think is a good change.

Sources/Plasma/Apps/CMakeLists.txt
dpogue: I'm definitely open to suggestions here, but don't really have any better ideas myself 😕

Hoikas:
These commits should probably be added to `.git-blame-ignore-revs`

dpogue:
> These commits should probably be added to `.git-blame-ignore-revs`

I can do that, but I'm unclear on whether that's necessary in the case of `git mv` to move files around. The file contents didn't change, and keeping knowledge of the path change seems necessary to handle blaming properly against older versions?

Hoikas (approved):
I don't think there are any objections here, so we should probably take that as approval. Let's get this done before the tools change again, causing conflicts.

PR #1499, opened 2023-10-04
Support XIB compiling for non-Xcode projects

dpogue:
XIB (XML Interface Builder) files describe windows and user interface elements and their properties on macOS (and other Apple platforms), and are authored using Xcode's Interface Builder drag-and-drop tool. As part of compiling an app, those XIB files need to be compiled from XML to a proprietary NIB format and included in the .app bundle as resources.

Xcode projects will handle this automatically when there's an XIB file in the list of files to process, but if you generate Makefiles with CMake then these files don't get compiled properly, and the source XML files get bundled up as resources instead. This leads to errors when launching the application because it can't find the compiled interface definitions.

The Xcode compiler tools provide an `ibtool` command-line tool that can compile XIB files to NIB files, but we need to teach CMake to invoke it. That is what this PR accomplishes.

Aside: We'll eventually have similar issues with Metal shaders, which also need their own compilation process involving special tools, but we can deal with that when we get there.


colincornaby:
Glad someone is looking at this. Thanks!

cmake/XcodeHelpers.cmake
```diff
@@ -0,0 +1,29 @@
...
+    )
+
+    # Xcode will automatically run ibtool to compile the XIB files into NIB
+    # resources, but if we're generating Makefiles or Ninja projects then we
+    # need to handle that ourselves...
+    if (CMAKE_GENERATOR STREQUAL "Xcode")
+        target_sources(${TARGET} PRIVATE ${_pxib_SOURCES})
+    else()
+        find_program(IBTOOL ibtool HINTS "/usr/bin" "${OSX_DEVELOPER_ROOT}/usr/bin")
+        if (${IBTOOL} STREQUAL "IBTOOL-NOTFOUND")
```
Hoikas, line 16:
```suggestion
        if (NOT IBTOOL)
```

or use the `REQUIRED` argument on the `find_program` call

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -109,28 +114,35 @@ elseif(APPLE)
...
 
+plasma_executable(plClient CLIENT INSTALL_PDB
+    SOURCES
+        ${plClient_SOURCES} ${plClient_HEADERS}
+        ${plClient_TEXT} ${plClient_RESOURCES}
+)
+
 if(APPLE)
-    plasma_executable(plClient CLIENT INSTALL_PDB SOURCES ${plClient_SOURCES} ${plClient_HEADERS} ${RESOURCES})
+    target_xib_resources(plClient SOURCES ${plClient_IBSOURCES})
```
Hoikas, line 143: Alternatively, you could process the XIBs in the `plasma_executable` command, which would be more ideal in terms of limiting the platform-specific branches in consuming code.

cmake/XcodeHelpers.cmake
```diff
@@ -0,0 +1,29 @@
...
+            message(SEND_ERROR "Could not find Xcode's ibtool to process .xib files")
+        endif()
+
+        foreach(XIBFILE ${_pxib_SOURCES})
```
Hoikas, line 20:
```suggestion
        foreach(XIBFILE IN LISTS _pxib_SOURCES)
```

Sources/Plasma/Apps/plClient/CMakeLists.txt
dpogue: you mean adding another keyword like `plasma_exectuable(plClient CLIENT INSTALL_PDB SOURCES ... IBSOURCES ...)`? Or trying to parse out `.xib` files from the list of provided sources?

Hoikas: I was thinking more of the latter.

dpogue:
FYI I took a stab at doing this as part of `plasma_executable` on another branch to try it, and it sorta works except now it's bundling both the xib and the nib files into the Resources folder, and we only want the nib compiled output.

I'm wondering if I can somehow set the `MACOSX_PACKAGE_LOCATION` source property on the xibs in a way that excludes them as resources even when they are listed as `RESOURCES` in the target properties...

dpogue:
This should be ready for re-review now. It appears to work both with Xcode (in CI) and with make (locally), as far as the XIB files are concerned.

colincornaby:
It looks like this avoids taking over for the Xcode generator. Has this been tested with the Xcode generator? If it's non disruptive to the Xcode generator, I'd feel better about having this merged in first.

dpogue:
> It looks like this avoids taking over for the Xcode generator. Has this been tested with the Xcode generator?

In the Xcode case, this lets Xcode handle the XIB files the way it normally would. CI shows the nibs being put in the Resources folder as expected.

colincornaby:
I'd be fine with this being merged first then. I feel like there might still be some review left on Metal and I think getting this merged in wouldn't be too painful.

PR #1500, opened 2023-10-08
Fix a couple of endianness issues with `plNetAddress`

dgelessus:
This fixes auth server reconnects for servers that send `Auth2Cli_ServerAddr` messages (i. e. Cyan's server and MOSS). Previously, the received IPv4 address was byte-swapped, making the reconnect always fail.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
@@ -1861,7 +1861,7 @@ bool RecvMsg<Auth2Cli_ServerAddr>(const uint8_t in[], unsigned, void*)
     hsLockGuard(s_critsect);
     if (s_active) {
         s_active->token = msg.token;
-        s_active->addr.SetHost(msg.srvAddr);
+        s_active->addr.SetHost(hsToBE32(msg.srvAddr));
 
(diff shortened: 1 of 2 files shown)


Sources/Plasma/PubUtilLib/plNetGameLib/Private/plNglAuth.cpp
```diff
@@ -1861,7 +1861,7 @@ bool RecvMsg<Auth2Cli_ServerAddr>(const uint8_t in[], unsigned, void*)
     hsLockGuard(s_critsect);
     if (s_active) {
         s_active->token = msg.token;
-        s_active->addr.SetHost(msg.srvAddr);
+        s_active->addr.SetHost(hsToBE32(msg.srvAddr));
```
dpogue, line 1864: I'm guessing we need to byteswap here because the server is (incorrectly) sending the address in LE rather than network order?

dgelessus: Well, it's "correct" in the sense that it's consistent with the rest of the auth server protocol - IPv4 addresses are treated as integers, so they are transmitted little-endian. It's definitely a weird choice though.

An extra twist is that the NetCli receive machinery also does an implicit endianness conversion before the message gets here. So the IPv4 address is received little-endian, then implicitly converted to host byte order, and finally explicitly converted to big-endian. Fun!

dpogue (approved):
This looks okay to me

PR #1501, opened 2023-10-08
Remove a few unneeded `#pragma`s

dgelessus:


Hoikas (approved):

Sources/Plasma/PubUtilLib/plSDL/plStateVariable.cpp
```diff
@@ -1775,9 +1775,6 @@ ST::string plSimpleStateVariable::GetKeyName(int idx) const
     return "(nil)";
 }
 
-#ifdef _MSC_VER
-#   pragma optimize( "g", off )    // disable float optimizations
```
zrax, line 1779: Just for curiosity's sake, do we know why these were added in the first place?  Was there a floating point bug or inaccuracy these were trying to avoid?

zrax (approved):

Sources/Plasma/PubUtilLib/plSDL/plStateVariable.cpp
dgelessus: Yeah, my best guess is that some old version of MSVC would optimize/inline some floating-point operations incorrectly. I don't know for sure though. I'm mostly just assuming that whatever issue it was has been fixed in current MSVC versions.

The MSVC docs for [`/Og` (Global Optimizations)](https://learn.microsoft.com/en-us/cpp/build/reference/og-global-optimizations?view=msvc-170) say that these optimizations are now enabled by default and the option is deprecated. If there were still any issues with it, I hope we would have noticed them by now.

PR #1507, opened 2023-10-17
Fix ahnyQuab ogg decompression crashes.

### LINKED ISSUE #1506: Patcher crashes on OGG files
Patching is sometimes crashing. To reproduce:

- Delete all files in an Uru installation except for the server.ini.
- Run plFilePatcher (or the normal game client) on that directory.
- The game will be redownloaded either completely or partially
- Whichever client is used will throw exceptions within plOGGCodec for various OGG files

It's not clear if this is a server issue or a client issue, but I'm leaning client issue because:
- The build from August 2022 that was shared with Apple DTS is still ok.
- These OGG files have not changed.
- The MD5 hash of the OGG files matches existing downloaded installations.
- The OGG files playback on disk.

This bug was reproduced in ca80c0d8c16daf291ead344de168012b0c0ed259 - and I do not have a good commit hash that works so far.

Hoikas:
This issue was introduced when `hsStream::Close()` was removed in favor of RAII in #1448. Unfortunately, the patcher is handing off files before it deletes the patcher stream, which causes race conditions around zlib decompression not being complete. We need to explicitly close the underlying stream before performing any handoffs to prevent crashes.

Fixes #1506

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -264,6 +264,13 @@ public:
         return retVal;
     }
 
+    void Close()
+    {
+        if (hsCheckBits(fFlags, kFlagZipped))
+            plZlibStream::Close();
+        fOutput.reset();
+    }
+
     uint32_t Write(uint32_t count, const void* buf) override
     {
         // tick whatever progress bar we have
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -264,6 +264,13 @@ class pfPatcherStream : public plZlibStream
         return retVal;
     }
 
+    void Close()
```
dpogue, line 267: is this `override`?

Hoikas: There is no more `virtual Close();` as of #1448.

dpogue: Ah okay

dpogue (approved):

Deledrius (approved):

dpogue:
Can confirm that this allowed me to patch the entire client data set from TrollLand with no issues.

PR #1511, opened 2023-10-20
simple dx9 code cleanups

jfmherokiller:
these changes make the code a bit easier to read at a glance

### DIFF
--- a/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
@@ -9162,8 +9162,7 @@ HRESULT plDXPipeline::ISetShaders(plShader* vShader, plShader* pShader)
         // This is truly obnoxious, but D3D insists that, while using the progammable pipeline,
         // all stages be set up like this, not just the ones we're going to use. We have to
         // do this if we have either a vertex or a pixel shader. See below. Whatever. mf
-        int i;
-        for( i = 0; i < 8; i++ )
+        for(int i = 0; i < 8; i++ )
         {
             fD3DDevice->SetTextureStageState(i, D3DTSS_TEXCOORDINDEX, fLayerUVWSrcs[i] = i);
             fD3DDevice->SetTextureStageState(i, D3DTSS_TEXTURETRANSFORMFLAGS, fLayerXformFlags[i] = 0);
@@ -9189,8 +9188,7 @@ HRESULT plDXPipeline::ISetShaders(plShader* vShader, plShader* pShader)
 
         if( !vShader )
         {
-            int i;
-            for( i = 0; i < 8; i++ )
+            for(int i = 0; i < 8; i++ )
             {
                 fD3DDevice->SetTextureStageState(i, D3DTSS_TEXCOORDINDEX, fLayerUVWSrcs[i] = i);
                 fD3DDevice->SetTextureStageState(i, D3DTSS_TEXTURETRANSFORMFLAGS, fLayerXformFlags[i] = 0);
--- a/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
@@ -10591,8 +10591,8 @@ struct plShadowVertStruct
 void plDXPipeline::IReleaseBlurVBuffers()
 {
     const uint32_t kVSize = sizeof(plShadowVertStruct);
-    int i;
-    for( i = 0; i < kMaxRenderTargetNext; i++ )
+
+    for(int i = 0; i < kMaxRenderTargetNext; i++ )
     {
         if (fBlurVBuffers[i])
         {
--- a/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
@@ -6380,11 +6380,13 @@ void    plDXPipeline::IHandleTextureMode(plLayerInterface* layer)
                 if( fLayerState[0].fBlendFlags & hsGMatState::kBlendNoTexColor )
                     fD3DDevice->SetTextureStageState( 0, D3DTSS_COLOROP, D3DTOP_SELECTARG2 );
                 else
-                    fD3DDevice->SetTextureStageState( 0, D3DTSS_COLOROP, D3DTOP_MODULATE );   
-                fD3DDevice->SetTextureStageState( 0, D3DTSS_COLORARG1, 
-                    fLayerState[0].fBlendFlags & hsGMatState::kBlendInvertColor 
-                        ? D3DTA_TEXTURE | D3DTA_COMPLEMENT 
-                        : D3DTA_TEXTURE);
+                    fD3DDevice->SetTextureStageState( 0, D3DTSS_COLOROP, D3DTOP_MODULATE );
+
+                if (fLayerState[0].fBlendFlags & hsGMatState::kBlendInvertColor) {
+                    fD3DDevice->SetTextureStageState(0, D3DTSS_COLORARG1, D3DTA_TEXTURE | D3DTA_COMPLEMENT);
+                } else {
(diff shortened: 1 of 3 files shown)


colincornaby:
Going to leave some feedback in since I do spend a lot of time reading and doing some work on this pipeline. I'll also add that in since code style is a preference - I'll add my feedback but it's not necessarily representative of everyone.

Two general things:
- Any code changes to this file need to be pretty well reviewed in since this is an active pipeline - and may need to be retested depending on the scope of the changes. So I _personally_ tend to be more skeptical of changes made for style only reasons. The original code might not be messy, but it also is proven to work well. (Unless it hasn't, and in that case then it might need some updating.)
- It looks like a lot more commits have been added to this PR in the time since it was opened. There are a few PRs like that (like the Metal one) - but the commits should probably reach a conclusion or this PR should be brought back to a draft. It's more overhead to keep re-reviewing as commits are added.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -753,7 +753,7 @@ void    plDXPipeline::IInitDeviceState()
     fD3DDevice->SetRenderState( D3DRS_MULTISAMPLEANTIALIAS, ( fSettings.fD3DCaps & kCapsFSAntiAlias ) ? TRUE : FALSE );
     fD3DDevice->SetRenderState( D3DRS_ANTIALIASEDLINEENABLE,        FALSE );
 
-    fD3DDevice->SetRenderState( D3DRS_DITHERENABLE,     FALSE );
+    fD3DDevice->SetRenderState( D3DRS_DITHERENABLE,     TRUE );
```
colincornaby, line 756: Not sure if this is supposed to be here - but this bool has been flipped which changes the rendering features.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -4914,26 +4914,26 @@ void plDXPipeline::IRestoreSpanLights()
...
 
-        fD3DDevice->SetLight( ref->fD3DIndex, &light );
-        ref->fScale = scale;
-    }
+    light.Specular.r *= scale;
+    light.Specular.g *= scale;
+    light.Specular.b *= scale;
+
+    fD3DDevice->SetLight( ref->fD3DIndex, &light );
+    ref->fScale = scale;
```
colincornaby, line 4936: There's a few of these in the source code. I don't know if H'uru has a stance on this. Early return vs single exit is its own debate in industry. Because other code sections also use early return - my hunch is it won't be an issue.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -6410,20 +6414,25 @@ void    plDXPipeline::IHandleTextureMode(plLayerInterface* layer)
...
+            if (fLayerState[0].fBlendFlags & hsGMatState::kBlendInvertAlpha) {
+                fD3DDevice->SetTextureStageState(0, D3DTSS_ALPHAARG1, D3DTA_TEXTURE | D3DTA_COMPLEMENT);
+            } else {
+                fD3DDevice->SetTextureStageState(0, D3DTSS_ALPHAARG1, D3DTA_TEXTURE);
+            }
+            if (fLayerState[0].fBlendFlags & hsGMatState::kBlendInvertVtxAlpha) {
+                fD3DDevice->SetTextureStageState(0, D3DTSS_ALPHAARG2, D3DTA_DIFFUSE | D3DTA_COMPLEMENT);
+            } else {
+                fD3DDevice->SetTextureStageState(0, D3DTSS_ALPHAARG2, D3DTA_DIFFUSE | 0);
+            }
```
colincornaby, line 6435: Going to weigh in on all the ternary operator changes here:
- I think there is a case that nested ternary operators are hard to read and should be split up.
- I think single level ternary operators are ok - and I think that removing them in favor of if statements is a regression.
- I'm still thinking about the if statements. I'm not sure it's totally clear either and I had to check it several times to make sure it was ok. Nested if statements are their own readability problem. It's possible that maybe the second level of the if statement should be a ternary operator.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
```diff
@@ -10584,8 +10591,8 @@ struct plShadowVertStruct
 void plDXPipeline::IReleaseBlurVBuffers()
 {
     const uint32_t kVSize = sizeof(plShadowVertStruct);
-    int i;
-    for( i = 0; i < kMaxRenderTargetNext; i++ )
+
+    for(int i = 0; i < kMaxRenderTargetNext; i++ )
```
colincornaby, line 10595: I wouldn't worry too much about these statements. The possibility of confusion of the reader is low - and this sort of churn is kind of an issue generally for people who want to rebase. But that's just my opinion. Generally it feels like H'uru doesn't take on really low priority code style issues to reduce churn. But that's just my opinion.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXRenderTargetRef.h
```diff
@@ -78,8 +78,19 @@ class plDXRenderTargetRef: public plDXTextureRef
...
+            if constexpr (std::is_same<ColorType, IDirect3DSurface9*>::value) {
+                fD3DColorSurface = color;
+                fD3DTexture = nullptr;
+            }
+            else if constexpr (std::is_same<ColorType, IDirect3DTexture9*>::value) {
+                fD3DTexture = color;
+                fD3DColorSurface = nullptr;
+            }
+            fD3DDepthSurface = depth;
+        }
```
colincornaby, line 93: Is there a reason DepthType is part of the template if there is only one possible DepthType? I'm still thinking over the use of a template here. The double functions were weird, but it's was also nice to have absolute types the arguments needed to be. Even if I had to read the functions a few times.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
jfmherokiller: oh thanks for catching that i had no clue that got lost in my edits

Sources/Plasma/FeatureLib/pfDXPipeline/plDXRenderTargetRef.h
jfmherokiller: simplified the template in the latest commit

Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp
colincornaby: I'll dig in a bit more on why I think the ternary operator changes are a regression here - and why the original author (who I'm guessing is Cyan) probably used them.

Ternary operators and if/then/else are _not_ implicitly the same thing. A ternary operator guarantees an outcome will be reached, while if/then/else does not.

So when I read:
```
fD3DDevice->SetTextureStageState( 0, D3DTSS_ALPHAARG1, 
                fLayerState[0].fBlendFlags & hsGMatState::kBlendInvertAlpha 
                    ? D3DTA_TEXTURE | D3DTA_COMPLEMENT 
                    : D3DTA_TEXTURE);
```

I _know_ SetTextureStageState state will be called with something. It's guaranteed that SetTextureStageState will still be called. It may take me a bit of grok'ing to figure out what the outcome will be, but I know there will be an outcome.

If/then/else loses that guarantee which is probably why it was not originally used. You have the overhead of grok'ing if an if/then/else will always result in an outcome regardless of the input, which is more complicated and introduces higher risk of error. If/then/else also repeats the same statement over and over - which increases risk that if the programmer needs to make changes they might forget to update one of the if/then branches.

I think there are more efficient ways of breaking up the ternary operators. A ternary operator could be split over several lines if it's nested. And as I said in the original comment - I think un-nested ternary operators should not be removed.

jfmherokiller:
i am going close this because i noticed it causes heavy frame drops and i cant fingure out why. And I wont break the pipeline that horrificly unless i have some kind of ultimate fix in the works.

PR #1513, opened 2023-10-20
Move most pnUtils code into more appropriate places

dgelessus:
pnUtils is a pretty random mix of unrelated things. This PR cleans things up a bit by moving code to more appropriate places (plEncryption.h/.cpp into pnEncryption) and removing duplicate APIs (`TimeGetMs` replaced by `hsTimer::GetMilliSeconds`).

The only thing left in pnUtils now is the UTF-16 `StrCopy`. I left that where it is for now, because it's not trivial to replace with `ST::string` or standard C++ APIs and it's still used in a couple of different places. (I could move it into CoreLib, but I don't want *more* weird string functions implicitly pulled in everywhere...)


Sources/Plasma/NucleusLib/pnAsyncCoreExe/Pch.h
```diff
@@ -68,10 +68,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "Private/pnAceInt.h"
 #include "hsThread.h"
+#include "hsTimer.h"
```
Hoikas, line 71: This means that we need to link against `pnNucleusInc`.

dpogue:
> (I could move it into CoreLib, but I don't want _more_ weird string functions implicitly pulled in everywhere...)

CoreLib is where I had begrudgingly settled on `StrCopy` landing when I thought about doing this change myself.

Part of me would like to see plEncryption turned into something more like plRC4Encryption (and clean it up a bit to removed the unimplemented RSA code and maybe turn it into a class instance), but that could be work for another PR.

Hoikas (approved):

PR #1326, opened 2023-02-17
Metal Rendering on macOS

colincornaby:
Update:

Welcome back! Now that the Mac client is merged, we're just down to the Metal bits.

There are still some major things missing from this PR that people should be aware of. This will be a milestone - but not quite the endpoint yet for work.

- Resolution switching does not work in this PR. There is an attempt to implement this in #1244. Some things are going to be wonky because the client doesn't know the current display mode. The Mysterium demo included #1244.
- The Mac client cannot patch itself and uses the Windows patch data. Being dealt with elsewhere.
- There are some build issues that can be worked around on Apple Silicon.
- Cross compiling between architectures still does not work.
- The shaders are direct ports of the HLSL 1.1 assembly. They're not written for readability - they're meant to give the exact same results as DX. I have a branch going with ongoing work to sanitize this code. The code also backs off of exact numerical matching and implements things like fast sin instead of Cyan's sin approximation code. That may not end up as part of this PR.

-----

This PR is very much a draft. Getting this out there so people can look and give early feedback. A few known things:

- Formatting is a bit of a mess. Still looking at Clang format to clean things up. Some of this was written before I understood the projects conventions.
- The Mac client code itself is a bit of a mess. It still needs more cleanup.
- There is a shader cache subsystem that I'm not completely happy with. I haven't quite figured out how to clean it up. It's a giant hash table that takes shader attributes and retrieves/caches the right shader program based on that. I think maybe it could be done smarter. And possibly also have a shared implementation with GL.
- The Mac client can't do executable patching yet.
- There's no launcher with the Mac client. It's all one app. So far - this is intended.
- The Mac client can't patch unless the account has admin privs. There is no privilege escalation built in.
- GPU based character skinning is not yet in this branch (looking more likely I'll pull it in.)
- Resolution controls and frame buffer scaling is not yet in this branch, so Retina support is still shaky. (Parts of this are being considered on the backing-scale-changes branch.

There are probably more things I've forgotten as well. Again - early draft PR.

PR #1515, opened 2023-10-23
this is a set of modifications to modernize parts of the code

jfmherokiller:
feel free to poke around here. this draft is more an experiment then an actual draft atm.

### DIFF
--- a/Sources/Plasma/Apps/plClient/win32/plClient_Win.cpp
+++ b/Sources/Plasma/Apps/plClient/win32/plClient_Win.cpp
@@ -150,7 +150,7 @@ void plClient::InitDLLs()
 {
     hsStatusMessage("Init dlls client\n");
     std::vector<plFileName> dlls = plFileSystem::ListDir("ModDLL", "*.dll");
-    for (auto& dll : dlls)
+    for (const auto& dll : dlls)
     {
         HMODULE hMod = LoadLibraryW(dll.WideString().data());
         if (hMod)
--- a/Sources/Plasma/Apps/plClient/win32/plClient_Win.cpp
+++ b/Sources/Plasma/Apps/plClient/win32/plClient_Win.cpp
@@ -150,9 +150,9 @@ void plClient::InitDLLs()
 {
     hsStatusMessage("Init dlls client\n");
     std::vector<plFileName> dlls = plFileSystem::ListDir("ModDLL", "*.dll");
-    for (auto iter = dlls.begin(); iter != dlls.end(); ++iter)
+    for (auto& dll : dlls)
     {
-        HMODULE hMod = LoadLibraryW(iter->WideString().data());
+        HMODULE hMod = LoadLibraryW(dll.WideString().data());
         if (hMod)
         {
             pInitGlobalsFunc initGlobals = (pInitGlobalsFunc)GetProcAddress(hMod, "InitGlobals");
--- a/Sources/Plasma/Apps/plPythonPack/main.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/main.cpp
@@ -213,9 +213,9 @@ void FindFiles(std::vector<plFileName> &filenames, std::vector<plFileName> &path
     // Get the names of all the python files
     std::vector<plFileName> pys = plFileSystem::ListDir(path, "*.py");
 
-    for (auto iter = pys.begin(); iter != pys.end(); ++iter)
+    for (const auto& py : pys)
     {
-        filenames.push_back(iter->GetFileName());
+        filenames.push_back(py.GetFileName());
         pathnames.push_back(path);
     }
 }
@@ -223,8 +223,9 @@ void FindFiles(std::vector<plFileName> &filenames, std::vector<plFileName> &path
 void FindSubDirs(std::vector<plFileName> &dirnames, const plFileName &path)
 {
     std::vector<plFileName> subdirs = plFileSystem::ListSubdirs(path);
-    for (auto iter = subdirs.begin(); iter != subdirs.end(); ++iter) {
-        ST::string name = iter->GetFileName();
+    for (const auto& subdir : subdirs)
+    {
+        ST::string name = subdir.GetFileName();
         if (s_ignoreSubdirs.find(name) == s_ignoreSubdirs.end())
             dirnames.push_back(name);
     }
@@ -234,15 +235,15 @@ void FindPackages(std::vector<plFileName>& fileNames, std::vector<plFileName>& p
 {
     std::vector<plFileName> packages;
     FindSubDirs(packages, path);
-    for (int i = 0; i < packages.size(); i++)
+    for (const auto& package : packages)
     {
         ST::string packageName;
         if (!parent_package.empty())
             packageName = parent_package + ".";
-        packageName += packages[i].AsString();
+        packageName += package.AsString();
         std::vector<plFileName> packageFileNames;
         std::vector<plFileName> packagePathNames;
-        plFileName packagePath = plFileName::Join(path, packages[i]);
+        plFileName packagePath = plFileName::Join(path, package);
         FindFiles(packageFileNames, packagePathNames, packagePath);
 
         // Check for the magic file to make sure this is really a package...
--- a/Sources/Plasma/Apps/plPythonPack/main.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/main.cpp
@@ -215,7 +215,7 @@ void FindFiles(std::vector<plFileName> &filenames, std::vector<plFileName> &path
 
     for (const auto& py : pys)
     {
-        filenames.push_back(py.GetFileName());
+        filenames.emplace_back(py.GetFileName());
         pathnames.push_back(path);
     }
 }
--- a/Sources/Plasma/Apps/plPythonPack/main.cpp
+++ b/Sources/Plasma/Apps/plPythonPack/main.cpp
@@ -225,7 +225,7 @@ void FindSubDirs(std::vector<plFileName> &dirnames, const plFileName &path)
     std::vector<plFileName> subdirs = plFileSystem::ListSubdirs(path);
     for (const auto& subdir : subdirs)
     {
-        ST::string name = subdir.GetFileName();
+        const ST::string& name = subdir.GetFileName();
         if (s_ignoreSubdirs.find(name) == s_ignoreSubdirs.end())
             dirnames.push_back(name);
     }
--- a/Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
+++ b/Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
@@ -151,9 +151,17 @@ protected:
     bool    ICheckCubicRenderTargets( IDirect3D9 *pD3D, UINT iAdapter, D3DDEVTYPE deviceType, D3DEnum_ModeInfo *modeInfo );
     HRESULT IConfirmDevice( D3DCAPS9* pCaps, DWORD dwBehavior, D3DFORMAT format );
 
-    static const uint8_t kNumDisplayFormats;
-    static const D3DFORMAT kDisplayFormats[];
-
+    static constexpr D3DFORMAT kDisplayFormats[] =
+    {
+            D3DFMT_A1R5G5B5,
+            D3DFMT_A2B10G10R10,
+            D3DFMT_A8R8G8B8,
+            D3DFMT_R5G6B5,
+            D3DFMT_X1R5G5B5,
+            D3DFMT_X8R8G8B8,
+    };
+    static constexpr uint8_t kNumDisplayFormats = std::end(kDisplayFormats) - std::begin(kDisplayFormats);
+    static_assert(kNumDisplayFormats == 6);
 public:
     hsGDirect3DTnLEnumerate();
     virtual ~hsGDirect3DTnLEnumerate();
--- a/Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
+++ b/Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
@@ -160,8 +160,7 @@ protected:
             D3DFMT_X1R5G5B5,
             D3DFMT_X8R8G8B8,
     };
-    static constexpr uint8_t kNumDisplayFormats = std::end(kDisplayFormats) - std::begin(kDisplayFormats);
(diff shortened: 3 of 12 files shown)


jfmherokiller:
using static constexpr seems to incrase compile time but atleast under my tests also provides some marginal speedups

Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
```diff
@@ -151,9 +151,17 @@ class hsGDirect3DTnLEnumerate
...
+    {
+            D3DFMT_A1R5G5B5,
+            D3DFMT_A2B10G10R10,
+            D3DFMT_A8R8G8B8,
+            D3DFMT_R5G6B5,
+            D3DFMT_X1R5G5B5,
+            D3DFMT_X8R8G8B8,
+    };
+    static constexpr uint8_t kNumDisplayFormats = std::end(kDisplayFormats) - std::begin(kDisplayFormats);
+    static_assert(kNumDisplayFormats == 6);
```
zrax, line 164:
```suggestion
    static constexpr uint8_t kNumDisplayFormats = std::size(kDisplayFormats);
```

Sources/Plasma/Apps/plClient/win32/plClient_Win.cpp
```diff
@@ -150,9 +150,9 @@ void plClient::InitDLLs()
 {
     hsStatusMessage("Init dlls client\n");
     std::vector<plFileName> dlls = plFileSystem::ListDir("ModDLL", "*.dll");
-    for (auto iter = dlls.begin(); iter != dlls.end(); ++iter)
+    for (auto& dll : dlls)
```
zrax, line 153:
```suggestion
    for (const auto& dll : dlls)
```

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
```diff
@@ -45,17 +45,18 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
-static const float kAlmostZero = 1.e-5f;
-static const float kPastZero = -kAlmostZero;
-static const float kPastOne = 1.f + kAlmostZero;
-static const float kAlmostOne = 1.f - kAlmostZero;
-static const float kAlmostZeroSquared = kAlmostZero*kAlmostZero;
+static constexpr float kAlmostZero = 1.e-5f;
+static constexpr float kPastZero = -kAlmostZero;
+static constexpr float kPastOne = 1.f + kAlmostZero;
+static constexpr float kAlmostOne = 1.f - kAlmostZero;
+static constexpr float kAlmostZeroSquared = kAlmostZero*kAlmostZero;
```
colincornaby, line 52: I don't think changing a const to a constexpr will have any change on performance. Evaluating at compile time might actually slightly increase memory use because the value might be encoded into the binary multiple times.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
```diff
@@ -45,17 +45,18 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
 static inline hsVector3 Cross(const hsScalarTriple& p0, const hsScalarTriple& p1)
 {
-    return hsVector3(p0.fY * p1.fZ - p0.fZ * p1.fY, 
-                    p0.fZ * p1.fX - p0.fX * p1.fZ, 
-                    p0.fX * p1.fY - p0.fY * p1.fX);
+    return {p0.fY * p1.fZ - p0.fZ * p1.fY, 
+    	       p0.fZ * p1.fX - p0.fX * p1.fZ, 
+    	       p0.fX * p1.fY - p0.fY * p1.fX
+    };
```
colincornaby, line 59: This should probably remain an `hsVector3` for type safety and clarity. I don't think there's any performance impact here.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXTextFont.cpp
```diff
@@ -63,10 +63,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 //// Local Stuff //////////////////////////////////////////////////////////////
 
-static const long PLD3D_FONTFVF = D3DFVF_XYZ | D3DFVF_DIFFUSE 
+static constexpr long PLD3D_FONTFVF = D3DFVF_XYZ | D3DFVF_DIFFUSE 
                                 | D3DFVF_TEX1 | D3DFVF_TEXCOORDSIZE3(0);
 
-static D3DMATRIX d3dIdentityMatrix{ 1.0f, 0.0f, 0.0f, 0.0f,
+static constexpr D3DMATRIX d3dIdentityMatrix{ 1.0f, 0.0f, 0.0f, 0.0f,
```
colincornaby, line 69: I don't know if the identity matrix should be constexpr. Something being both static and constexpr may already be a bit of a paradox. But the identity matrix is loaded as a buffer from a pointer, which means it can't actually be constexpr easily by a compiler. And it might be more optimal just to have one copy of the buffer in memory anyway.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
jfmherokiller: fair enough i just found it as duplicating the type since the function already returns it.

colincornaby: Always better to use the defined constructor for an object instead of pushing in memory directly. Even if it's a bit duplicative with the type name.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
jfmherokiller: constexpr in this case only guarantees that the value is computed at compile time instead of runtime. And since its declared static it will only be compiled once.

colincornaby: constexpr doesn't actually guarantee that it will be evaluated at compile time. Like inline, it's a hint that the compiler may ignore.

Here's it's just a little strange because all these numbers are constants. There's nothing that could be done at runtime.  `1.f + kAlmostZero` is a constant value - not something that gets done at runtime.

colincornaby: Sorry for the extra comment - was thinking about it this morning a bit more...

Maybe a better way to phrase my question: In since there is a performance claim here - was the previous way identity matrices and numerical constants were defined not compile time constant? The reason the performance is important here is because if there is an issue in the way we're defining our matrices and numerical constants - we need to fix that across the engine.

If constexpr means the same thing as const in this context - that doesn't mean I'd ask for it to be changed back. But I want to make sure we understand what is going on here for future work and PRs.

jfmherokiller: i will admit this is more a semantic change then a functional one as far as goldbolt is concerned.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
colincornaby: @dgelessus - I see your emoji. I'm poking at this and I see C++ is policing the number of elements in the struct.  However - this struct does have a defined constructor and it does seem weird to go around it.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
colincornaby: My understanding of constexpr is it usually only adds behavior to constexpr functions. The changes in the iterator are actually a really good example of constexpr functions.

In C - unless the compiler is just very strange - numeric constants and C arrays should evaluate at compile time. C arrays are just buffers so they'll be stored in the binary and loaded into memory with the binary. Numerics should be computed at compile time. Something like 'const float halfPi = M_PI/2.f` will evaluate at compile time.

Just be aware if you're looking for performance that there is a set of rules about whats precompiled that predates constexpr. Constexpr implies const - so it hopefully shouldn't hurt anything here or in the identity matrices. But under typical C compiler rules this should all be evaluated at compile time.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
dgelessus: Ah, I wanted to explain later and forgot, sorry. As I understand it, the brace syntax in this case *will* use the constructor defined in the class, despite the brace syntax. According to cppreference, the C-style field by field initialization is called [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization), which never applies to classes with user-defined constructors. Instead, the brace syntax does [list initialization](https://en.cppreference.com/w/cpp/language/list_initialization) here, which in this case calls the normal constructor, exactly like the previous code.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXTextFont.cpp
Hoikas: `static constexpr` is perfectly valid. Sometimes you have things that can be used at both compile-time and runtime. In this case, we aren't doing any compile-time matrix math, so it probably makes more sense to just leave this one alone. I doubt that the generated machine code is any different since we're just going to be pushing a pointer onto the stack.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
Hoikas: Changing these to `static constexpr` is ideal.

Hoikas:
You may revert or remove [843c93b](https://github.com/H-uru/Plasma/pull/1515/commits/843c93b59243e2f750ca319b52d7324909855381)

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
colincornaby: @dgelessus Thanks! I'll review the docs.

jfmherokiller:
dropped the commit and with that i think ill move to more cleanups in other places

Sources/Plasma/FeatureLib/pfDXPipeline/plDXTextFont.cpp
jfmherokiller: i can revert this if desired

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -213,18 +213,19 @@ void FindFiles(std::vector<plFileName> &filenames, std::vector<plFileName> &path
     // Get the names of all the python files
     std::vector<plFileName> pys = plFileSystem::ListDir(path, "*.py");
 
-    for (auto iter = pys.begin(); iter != pys.end(); ++iter)
+    for (const auto& py : pys)
     {
-        filenames.push_back(iter->GetFileName());
+        filenames.push_back(py.GetFileName());
```
Hoikas, line 218:
```suggestion
        filenames.emplace_back(py.GetFileName());
```

While we're here, let's make this more technically correct.

Sources/Plasma/Apps/plPythonPack/main.cpp
```diff
@@ -213,18 +213,19 @@ void FindFiles(std::vector<plFileName> &filenames, std::vector<plFileName> &path
...
 }
 
 void FindSubDirs(std::vector<plFileName> &dirnames, const plFileName &path)
 {
     std::vector<plFileName> subdirs = plFileSystem::ListSubdirs(path);
-    for (auto iter = subdirs.begin(); iter != subdirs.end(); ++iter) {
-        ST::string name = iter->GetFileName();
+    for (const auto& subdir : subdirs)
+    {
+        ST::string name = subdir.GetFileName();
```
Hoikas, line 228:
```suggestion
        const ST::string& name = subdir.GetFileName();
```

Let's defer the copy until we're really sure that we need it.

Sources/Plasma/PubUtilLib/plMath/plTriUtils.cpp
Hoikas, line 59: Please fix the indentation here.

Sources/Tools/plFilePatcher/main.cpp
```diff
@@ -55,7 +55,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #if HS_BUILD_FOR_WIN32
 #   include "hsWindows.h"
 #   include <io.h>
-#   define isatty _isatty
+#   define ttycheck(fileinfo) (GetFileType(fileinfo) == FILE_TYPE_CHAR)
```
Hoikas, line 58: We prefer to use standards compliant mechanisms. AFAIK, `isatty` is part of the POSIX standard. The Windows SDK just emits an erroneous deprecation warning with `isatty` and wants `_isatty` (which is just the Microsoft way of making life more difficult for cross-platform development). Is there a particular bug that is fixed by using `(GetFileType(fileinfo) == FILE_TYPE_CHAR)` instead of `_isatty`?

Hoikas (commented):
This PR covers an odd set of topics. I think we would prefer to see changesets grouped by topic where possible.

Sources/Tools/plFilePatcher/main.cpp
jfmherokiller: yes it does if the user attempts compile and run the game using msys2 or cygwin the value of _isatty can be invalid.

Hoikas: In that case, you should probably fix the preprocessor checks to detect that situation to use `isatty`.

jfmherokiller: shouldnt HS_BUILD_FOR_WIN32 handle that?

dpogue: `HS_BUILD_FOR_WIN32` will be defined for MSVC, clang-cl, and mingw

In this case, you'll need an extra check for mingw to avoid aliasing isatty

Hoikas: This still seems to be using non-standards complaint naming, unfortunately. We prefer the POSIX standard `isassty`.

Hoikas:
Due to the changes to pfDXPipeline, this will be deferred until after #1326 is merged.

jfmherokiller:
fine by me I just hope the mac port lands successfully

PR #1518, opened 2023-11-01
modernize shader assembler

jfmherokiller:
decided to clean up the shader assember a bit.

### DIFF
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -250,8 +250,7 @@ int main(int argc, char* argv[])
     }
 
     try {
-	    for (int i = 0; i < numNames; i++ )
-	    {
+	    for (int i = 0; i < numNames; i++) {
 		    plDXShaderAssembler ass;
 		    IAssShader(ass, nameList[i].data());
         }
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -101,7 +101,7 @@ void ICreateHeader(const ST::string& varName, const plFileName& fileName, FILE* 
     hsSsize_t byteLen = shader->GetBufferSize();
     hsSsize_t quadLen = byteLen >> 2;
 
-    auto* codes = static_cast<unsigned char*>(shader->GetBufferPointer());
+    auto* codes = reinterpret_cast<unsigned char*>(shader->GetBufferPointer());
 
     ST::printf(fp, "static const uint32_t {}byteLen = {};\n\n", varName, byteLen);
     ST::printf(fp, "static const uint8_t {}Codes[] = {{\n", varName);
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -181,7 +181,7 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
 // vs_ => "vs_
 // ps_ => "ps_
 // , => ",
-static constexpr std::string_view kEnumNames[] = {
+static constexpr std::string_view kEnumNames[] {
     "vs_WaveFixedFin6",
     "ps_WaveFixed",
     "vs_CompCosines",
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -161,7 +161,8 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
         return;
 
     hsUNIXStream inFp;
-    if (AttemptToOpenInputFile(inFile, inFp)) return;
+    if (AttemptToOpenInputFile(inFile, inFp))
+        return;
 
     uint32_t shaderCodeLen = inFp.GetEOF();
     auto shaderCode = std::make_unique<char[]>(shaderCodeLen);
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -207,20 +207,38 @@ static constexpr std::string_view kEnumNames[] = {
     "vs_GrassShader",
     "ps_GrassShader"
 };
+int main(int argc, char* argv[])
+{
+    std::vector<std::string_view> allArgs(argv, argv + argc);
+    //remove command
+    allArgs.erase(allArgs.begin());
 
+    if( allArgs.empty() )
+    {
+        ST::printf("{} <file0> <file1> ...\n", argv[0]);
+        return 0;
+    }
+
+
+    const std::string_view* nameList = nullptr;
+    int numNames = 0;
+    if (allArgs.at(0).compare("all") == 0)
+    {
         nameList = kEnumNames;
         numNames = std::size(kEnumNames);
     }
     else
     {
-        nameList = argv+1;
-        numNames = argc-1;
+
+        nameList = allArgs.data();
+        numNames = allArgs.size();
     }
 
     try {
-        plDXShaderAssembler ass;
-        for (int i = 0; i < numNames; i++ ) {
-            IAssShader(ass, nameList[i]);
+	    for (int i = 0; i < numNames; i++ )
+	    {
+		    plDXShaderAssembler ass;
+		    IAssShader(ass, nameList[i].data());
         }
     } catch (const plDXShaderError& error) {
         fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
--- a/Sources/Tools/plShaderAssembler/main.cpp
+++ b/Sources/Tools/plShaderAssembler/main.cpp
@@ -14,7 +14,7 @@
 #include <string_theory/format>
 #include <string_theory/stdio>
 
-class plDXShaderError : std::exception
+class plDXShaderError
 {
     ST::string fError;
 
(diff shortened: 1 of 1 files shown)


Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -83,9 +83,9 @@ std::shared_ptr<ID3DBlob> plDXShaderAssembler::AssShader(const char* shader, uns
     ID3DBlob* errors = nullptr;
     hsCOMError hr = fFuncPtr(shader, shaderLen, nullptr, nullptr, nullptr, 0, &compiled, &errors);
     if (SUCCEEDED(hr))
-        return std::shared_ptr<ID3DBlob>(compiled, IRelease);
+        return {compiled, IRelease};
 
-    ST::string strErrors(errors ? reinterpret_cast<const char*>(errors->GetBufferPointer()) : "Unspecified");
+    ST::string strErrors(errors ? static_cast<const char*>(errors->GetBufferPointer()) : "Unspecified");
```
Hoikas, line 88: The previous cast looks correct to me.

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -133,21 +155,12 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
...
     ST::printf("Processing {} into {}\n", name, outFile);
+
     hsUNIXStream outFp;
-    if (!outFp.Open(outFile, "w")) {
-        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
-        ST::printf(stderr, "Error opening file {} for output\n", outFile);
-        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
-        return;
-    }
+    if (AttemptToOpenOutFile(outFile, outFp)) return;
```
Hoikas, line 160: Plasma Code Style prohibits this formatting.

```suggestion
    if (AttemptToOpenOutFile(outFile, outFp))
        return;
```

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -133,21 +155,12 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
...
 
     hsUNIXStream inFp;
-    if (!inFp.Open(inFile, "r")) {
-        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
-        ST::printf(stderr, "Error opening file {} for input\n", inFile);
-        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
-        return;
-    }
+    if (AttemptToOpenInputFile(inFile, inFp)) return;
```
Hoikas, line 163:
```suggestion
    if (AttemptToOpenInputFile(inFile, inFp))
        return;
```

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -117,14 +117,36 @@ void ICreateHeader(const ST::string& varName, const plFileName& fileName, FILE*
...
+	if (!outFp.Open(outFile, "w")) {
+		fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
+		ST::printf(stderr, "Error opening file {} for output\n", outFile);
+		fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
+		return true;
+	}
+	return false;
+}
+
+static bool AttemptToOpenInputFile(plFileName inFile, hsUNIXStream& inFp)
```
Hoikas, line 139:
```suggestion
static bool AttemptToOpenInputFile(const plFileName& inFile, hsUNIXStream& inFp)
```

No need to copy the file name. For bonus modernization, instead of returning a success value and taking a result reference, you could return an `std::optional<hsUNIXStream>`.

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -117,14 +117,36 @@ void ICreateHeader(const ST::string& varName, const plFileName& fileName, FILE*
...
-    fputs("\n\n", fp);
+    ST::printf(fp, "\t];");
+    ST::printf(fp, "\n\n");
 
     ST::printf(fp, "static const plShaderDecl {}Decl(\"{}\", {}, {}byteLen, {}Codes);\n\n",
                varName, fileName, varName, varName, varName);
     ST::printf(fp, "static const plShaderRegister {}Register(&{}Decl);\n\n", varName, varName);
 }
 
+static bool AttemptToOpenOutFile(plFileName outFile, hsUNIXStream& outFp)
```
Hoikas, line 128:
```suggestion
static bool AttemptToOpenOutFile(const plFileName& outFile, hsUNIXStream& outFp)
```

No need to copy the file name. For bonus modernization, instead of returning a success value and taking a result reference, you could return an `std::optional<hsUNIXStream>`.

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -117,14 +117,36 @@ void ICreateHeader(const ST::string& varName, const plFileName& fileName, FILE*
...
+static bool AttemptToOpenInputFile(plFileName inFile, hsUNIXStream& inFp)
+{
+	if (!inFp.Open(inFile, "r")) {
+		fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
+		ST::printf(stderr, "Error opening file {} for input\n", inFile);
+		fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
+		return true;
+	}
+	return false;
+}
```
Hoikas, line 148: This doesn't really feel like an improvement because you've basically just copied and pasted the code to another location and caused some extra pass-by-value copying. I think it would be better if you condensed these to one function.

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -162,78 +175,83 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
         fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
     }
 }
-
+// Just copy in the enum and use Replace on
+// vs_ => "vs_
+// ps_ => "ps_
+// , => ",
+static constexpr std::string_view kEnumNames[] = {
```
Hoikas, line 182:
```suggestion
static constexpr std::string_view kEnumNames[] {
```

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -162,78 +175,83 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
...
+    "ps_GrassShader"
+};
 int main(int argc, char* argv[])
 {
-    if( argc < 2 )
+    std::vector<std::string_view> allArgs(argv, argv + argc);
+    //remove command
+    allArgs.erase(allArgs.begin());
+
+    if( allArgs.empty() )
```
Hoikas, line 229: This honestly just makes things seem more complicated to me...

Sources/Tools/plShaderAssembler/main.cpp
```diff
@@ -162,78 +175,83 @@ static void IAssShader(const plDXShaderAssembler& ass, const char* name)
...
+        nameList = allArgs.data();
+        numNames = allArgs.size();
     }
 
     try {
-        plDXShaderAssembler ass;
-        for (int i = 0; i < numNames; i++ ) {
-            IAssShader(ass, nameList[i]);
+	    for (int i = 0; i < numNames; i++ )
+	    {
```
Hoikas, line 252:
```suggestion
	    for (int i = 0; i < numNames; i++) {
```

Sources/Tools/plShaderAssembler/main.cpp
jfmherokiller: thanks for the idea i merged them in the latest commit

Sources/Tools/plShaderAssembler/main.cpp
jfmherokiller: attempted to have the function an std::optional like so

```
std::optional<hsUNIXStream> AttemptToOpenFile(const plFileName& file) {
    hsUNIXStream inFp;
    if (!inFp.Open(file, "rw")) {
        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
        ST::printf(stderr, "Error opening file {}\n", file);
        fputs("%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%\n", stderr);
        return std::nullopt; 
    }
    return inFp;
}
```
but the class hsUNIXStream doesnt seem to allow it.

Sources/Tools/plShaderAssembler/main.cpp
jfmherokiller: renamed it in a later commit, but its use is so that the argument array is of the same type as the constexpr

Sources/Tools/plShaderAssembler/main.cpp
Hoikas: You may have to write a move constructor for `hsUNIXStream`.

dgelessus: Other functions that return streams currently use `std::unique_ptr<hsStream>` or similar, which doesn't require move support.

Hoikas: That's true, but here @jfmherokiller's snippet is moving an `hsUNIXStream` constructed on the stack into storage owned by the returned `std::optional<hsUNIXStream>`.

jfmherokiller: using `std::optional<hsUNIXStream> inFp;.` seems very odd to me from a readability standpoint.

PR #1520, opened 2023-11-03
Enable macOS arm64 builds in CI

dpogue:
This requires using the host platform's Python interpreter rather than the target platform's interpreter, which runs as part of the build process for resource.dat generation.

Also have to disable running tests when cross-compiling because the host system can't run the test executables compiled for the target system.

### DIFF
--- a/Scripts/Ports/python3/vcpkg-cmake-wrapper.cmake
+++ b/Scripts/Ports/python3/vcpkg-cmake-wrapper.cmake
@@ -59,7 +59,7 @@ if(_PythonFinder_WantLibs)
         find_program(
             @PythonFinder_PREFIX@_EXECUTABLE
             NAMES "python" "python@PYTHON_VERSION_MAJOR@.@PYTHON_VERSION_MINOR@"
-            PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/tools/python3"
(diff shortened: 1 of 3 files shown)


Hoikas (approved):

Scripts/Ports/python3/vcpkg-cmake-wrapper.cmake
```diff
@@ -59,7 +59,7 @@ if(_PythonFinder_WantLibs)
         find_program(
             @PythonFinder_PREFIX@_EXECUTABLE
             NAMES "python" "python@PYTHON_VERSION_MAJOR@.@PYTHON_VERSION_MINOR@"
-            PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_TARGET_TRIPLET}/tools/python3"
+            PATHS "${_VCPKG_INSTALLED_DIR}/${VCPKG_HOST_TRIPLET}/tools/python3"
```
Hoikas, line 62: Actually, now that I'm thinking about it... We need to bump the port-version in vcpkg.json to ensure that this gets deployed.

Hoikas (changes requested):

Scripts/Ports/python3/vcpkg-cmake-wrapper.cmake
dpogue: Good catch! Pushed a fix

Hoikas (approved):

PR #1522, opened 2023-11-05
Set custom names for all of our threads

dgelessus:
This is helpful when debugging, especially with all of the IO worker threads created by the asio-based socket code.

The APIs for setting thread names are really platform-specific, so this adds a wrapper function that supports Windows, Linux, and Apple platforms. (Only the Windows version has been tested so far.)

### DIFF
--- a/Sources/Plasma/CoreLib/hsThread_Unix.cpp
+++ b/Sources/Plasma/CoreLib/hsThread_Unix.cpp
@@ -83,10 +83,7 @@ void hsThread::SetThisThreadName(const ST::string& name)
     hsAssert(res == 0, "Failed to set thread name");
 #elif defined(HS_BUILD_FOR_LINUX)
     // On Linux, thread names must fit into 16 bytes, including the terminator.
-    char buf[16];
-    strncpy(buf, name.c_str(), sizeof(buf));
-    buf[sizeof(buf) - 1] = '\0';
-    int res = pthread_setname_np(pthread_self(), buf);
+    int res = pthread_setname_np(pthread_self(), name.left(15).c_str());
     hsAssert(res == 0, "Failed to set thread name");
 #endif
     // Because this is just a debugging help, do nothing by default (sorry, BSDs).
--- a/Sources/Plasma/CoreLib/hsThread_Unix.cpp
+++ b/Sources/Plasma/CoreLib/hsThread_Unix.cpp
@@ -75,6 +75,23 @@ int clock_gettime(int clocktype, struct timespec* ts)
 
 /////////////////////////////////////////////////////////////////////////////
 
+void hsThread::SetThisThreadName(const ST::string& name)
+{
+#if defined(HS_BUILD_FOR_APPLE)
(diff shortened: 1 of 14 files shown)


Sources/Plasma/CoreLib/hsThread_Unix.cpp
```diff
@@ -75,6 +75,23 @@ int clock_gettime(int clocktype, struct timespec* ts)
...
+    // The Apple version doesn't take a thread argument and always operates on the current thread.
+    int res = pthread_setname_np(name.c_str());
+    hsAssert(res == 0, "Failed to set thread name");
+#elif defined(HS_BUILD_FOR_LINUX)
+    // On Linux, thread names must fit into 16 bytes, including the terminator.
+    char buf[16];
+    strncpy(buf, name.c_str(), sizeof(buf));
+    buf[sizeof(buf) - 1] = '\0';
+    int res = pthread_setname_np(pthread_self(), buf);
+    hsAssert(res == 0, "Failed to set thread name");
```
jfmherokiller, line 90: could a compiler assert by thrown if the name is longer then the limited number of charters?

dgelessus: Good question, I'm not really sure. But in the general case, we don't want that to be a hard error - e. g. code that's specific to Windows or Apple can use longer thread names without issues.

jfmherokiller: well i was thinking of some kind of static_assert that possibly called on a macro imbeded in the headers if such macro exists. 

also the static assert would pretty much be ifdefed to linux

jfmherokiller: After doing some research it seems like the limit is imposed by the linux kernel itself and requires a complete recompilation to change.

Sources/Plasma/CoreLib/hsThread_Unix.cpp
```diff
@@ -75,6 +75,23 @@ int clock_gettime(int clocktype, struct timespec* ts)
...
+#if defined(HS_BUILD_FOR_APPLE)
+    // The Apple version doesn't take a thread argument and always operates on the current thread.
+    int res = pthread_setname_np(name.c_str());
+    hsAssert(res == 0, "Failed to set thread name");
+#elif defined(HS_BUILD_FOR_LINUX)
+    // On Linux, thread names must fit into 16 bytes, including the terminator.
+    char buf[16];
+    strncpy(buf, name.c_str(), sizeof(buf));
+    buf[sizeof(buf) - 1] = '\0';
+    int res = pthread_setname_np(pthread_self(), buf);
```
Hoikas, line 89:
```suggestion
    int res = pthread_setname_np(pthread_self(), name.left(15).c_str());
```

dgelessus: Right, this isn't `wchar_t`, so we can do that 😄

Hoikas (approved):

PR #1523, opened 2023-11-05
Mac OS X 10.5 compiler fixes

dpogue:
This is not a full set of changes to make everything build on 10.5 PowerPC (because everything does not build on 10.5 PowerPC) but this is a hopefully uncontroversial set of small fixes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,10 @@ if(APPLE)
         if (PLASMA_MAC_UNIVERSAL)
             set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
         endif()
+
+        if(CMAKE_SYSTEM_VERSION VERSION_LESS 10) # Darwin 10 == Mac OS X 10.6
+            set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+        endif()
     endif(CMAKE_SYSTEM_NAME MATCHES "Darwin")
 endif(APPLE)
 
--- a/Sources/Plasma/CoreLib/CMakeLists.txt
+++ b/Sources/Plasma/CoreLib/CMakeLists.txt
@@ -21,7 +21,7 @@ set(CoreLib_SOURCES
     plLoadMask.cpp
     plProduct.cpp
     plViewTransform.cpp
-    hsWindows.cpp
+    $<$<PLATFORM_ID:Windows>:hsWindows.cpp>
 )
 
 if(CMAKE_USE_WIN32_THREADS_INIT)
--- a/Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
+++ b/Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
@@ -43,6 +43,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <cmath>
 #include <string_theory/format>
 
+#ifdef HS_BUILD_FOR_APPLE
+#   include <AvailabilityMacros.h>
+#   include <sys/time.h>
+#endif
+
 #include "plUnifiedTime.h"
 
 #include "hsStream.h"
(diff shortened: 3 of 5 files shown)


Deledrius (approved):

CMakeLists.txt
```diff
@@ -60,6 +60,10 @@ if(APPLE)
         if (PLASMA_MAC_UNIVERSAL)
             set(CMAKE_OSX_ARCHITECTURES "$(ARCHS_STANDARD)")
         endif()
+
+        if(CMAKE_SYSTEM_VERSION VERSION_LESS 10)
```
dgelessus, line 64: What is "10" here? Does `CMAKE_SYSTEM_VERSION` use the Darwin kernel version number (i. e. "10" is Mac OS X 10.6 Snow Leopard)?

Sources/Plasma/CoreLib/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ set(CoreLib_SOURCES
     plLoadMask.cpp
     plProduct.cpp
     plViewTransform.cpp
-    hsWindows.cpp
+    $<$<PLATFORM_ID:Windows>:hsWindows.cpp>
```
dgelessus, line 24: In that case, the `#ifdef HS_BUILD_FOR_WIN32` ... `#endif` should be removed from hsWindows.cpp.

Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
```diff
@@ -107,10 +112,22 @@ void plUnifiedTime::ToCurrentTime()
     struct timespec ts;
 
 #if defined(HS_BUILD_FOR_APPLE)
-    // timespec_get is only supported since macOS 10.15,
-    // but clock_gettime exists since macOS 10.12.
-    int res = clock_gettime(CLOCK_REALTIME, &ts);
-    hsAssert(res == 0, "clock_gettime failed");
+#   ifdef HAVE_BUILTIN_AVAILABLE
+    if (__builtin_available(macOS 10.12, *)) {
```
dgelessus, line 116: Since we don't actually need nanosecond precision, you can just use `gettimeofday` unconditionally and skip the availability check. https://github.com/H-uru/Plasma/pull/1396#discussion_r1208455783

CMakeLists.txt
dpogue: Yes, it's the Darwin version... maybe a comment is useful here

Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
dpogue: There are a few sources (including code-scanning tools) asserting that `gettimeofday` is semi-deprecated and should not be used in new code (but is part of the POSIX standard so it won't actually get removed). Since we already have the `clock_gettime` implementation and it's easy to guard, it seems worth keeping for newer systems.

dgelessus: Ugh... In that case, could we at least remove the `clock_gettime` branch? If we have to add an availability check anyway, we can use `timespec_get` where available, and fall back to `gettimeofday` for less recent macOSes/SDKs.

dpogue: Sure. Remind me, `timespec_get` is 10.15 and newer?

dgelessus: Yep:

https://github.com/H-uru/Plasma/blob/1c4209cb2e5b5852bfec11d0cab7a5be6eaf72d2/Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp#L110-L111

dgelessus (approved):

Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
```diff
@@ -107,10 +112,21 @@ void plUnifiedTime::ToCurrentTime()
     struct timespec ts;
 
 #if defined(HS_BUILD_FOR_APPLE)
-    // timespec_get is only supported since macOS 10.15,
-    // but clock_gettime exists since macOS 10.12.
-    int res = clock_gettime(CLOCK_REALTIME, &ts);
-    hsAssert(res == 0, "clock_gettime failed");
+#   if defined(HAVE_BUILTIN_AVAILABLE) && MAC_OS_X_VERSION_MIN_REQUIRED >= 101500
```
Hoikas, line 115:
```suggestion
#if defined(HAVE_BUILTIN_AVAILABLE) && MAC_OS_X_VERSION_MIN_REQUIRED >= 101500
```

Sources/Plasma/PubUtilLib/plUnifiedTime/plUnifiedTime.cpp
```diff
@@ -107,10 +112,21 @@ void plUnifiedTime::ToCurrentTime()
...
+    if (__builtin_available(macOS 10.15, *)) {
+        // timespec_get is only supported since macOS 10.15
+        int res = timespec_get(&ts, TIME_UTC);
+        hsAssert(res != 0, "timespec_get failed");
+    } else
+#   endif
```
Hoikas, line 121:
```suggestion
#endif // defined(HAVE_BUILTIN_AVAILABLE) && MAC_OS_X_VERSION_MIN_REQUIRED >= 101500
```

Hoikas (approved):

PR #1526, opened 2023-11-14
Update macOS iconset

patmauro:
Makes a few improvements to the macOS iconset for readability:
- Reverts to original Cyan book graphic(s), rather than our upscaled version, at sizes 256 and below, as these original graphics are _already_ optimized to read best at their respective sizes.
- The smallest icon (16x16) uses a special variant with a much bigger book against a simple gradient rather than a smushed re-use of the same background. This will read much better as an icon at this size.
- The 3 largest icons in the icon set are unchanged, and thus are not included in this PR.

Note: I noticed that a 512px icon a little more accurate to Cyan's original asset may be possible to generate with a little work. However, this doesn't seem strictly necessary at this time; the assets we have in place are more than sufficient. But I may take this as a follow-up item in the future.


colincornaby:
These changes look good. I especially like the extra detail on the smaller icon.

colincornaby (approved):

colincornaby:
@patmauro - Hey, sorry to dig up an old PR. Do you by chance have a version of this icon without the round rect frame? (I.E. just a giant edge to edge image.) I need a 1024x1024 one for iOS. iOS doesn't include the rounded edges.

patmauro:
> @patmauro - Hey, sorry to dig up an old PR. Do you by chance have a version of this icon without the round rect frame? (I.E. just a giant edge to edge image.) I need a 1024x1024 one for iOS. iOS doesn't include the rounded edges.

Should still have all the source files, yeah - only thing is I'm not 100% sure there's a version of the book asset that's large enough to fill edgeless 1024x1024. Let me see what we can do.

patmauro:
Note: With macOS Tahoe on the way, the dreaded "full redesign of macOS design language for no reason" is indeed on the horizon - however, luckily I think we are still in compliance here, since our icon doesn't feature any of the "external" elements that are now discouraged, so it _should_ cleanly adapt to the new icon style. That being said, let's keep an eye on this as we get closer to the Tahoe release and see if anything needs to be re-assessed here.

dpogue:
If needed, we can take the iOS Dark Mode icon (which is just the book with transparency) and drop that in as to Icon Composer as a layer. That gives us a semi-decent glass icon:

Icon-iOS-Default-83 5x83 5@2x Icon-iOS-Dark-83 5x83 5@2x Icon-iOS-ClearLight-83 5x83 5@2x Icon-iOS-ClearDark-83 5x83 5@2x

Just make sure to disable the glass effects on the book otherwise it looks awful 😛

colincornaby:
I have built and run Plasma on Tahoe - and verified the icon does get the automatic effect.

I tinkered with it in Icon Composer and I don't know if I have any more thoughts right now. I looked at keeping the background for the normal icon but maybe using a different treatment on the rest.

Happy to see brainstorming starting though. It would be nice to get a new icon out as we keep the Mac client updated.

PR #1527, opened 2023-11-14
Handle xcasset files when building outside of Xcode

dpogue:
Yes, this somehow manages to be worse than the XIB files because actool generates a partial plist output file that needs to be merged with the application plist in order to provide the correct icon.


Hoikas (approved):

Hoikas:
Does this look ok, @colincornaby?

colincornaby:
This looks fine.

colincornaby (approved):

PR #1529, opened 2023-11-22
Bump vcpkg to 2023.11.20

dpogue:
Also fix the Windows Python build after doing so.

## Updated Dependencies

* **asio**
    `1.28.0#1` -> `1.28.2`

* **curl**
    `8.2.1` -> `8.4.0`

* **fontconfig**
    `2.14.2` -> `2.14.2#1`

* **freetype**
    `2.12.1#3` -> `2.12.1#4`

* **libjpeg-turbo**
    `2.1.5.1#2` -> `3.0.1`

* **libvpx**
    `1.12.0#1` -> `1.13.1`

* **python3**
    `3.10.7#8` -> `3.10.7#9`
    Just a config change to fix Windows linking with the newer version of libffi in vcpkg.

* **zlib**
    `1.2.13` -> `1.3`


dpogue:
Weird that macOS fontconfig is failing here, because this branch built successfully in my fork... I'll try to investigate further 😕

dpogue:
This should be good to go now

Hoikas:
`distutils` was removed in Python 3.12....

dpogue:
> `distutils` was removed in Python 3.12....

yeah, and it looks like the macOS runners are updated to Python 3.12, but something in fontconfig's build process requires it 😕

Hoikas:
Should probably use actions/setup-python to force Python 3.11

Hoikas:
Well, that seems to have broken the Linux build. Fantastic. I guess installing the setuptools package from pip will be ok.

dpogue:
`setup-python` is probably doing something to inject itself forcefully into the path, and then interfering with finding the vcpkg python 🤦🏼‍♂️

Hoikas:
Probably so. That's arguably a vcpkg bug (or a bug in the python3 port) because vcpkg is *supposed* to insulate itself from the environment. <_<

Hoikas (approved):

PR #1530, opened 2023-11-29
Shader sanitization

colincornaby:
Plasma's shaders were written in assembly - which also carried to the Metal implementation.

Goals in this branch:

- The shaders should be written in clear code with proper variable names and functions where applicable.
- Algorithms should be documented.
- Any fast math assembly should be replaced with modern fast math functions where available. It is ok to drift from the HLSL 1.1 fast math code where imperceptible.
- Shaders should be clearly reviewable and easy to port to other languages.

For water shaders in particular, Chapter 1 of GPU Gems is a good reference:
https://developer.nvidia.com/gpugems/gpugems/part-i-natural-effects/chapter-1-effective-water-simulation-physical-models

There is a C code version of a water shader available here:
https://http.download.nvidia.com/developer/GPU_Gems/CD_Image/Index.html

Please note the shader supplied in GPU Gems does not appear to be the exact same shader that shipped with Uru.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/Water.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/Water.h
@@ -1,0 +1,14 @@
+//
+//  Water.h
+//  Plasma
+//
+//  Created by Colin Cornaby on 12/29/22.
+//
+
+#ifndef Water_h
+#define Water_h
+#include <metal_stdlib>
+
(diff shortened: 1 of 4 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/Water.h
```diff
@@ -0,0 +1,14 @@
+//
+//  Water.h
+//  Plasma
+//
+//  Created by Colin Cornaby on 12/29/22.
+//
```
colincornaby, line 6: I see the wrong headers made it in. I'll correct eventually - but this PR is just a draft right now.

PR #1537, opened 2023-12-03
Volatile dynamic buffers in Metal

colincornaby:
Metal wasn't honoring that dynamic buffers should be volatile like the DX pipeline was. On modern systems - this should be pretty low impact. But it's still nice to implement.

Plasma refreshes its buffers every frame which can make volatile buffers a little niche. However - Metal Plasma triple buffers which means this could allow a buffer not currently being presented to be freed.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
@@ -162,6 +162,7 @@ public:
     // Currently requires a CA drawable and not a Metal drawable. In since CA drawable is only abstract implementation I know about, not sure where we would find others?
     void                CreateNewCommandBuffer(CA::MetalDrawable* drawable);
     MTL::CommandBuffer* GetCurrentCommandBuffer();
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -162,6 +162,7 @@ class plMetalDevice
     // Currently requires a CA drawable and not a Metal drawable. In since CA drawable is only abstract implementation I know about, not sure where we would find others?
     void                CreateNewCommandBuffer(CA::MetalDrawable* drawable);
     MTL::CommandBuffer* GetCurrentCommandBuffer();
+    MTL::CommandBuffer* GetCurrentDrawableCommandBuffer() { return fCurrentCommandBuffer; }
```
Hoikas, line 165:
```suggestion
    MTL::CommandBuffer* GetCurrentDrawableCommandBuffer() const { return fCurrentCommandBuffer; }
```

colincornaby:
@Hoikas - I'm done with this branch - do you want me to merge it in?

Hoikas:
Yeah, let's go ahead and merge this.

PR #1539, opened 2023-12-03
Modernize NucleusLib async callbacks using lambdas

dgelessus:
Trying to improve the "callback hell" in the low-/middle-level network code. The AsyncCore APIs relied heavily on C-style callback functions with untyped `void*` context parameters, which made it difficult to follow the execution and data flow in NetGameLib.

This PR migrates almost all of the async callbacks to use `std::function` (where they didn't already) and the callers to use lambdas for capturing context variables. This obsoletes the `void*` parameters and removes lots of manual casts.

AsyncSocket is a special case: its callback function was really multiple distinct callbacks wrapped in a generic signature, so I replaced it with a "callbacks object" with multiple virtual methods that the caller has to implement. This is a very "Java-ish" solution and is incompatible with lambdas, but works well with how the AsyncSocket API is used. It also clarifies the different callbacks' signatures and helps with seeing where each one is called.

The PR diff for the NetGameLib files is a bit noisy, because I changed some functions to instance methods and had to rename a couple of fields. The individual commit diffs are more readable there (e34454cfd626a1489310a43a9ab21348a022eac1 is the noisy refactor).

### DIFF
--- a/Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcIo.h
+++ b/Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcIo.h
@@ -165,7 +165,7 @@ void AsyncSocketEnableNagling (
 typedef std::function<void(const std::vector<plNetAddress>& /* addrs */)> FAsyncLookupProc;
 
 void AsyncAddressLookupName (
-    FAsyncLookupProc    lookupProc,
-    const ST::string &  name,
-    unsigned            port
+    const ST::string& name,
+    unsigned port,
+    FAsyncLookupProc lookupProc
 );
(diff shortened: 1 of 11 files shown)


Sources/Plasma/NucleusLib/pnAsyncCore/Private/pnAcIo.h
```diff
@@ -103,12 +97,14 @@ struct AsyncNotifySocketWrite : AsyncNotifySocketRead {
...
-    \param sock
-    \param code
-    \param notify
-*/
-using FAsyncNotifySocketProc = std::function<bool(AsyncSocket, EAsyncNotifySocket, AsyncNotifySocket*)>;
+class AsyncNotifySocketCallbacks
+{
+public:
+    virtual void AsyncNotifySocketConnectFailed(plNetAddress remoteAddr) = 0;
+    virtual bool AsyncNotifySocketConnectSuccess(AsyncSocket sock, plNetAddress localAddr, plNetAddress remoteAddr) = 0;
```
Hoikas, line 75:
```suggestion
    virtual bool AsyncNotifySocketConnectSuccess(AsyncSocket sock, const plNetAddress& localAddr, const plNetAddress& remoteAddr) = 0;
```

Hoikas (approved):

PR #1542, opened 2023-12-14
Dynamically checking display depth for Mac client

colincornaby:
@Hoikas - This is a possible optimization for older GPUs.

I'm not happy about the approach - the way this is implemented is promoting the color depth of the entire window instead of just the Metal output. I'd like to figure out if there is a better way to do this - so I'm leaving this as a draft until the implementation and the potential improvements are better understood.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
@@ -63,6 +63,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 @interface PLSView ()
 
 @property NSTrackingArea* mouseTrackingArea;
+@property id windowDidChangeObserver;
+
 #if PLASMA_PIPELINE_METAL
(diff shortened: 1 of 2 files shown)


Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
```diff
@@ -283,4 +293,28 @@ - (void)resizeDrawable:(CGFloat)scaleFactor
...
+            }
+        } else {
+            if (self.metalLayer.pixelFormat != MTLPixelFormatBGR10A2Unorm)
+            {
+                self.metalLayer.pixelFormat = MTLPixelFormatBGR10A2Unorm;
+            }
+        }
+#endif
+    }
+}
```
Hoikas, line 318: Is this the correct brace newline style?

colincornaby:
I'm going to close this. I've seen some Apple samples even default to 16 bit color without checking the monitor modes. So I don't think this is a serious performance concern.

PR #1543, opened 2023-12-14
Make `Python.Cheat` command more usable by not parsing its argument

dgelessus:
The argument is now always treated as a plain string and passed directly to the cheat. This is the behavior expected by all the functions in xCheat.py, which parse the string argument manually where needed.

The previous implementation evaluated the argument value as a Python expression. This theoretically allowed passing non-string arguments, but none of the cheats use this (and it's also a potential code execution issue). If no argument was passed, the function was called with no arguments, which is also not supported by any of the cheats.

In practice, this required unintuitive syntax for calling the cheats correctly, for example:

```
Python.Cheat GetPlayerID None
Python.Cheat GetSDL "'FirstWeekClothing'"
```

With the new argument handling, these have been simplified to:

```
Python.Cheat GetPlayerID
Python.Cheat GetSDL FirstWeekClothing
```

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
@@ -1724,50 +1724,18 @@ bool PythonInterface::RunPYC(PyObject* code, PyObject* module)
     return true;
 }
 
-/////////////////////////////////////////////////////////////////////////////
-//
-//  Function   : RunFunction
-//  PARAMETERS : module - module name to run 'name' in
-//             : name - name of function
-//             : args - tuple with arguments
-//
-//
-PyObject* PythonInterface::RunFunction(PyObject* module, const char* name, PyObject* args)
-{
-    if (module == nullptr)
-        return nullptr;
-
-    PyObject* function = PyObject_GetAttrString(module, name);
-
-    PyObject* result = nullptr;
-    if (function != nullptr)
-    {
-        result = PyObject_Call(function, args, nullptr);
-        Py_DECREF(function);
-    }
-
-    return result;
-}
-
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -1749,44 +1749,28 @@ PyObject* PythonInterface::RunFunction(PyObject* module, const char* name, PyObj
...
+                PyObject* callResult = RunFunction(moduleObj, name, argsObj);
+                if (callResult) {
+                    result = true;
+                    Py_DECREF(callResult);
+                }
+                Py_DECREF(argsObj);
             }
-
-            Py_DECREF(argsObj);
+            Py_DECREF(argObj);
```
Hoikas, line 1768: There is quite a lot of opportunity for simplification here. At minimum, you could make the argument object creation less tedious by doing `PyObject* argObj plPython::ConvertFrom(plPython::ToTuple, arg);`. Even better would be to offload all of this to the new Python vectorcall stuff, eg:
```C++
pyObjectRef moduleObj = ImportModule(module);
if (moduleObj) {
    pyObjectRef funcObj = PyObject_GetAttrString(moduleObj, name);
    if (funcObj)
        return (bool)plPython::CallObject(funcObj, args);
}
return false;
```
which will remove the `tp_call` translation penalty.

dgelessus: Good idea - the code more straightforward with `plPython::CallObject`. (The performance benefits from vectorcall don't really matter here, because this API is used only by `Python.Cheat` and nothing else.)

Hoikas (approved):

Issue #1544, opened 2023-12-18
Plasma Antialiasing Support + Metal Performance


colincornaby:
This issue is a summary of the findings while investigating issues with Plasma on Intel Iris 5100 hardware first noted by @Hoikas. Specifically - Plasma with max antialiasing performed far worse than it did on Windows.

### Antialiasing in Plasma

Plasma supports four antialiasing modes - with a slider used to change between them in graphics options.
| 1 | 2 | 4 | 6 |
|--|--|--|--|
| 1x (AA Off) | 2x | 4x | 6x |

### 6x antialiasing

Plasma's support of 6x antialiasing has been a bit of a mystery. Typical antialiasing modes are 2x/4x/8x.

It appears 6x antialiasing was a Radeon specific feature back around the time Uru was released. References have been found to it for:
- Radeon 9000 series hardware
- Radeon X000 series hardware
- Radeon X1000 series hardware

I could not find any mention of 6x antialiasing on the Radeon 2000 series or newer. It's unclear if 6x antialiasing continued to be supported on AMD GPUs. Additionally - no other vendor ever supported 6x antialiasing. 

The D3D pipeline will automatically pick the next lowest antialiasing setting is one is not available. This means on most GPUs - the AA settings look like this.

| 1 | 2 | 4 | 6 |
|--|--|--|--|
| 1x (AA Off) | 2x | 4x | 4x |

_On most hardware - the two highest settings are identical._

Metal also does not support 6x AA on AMD hardware. It's unclear if any other API besides D3D9 does.

### Intel GPUs

2x antialiasing was not supported until the Intel 6000 series of graphics. That means on 5000 series graphics and lower - the AA settings look like this under D3D:

| 1 | 2 | 4 | 6 |
|--|--|--|--|
| 1x (AA Off) | 1x (AA Off) | 4x | 4x |

On older Intel hardware, the first two AA settings are identical, and the last two AA settings are identical.

### Metal Pipeline

The Metal pipeline will select 8x antialiasing where it sees 6x antialiasing - so it's AA settings look like the following:
| 1 | 2 | 4 | 6 |
|--|--|--|--|
| 1x (AA Off) | 2x | 4x | 8x |

This means it's selecting very different AA settings than the D3D pipeline. For example, on older Intel hardware (such as the Iris 5100) - this is how it compares to Windows.

| Platform | 1 | 2 | 4 | 6 |
|--|--|--|--|--|
[...]

PR #1545, opened 2023-12-26
Matching MSAA behavior from D3D

colincornaby:
This will make the max AA setting functionally unavailable as 8x will be out of range. However - this will be more consistant with Windows where the max AA setting is typically interperated as 4x or 6x.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
@@ -186,27 +186,14 @@ void plMetalDevice::SetMaxAnsiotropy(uint8_t maxAnsiotropy)
 
 void plMetalDevice::SetMSAASampleCount(uint8_t sampleCount)
 {
-    // Plasma has some MSAA levels that don't completely correspond to what Metal can do
-    // Best fit them to levels Metal can do. Once they are best fit see if the hardware
-    // is capable.
-
-    uint8_t actualSampleCount = 1;
-    if (sampleCount == 6) {
-        actualSampleCount = 8;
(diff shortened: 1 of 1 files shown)


dpogue (approved):

Hoikas:
Does this close #1544?

colincornaby:
> Does this close #1544?

It depends. We probably need to decide.

This technically creates new issues. The high MSAA setting will do nothing on Metal hardware because 6x MSAA is a proprietary AMD extension that may only be available under Direct3D. (Maybe OpenGL on some platforms too?)

Also the 2x MSAA setting is unavailable on some hardware because like the Intel 5100, meaning the low MSAA setting does something.

A few changes have been proposed:
- That we allow 8x MSAA. (I think this is really unplayable on some slower hardware.)
- That we remove 6x MSAA and treat it as an 8x MSAA request. (I'm personally not in favor of this. I think we should keep 6x MSAA, and 8x MSAA is simply awful on Intel 5100 GPUs.)
- That we replace the MSAA with a drop down so we don't have these slider notches that don't do anything.

If we're fine with the following MSAA slider configurations in Metal:
- 1st notch is no MSAA
- 2nd notch is 2x MSAA if the hardware supports it, otherwise no MSAA
- Third notch is 4x MSAA
- Fourth notch is 4x MSAA (because 6x is not supported)

...then we're ok and the issue is closed.

colincornaby:
Another reason to consider keeping #1544 open is that it also affects the D3D renderer. The D3D renderer also surfaces MSAA modes that are unavailable for the hardware.

Hoikas:
I'm going to go ahead and merge this because it's an objective improvement and doesn't preclude other fixes.

PR #1549, opened 2024-01-02
Remove old debug-only fields from `plKeyData`

dgelessus:
The debugging message in the destructor was unconditionally disabled and no longer compiles if manually enabled. The debugging check in `plKeyImp::SetObjectPtr` can be rewritten to not require any extra fields (also making it a bit faster by removing the string comparison).

This uncovered missing includes in multiple places due to plKey.h not including `<string_theory/string>` anymore.


Hoikas (approved):

dgelessus:
There seems to be another missing include that only became noticeable on macOS - will fix

Hoikas (approved):

PR #1550, opened 2024-01-03
Screen more kinds of unwanted network messages

dgelessus:
Like H-uru/dirtsand#191, but on the client side. This expands `plNetClientMsgScreener` to also screen messages nested inside other messages, as can happen for avatar messages.

Also disallows receiving any kind of `plInputEventMsg` over the network - particularly `plControlEventMsg`, which can execute console commands, like the already blocked `plConsoleMsg`. This should have no effect on normal gameplay, but will probably break bot effects that modify other clients' fog settings (I haven't tested this though). If this is a problem, we could selectively allow "known safe" console commands instead of blocking everything.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plAvTaskBrain.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plAvTaskBrain.cpp
@@ -107,7 +107,7 @@ void plAvTaskBrain::DumpDebug(const char *name, int &x, int&y, int lineHeight, p
 
 // GetBrain ------------------------
 // ---------
-plArmatureBrain *plAvTaskBrain::GetBrain()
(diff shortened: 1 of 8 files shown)


dgelessus:
Broadcast by type could work as a stopgap until we have proper receivers for the fog messages.

In any case, I agree that would be the right solution in the long term. But the bots are currently set up to use networked console commands, and because they're deployed on Cyan's shard, the receiving clients will generally *only* understand networked console commands and not any new fog messages we might add. So for the time being, I think we're stuck with networked console commands...

Plus, there are other renderer settings that aren't covered by `plFogEnvironment`, like the clear color, which the bots also want to change.

Hoikas (approved):

PR #1551, opened 2024-01-08
Per pixel lighting support in Metal

colincornaby:
This PR adds Metal per pixel lighting support to the shaders - controlled by a preprocessor flag. Per pixel lighting defaults to off in since it's an alternative rendering mode that does not match the original engine. While per pixel lighting seems correct - it should be regarded as experimental. Per pixel shading enables a shader variant for the "fixed" pipeline shader. This shader variant is controlled by a shader compile time flag. However - this flag is a constant function which means evaluation of it can be deferred to when the shader is compiled from Metal byte code into vender specific assembly. I.E. - Metal can wait to apply the flag until the final runtime shader compilation.

Metal was originally ported from the OpenGL fork which used per pixel lighting. Metal switched back to per vertex for several reasons:
- The fragment shader function was trying to emulate the Direct3D fixed function pipeline - and because of high register usage was running into a lot of register overflows. The additional registers required by per pixel rendering was causing extreme performance issues. The shader has moved most of this register usage into compile time constants - so there are enough registers free now to do per pixel lighting.
- Most of Uru uses lower complexity meshes or light maps. Bump mapping was never in Uru. So the benefits of per pixel lighting were less - and per pixel lighting was deprioritized.

Please look over this closely - I want to make sure the per vertex path has remained unchanged. Not in a rush to get this in.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -100,7 +100,7 @@ constant const uint32_t miscFlags8 [[ function_constant(FunctionConstantLayerFla
 constant const uint8_t sourceTypes[MAX_BLEND_PASSES] = { sourceType1, sourceType2, sourceType3, sourceType4, sourceType5, sourceType6, sourceType7, sourceType8};
 constant const uint32_t blendModes[MAX_BLEND_PASSES] = { blendModes1, blendModes2, blendModes3, blendModes4, blendModes5, blendModes6, blendModes7, blendModes8};
 constant const uint32_t miscFlags[MAX_BLEND_PASSES] = { miscFlags1, miscFlags2, miscFlags3, miscFlags4, miscFlags5, miscFlags6, miscFlags7, miscFlags8};
-    constant const uint8_t passCount = (sourceType1 > 0) + (sourceType2 > 0) + (sourceType3 > 0) + (sourceType4 > 0) + (sourceType5 > 0) + (sourceType6 > 0) + (sourceType7 > 0) + (sourceType8 > 0);
+constant const uint8_t passCount = (sourceType1 > 0) + (sourceType2 > 0) + (sourceType3 > 0) + (sourceType4 > 0) + (sourceType5 > 0) + (sourceType6 > 0) + (sourceType7 > 0) + (sourceType8 > 0);
     
 constant const bool has2DTexture1 = (sourceType1 == PassTypeTexture && hasLayer1);
 constant const bool has2DTexture2 = (sourceType2 == PassTypeTexture && hasLayer2);
@@ -172,31 +172,20 @@ typedef struct
     float4 position [[position, invariant]];
     float3 texCoord1;
 } ShadowCasterInOut;
-
-vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
-                                       constant VertexUniforms & uniforms   [[ buffer(    VertexShaderArgumentFixedFunctionUniforms) ]],
-                                       constant plMetalLights & lights      [[ buffer(VertexShaderArgumentLights) ]],
-                                       constant float4x4 & blendMatrix1     [[ buffer(VertexShaderArgumentBlendMatrix1), function_constant(temp_hasOnlyWeight1) ]])
+    
+half4 calcLitMaterialColor(constant plMetalLights & lights,
+                          const half4 materialColor,
+                          constant plMaterialLightingDescriptor & materialLighting,
+                          const float4 position,
+                          const float3 normal)
 {
-    ColorInOut out;
-    // we should have been able to swizzle, but it didn't work in Xcode beta? Try again later.
-    const half4 inColor = half4(in.color.b, in.color.g, in.color.r, in.color.a) / half4(255.f);
-
-    const half3 MAmbient = mix(inColor.rgb, uniforms.ambientCol, uniforms.ambientSrc);
-    const half4 MDiffuse = mix(inColor, uniforms.diffuseCol, uniforms.diffuseSrc);
-    const half3 MEmissive = mix(inColor.rgb, uniforms.emissiveCol, uniforms.emissiveSrc);
-
     half3 LAmbient = half3(0.h, 0.h, 0.h);
     half3 LDiffuse = half3(0.h, 0.h, 0.h);
-
-    const float3 Ndirection = normalize(float4(in.normal, 0.f) * uniforms.localToWorldMatrix).xyz;
-
-    float4 position = float4(in.position, 1.f) * uniforms.localToWorldMatrix;
-    if (temp_hasOnlyWeight1) {
-        const float4 position2 = blendMatrix1 * float4(in.position, 1.f);
-        position = (in.weight1 * position) + ((1.f - in.weight1) * position2);
-    }
-
+    
+    const half3 MAmbient = mix(materialColor.rgb, materialLighting.ambientCol, materialLighting.ambientSrc);
+    const half4 MDiffuse = mix(materialColor, materialLighting.diffuseCol, materialLighting.diffuseSrc);
+    const half3 MEmissive = mix(materialColor.rgb, materialLighting.emissiveCol, materialLighting.emissiveSrc);
+    
     for (size_t i = 0; i < lights.count; i++) {
         constant const plMetalShaderLightSource *lightSource = &lights.lampSources[i];
         if (lightSource->scale == 0.0h)
@@ -231,16 +220,35 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
         }
 
         LAmbient.rgb = LAmbient.rgb + half3(direction.w * (lightSource->ambient.rgb * lightSource->scale));
-        const float3 dotResult = dot(Ndirection, direction.xyz);
+        const float3 dotResult = dot(normal, direction.xyz);
         LDiffuse.rgb = LDiffuse.rgb + MDiffuse.rgb * (lightSource->diffuse.rgb * lightSource->scale) * half3(max(0.f, dotResult) * direction.w);
     }
 
-    const half3 ambient = (MAmbient.rgb) * clamp(uniforms.globalAmb.rgb + LAmbient.rgb, 0.h, 1.h);
+    const half3 ambient = (MAmbient.rgb) * clamp(materialLighting.globalAmb.rgb + LAmbient.rgb, 0.h, 1.h);
     const half3 diffuse = clamp(LDiffuse.rgb, 0.h, 1.h);
-    const half4 material = half4(clamp(ambient + diffuse + MEmissive.rgb, 0.h, 1.h),
-                                 abs(uniforms.invVtxAlpha - MDiffuse.a));
+    return clamp(half4(ambient + diffuse + MEmissive.rgb, MDiffuse.a), 0.h, 1.h);
+}
 
-    out.vtxColor = half4(material.rgb, abs(uniforms.invVtxAlpha - MDiffuse.a));
+vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
+                                       constant VertexUniforms & uniforms   [[ buffer(    VertexShaderArgumentFixedFunctionUniforms) ]],
+                                       constant plMaterialLightingDescriptor & materialLighting   [[ buffer(    VertexShaderArgumentMaterialLighting) ]],
+                                       constant plMetalLights & lights      [[ buffer(VertexShaderArgumentLights) ]],
+                                       constant float4x4 & blendMatrix1     [[ buffer(VertexShaderArgumentBlendMatrix1), function_constant(temp_hasOnlyWeight1) ]])
+{
+    ColorInOut out;
+    const half4 inColor = half4(in.color.b, in.color.g, in.color.r, in.color.a) / half4(255.f);
+
+    const float3 Ndirection = normalize(float4(in.normal, 0.f) * uniforms.localToWorldMatrix).xyz;
+
+    float4 position = float4(in.position, 1.f) * uniforms.localToWorldMatrix;
+    if (temp_hasOnlyWeight1) {
+        const float4 position2 = blendMatrix1 * float4(in.position, 1.f);
+        position = (in.weight1 * position) + ((1.f - in.weight1) * position2);
+    }
+
+    out.vtxColor = calcLitMaterialColor(lights, inColor, materialLighting, position, Ndirection);
+    out.vtxColor.a = abs(uniforms.invVtxAlpha - out.vtxColor.a);
+    
     const float4 vCamPosition = position * uniforms.worldToCameraMatrix;
     
     // Fog
@@ -612,6 +620,7 @@ fragment half4 shadowFragmentShader(ShadowCasterInOut in [[stage_in]])
 
 vertex ColorInOut shadowCastVertexShader(Vertex in                              [[ stage_in ]],
                                          constant VertexUniforms & uniforms     [[ buffer(    VertexShaderArgumentFixedFunctionUniforms) ]],
+                                         constant plMaterialLightingDescriptor & materialLighting   [[ buffer(    VertexShaderArgumentMaterialLighting) ]],
                                          constant plShadowState & shadowState   [[ buffer(VertexShaderArgumentShadowState) ]])
 {
     ColorInOut out;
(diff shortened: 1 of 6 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -445,7 +472,7 @@ fragment half4 pipelineFragmentShader(ColorInOut in [[stage_in]],
             }
         }
         
-        currentColor = half4(in.vtxColor.rgb, 1.0h) * currentColor;
```
colincornaby, line 448: Might be a good time to ask about this one. After blending the layers - the shader again multiplies by the vertex +lighting color. This was carried over from the original GL shaders. I know if this isn't here the output is wrong. But I'm not sure why.

colincornaby:
I brought this branch up to date with master - and now it's acting strangely. For example - it's hitting asserts while trying to load the server.ini. I'll dig a bit and see whats going on. I'm pretty sure this behavior is newly unique to this branch. Last time I was in here it was working.

colincornaby:
The issue I was befuddled by is addressed in #1591.

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
dpogue: Skimming through the code, my guess is this:
`currentColor` starts out set to the lighting colour with no material values, but then on the first layer is overwritten to the layer colours of the first layer (in `blendFirst`) with subsequent layers blending into that. Since we've lost the lighting colours at that point, we need to multiply them back in.

dpogue:
FWIW I'm pretty sure OpenGL is also doing per-vertex lighting to match DirectX.

I might have been doing that in the fragment shader at one point while I was getting it set up, but it's in the vertex shader in the current version of the GL branch.

colincornaby:
Re-stating discussions elsewhere here for the status on this branch.

- While evaluating the lighting pipeline - it was found Metal is not properly handling the bump maps.
- Plasma calls them "bump maps" but they actually seem to be normal maps.
- Normal maps typically require per pixel lighting. In the D3D renderer there is a algorithm that attempts to bake the lighting algorithm into the texturing pipeline.
- This algorithm has significant issues - but would have been a reasonable approximation in the early 2000s.

Relative to this PR:

- We're not going to reimplement Cyan's algorithm in Metal. We have the ability to do proper normal mapping on Metal hardware.
- We will not ship per pixel lighting by default - but meshes can opt into it.
- Meshes with normal mapping will be opted into per pixel lighting. A normal mapping path in the shaders will be implemented in a future PR.
- The Metal pipeline will render normal maps differently than Windows - but the quality should be much higher.

We'll want to move ahead with this PR to get normal mapping into Metal.

colincornaby:
@dpogue - Checking back in to see if you have any thoughts here. We'll need this for bump mapping (even if we don't turn it on for everything.) Not high priority, just didn't want it to get lost.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -151,11 +151,22 @@ bool plRenderTriListFunc::RenderPrims() const
 
     size_t uniformsSize = offsetof(VertexUniforms, uvTransforms) + sizeof(UVOutDescriptor) * fDevice->fPipeline->fCurrNumLayers;
     fDevice->CurrentRenderCommandEncoder()->setVertexBytes(fDevice->fPipeline->fCurrentRenderPassUniforms, sizeof(VertexUniforms), VertexShaderArgumentFixedFunctionUniforms);
+    
+    fDevice->CurrentRenderCommandEncoder()->setVertexBytes(&fDevice->fPipeline->fCurrentRenderPassMaterialLighting, sizeof(plMaterialLightingDescriptor), VertexShaderArgumentMaterialLighting);
+    if (PLASMA_PER_PIXEL_LIGHTING)
+    {
+        fDevice->CurrentRenderCommandEncoder()->setFragmentBytes(&fDevice->fPipeline->fCurrentRenderPassMaterialLighting, sizeof(plMaterialLightingDescriptor), FragmentShaderArgumentMaterialLighting);
+    }
```
dpogue, line 159: Doesn't match our coding style, but we probably also want to make this controlled by a pipeline value rather than a define (even if we just hardcode that pipeline value to the value of the define for now)

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -151,11 +151,22 @@ bool plRenderTriListFunc::RenderPrims() const
...
     plMetalLights* lights = &fDevice->fPipeline->fLights;
     size_t         lightSize = offsetof(plMetalLights, lampSources) + (sizeof(plMetalShaderLightSource) * lights->count);
 
-    fDevice->CurrentRenderCommandEncoder()->setVertexBytes(lights, sizeof(plMetalLights), VertexShaderArgumentLights);
+    if (PLASMA_PER_PIXEL_LIGHTING)
+    {
+        fDevice->CurrentRenderCommandEncoder()->setFragmentBytes(lights, sizeof(plMetalLights), FragmentShaderArgumentLights);
+    } else {
+        fDevice->CurrentRenderCommandEncoder()->setVertexBytes(lights, sizeof(plMetalLights), VertexShaderArgumentLights);
+    }
```
dpogue, line 169: Doesn't match our coding style, but we probably also want to make this controlled by a pipeline value rather than a define (even if we just hardcode that pipeline value to the value of the define for now)

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -231,16 +225,46 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
...
+        out.worldPos = position;
+        out.normal = Ndirection;
+    }
+    
+    if (perPixelLighting)
+    {
+        out.vtxColor = inColor;
+    } else {
+        out.vtxColor = calcLitMaterialColor(lights, inColor, materialLighting, position, Ndirection);
+    }
```
dpogue, line 266: I don't know how strict our coding style has been within Metal shaders, but the opening brace should be on the same line as the `if`

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.cpp
```diff
@@ -110,6 +110,8 @@ void plMetalMaterialPassPipelineState::GetFunctionConstants(MTL::FunctionConstan
     constants->setConstantValues(&fFragmentShaderDescription.fPassTypes, MTL::DataTypeUChar, NS::Range(FunctionConstantSources, 8));
     constants->setConstantValues(&fFragmentShaderDescription.fBlendModes, MTL::DataTypeUInt, NS::Range(FunctionConstantBlendModes, 8));
     constants->setConstantValues(&fFragmentShaderDescription.fMiscFlags, MTL::DataTypeUInt, NS::Range(FunctionConstantLayerFlags, 8));
+    bool perPixelLighting = PLASMA_PER_PIXEL_LIGHTING;
+    constants->setConstantValue(&perPixelLighting, MTL::DataTypeBool, FunctionConstantPerPixelLighting);
```
dpogue, line 114: TODO: move the calculation of this to the pipeline so that it can (later) be dynamic per mesh

dpogue (commented):
Overall I think this is okay. Admittedly I only skimmed the actual shader code implementation details.

I think we should do the work now to refactor this a bit so that it can be turned on/off dynamically per mesh, even if we don't actually hook up the dynamic choice yet.

Otherwise, just a few code style issues.

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
colincornaby: Because Metal Shaders are C++ - I'm assuming we should apply the same coding style.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -151,11 +151,22 @@ bool plRenderTriListFunc::RenderPrims() const
 
     size_t uniformsSize = offsetof(VertexUniforms, uvTransforms) + sizeof(UVOutDescriptor) * fDevice->fPipeline->fCurrNumLayers;
     fDevice->CurrentRenderCommandEncoder()->setVertexBytes(fDevice->fPipeline->fCurrentRenderPassUniforms, sizeof(VertexUniforms), VertexShaderArgumentFixedFunctionUniforms);
+    
+    fDevice->CurrentRenderCommandEncoder()->setVertexBytes(&fDevice->fPipeline->fCurrentRenderPassMaterialLighting, sizeof(plMaterialLightingDescriptor), VertexShaderArgumentMaterialLighting);
+    if (PLASMA_PER_PIXEL_LIGHTING)
+    {
```
Hoikas, line 157:
```suggestion
    if (PLASMA_PER_PIXEL_LIGHTING) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -151,11 +151,22 @@ bool plRenderTriListFunc::RenderPrims() const
...
+        fDevice->CurrentRenderCommandEncoder()->setFragmentBytes(&fDevice->fPipeline->fCurrentRenderPassMaterialLighting, sizeof(plMaterialLightingDescriptor), FragmentShaderArgumentMaterialLighting);
+    }
 
     plMetalLights* lights = &fDevice->fPipeline->fLights;
     size_t         lightSize = offsetof(plMetalLights, lampSources) + (sizeof(plMetalShaderLightSource) * lights->count);
 
-    fDevice->CurrentRenderCommandEncoder()->setVertexBytes(lights, sizeof(plMetalLights), VertexShaderArgumentLights);
+    if (PLASMA_PER_PIXEL_LIGHTING)
+    {
```
Hoikas, line 165:
```suggestion
    if (PLASMA_PER_PIXEL_LIGHTING) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.cpp
```diff
@@ -267,23 +269,21 @@ void plMetalRenderSpanPipelineState::ConfigureBlendMode(const uint32_t blendMode
...
-    MTL::Function* function = library->newFunction(
-                                         NS::String::string("pipelineVertexShader", NS::ASCIIStringEncoding),
-                                         MakeFunctionConstants(),
-                                         &error)
-                                  ->autorelease();
-    return function;
+    MTL::Function* function = library->newFunction(NS::String::string("pipelineVertexShader", NS::ASCIIStringEncoding),
+                                                   MakeFunctionConstants(),
+                                                   &error);
+    assert(!error);
```
Hoikas, line 275: This should be using `hsAssert`

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.cpp
```diff
@@ -267,23 +269,21 @@ void plMetalRenderSpanPipelineState::ConfigureBlendMode(const uint32_t blendMode
...
-    return library->newFunction(
-                      NS::String::string("pipelineFragmentShader", NS::ASCIIStringEncoding),
-                      MakeFunctionConstants(),
-                      (NS::Error**)nullptr)
-        ->autorelease();
+    NS::Error* error = nullptr;
+    MTL::Function* function = library->newFunction(NS::String::string("pipelineFragmentShader", NS::ASCIIStringEncoding),
+                                                   MakeFunctionConstants(),
+                                                   &error);
+    assert(!error);
```
Hoikas, line 285: `hsAssert`

Hoikas (commented):
Needs code style cleanups

colincornaby:
With the latest changes I added some better detection of redundant bindings. I can see the redundant bindings in the Metal debugger, and it's been on my list of things to address. Unnecessary draw commands create a larger buffer to send to the GPU - so they should be avoided.

However - right now I'm using memcmp to compare struct values. I'd rather use a bool that marks when something is dirty and needs to be sent to the GPU. I've left notes behind in the code and will continue improving this in the future.

colincornaby:
FYI there are some outstanding bugs here that I've found while working on bump mapping. Preliminary fixes were made in a bump mapping branch. I'll bring the fixes back here and harden them.

Hoikas:
I'm converting this to a draft because there appears to work in progress and to clarify the status of this PR from the main listing.

colincornaby:
This was important to bump mapping - I will put this on my list of things to promote up to a real review.

colincornaby:
I'm working on getting this one current - but unfortunately I don't remember what the bug was that I discovered in the bump mapping branch that needed to have the fix back ported. I'm looking through the bump mapping branch for changes that stand out.

colincornaby:
I'm going to go ahead and bring this one back up for review. Possible issues were either patched in other PRs - and I added a few more fixes in this commit.

I don't know if that's the problem I originally found, but I've been testing and and reviewing the code. I'll do several more passes.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.h
```diff
@@ -50,6 +50,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plMetalDevice.h"
 #include "plSurface/plShaderTable.h"
 
+#ifndef PLASMA_FORCE_PER_PIXEL_LIGHTING
+#define PLASMA_FORCE_PER_PIXEL_LIGHTING 0
+#endif
```
dpogue: Is this define used anywhere now, or left over from the earlier version where it was a compile-time decision?

dpogue: oh, I see this is used in plMetalPipeline.cpp... maybe it makes sense to move the definition there?

colincornaby: This is used to force per pixel lighting as always on for all surfaces, which is handy for debugging.

colincornaby: I've moved the preprocessor define into the plMetalPipeline header. Although now I'm wondering if it should go into the cpp...

dpogue: If that's the only place it's used, the cpp file might make more sense

colincornaby: I'll move it to the cpp.

Eventually I'd like to make it a CMake option, but in since Metal is the only renderer that supports full per pixel right now it's not a priority.

colincornaby: Ok, it's moved to the cpp.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1662,6 +1673,28 @@ bool plMetalPipeline::IHandleMaterialPass(hsGMaterial* material, uint32_t pass,
     return true;
 }
 
+void plMetalPipeline::IBindLights()
+{
+    size_t         lightSize = offsetof(plMetalLights, lampSources) + (sizeof(plMetalShaderLightSource) * fLights.count);
+
+    if ( !(fState.fBoundLights.has_value() && memcmp(&fState.fBoundLights, &fLights, lightSize) == 0) ) {
```
```suggestion
    if (!(fState.fBoundLights.has_value() && memcmp(&fState.fBoundLights, &fLights, lightSize) == 0)) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1662,6 +1673,28 @@ bool plMetalPipeline::IHandleMaterialPass(hsGMaterial* material, uint32_t pass,
...
+        memcpy(&fState.fBoundLights, &fLights, lightSize);
+        if (fLightingPerPixel) {
+            fDevice.CurrentRenderCommandEncoder()->setFragmentBytes(&fLights, lightSize, FragmentShaderArgumentLights);
+        } else {
+            fDevice.CurrentRenderCommandEncoder()->setVertexBytes(&fLights, lightSize, VertexShaderArgumentLights);
+        }
+    }
+    
+    if ( !(fState.fBoundMaterialProperties.has_value() && fState.fBoundMaterialProperties == fCurrentRenderPassMaterialLighting) ) {
```
```suggestion
    if (!(fState.fBoundMaterialProperties.has_value() && fState.fBoundMaterialProperties == fCurrentRenderPassMaterialLighting)) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -2418,6 +2451,18 @@ void plMetalPipeline::IScaleLight(size_t i, float scale)
     fLights.lampSources[i].scale = scale;
 }
 
+void plMetalPipeline::ISetEnablePerPixelLighting(const bool enable)
+{
+    if (fLightingPerPixel != enable)
+    {
```
```suggestion
    if (fLightingPerPixel != enable) {
```

dpogue (approved):

PR #1554, opened 2024-01-09
Try moving Metal Layer handling to the pipeline

dpogue:
The goal here is to make the pipeline implementation responsible for constructing the renderer layer, so that everything related to Metal can be handled by the Metal pipeline and (eventually) everything related to GL can be handled by the GL pipeline. The client just has a layer as part of its view and doesn't need to care what renderer is being used.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
@@ -41,7 +41,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *==LICENSE==*/
 
 #import "PLSView.h"
-#include <Metal/Metal.h>
 #include <QuartzCore/QuartzCore.h>
 #include "plMessage/plInputEventMsg.h"
 
@@ -63,35 +62,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 @interface PLSView ()
 
 @property NSTrackingArea* mouseTrackingArea;
-#if PLASMA_PIPELINE_METAL
-@property(weak) CAMetalLayer* metalLayer;
-#endif
 
 @end
 
 @implementation PLSView
 
-// MARK: View setup
-- (id)initWithFrame:(NSRect)frameRect
-{
-    self = [super initWithFrame:frameRect];
-#if PLASMA_PIPELINE_METAL
-    CAMetalLayer* layer = [CAMetalLayer layer];
-    layer.contentsScale = [[NSScreen mainScreen] backingScaleFactor];
-    layer.maximumDrawableCount = 3;
-    layer.pixelFormat = MTLPixelFormatBGR10A2Unorm;
-    self.layer = self.metalLayer = layer;
-#endif
-    self.layer.backgroundColor = NSColor.blackColor.CGColor;
-    return self;
-}
-
 - (BOOL)acceptsFirstResponder
-{
-    return YES;
-}
-
-- (BOOL)wantsLayer
 {
     return YES;
 }
@@ -275,9 +251,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
         return;
     }
 
-#if PLASMA_PIPELINE_METAL
-    _metalLayer.drawableSize = newSize;
-#endif
+    self.layer.contentsScale = scaleFactor;
+
+    if ([self.layer isKindOfClass:[CAMetalLayer class]]) {
+        ((CAMetalLayer*)self.layer).drawableSize = newSize;
+    }
+
     [self.delegate renderView:self
           didChangeOutputSize:newSize
                         scale:scaleFactor];
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -42,9 +42,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 // System Frameworks
 #import <Cocoa/Cocoa.h>
-#ifdef PLASMA_PIPELINE_GL
-#import <OpenGL/gl.h>
-#endif
 #ifdef PLASMA_PIPELINE_METAL
 #import <Metal/Metal.h>
 #endif
@@ -82,17 +79,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plNetGameLib/plNetGameLib.h"
 #include "plProduct.h"
 
-// Until a pipeline is integrated with macOS, need to import the
-// abstract definition.
-#include "plPipeline/pl3DPipeline.h"
-
 void PumpMessageQueueProc();
 
 extern bool gDataServerLocal;
 extern bool gPythonLocal;
 extern bool gSDLLocal;
-
-bool NeedsResolutionUpdate = false;
 
 std::vector<ST::string> args;
 
@@ -111,7 +102,6 @@ std::vector<ST::string> args;
 @property CVDisplayLinkRef displayLink;
 @property dispatch_queue_t renderQueue;
 @property CALayer* renderLayer;
-@property(weak) PLSView* plsView;
 @property PLSPatcherWindowController* patcherWindow;
 @property NSModalSession currentModalSession;
 @property PLSPatcher* patcher;
@@ -191,16 +181,18 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
                                                        defer:NO];
     window.backgroundColor = NSColor.blackColor;
 
-    PLSView* view = [[PLSView alloc] init];
-    self.plsView = view;
-    window.contentView = view;
-    [window setDelegate:self];
-    
-    gClient.SetClientWindow((__bridge void *)view.layer);
+    PLSView* plView = [[PLSView alloc] init];
+    window.contentView = plView;
+
+    gClient.SetClientWindow((__bridge void *)plView.layer);
     gClient.SetClientDisplay((hsWindowHndl)NULL);
 
     self = [super initWithWindow:window];
     self.window.acceptsMouseMovedEvents = YES;
+
+    plView.delegate = self;
+    [window setDelegate:self];
+
     return self;
 }
 
@@ -209,7 +201,7 @@ dispatch_queue_t loadingQueue = dispatch_queue_create("", DISPATCH_QUEUE_SERIAL)
 - (void)startRunLoop
 {
     [[NSRunLoop currentRunLoop] addPort:[NSMachPort port] forMode:@"PlasmaEventMode"];
-    [self.plsView setBoundsSize:self.plsView.bounds.size];
+    [self.window.contentView setBoundsSize:self.window.contentView.bounds.size];
     auto* msg = new plDisplayScaleChangedMsg(self.window.backingScaleFactor);
     msg->Send();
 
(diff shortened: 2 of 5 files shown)


Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -354,7 +359,11 @@ class plPipeline : public plCreatable
     plDisplayMode fDesktopParams;
 
     virtual size_t GetViewStackSize() const = 0;
-    
+
+#ifdef HS_BUILD_FOR_MACOS
+    virtual CALayer* GetRenderLayer() { return nullptr; }
```
Hoikas, line 364:
```suggestion
    virtual CALayer* GetRenderLayer() const { return nullptr; }
```

Looks like a simple getter that should be a const member.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -4372,6 +4372,14 @@ uint32_t plMetalPipeline::IGetBufferFormatSize(uint8_t format) const
     return size;
 }
 
+CALayer* plMetalPipeline::GetRenderLayer()
+{
+    CA::MetalLayer* layer = CA::MetalLayer::layer();
+    layer->setPixelFormat(MTL::PixelFormatBGR10A2Unorm);
```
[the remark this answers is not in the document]
colincornaby: contentsScale is independent from Metal so it probably should go into the client. contentsScale should apply to both GL and Metal evenly. Or whatever else comes along in the future.

colincornaby:
I'll add my misgivings here that I've said elsewhere - but also admit what is proposed might be the best way.

CAMetalLayer is _not_ part of Metal - it's part of the window server API. Which is why I originally bumped it into the client. The client has a 1:1 relationship with the window server. If we wanted to support a new type of window server we'd write a new client.

Apple's window servers have all agreed on a common layer for Metal so far. So that's at least good. But they have not agreed on a common OpenGL layer (which may also be due to their OpenGL and OpenGL ES vendor specific APIs being vastly different.) So we're starting to push more window server specific details into the pipeline, which is kind of meh.

For both Metal and OpenGL on Apple platforms - there are higher level abstractions such as MLKView and NSOpenGLView. NSOpenGLView has fallen out of favor, but MTKView is still widely used and wouldn't be compatible with this change. MTKView wants to supply its own layer. However - I don't have any plans to run Plasma under MTKView on any platform.

Another misgiving on iOS it might be preferred that the view create it's layer directly via layerClass, which this would prevent.
https://developer.apple.com/documentation/uikit/uiview/1622626-layerclass?language=objc

But my understanding is all these issues are shared by the GL pipeline as well in a much larger way. GL has to communicate with multiple window servers from multiple vendors to jumpstart a GL viewport and context.

I guess one question I would have is: At least for GL - the GL layer will _only_ work on macOS. Is there a disadvantage to leaving the layer type selection within the macOS client?

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
```diff
@@ -63,39 +62,16 @@
...
 
-// MARK: View setup
-- (id)initWithFrame:(NSRect)frameRect
-{
-    self = [super initWithFrame:frameRect];
-#if PLASMA_PIPELINE_METAL
-    CAMetalLayer* layer = [CAMetalLayer layer];
-    layer.contentsScale = [[NSScreen mainScreen] backingScaleFactor];
-    layer.maximumDrawableCount = 3;
-    layer.pixelFormat = MTLPixelFormatBGR10A2Unorm;
```
colincornaby, line 82: Previously we've talked about inspecting the screen bit depth and setting the layer pixel format based on that. I.E. don't set a 10 bit color depth if the screen isn't 10 bit. I don't know if that change is moving ahead - but moving that out of the client would complicate things. The answer might need to be some sort of engine level flag for 10 bit color. In theory - Direct3D 9 supports 10 bit color output as well.

I've also seen Apple just set this to 16 bit in their examples (which is what our fragment shaders output in Metal) and call it a day. But memory usage would be higher.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -209,7 +201,7 @@ - (id)init
 - (void)startRunLoop
 {
     [[NSRunLoop currentRunLoop] addPort:[NSMachPort port] forMode:@"PlasmaEventMode"];
-    [self.plsView setBoundsSize:self.plsView.bounds.size];
```
colincornaby, line 212: Odd. But I think maybe this line was here to get the view to just enter setBoundsSize and do setup.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -457,20 +449,27 @@ - (void)startClient
 {
     PF_CONSOLE_INITIALIZE(Audio)
 
-    self.plsView.delegate = self;
+    self.renderLayer = gClient->GetPipeline()->GetRenderLayer();
+    _renderLayer.backgroundColor = NSColor.blackColor.CGColor;
+
+    self.window.contentView.layer = _renderLayer;
+    self.window.contentView.wantsLayer = YES;
```
colincornaby, line 456: Not sure how I feel. PLSView was meant to be self contained and wrap all details about the layer. Mostly.

dpogue:
It's mostly a separation of concerns thing: the client shouldn't need to care what renderer is being used and handle special cases for each renderer.

Ideally everything related to a given renderer should be implemented in its pipeline implementation, and that seems to work fine everywhere except macOS where they want different UI things used for different renderer types :(

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
dpogue: There's some nuance here that I don't entirely understand about layer-hosted views and layer-backed views and the order of setting `layer` and `wantsLayer`

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -457,20 +449,27 @@ - (void)startClient
...
-    
+
+#ifdef PLASMA_PIPELINE_METAL
+    if ([_renderLayer respondsToSelector:@selector(device)]) {
+        [_renderLayer addObserver:self
+                           forKeyPath:@"device"
+                              options:NSKeyValueObservingOptionNew | NSKeyValueObservingOptionInitial
+                              context:DeviceDidChangeContext];
+    }
+#endif
```
colincornaby, line 471: This is a good change. I think that this probably does extra checking. It could simply check the layer type and then check for device - in since CAMetalLayer already has device defined. The respondsToSelector would be unnecessary (wouldn't want some other device function floating in unintentionally in a different layer type.) The ifdef might be redundant in since it's a dynamic check, but we could skip the dynamic check if Metal wasn't present.

Longer term - we need a common code path for GL and Metal to handle screen and device changes. I think screen right now is all CGDisplay based so that's agnostic to the underlying API.

dpogue: As far as I can tell, the only thing this is currently being used for it to print the Metal device name in the window title for debugging purposes, so I'm hopeful it can be dropped entirely at some point in the future.

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -60,6 +60,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #define DEFAULT_SHADOWS 0
 #define DEFAULT_PLANARREFLECTIONS 0
 
+#if __OBJC__
+    @class CALayer;
+#else
+    class CALayer;
+#endif
```
colincornaby, line 67: Now I'm curious, but I bet I know the answer... Another reason CALayer was kept out of the renderer was to prevent Obj-C from creeping in. This forward declaration seems to be part of the abstraction against that. Metal can allocate a CALayer through Metal-cpp. Wouldn't this introduce Obj-C (or, I'm guessing Obj-C msg sending) into the GL renderer itself?

One risk of putting it into the pipeline is that a lot of stuff includes the pipeline, so you risk Obj-C leakage. I'm wondering if maybe there needs to be a secondary class that just pairs pipelines with layers, and that way the layer integration and all the mess that goes with can live outside of the pipeline.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
colincornaby: Strangely - this has been useful. On Intel MacBook Pros I've run into situations in which Plasma is running on the wrong GPU, so it's a nice sanity check. However it could be maybe re-implemented against the enumeration code that already exists in the engine and may already be storing the same data.

There's another reason this is a helpful debug hint. Plasma may need to switch GPUs mid run (which can be supported with Plasma's GPU crash recovery mechanisms.) So double checking the GPU device would be helpful for proofing that out. But I haven't gotten eGPU ejection or Plasma-follows-your-current-displays-GPU yet.

Sources/Plasma/NucleusLib/inc/plPipeline.h
dpogue: Maybe this is why I'm playing with subclassing CAOpenGLLayer entirely in C++ using objc runtime APIs (🙊), but in a class that's only included when compiling for Darwin.

The alternative is that the client needs to have a CAOpenGLLayer subclass that knows internal state of the GL pipeline like the GL context object, and that feels like it is prone to breaking due to a fuzzy API boundary between the two (whereas CAOpenGLLayer's expected API is well understood)

colincornaby:
I'm going to put in some more time on this review this weekend. Still thinking it over. But I need to put in some work to get some more concrete thoughts on this.

I think maybe one thing I was thinking before is that PLSView would be the glue code that matched the renderer to a layer. The view is macOS specific, but we're already in a decent amount of macOS specificity. Layers aren't technically part of SwiftUI, so it provided some level of abstraction and swappability for whenever Apple gets around to replacing layers. (For now, Metal and OpenGL still require layers to operate on macOS.)

I know I've run into issues with this approach on macOS, and you may be as well. Metal needs to talk to the layer to set the framebuffer size for resolution switches, and it also needs to make a request of the layer to swap framebuffers and get a new back buffer. I used to try to abstract that more to with interfaces back to the client - but recently I think I've been allowing the pipeline to talk directly to the layer.

I don't know what you're thinking with OpenGL - but there are probably similar issues. OpenGL also comes with the complication that context creation is managed by the layer as well. I think macOS takes a reasonable position of setting the current context within a callback. OpenGL doesn't have a cross platform context type built in so it's reasonable for macOS to abstract that away. Trying to work with contexts directly will introduce a whole bunch of platform specific code in the renderer. However - not all window managers agree on that. Some window managers encourage direct context manipulation.

If it's helpful in general or here - it may be beneficial to not manage contexts within the GL renderer. I think the context problem is kind of similar to this problem. Contexts are platform specific and will introduce platform code into the renderer - and the client itself is already a pretty good heap of platform code. Multiple contexts also shouldn't be needed - we should be ok with the system provided context. Multiple contexts are only needed for rendering in multiple threads at once. In theory - things like shadow buffers and other offscreen buffers could be rendered in a multithreaded manner. But Plasma has been pretty serial so far.

I'll think about it more. I'm still stuck on PLSView being meant to solve this, but PLSView not necessarily solving it in practice.

PR #1560, opened 2024-01-30
Use M1 macOS runners for arm64 builds

dpogue:
Build times seem to be cut in half using the M1 runners, and this allows us to properly install and build with Qt on arm64 builds.

Worth noting, I also tried running the x86_64 builds on M1 and it failed to compile Python. That might be something we need to look at in the future. For now, I've kept x86_64 running on Intel macOS 12.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,13 +245,13 @@ jobs:
 
 
   macos:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.platform.runner }}
     name: ${{ matrix.platform.str }}-${{ matrix.cfg.str }}
     strategy:
       matrix:
         platform:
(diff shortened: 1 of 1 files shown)


colincornaby:
Probably not crucial - but are we able to lipo together a proper universal build now? I've culled all the dylibs so we should just be down to lipo'ing the one executable.

dpogue:
> are we able to lipo together a proper universal build now?

Probably. I think we'd keep the build process the same as it is now (build for x86_64 and arm64 separately), and then have a followup action that runs after both those builds have finished that grabs the outputs and lipos them together. I guess this step is also where we'd need to handle the code signing stuff...

Hoikas (approved):

colincornaby (approved):

PR #1562, opened 2024-02-04
Fix Linux CI.

Hoikas:
Per @dpogue, the problem is https://gitlab.kitware.com/cmake/cmake/-/issues/25650. Long story short, CMake broke PCH and Unity on CMake 3.28.2, so disable that combination when it's used.


dpogue (approved):
This seems to work, so let's merge it

PR #1563, opened 2024-02-05
Missing header includes & MinGW fixes

dpogue:
The first commit is a bunch of missing includes that are masked by PCH and Unity builds. Without those enabled, the project will fail to compile without these fixes.

The second commit is a few fixes to get some things building with MinGW again (I haven't been able to get MinGW Python working locally, so I can't promise that it will fix the whole engine build)

### DIFF
--- a/Sources/Plasma/Apps/plCrashHandler/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plCrashHandler/CMakeLists.txt
@@ -18,14 +18,18 @@ target_link_libraries(
         plWinDpi
         pfCrashHandler
 
-        $<$<PLATFORM_ID:Windows>:Comctl32>
-        $<$<PLATFORM_ID:Windows>:UxTheme>
+        $<$<PLATFORM_ID:Windows>:comctl32>
+        $<$<PLATFORM_ID:Windows>:uxtheme>
 )
 target_compile_definitions(
     plCrashHandler
     PRIVATE
         $<$<PLATFORM_ID:Windows>:UNICODE>
 )
+
+if(MINGW)
+    target_link_options(plCrashHandler PRIVATE -municode)
+endif()
 
 source_group("Source Files" FILES ${plCrashHandler_SOURCES})
 source_group("Resource Files" FILES ${plCrashHandler_RESOURCES})
--- a/Sources/Plasma/FeatureLib/pfPython/pyGameCliGlue.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyGameCliGlue.cpp
@@ -40,11 +40,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
+#include <Python.h>
 #include "pyGameCli.h"
 
 #include "plPythonConvert.h"
 
-#include <Python.h>
 
 // ===========================================================================
 
--- a/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
+++ b/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
@@ -47,7 +47,25 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <variant>
 
 #include "hsWindows.h"
-#include <ShellScalingApi.h>
+
+#ifdef HAVE_SHELLSCALINGAPI
+#   include <ShellScalingApi.h>
+#else
+    typedef enum
+    {
+        PROCESS_DPI_UNAWARE = 0,
+        PROCESS_SYSTEM_DPI_AWARE = 1,
+        PROCESS_PER_MONITOR_DPI_AWARE = 2
+    } PROCESS_DPI_AWARENESS;
+
+    typedef enum
+    {
+        MDT_EFFECTIVE_DPI = 0,
+        MDT_ANGULAR_DPI = 1,
+        MDT_RAW_DPI = 2,
+        MDT_DEFAULT = MDT_EFFECTIVE_DPI
+    } MONITOR_DPI_TYPE;
+#endif
 
 #include "plStatusLog/plStatusLog.h"
 
--- a/Sources/Tools/plNetLog/ChunkBuffer.h
+++ b/Sources/Tools/plNetLog/ChunkBuffer.h
@@ -50,6 +50,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <QString>
 #include <QMutex>
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+using qtMutexLocker = QMutexLocker<QMutex>;
+#else
+using qtMutexLocker = QMutexLocker;
+#endif
+
 class ChunkBuffer
 {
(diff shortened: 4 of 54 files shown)


Sources/Tools/plNetLog/ChunkBuffer.h
```diff
@@ -149,7 +149,7 @@ class ChunkBuffer
     QMutex              m_mutex;
     std::list<Buffer>   m_chunks;
 
-    void waitOnData(QMutexLocker& lock);
+    void waitOnData(QMutexLocker<QMutex>& lock);
```
zrax, line 152: It looks like `QMutexLocker` was changed to a template only in Qt6, so this change will break Qt5.

dpogue: Ah, good catch, I guess we'll need some conditional macro alias for it :\

dgelessus (commented):
Thanks, it's always good to squash these include issues. Good to know that precompiled headers can also mask missing includes. I was aware of unity builds doing this (which is why I keep them disabled locally), but didn't know that PCHs can lead to the same problem.

Sources/Plasma/FeatureLib/pfPython/pyGameCliGlue.cpp
```diff
@@ -40,11 +40,11 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
+#include <Python.h>
 #include "pyGameCli.h"
 
 #include "plPythonConvert.h"
 
-#include <Python.h>
```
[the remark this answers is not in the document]
dgelessus: Hmm, pyGlueHelpers.h looks like it was meant to work without Python.h included, based on the manual definitions of `PyObject`, etc., but the `#if PY_VERSION_HEX` breaks this principle.

I assume the goal was to avoid indirectly including Python.h in all of the pfPython header files. I guess the proper fix would be to split pyGlueHelpers.h into two headers (one that works without Python.h, for use by other headers, and one that includes Python.h, for use by the implementation code), but that would be a larger change obviously.

Sources/Plasma/Apps/plCrashHandler/CMakeLists.txt
```diff
@@ -18,15 +18,19 @@ target_link_libraries(
...
+        $<$<PLATFORM_ID:Windows>:comctl32>
+        $<$<PLATFORM_ID:Windows>:uxtheme>
 )
 target_compile_definitions(
     plCrashHandler
     PRIVATE
         $<$<PLATFORM_ID:Windows>:UNICODE>
 )
 
+if (MINGW)
```
Hoikas, line 30:
```suggestion
if(MINGW)
```

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
```diff
@@ -47,10 +47,29 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
 #include "hsWindows.h"
-#include <ShellScalingApi.h>
+#if __has_include(<ShellScalingApi.h>)
+#   include <ShellScalingApi.h>
+#endif
 
 #include "plStatusLog/plStatusLog.h"
 
+#ifndef DPI_ENUMS_DECLARED
```
Hoikas, line 56: Is this defined in `ShellScalingApi.h`?

dpogue: Yes, I don't love using something that's not documented API, but MinGW doesn't ship the ShellScalingApi header.

I could alternatively define these in the `#else` block of the `#if __has_include` if you prefer that

Hoikas: Yeah, I feel like that might be a little more obvious.

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
```diff
@@ -47,7 +47,24 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <variant>
 
 #include "hsWindows.h"
-#include <ShellScalingApi.h>
+#if __has_include(<ShellScalingApi.h>)
```
dgelessus, line 50: For anyone else wondering, `__has_include` was standardized in C++17, which is our current minimum, so it's safe to use unconditionally.

So far, it seems that H'uru has preferred CMake configure-time checks instead of `__has_include`. Do we want to do that instead for consistency?

(Personally I have no issue with `__has_include` though, and I actually prefer it in most cases, because a configure-time check is more work to maintain and usually much slower.)

Sources/Tools/plNetLog/ChunkBuffer.h
```diff
@@ -50,6 +50,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <QString>
 #include <QMutex>
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+#define Q_MUTEX_LOCKER_TYPE QMutexLocker<QMutex>
+#else
+#define Q_MUTEX_LOCKER_TYPE QMutexLocker
+#endif
```
dgelessus, line 57: This could be done with `typedef`/`using`, right?

```suggestion
#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
using = Q_MUTEX_LOCKER_TYPE QMutexLocker<QMutex>;
#else
using Q_MUTEX_LOCKER_TYPE = QMutexLocker;
#endif
```

dgelessus (commented):
Small things, otherwise LGTM.

(Execpt for the Python.h dependency/order issue discussed above, but that's a bigger fix and can be done separately IMO.)

Sources/Tools/plNetLog/ChunkBuffer.h
dpogue: Probably... let me give it a shot

dpogue:
> Execpt for the Python.h dependency/order issue discussed above, but that's a bigger fix and can be done separately IMO.

The Python stuff is all consistent with the rest of the existing files now, but there's definitely room for future improvements to be done as a follow-up

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
Hoikas: It would probably be better to, for consistency's sake, do configure time checks.

Sources/Tools/plNetLog/ChunkBuffer.h
```diff
@@ -50,6 +50,12 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <QString>
 #include <QMutex>
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+using Q_MUTEX_LOCKER_TYPE = QMutexLocker<QMutex>;
+#else
+using Q_MUTEX_LOCKER_TYPE = QMutexLocker;
+#endif
```
Hoikas, line 57: This should probably look more like a type than a define (eg not be all caps and/or snake case).

Hoikas (approved):

PR #1570, opened 2024-02-26
Fix missing Anim FFwd Responder Cmd.

Hoikas:
Fixes the Animation FFwd command missing from the list of potetial responder commands in the Max plugin. How long has this been broken?????????


dgelessus (approved):
Could we add a static assert to ensure that `kNumTypes` matches up with the length of `static int oldTypes[]` below?

Deledrius (approved):

Deledrius (approved):

PR #1572, opened 2024-03-02
Add support for new DynamicCamMap checkbox.

Hoikas:
This is the companion piece to H-uru/moul-assets#251. There is no longer any need to manually localize the options menu :tada:. I didn't bother with improving the quality of tangential Python code. All of the options menu Python is terrible, and I didn't want to get bogged down with that.

### DIFF
--- a/Scripts/Python/plasma/Plasma.py
+++ b/Scripts/Python/plasma/Plasma.py
@@ -827,6 +827,10 @@ def PtStartScreenCapture(selfKey,width=800,height=600):
     """Starts a capture of the screen"""
     pass
 
+def PtSupportsPlanarReflections() -> bool:
+    """Returns if planar reflections are supported"""
+    ...
+
 def PtToggleAvatarClickability(on):
     """Turns on and off our avatar's clickability"""
     pass
--- a/Scripts/Python/xIniDisplay.py
+++ b/Scripts/Python/xIniDisplay.py
@@ -62,9 +62,10 @@ kGraphicsQualityLevel = "Quality.Level"
 kGraphicsShadows = "Graphics.Shadow.Enable"
 kGraphicsVerticalSync = "Graphics.EnableVSync"
 kGraphicsShadowQuality = "Graphics.Shadow.VisibleDistance"
+kGraphicsDynamicReflections = "Graphics.EnablePlanarReflections"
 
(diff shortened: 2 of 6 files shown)


Scripts/Python/plasma/Plasma.py
```diff
@@ -827,6 +827,10 @@ def PtStartScreenCapture(selfKey,width=800,height=600):
     """Starts a capture of the screen"""
     pass
 
+def PtSupportsPlanarReflections() -> bool:
+    """Returns if planar reflections are supported"""
+    ...
```
dgelessus, line 832: For consistency with the rest of the file, and because this isn't a .pyi file:

```suggestion
    pass
```

Scripts/Python/xIniDisplay.py
```diff
@@ -62,9 +62,10 @@
 kGraphicsShadows = "Graphics.Shadow.Enable"
 kGraphicsVerticalSync = "Graphics.EnableVSync"
 kGraphicsShadowQuality = "Graphics.Shadow.VisibleDistance"
+kGraphicsDynamicReflections = "Graphics.EnablePlanarReflections"
```
dgelessus, line 65: For anyone else wondering: this console command has already existed since forever.

dgelessus (commented):
LGTM at first glance (but haven't done any local testing yet)

Scripts/Python/plasma/Plasma.py
Hoikas: This is the way that I've been doing all the rest of the new/updated API functions to ease the (eventual? hopeful?) transition to pyi. Example:

https://github.com/H-uru/Plasma/blob/5807230e6e43e1f80e5b24c4878b468330710f2a/Scripts/Python/plasma/Plasma.py#L879-L883

dgelessus: Ah, alright then - I guess I never noticed any of those yet. Yes, it would be nice to convert these to proper .pyi files at some point.

Deledrius (approved):

Hoikas:
There seems to be a problem in the key map dialog where none of the key bindings are displaying. Investigating...

Hoikas:
Key bindings dialog is now working correctly.

Deledrius (approved):

PR #1573, opened 2024-03-09
Key Map dialog improvements.

Hoikas:
Previously, it was possible to get the key map chronicle into an invalid state with the same key bound to multiple controls. You can reproduce that by binding UpArrow to "start chat" then rebinding UpArrow to "move forward". The chronicle will then have two entries for UpArrow in it. The bindings will behave as you expect until you quit and restart Uru, at which point UpArrow will be bound to "start chat" again. The only way to fix this is by resetting the key map to default.

To fix this, I cleaned up the code to make the ownership of the key map more clear. Previously, the code half-handedly updated the vault chronicle value and the key map. Now, we bind any input directly, and completely rewrite the chronicle each time a key is bound. In this way, the UI display, key bindings, and key chronicle all have a single source of truth.

Further, you can now unbind keys by pressing the current key binding. This is useful if, for example, you don't want "start chat" to be bound to anything. Previously, you would have needed to bind the "start chat" key to some other control, then bind that other control to what you actually wanted.

With this fixing the major keymap bugs and gotchas, we could probably close #1067.

### DIFF
--- a/Scripts/Python/xOptionsMenu.py
+++ b/Scripts/Python/xOptionsMenu.py
@@ -1810,8 +1810,12 @@ class xOptionsMenu(ptModifier):
         km = ptKeyMap()
         newKeyStr = km.convertVKeyToChar(vkey, modifiers)
 
-        # This will cause any previous uses of the key to be unbound.
+        # If this is the same key as before, unmap the binding.
         controlCode = kControlCodes[controlCodeId]
+        if self.IGetBoundKey(controlCode, keyIdx=0 if isPrimary else 1) == newKeyStr:
+            newKeyStr = "(unmapped)"
+
+        # This will cause any previous uses of the key to be unbound.
         if isinstance(controlCode, str):
             PtDebugPrint(f"xOptionsMenu.ISetKeyMapping(): Binding {newKeyStr=} to console command {controlCode=}")
             km.bindKeyToConsoleCommand(newKeyStr, controlCode)
--- a/Scripts/Python/xOptionsMenu.py
+++ b/Scripts/Python/xOptionsMenu.py
@@ -301,7 +301,7 @@ gKM1ControlCodesRow2 = {
     kKMEditLine18Row2: _KeyLine(None, False, False),
 }
 
-defaultControlCodeBinds = {
+kDefaultControlCodeBinds = {
     PlasmaControlKeys.kKeyMoveForward: ("UpArrow", "(unmapped)"),
     PlasmaControlKeys.kKeyMoveBackward: ("DownArrow", "(unmapped)"),
     PlasmaControlKeys.kKeyRotateLeft: ("LeftArrow", "(unmapped)"),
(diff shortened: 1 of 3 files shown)


Scripts/Python/xOptionsMenu.py
```diff
@@ -1834,23 +1765,20 @@ def LoadKeyMap(self):
...
-                controlStr = km.convertControlCodeToString(control_code)
-                SubArray = KeyMapArray[counter].split("$")
-                key1 = SubArray[0]
-                key2 = SubArray[1]
-                PtDebugPrint("Binding " + key1 + " & " + key2 + " to " + controlStr)
-                km.bindKey(key1,key2,controlStr)
+                controlStr = km.convertControlCodeToString(controlCode)
+                keys = mappedKey.split("$")
+                PtDebugPrint(f"xOptionsMenu.LoadKeyMap(): Binding {keys=} to {controlStr=}", level=kWarningLevel)
+                km.bindKey(*keys, controlStr)
```
dgelessus, line 1781: To make this more robust in case the chronicle text isn't formatted correctly:

```suggestion
                key1, _, key2 = mappedKey.partition("$")
                PtDebugPrint(f"xOptionsMenu.LoadKeyMap(): Binding {key1=} & {key2=} to {controlStr=}", level=kWarningLevel)
                km.bindKey(key1, key2, controlStr)
```

Sources/Plasma/NucleusLib/pnInputCore/plInputMap.cpp
```diff
@@ -659,7 +659,7 @@ ST::string plKeyMap::GetStringUnmapped()
         default:
             break;
     }
-    return ST_LITERAL("(unmapped)");
+    return ST_LITERAL("(Unmapped)");
```
dgelessus, line 662: This is inconsistent with the hardcoded uses of `(unmapped)` all over the place :/

Also, this value ends up in the `KeyMap` chronicle as far as I can tell. Are we sure that this change won't break anything when switching from/to older clients? It's *probably* fine because "unmapped" is the default case when it can't translate a string back to a `plKeyDef`, but still, I wouldn't change this unless there's a good reason...

I wouldn't worry about consistency with the other languages. The capitalization there is non-standard anyway (at least for French and German, and probably Italian and Spanish too) - they should actually all be lowercase with a space in between. I assume the CamelCase is a workaround because the `KeyMap` chronicle uses spaces as separators, so an individual key name cannot safely contain a space.

Scripts/Python/xOptionsMenu.py
```diff
@@ -702,80 +698,15 @@ def OnGUINotify(self,id,control,event):
...
-                        KeyMapString = self.getChronicleVar("KeyMap")
-                        KeyMapArray = KeyMapString.split()
-                        KeyMapArray[(kmID-300)] = key1 + "$" + key2 + " "
-                        for key in KeyMapArray:
-                            NewKeyMapString += key + " "
-                        self.setNewChronicleVar("KeyMap", NewKeyMapString.rstrip())
-                    # lose the focus when done
-                    KeyMapDlg.dialog.noFocus()
-                    # force writing the keymap
-                    km.writeKeyMap()
```
dgelessus, line 735: The new version no longer calls `km.writeKeyMap()` anywhere. I assume this is intentional?

Perhaps we should just remove `ptKeyMap.writeKeyMap` and the underlying `plInputInterfaceMgr::WriteKeyMap()`. AFACIT that code did nothing useful anyway. It tries to write a file init/keyboard.fni, but MOULa installs normally don't have an init folder, so that always fails silently.

dgelessus:
> Previously, you would have needed to bind the "start chat" key to some other control, then bind that other control to what you actually wanted.

I always forget that Americans don't have that extra key next to the left Shift key that Uru interprets as `(unmapped)` 😄

> With this fixing the major keymap bugs and gotchas, we could probably close #1067.

I don't think so. That PR mainly implements something different, namely saving the key map to a local config file (in a way that actually works, unlike `ptKeyMap.writeKeyMap`) so it's used as the default for new avatars (where there's no `KeyMap` chronicle yet).

Sources/Plasma/NucleusLib/pnInputCore/plInputMap.cpp
Hoikas: I had to chase down all of this to make sure that things didn't break if we change languages... What calls to bind a key does is check the current language for the key. If it's not found, it goes through all the other languages, looking for that key. If it's still not found, it gives up and treats it as `plKeyCombo::kUmapped`.

Scripts/Python/xOptionsMenu.py
Hoikas: Right, I intentionally removed this call because it has no effect. I didn't want to remove the `writeKeyMap` code in this PR, though.

dpogue:
Out of scope for this PR, but there's an open question around how to handle internationalization of keymaps particularly across different keyboard layouts when targeting non-Windows environments: https://github.com/H-uru/Plasma/discussions/786 (mostly linking this to hopefully get some more thoughts in that discussion)

dpogue (approved):

PR #1580, opened 2024-04-07
fix Win7/8 resolution

Filtik:
In Windows 7 and Windows 8 the intro is very small and no mouse cursor is visible.
This fixes the problem.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.cpp
+++ b/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.cpp
@@ -160,7 +160,7 @@ UINT plWinDpi::GetDpi(HWND hWnd) const
     auto dpiForMonitor = fGetDpiForMonitor(monitor, MDT_EFFECTIVE_DPI, &dpiX, &dpiY);
     if (dpiForMonitor.has_value()) {
         if (SUCCEEDED(dpiForMonitor.value()))
-            return UINT(float(dpiY));
+            return dpiY;
         LogRed("Per-Monitor DPI failed: {}", hsCOMError(dpiForMonitor.value()));
     }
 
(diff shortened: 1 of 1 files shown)


dpogue (commented):
This change does appear to be correct. After fetching the DPI value, we do `MulDiv(metricValue, dpiValue, 96)`, so I think the current (buggy) behaviour is dividing by 96 twice in some cases.

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.cpp
```diff
@@ -160,15 +160,15 @@ UINT plWinDpi::GetDpi(HWND hWnd) const
     auto dpiForMonitor = fGetDpiForMonitor(monitor, MDT_EFFECTIVE_DPI, &dpiX, &dpiY);
     if (dpiForMonitor.has_value()) {
         if (SUCCEEDED(dpiForMonitor.value()))
-            return UINT(float(dpiY) / 96.0f);
+            return UINT(float(dpiY));
```
Hoikas, line 163:
```suggestion
            return dpiY;
```

Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.cpp
```diff
@@ -160,15 +160,15 @@ UINT plWinDpi::GetDpi(HWND hWnd) const
...
         LogRed("Per-Monitor DPI failed: {}", hsCOMError(dpiForMonitor.value()));
     }
 
     // Legacy DPI computation... This is both slow and lacks much knowledge about DPI.
     HDC hdc = GetDC(hWnd);
     int ydpi = GetDeviceCaps(hdc, LOGPIXELSY);
     ReleaseDC(hWnd, hdc);
-    return UINT(float(ydpi) / 96.0f);
+    return UINT(float(ydpi));
```
Hoikas, line 171:
```suggestion
    return UINT(ydpi);
```

Hoikas (changes requested):

Hoikas (approved):

Hoikas:
When you have a chance, please rebase this onto master so that we can get a proper CI run.

PR #1581, opened 2024-04-08
Move shaders into shader library target

colincornaby:
This PR does several things:

- It moves the shaders into a proper Metal shader library target when using the Xcode generator.
- It compiles the shaders twice - once for Metal 2.1 (macOS 10.14) and once for Metal 2.3 (macOS 11.0.) The Metal 2.3 build of the shaders can take advantage of special features when running on Apple Silicon.
- plMetalDevice now dynamically loads the correct compiled shader archive based on the host version of macOS
- The shader library is now cached instead of being re-created each time a shader is needed. (This change is cleanup thats long overdue.)

I'd like some feedback on the implementation in the CMake - and feedback on this change requiring CMake 3.28.


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceLibraryLoader.mm
```diff
@@ -0,0 +1,60 @@
...
+#include <Foundation/Foundation.h>
+#include <Metal/Metal.h>
+#include "plMetalDevice.h"
+
+void plMetalDevice::LoadLibrary()
+{
+    NS::Error* error;
+    if (@available(macOS 11, *))
+    {
+        NSURL* shaderURL = [NSBundle.mainBundle URLForResource:@"pfMetalPipelineShadersMSL23" withExtension:@"metallib"];
```
dpogue, line 52: In theory, this is roughly equivalent to:
```c++
plFileName::Join(NSBundle::mainBundle()->resourceURL(), ST_LITERAL("pfMetalPipelineShadersMSL23.metallib")
```

(which I realize is not the "correct" way to do things, but does workaround the bizarre oversight in the metal-cpp library)

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -188,6 +188,8 @@ class plMetalDevice
     void EncodeBlur(MTL::CommandBuffer* commandBuffer, MTL::Texture* texture, float sigma);
 
     MTL::PixelFormat GetFramebufferFormat() const { return fFramebufferFormat; };
+    
+    MTL::Library* GetShaderLibrary() { return fShaderLibrary; }
```
Hoikas, line 192:
```suggestion
    MTL::Library* GetShaderLibrary() const { return fShaderLibrary; }
```

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -116,6 +116,10 @@ elseif(APPLE)
         Mac-Cocoa/PLSLoginWindowController.xib
         Mac-Cocoa/PLSPatcherWindowController.xib
     )
+    list(APPEND plClient_SHADERTARGETS
+        pfMetalPipelineShadersMSL21
+        pfMetalPipelineShadersMSL23
+    )
```
Hoikas, line 122: This should probably be pulled in transitively from `pfMetalPipeline` target.

Hoikas (commented):
You probably need to assert in the metal pipeline that the CMake version is at the minimum required for this feature to work.

Sources/Plasma/Apps/plClient/CMakeLists.txt
colincornaby: I moved what I could but this specific part (the XCODE_EMBED_RESOURCES target property) is not transitive.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceLibraryLoader.mm
colincornaby: I'll probably leave it the way it is for now (we're already doing something similar for the performance shaders.) But I'll revisit if Apple fixes the missing functions.

colincornaby:
> You probably need to assert in the metal pipeline that the CMake version is at the minimum required for this feature to work.

What would be the best way to do that? A different cmake_minimum_required or some sort of different check?

Hoikas:
> What would be the best way to do that? A different cmake_minimum_required or some sort of different check?

We probably may not want to use `cmake_ minimum_required()` because this has [side effects related to setting policies](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings), and we may not want to have different policy scopes playing with our mind. It may be better to test and error if the version is too low:
```cmake
if(${CMAKE_VERSION} VERSION_LESS 3.9)
    message(FATAL_ERROR "boo")
endif()
```

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/GammaCorrection.metal
```diff
@@ -79,3 +79,17 @@ fragment half4 gammaCorrectFragment(GammaVertexOut in               [[stage_in]]
...
+
+#if defined(__METAL_IOS__) || __METAL_VERSION__ >= 230
+fragment half4 gammaCorrectFragmentInPlace(GammaVertexOut in        [[stage_in]],
+                                                 half4 color        [[color(0)]],
+                                 texture1d_array<ushort> LUT        [[texture(1)]])
+{
+    return {
+        half(float(LUT.sample(lutSampler, color.r, 0).x)/USHRT_MAX),
+        half(float(LUT.sample(lutSampler, color.g, 1).x)/USHRT_MAX),
+        half(float(LUT.sample(lutSampler, color.b, 2).x)/USHRT_MAX),
```
Hoikas, line 91:
```suggestion
        half(float(LUT.sample(lutSampler, color.r, 0).x) / USHRT_MAX),
        half(float(LUT.sample(lutSampler, color.g, 1).x) / USHRT_MAX),
        half(float(LUT.sample(lutSampler, color.b, 2).x) / USHRT_MAX),
```

Hoikas:
Looks like this PR needs to be rebased to build.

colincornaby:
I went ahead and rebased pulling in @dpogue's latest Mac fixes. Waiting to see if the build is green.

colincornaby:
It looks like builds are passing again on this PR.

Hoikas (approved):
This looks fine to me, but I'd like to hear from @dpogue as well.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceLibraryLoader.mm
```diff
@@ -0,0 +1,60 @@
...
+        fShaderLibrary = fMetalDevice->newLibrary(static_cast<NS::URL*>(shaderURL), &error);
+    }
+    else
+    {
+        NSURL* shaderURL = [NSBundle.mainBundle URLForResource:@"pfMetalPipelineShadersMSL21" withExtension:@"metallib"];
+        fShaderLibrary = fMetalDevice->newLibrary(static_cast<NS::URL*>(shaderURL), &error);
+    }
```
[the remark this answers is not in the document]
dpogue: This maybe still needs to be resolved?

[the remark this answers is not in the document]
colincornaby: Re-reviewed - there is an assert present for dealing with this error on line 69.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceLibraryLoader.mm
```diff
@@ -0,0 +1,70 @@
...
+    else
+#endif
+    {
+        NSURL* shaderURL = [NSBundle.mainBundle URLForResource:@"pfMetalPipelineShadersMSL21" withExtension:@"metallib"];
+        fShaderLibrary = fMetalDevice->newLibrary(static_cast<NS::URL*>(shaderURL), &error);
+    }
+    hsAssert(error == nil,  "Unexpected error loading Metal shader library");
+}
```
Hoikas, line 70: Missed on the first review, but this bracing style is incorrect.

colincornaby: Is there a preference for how the else should be braced in since it's interrupted by the preprocessor logic?

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/GammaCorrection.metal
```diff
@@ -79,3 +79,17 @@ fragment half4 gammaCorrectFragment(GammaVertexOut in               [[stage_in]]
         1.0
     };
 }
+
+#if defined(__METAL_IOS__) || __METAL_VERSION__ >= 230
+fragment half4 gammaCorrectFragmentInPlace(GammaVertexOut in        [[stage_in]],
```
dpogue, line 84: Where does this function actually get used? I see the binding to `gammaCorrectFragment` but not conditionally to `gammaCorrectFragmentInPlace`

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
```diff
@@ -1241,12 +1240,9 @@ MTL::SamplerState* plMetalDevice::SampleStateForClampFlags(hsGMatState::hsGMatCl
...
 {
     MTL::RenderPipelineDescriptor* gammaDescriptor = MTL::RenderPipelineDescriptor::alloc()->init();
-    MTL::Library*                  library = fMetalDevice->newDefaultLibrary();
-
-    gammaDescriptor->setVertexFunction(library->newFunction(MTLSTR("gammaCorrectVertex"))->autorelease());
-    gammaDescriptor->setFragmentFunction(library->newFunction(MTLSTR("gammaCorrectFragment"))->autorelease());
 
-    library->release();
+    gammaDescriptor->setVertexFunction(fShaderLibrary->newFunction(MTLSTR("gammaCorrectVertex"))->autorelease());
+    gammaDescriptor->setFragmentFunction(fShaderLibrary->newFunction(MTLSTR("gammaCorrectFragment"))->autorelease());
```
dpogue, line 1245: Need the conditional here to bind to `gammaCorrectFragmentInPlace`?

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/GammaCorrection.metal
colincornaby: There's a follow up PR to actually use the function. This PR was busy enough I didn't want to bring it in. I did want to have an Metal 2.3 only function present though to test that optional Metal 2.3 paths were actually working.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
colincornaby: Unfortunately not that simple. I do have a branch though that makes the necessary changes.

https://github.com/colincornaby/Plasma/commit/4d9d984ea4c64505365dfa5724c6db539b28dbb9

colincornaby: Just checking in on how you're feeling about this @dpogue - I can bring that other branch in but it might be nice to avoid the churn and leave that to a secondary PR.

dpogue: Yeah, that sounds fine to me.

dpogue (approved):
Approved from my side, pending sorting out the bracing comment from Hoikas

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceLibraryLoader.mm
colincornaby: @Hoikas - Just checking in again here. I think this is the last issue holding up the PR.

Hoikas: I don't see a preference listed on the [Plasma Code Style](https://github.com/H-uru/Plasma/wiki/Preferred-Code-Style) wiki page, but my thought would be something like:

```C++
#ifdef FOO
    if (foo) {
        // bar
    } else
#endif
    {
        // meow
    }
```

But that looks pretty bad. I suppose, in light of the preprocessor shenanigans, this is ok.

colincornaby: Thanks. I've updated this chunk to reflect the proper code style. I think thats the last outstanding suggestion here.

dpogue (approved):

PR #1585, opened 2024-04-27
Sanitizer fixes and plProduct static init fix

dpogue:
* plDetectorLog was initializing before plProduct had initialized its static strings, leading to some weird stuff where the User Data Folder path was missing "Uru Live"

* Was running with AddressSanitizer to try to track down some GL Pipeline issues, and ended up tracking down a bunch of other issues and warnings first.

  * It is (apparently) undefined behaviour to cast non-memory-aligned values to int types, which we were doing in a few spots (and are still doing in a bunch more spots). This fixes the easy ones in plUUID and hsStream. Also, `memcpy` is a built-in now so there should be no speed benefit to doing casting.
  * `fRegisteredForTime` (and a few other values) were never being zero initialized in plTransitionMgr
  * plBufferedFileReader's `fBuffer` is created with `new[]` but was being released with `delete`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,8 @@ jobs:
       - name: Test
         run: |
           cmake --build build --target check --config "${{ matrix.cfg.type }}" -j 4
+        env:
+          CTEST_OUTPUT_ON_FAILURE: 1
 
   max:
     # Can only run if we have a token for our super seekrit Max SDK repo. Sad.
@@ -247,6 +249,8 @@ jobs:
       - name: Test
         run: |
           cmake --build build --target check -j 4
+        env:
+          CTEST_OUTPUT_ON_FAILURE: 1
 
       - name: Install
         run: |
@@ -326,9 +330,10 @@ jobs:
           cmake --build build --config "${{ matrix.cfg.type }}" -j 2
 
       - name: Test
-        if: matrix.platform.arch == 'x86_64'
         run: |
           cmake --build build --target check --config "${{ matrix.cfg.type }}" -j 2
+        env:
+          CTEST_OUTPUT_ON_FAILURE: 1
 
       - name: Install
         run: |
--- a/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
+++ b/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
@@ -50,41 +50,28 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ 
 
 struct plUUIDHelper
 {
-    typedef struct _GUID {
-      uint32_t Data1;
-      uint16_t Data2;
-      uint16_t Data3;
-      uint8_t  Data4[8];
-    } GUID;
-    
-    static inline void CopyToPlasma( plUUID * dst, const uuid_t & src )
+    static inline void CopyToPlasma(plUUID* dst, const uuid_t& src)
     {
-        hsAssert( sizeof(uuid_t)==sizeof(dst->fData), "sizeof(uuid_t)!=sizeof(plUUID)" );
-        memcpy( (void*)dst->fData, (const void *)src, sizeof( plUUID ) );
-        
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst->fData);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        memcpy(dst->fData, src, sizeof(plUUID));
+
+        // plUUIDs need to always be encoded in little endian, but the uuid_t struct is big endian
+        // This isn't guaranteed to be 32-bit-aligned, so casting to uint16_t/uint32_t pointers is UB
+        std::swap(dst->fData[0], dst->fData[3]);
+        std::swap(dst->fData[1], dst->fData[2]);
+        std::swap(dst->fData[4], dst->fData[5]);
+        std::swap(dst->fData[6], dst->fData[7]);
     }
 
-    static inline void CopyToNative( uuid_t & dst, const plUUID * src )
+    static inline void CopyToNative(uuid_t& dst, const plUUID* src)
     {
-        hsAssert( sizeof(uuid_t)==sizeof(src->fData), "sizeof(uuid_t)!=sizeof(plUUID)" );
-        memcpy( (void*)dst, (const void *)src->fData, sizeof( plUUID ) );
-        
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        memcpy(dst, src->fData, sizeof(plUUID));
+
+        // uuid_t struct is big endian, but plUUIDs are always encoded in little endian
+        // This isn't guaranteed to be 32-bit-aligned, so casting to uint16_t/uint32_t pointers is UB
+        std::swap(dst[0], dst[3]);
+        std::swap(dst[1], dst[2]);
+        std::swap(dst[4], dst[5]);
+        std::swap(dst[6], dst[7]);
     }
 };
 
--- a/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
+++ b/Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
@@ -133,12 +133,12 @@ bool plUUID::FromString( const char * str )
     return true;
 }
 
-bool plUUID::ToString( ST::string & out ) const
+bool plUUID::ToString(ST::string& out) const
 {
     uuid_t g;
-    plUUIDHelper::CopyToNative( g, this );
+    plUUIDHelper::CopyToNative(g, this);
     char buf[40];
-    uuid_unparse( g, buf );
+    uuid_unparse_lower(g, buf);
     out = buf;
     return true;
 }
--- a/cmake/hsBuildInfo.inl.cmake
+++ b/cmake/hsBuildInfo.inl.cmake
@@ -7,14 +7,14 @@ namespace plProduct
     constexpr uint32_t PRODUCT_BUILD_ID = @PRODUCT_BUILD_ID@;
     constexpr uint32_t PRODUCT_BUILD_TYPE = @PRODUCT_BUILD_TYPE@;
 
-    static const ST::string PRODUCT_CORE_NAME = ST_LITERAL("@PRODUCT_CORE_NAME@");
-    static const ST::string PRODUCT_SHORT_NAME = ST_LITERAL("@PRODUCT_SHORT_NAME@");
-    static const ST::string PRODUCT_LONG_NAME = ST_LITERAL("@PRODUCT_LONG_NAME@");
+    constexpr std::string_view PRODUCT_CORE_NAME = "@PRODUCT_CORE_NAME@";
+    constexpr std::string_view PRODUCT_SHORT_NAME = "@PRODUCT_SHORT_NAME@";
+    constexpr std::string_view PRODUCT_LONG_NAME = "@PRODUCT_LONG_NAME@";
     constexpr std::string_view PRODUCT_UUID = "@PRODUCT_UUID@";
 
-    static const ST::string GIT_REV = ST_LITERAL("@GIT_REV@");
-    static const ST::string GIT_TAG = ST_LITERAL("@GIT_TAG@");
+    constexpr std::string_view GIT_REV = "@GIT_REV@";
+    constexpr std::string_view GIT_TAG = "@GIT_TAG@";
 
-    static const ST::string VOLATILE_BUILD_DATE = ST_LITERAL("@BUILD_DATE@");
-    static const ST::string VOLATILE_BUILD_TIME = ST_LITERAL("@BUILD_TIME@");
+    constexpr std::string_view VOLATILE_BUILD_DATE = "@BUILD_DATE@";
+    constexpr std::string_view VOLATILE_BUILD_TIME = "@BUILD_TIME@";
 };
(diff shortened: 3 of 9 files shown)


Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst->fData);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(dst->fData), "sizeof(uuid_t) != sizeof(plUUID)");
+        memcpy((void*)dst->fData, (const void*)src, sizeof(plUUID));
```
dgelessus, line 56: The casts shouldn't be necessary here. Even in C++, every pointer can be implicitly converted to `void*` (as long as you obey `const`, etc.).

The `src` cast is especially worrying to me. `src` is a `uuid_t&` *reference*, not a pointer, so I'm not sure how `(const void*)src` could work at all...

```suggestion
        memcpy(dst->fData, &src, sizeof(plUUID));
```

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(src->fData), "sizeof(uuid_t) != sizeof(plUUID)");
+        memcpy((void*)dst, (const void*)src->fData, sizeof(plUUID));
```
dgelessus, line 69:
```suggestion
        memcpy(&dst, src->fData, sizeof(plUUID));
```

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst->fData);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(dst->fData), "sizeof(uuid_t) != sizeof(plUUID)");
+        memcpy((void*)dst->fData, (const void*)src, sizeof(plUUID));
+
+        // plUUIDs need to always be encoded in little endian, but the uuid_t struct is big endian
+        // This isn't memory-aligned, so casting to pointers is technically UB
```
dgelessus, line 59:
```suggestion
        // This isn't guaranteed to be 32-bit-aligned, so casting to uint16_t/uint32_t pointers is UB
```

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(src->fData), "sizeof(uuid_t) != sizeof(plUUID)");
+        memcpy((void*)dst, (const void*)src->fData, sizeof(plUUID));
+
+        // uuid_t struct is big endian, but plUUIDs are always encoded in little endian
+        // This isn't memory-aligned, so casting to pointers is technically UB
```
dgelessus, line 72:
```suggestion
        // This isn't guaranteed to be 32-bit-aligned, so casting to uint16_t/uint32_t pointers is UB
```

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-        
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst->fData);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(dst->fData), "sizeof(uuid_t) != sizeof(plUUID)");
```
Hoikas, line 55: This should be replaced with a single compile time `static_assert`.

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,30 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-        
-        /*
-         GUIDs are always internally encoded in little endian,
-         while the uuid_t structure is big endian.
-         */
-        GUID* guid = reinterpret_cast<GUID*>(dst);
-        guid->Data1 = hsSwapEndian32(guid->Data1);
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        hsAssert(sizeof(uuid_t) == sizeof(src->fData), "sizeof(uuid_t) != sizeof(plUUID)");
```
Hoikas, line 68: This should be replaced with a single compile time `static_assert`.

Sources/Plasma/PubUtilLib/plModifier/plDetectorLog.cpp
```diff
@@ -42,15 +42,21 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        return;
+
+#ifndef PLASMA_EXTERNAL_RELEASE
+    fLog = plStatusLogMgr::GetInstance().CreateStatusLog(
+        20,
+        "Detector.log",
+        plStatusLog::kFilledBackground | plStatusLog::kDeleteForMe |
+        plStatusLog::kDontWriteFile | plStatusLog::kAlignToTop);
 #endif
+}
```
Hoikas, line 59: It looks like this is addressing a side effect of `plProduct` not being initialized in `CreateStatusLog()`. Perhaps `plProduct` should make sure it is initialized when `CreateStatusLog()` calls into it. Otherwise, we still have potential static initialization issues.

dgelessus: As far as I can tell, `plProduct` doesn't need to be explicitly initialized in any way - it's all global constants (`static const`). The problem is that (roughly speaking) C++ doesn't specify in what order the global variables of different source files are initialized, so you cannot safely have one global initializer (e. g. `plDetectorLog::fLog`) use another global variable initialized in another file e. g. `plProduct::PRODUCT_LONG_NAME`).

Moving the initialization of `plDetectorLog::fLog` into a method is a sufficient fix, as far as I can tell. As long as the method is never called (even indirectly) from any global initializer, it's guaranteed that all global variables will have been initialized by the time the method can be called.

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
dpogue: There's actually already a static assert at the top of the file, so I'll just remove these hsAssert lines entirely

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
dpogue: `uuid_t` appears to be a typedef for `unsigned char[12]`

colincornaby:
Thanks - the address sanitizer stuff was bugging me for a while. Hopefully this is the same issues I was seeing.

dpogue:
I have dropped the change to plDetectorLog and instead made more of the plProduct strings use `std::string_view`.

Also fixed the x86_64 CI builds on macOS because `macos-latest` now points to macos-14 on arm64, and vcpkg doesn't want to build Python for x86_64 when the host is arm64 🙄

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -50,41 +50,28 @@ static_assert(sizeof(plUUID) == sizeof(uuid_t), "plUUID and uuid_t types differ
...
-        guid->Data2 = hsSwapEndian16(guid->Data2);
-        guid->Data3 = hsSwapEndian16(guid->Data3);
+        memcpy(dst->fData, src, sizeof(plUUID));
+
+        // plUUIDs need to always be encoded in little endian, but the uuid_t struct is big endian
+        // This isn't guaranteed to be 32-bit-aligned, so casting to uint16_t/uint32_t pointers is UB
+        std::swap(dst->fData[0], dst->fData[3]);
+        std::swap(dst->fData[1], dst->fData[2]);
+        std::swap(dst->fData[4], dst->fData[5]);
+        std::swap(dst->fData[6], dst->fData[7]);
```
Hoikas, line 62: While we're here, I wonder if we should add an unit test for uuids?

.github/workflows/ci.yml
```diff
@@ -266,7 +266,7 @@ jobs:
       matrix:
         platform:
           - { str: macos-arm64, arch: arm64, runner: "macos-14" }
-          - { str: macos-x64, arch: x86_64, runner: "macos-latest" }
+          - { str: macos-x64, arch: x86_64, runner: "macos-13" }
```
Hoikas, line 269: This is high impact enough that we should probably split it out into a separate PR for priority merging.

cmake/hsBuildInfo.inl.cmake
```diff
@@ -7,9 +7,9 @@ namespace plProduct
...
 
-    static const ST::string PRODUCT_CORE_NAME = ST_LITERAL("@PRODUCT_CORE_NAME@");
-    static const ST::string PRODUCT_SHORT_NAME = ST_LITERAL("@PRODUCT_SHORT_NAME@");
-    static const ST::string PRODUCT_LONG_NAME = ST_LITERAL("@PRODUCT_LONG_NAME@");
+    constexpr std::string_view PRODUCT_CORE_NAME = "@PRODUCT_CORE_NAME@";
+    constexpr std::string_view PRODUCT_SHORT_NAME = "@PRODUCT_SHORT_NAME@";
+    constexpr std::string_view PRODUCT_LONG_NAME = "@PRODUCT_LONG_NAME@";
     constexpr std::string_view PRODUCT_UUID = "@PRODUCT_UUID@";
 
     static const ST::string GIT_REV = ST_LITERAL("@GIT_REV@");
```
dgelessus, line 15: Should we convert the other `static const ST::string` variables here to `constexpr std::string_view` as well?

dgelessus (approved):
The code changes look good to me overall now. (I have no particular opinions on the things @Hoikas commented)

.github/workflows/ci.yml
dpogue: https://github.com/H-uru/Plasma/pull/1586

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
dpogue: Added some pnUUID unit tests, including one that specifically checks the endianness and passes both before and after this change

cmake/hsBuildInfo.inl.cmake
dpogue: Changed all of these to constexpr to avoid any surprises in the future about initialization ordering

Hoikas:
Looks like we have a test failing on macOS.

dpogue:
> Looks like we have a test failing on macOS.

The pnUUID test was failing on the string comparison due to case differences. I've worked around that for now by using `EXPECT_STRCASEEQ` but I wanted to raise it for discussion.

It seems that Linux and Windows stringify plUUID with lowercase letters. On macOS it's using uppercase letters. I suspect the plUUID `AsString()` is only used for debugging, but do we want to enforce lowercase consistency across all OSes?

dpogue:
Okay, this should be good to go now. I opted to enforce that stringified UUIDs are lowercase for consistency with existing Windows code.

Sources/Plasma/NucleusLib/pnUUID/pnUUID_Unix.cpp
```diff
@@ -133,13 +120,13 @@ bool plUUID::FromString( const char * str )
...
+bool plUUID::ToString(ST::string& out) const
 {
     uuid_t g;
-    plUUIDHelper::CopyToNative( g, this );
+    plUUIDHelper::CopyToNative(g, this);
     char buf[40];
-    uuid_unparse( g, buf );
-    out = buf;
+    uuid_unparse(g, buf);
+    out = ST::string(buf).to_lower();
```
dgelessus, line 129: There seems to be an `uuid_unparse_lower` function in `<uuid/uuid.h>` on both [Linux](https://github.com/util-linux/util-linux/blob/e3192bfd1dd129c70f5416e1464135d8cd22c956/libuuid/src/uuid.h#L108) and [macOS](https://github.com/apple-oss-distributions/xnu/blob/94d3b452840153a99b38a3a9659680b2a006908e/bsd/uuid/uuid.h#L74), which is defined to return lowercase specifically. Can we use that directly?

dpogue: Yep! Good suggestion

colincornaby:
I'll toss in that just for "technically correct"-ness - UUIDs should always be compared case insensitively.

https://datatracker.ietf.org/doc/html/rfc4122

> The hexadecimal values "a" through "f" are output as lower case characters and are case insensitive on input.

Apple's implementation looks out of line from the spec (I'd be curious as to why) as they output in upper case - but not in a way that should actually impact UUID use. Something to keep in mind for any sort of comparisons we do with UUIDs though.

dpogue:
> Something to keep in mind for any sort of comparisons we do with UUIDs though.

Any comparisons of UUIDs should be done bitwise by the plUUID class, this only came to light because I added a unit test that checked the stringified output against a known value.

Hoikas (approved):

PR #1587, opened 2024-05-30
Detect pthread_setname_np with CMake

dpogue:
In some non-standard setups that slip through our checks, this function doesn't exist and we should guard against calling it. Since it's only used for debugging, let's just use CMake to check if it's defined in the header rather than complicated conditionals in the code for all the different possible versions of this that might exist (or not exist).

### DIFF
--- a/Sources/Plasma/CoreLib/hsThread_Unix.cpp
+++ b/Sources/Plasma/CoreLib/hsThread_Unix.cpp
@@ -78,6 +78,7 @@ int clock_gettime(int clocktype, struct timespec* ts)
 
 void hsThread::SetThisThreadName(const ST::string& name)
 {
+#ifdef HAVE_PTHREAD_SETNAME_NP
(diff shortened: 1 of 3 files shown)


Sources/Plasma/CoreLib/hsThread_Unix.cpp
```diff
@@ -86,6 +87,7 @@ void hsThread::SetThisThreadName(const ST::string& name)
     // On Linux, thread names must fit into 16 bytes, including the terminator.
     int res = pthread_setname_np(pthread_self(), name.left(15).c_str());
     hsAssert(res == 0, "Failed to set thread name");
+#endif
 #endif
```
Hoikas, line 91: Since we now have two different preprocessor blocks, it would be nice to have comments at each `#endif` indicating which block is ending.

Hoikas (approved):

zrax (approved):

PR #1588, opened 2024-05-30
Don't try to enable Metal on older Macs

dpogue:
Metal was introduced in OS X 10.11 El Capitan, so the framework doesn't exist on older Apple systems. Instead of trying to unconditionally use it on Apple devices, check for the framework first.

Note: Metal still only works with the Xcode project generator, and not with Makefiles/Ninja because CMake doesn't yet know how to handle the metal shaders.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ find_package(ZLIB REQUIRED)
 
 if(APPLE)
     find_package(Security)
+    find_package(Metal)
 elseif(UNIX)
     find_package(LIBSECRET)
     find_package(Uuid REQUIRED)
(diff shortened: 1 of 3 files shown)


colincornaby:
I think technically the fix is to check the macOS SDK version. That would handle the edge case you mentioned. We only want to try to build Metal when the macOS 10.11 SDK is present.

Unfortunately that's sent me down a Stack Overflow rabbit hole on how to check for that. Let me dig a little deeper. Ideally we'd detect for the Metal SDK just like we would for the Direct3D SDK or any other library.

colincornaby (approved):

Hoikas (approved):

PR #1590, opened 2024-06-02
Optimized gamma correction on Apple Silicon

colincornaby:
This is a follow up to #1581. Previous pull request added an optional path for Metal 2.3 - this PR uses it.

In Metal - we have to manually do gamma correction. Traditionally - this has meant:

- Rendering the main render pass into an intermediate buffer we create.
- Ending the main render pass.
- Creating a new render pass that binds the actual framebuffer as the render target, and the intermediate buffer as a source texture.
- Executing a shader that passes through pixels from the intermediate buffer to the final buffer - doing the gamma correction.

Apple Silicon can read and write safely from a framebuffer - because each portion of the framebuffer is directly owned by a GPU core in its tile memory. So we don't need two passes, and we don't need an intermediate buffer. We render straight to the main framebuffer, and then execute a shader that reads each pixel value from the main buffer, and writes the corrected value back directly to the framebuffer. This improves performance (render pass changes on Apple Silicon can be expensive) and reduces memory use.

On macOS - Apple Silicon functionality requires shaders compiled against Metal 2.3. Metal 2.3 shaders also only run on macOS 11 - so we need to maintain Metal 2.1 shaders for earlier macOS versions. (macOS 11 is the first version to support Apple Silicon Macs, so we don't need to worry about an Apple Silicon Mac running Metal 2.1.)

Additionally - we need to do a check to make sure the GPU is an Apple GPU before attempting to load the optimized shader. This check has been surfaced to the device class as a tile memory capability.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
@@ -344,7 +344,7 @@ void plMetalDevice::BeginNewRenderPass()
         } else {
             renderPassDescriptor->colorAttachments()->object(0)->setTexture(fCurrentFragmentMSAAOutputTexture);
 
-            // if we need postprocessing, output to the main pass texture
+            // if we need postprocessing, output to the intermediate main pass texture
             // otherwise we can go straight to the drawable
             
             // We only need the intermediate texture for post processing on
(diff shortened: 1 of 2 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -259,20 +259,31 @@ class plMetalDevice
...
-    void BeginNewRenderPass();
-    void ReleaseSamplerStates();
-    void ReleaseFramebufferObjects();
-
-    // Blur states
-    std::unordered_map<float, NS::Object*> fBlurShaders;
+    // MARK:  - Device capabilities
+public:
+    /// Returns true if the device supports title memory features such as directly writable render buffers.
+    BOOL SupportsTileMemory() { return fSupportsTileMemory; }
```
dpogue, line 284: `const`

Also, is this ever used outside of this class? Does it need a public getter vs just accessing `fSupportsTileMemory` directly?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
```diff
@@ -344,7 +346,10 @@ void plMetalDevice::BeginNewRenderPass()
 
             // if we need postprocessing, output to the main pass texture
             // otherwise we can go straight to the drawable
```
dpogue, line 348: Is this comment still accurate (vs the updated one added below)?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
colincornaby: Probably does not need to be public. Knowing if the GPU supports tile memory could influence which shader the pipeline picks to do something - but so far most of that has been handled by the device. Especially because the device class has ended up doing a lot of the heavy lifting to make Metal look more like a traditional device to the pipeline.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
colincornaby: I think the comment is still accurate. I'll add a tad bit more language to clarify.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
colincornaby: I kept the getter just for cleanliness - but did mark it as inline. I also made it private.

My hunch is someday the pipeline might need it, but it should be fine making it private for now.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -259,20 +259,30 @@ class plMetalDevice
...
-    void ReleaseSamplerStates();
-    void ReleaseFramebufferObjects();
-
-    // Blur states
-    std::unordered_map<float, NS::Object*> fBlurShaders;
+    // MARK:  - Device capabilities
+private:
+    /// Returns true if the device supports tile memory features such as directly writable render buffers.
+    inline BOOL SupportsTileMemory() const { return fSupportsTileMemory; }
+    BOOL fSupportsTileMemory;
```
dpogue, line 285: This should be a `bool` rather than a `BOOL`

MTL::Device `supportsFamily` returns a `bool` in C++: https://github.com/bkaradzic/metal-cpp/blob/a63bd172ddcba73a3d87ca32032b66ad41ddb9a6/Metal/MTLDevice.hpp#L388

dpogue (approved):
LGTM

PR #1591, opened 2024-06-02
Fix a bad assert in hsRAMStream

dpogue:
This was added because UBSan was warning that we were calling memcpy with a null destination, and while that is technically true, it was also a no-op because it was copying 0 bytes. A null buffer gets properly resized with the data in the next step, so this is actually expected behaviour that is broken by asserting too early.

Instead, detect the cases where we're calling memcpy with actual data, and assert in those cases that the RAMStream buffer is non-null.

### DIFF
--- a/Sources/Tests/CoreTests/test_RAMStream.cpp
+++ b/Sources/Tests/CoreTests/test_RAMStream.cpp
@@ -1,0 +1,61 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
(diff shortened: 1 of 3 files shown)


dpogue:
I added a (pretty minimal) unit test here that would trigger the previous assert, and is working now with the modified asserts.

colincornaby (approved):

Sources/Tests/CoreTests/test_RAMStream.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+*==LICENSE==*/
+
+#include <gtest/gtest.h>
+
+#include "hsStream.h"
+
+TEST(hsRAMStream, initializeBufferOnFirstWrite)
+{
+    const char* str = "hsRAMStream initializeBufferOnFirstWrite";
```
Hoikas, line 49:
```suggestion
    constexpr std::string_view str = "hsRAMStream initializeBufferOnFirstWrite";
```

Sources/Tests/CoreTests/test_RAMStream.cpp
```diff
@@ -0,0 +1,61 @@
...
+    hsRAMStream s;
+
+    s.WriteSafeString(str);
+    EXPECT_EQ(s.GetPosition(), strlen(str) + 2);
```
Hoikas, line 53:
```suggestion
    EXPECT_EQ(s.GetPosition(), str.size() + 2);
```

Sources/Tests/CoreTests/test_RAMStream.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    s.WriteLE32(1);
+    EXPECT_EQ(s.GetPosition(), strlen(str) + 2 + 4);
```
Hoikas, line 56:
```suggestion
    EXPECT_EQ(s.GetPosition(), str.size() + 2 + 4);
```

Sources/Tests/CoreTests/test_RAMStream.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    s.Skip(-4);
+    s.WriteLE32(5);
+    EXPECT_EQ(s.GetPosition(), strlen(str) + 2 + 4);
```
Hoikas, line 60:
```suggestion
    EXPECT_EQ(s.GetPosition(), str.size() + 2 + 4);
```

Sources/Tests/CoreTests/test_RAMStream.cpp
```diff
@@ -0,0 +1,61 @@
...
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include <gtest/gtest.h>
+
```
Hoikas, line 44:
```suggestion
#include <string_view>

```

Hoikas (commented):
Some comments about what the unit test is doing would potentially be a good addition.

PR #1596, opened 2024-06-30
ModDLL & Linux hsGD3DeviceSelector fixes

dpogue:
* There's a static initializer ordering issue that sometimes crops up with the `hsG3DDeviceSelector` Device Enumerator vector on Linux which causes crashes (not that there's much of a client to crash at the moment). Easy enough to change the vector to a function static rather than a global static to work around the problem.

* Support loading ModDLLs on Linux and macOS using `dlopen` and `dlsym`.

### DIFF
--- a/Sources/Plasma/Apps/plClient/plClient.cpp
+++ b/Sources/Plasma/Apps/plClient/plClient.cpp
@@ -62,6 +62,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "pnDispatch/plDispatch.h"
 #include "pnDispatch/plDispatchLogBase.h"
+#include "pnFactory/plFactory.h"
 #include "pnKeyedObject/plFixedKey.h"
 #include "pnKeyedObject/plKey.h"
 #include "pnMessage/plAudioSysMsg.h"
@@ -158,6 +159,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pfPython/cyMisc.h"
 #include "pfPython/cyPythonInterface.h"
 
+#ifdef HS_BUILD_FOR_UNIX
+#    include <dlfcn.h> // For ModDLL loading
+#endif
 
 #define MSG_LOADING_BAR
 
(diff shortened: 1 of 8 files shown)


Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -373,6 +377,58 @@ bool plClient::Shutdown()
...
+#else
+        hsLibraryHndl mod = dlopen(iter->AsString().c_str(), RTLD_LAZY | RTLD_LOCAL);
+#endif
+
+        if (mod)
+        {
+#ifdef HS_BUILD_FOR_WIN32
+            pInitGlobalsFunc initGlobals = (pInitGlobalsFunc)GetProcAddress(mod, "InitGlobals");
+#else
+            pInitGlobalsFunc initGlobals = (pInitGlobalsFunc)dlsym(mod, "InitGlobals");
```
dgelessus, line 406: Can we use explicit `static_cast` or `reinterpret_cast` here? (I haven't checked which one is correct here - it might need to be `reinterpret_cast` for the conversion from `void*` to a function pointer...)

Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -373,6 +377,58 @@ bool plClient::Shutdown()
...
+}
+
+void plClient::ShutdownDLLs()
+{
+    for (hsLibraryHndl mod : fLoadedDLLs)
+    {
+#ifdef HS_BUILD_FOR_WIN32
+        BOOL ret = FreeLibrary(mod);
+        if (!ret)
+            hsStatusMessage("Failed to free lib\n");
```
dgelessus, line 423: No need for an explicit newline (though as far as I can tell it doesn't hurt either).

```suggestion
            hsStatusMessage("Failed to free lib");
```

Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -373,6 +377,58 @@ bool plClient::Shutdown()
...
+#else
+        dlclose(mod);
```
dgelessus, line 425: `dlclose` also has a status return and can *theoretically* fail...

```suggestion
        int ret = dlclose(mod);
        if (ret != 0)
            hsStatusMessage(ST::format("Failed to free lib: {}", dlerror()).c_str());
```

dgelessus (commented):
Hooray, I no longer need to manually patch `hsG3DDeviceSelector` every time I test on Linux 😄

Hoikas:
The last two commits should probably be in separate PRs.

dpogue:
> The last two commits should probably be in separate PRs.

Split to https://github.com/H-uru/Plasma/pull/1597

Hoikas (approved):

PR #1598, opened 2024-07-06
FindPhysX should search specific lib names first

dpogue:
This resolves an issue where PhysX libraries installed with generic names at a system level were being found preferentially ahead of the ones created by vcpkg.

### DIFF
--- a/cmake/FindPhysX.cmake
+++ b/cmake/FindPhysX.cmake
@@ -139,10 +139,15 @@ elseif(NOT VCPKG_TOOLCHAIN)
     set(_PHYSX_RELEASE_PATHS ${CMAKE_PREFIX_PATH})
 endif()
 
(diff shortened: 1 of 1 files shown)


Hoikas:
Have you tried modifying the `find_library()` calls to pass the `NO_SYSTEM_ENVIRONMENT_PATH` argument instead?

dpogue:
> Have you tried modifying the `find_library()` calls to pass the `NO_SYSTEM_ENVIRONMENT_PATH` argument instead?

But we'd only want to do that if `USE_VCPKG` is set, right?

Hoikas:
It looks like the find module is testing `if(VCPKG_TOOLCHAIN)` in places.

dpogue:
Okay, will give that a shot

Hoikas (approved):

Issue #1599, opened 2024-07-06
[macOS] Caps Lock does not turn on running


dpogue:
Turning <kbd>Caps Lock</kbd> on and off does not affect the movement speed on macOS. Holding <kbd>Shift</kbd> to run works fine.

The bit of extra complexity here is detecting the caps lock state on startup, and monitoring for changes to caps lock state while Uru does not have focus, but it seems like Cocoa events provide ways to do that.

colincornaby:
I'll take a look.

This also reminds me that there is an open question of optionally adopting the Game Controller framework on newer versions - which provides a more sane interface for games to get keyboard input. This would also be required for iOS/visionOS.

We'd need this code path for older macOS versions though. So I'll look at patching this path.

dpogue:
The fix in #1601 works well enough for me to consider this closed.

PR #1601, opened 2024-07-07
Fixing caps lock input on macOS

colincornaby:
This is for fixing #1599.

I don't see any facilities in Cocoa to query the current state of caps lock, only updates. It seems like that might be present in HID - but HID is gated by security permissions in recent macOS releases. So far I've considered HID out of bounds for macOS.

@dpogue - If you have any Cocoa APIs I don't know about I'd be happy to implement. Would also be happy to forward on to Apple for guidance. I didn't see anything during my research online - and it does not even seem like the Game Controller framework supports querying Caps Lock. But my investigation was pretty shallow.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSKeyboardEventMonitor.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSKeyboardEventMonitor.mm
@@ -152,6 +152,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
     if (keycode == kVK_Control) {
         down = (event.modifierFlags & NSEventModifierFlagControl) != 0;
     }
+    if (keycode == kVK_CapsLock) {
+        down = (event.modifierFlags & NSEventModifierFlagCapsLock) != 0;
+    }
 
     /*
      This gets weird.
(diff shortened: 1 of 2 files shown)


Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSKeyboardEventMonitor.mm
```diff
@@ -171,8 +177,11 @@ - (BOOL)processKeyEvent:(NSEvent*)event
     }
 
     @synchronized(self.view.layer) {
-        self.inputManager->HandleKeyEvent(
-            (plKeyDef)keycode, down, event.type == NSEventTypeFlagsChanged ? false : event.ARepeat);
+        // Caps lock modifer has sepcial handling that was earlier
```
dpogue, line 180: typo: sepcial -> special

dpogue (approved):
This looks alright to me

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSKeyboardEventMonitor.mm
colincornaby: Fixed.

PR #1602, opened 2024-07-08
Prevent display sleep while game is running on macOS

### LINKED ISSUE #1600: [macOS] Uru should (optionally) hold a wakelock assertion
Would be nice to have a "Prevent system from going to sleep" checkbox in the options while the game is running

dpogue:
Closes #1600

### DIFF
--- a/Sources/Plasma/Apps/plClient/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plClient/CMakeLists.txt
@@ -243,6 +243,8 @@ target_link_libraries(
         CURL::libcurl
         "$<$<PLATFORM_ID:Darwin>:-framework Cocoa>"
         "$<$<PLATFORM_ID:Darwin>:-framework QuartzCore>"
+        # Only need IOKit when targeting Mac OS X < 10.9
+        "$<$<AND:$<PLATFORM_ID:Darwin>,$<VERSION_LESS:${CMAKE_OSX_DEPLOYMENT_TARGET},10.9>>:-framework IOKit>"
         $<$<PLATFORM_ID:Darwin>:${LibArchive_LIBRARIES}>
 )
 target_include_directories(plClient PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -50,6 +50,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #endif
 #import <QuartzCore/QuartzCore.h>
 
+#if defined(HS_BUILD_FOR_MACOS) && MAC_OS_X_VERSION_MIN_REQUIRED < 1090
+#import <IOKit/pwr_mgt/IOPMLib.h>
+#endif
+
 // Cocoa client
 #import "NSString+StringTheory.h"
 #import "PLSKeyboardEventMonitor.h"
@@ -65,6 +69,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <unordered_set>
 
 // Plasma engine
+#include "hsDarwin.h"
 #include "plClient/plClient.h"
 #include "plClient/plClientLoader.h"
 #include "plCmdParser.h"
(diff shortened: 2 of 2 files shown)


dpogue:
> Was NSProcessInfo beginActivityWithOptions:reason: considered as an alternative?

I did not know that existed... all my investigation led to the IOKit API, which is also what's used internally by the `caffeinate` command-line tool.

One advantage of the IOKit code (besides it being C API and not ObjC 😛) is that it works back to OS X 10.5, whereas the NSProcessInfo code looks to be 10.7 and up only.

Do you know if there's an actual difference in implementation between the two, or does NSProcessInfo just call out to IOKit?

colincornaby:
beginActivity: seems to require 10.9 - which would be even higher. But - beginActivity also controls our QoS level in the operating system, which the IOKit call does not do. The reason I've been suspicious we might need this call is that our login window will not appear until the app is brought to the foreground - which could imply our run loop is getting squashed.

Application/backgrounding QoS was not introduced until later. Very possibly that release was 10.9. So the IOKit call would not cover that.

It's a bit of a mess, but:
- We should probably make the AppKit call in 10.9 and later because that would preserve our background QoS.
- Before that, we could make IOKit calls.

FWIW - I'd prefer if there was a compile path that did not include IOKit at all for later builds. It's not critical - but IOKit has bit by bit been getting deprecated or restricted. So it would be preferable if the modern builds didn't link IOKit just to reduce risk. Could just be that I'm overly cautious on IOKit use - but again, been seeing a lot of IOKit API getting deprecated.

Hoikas:
What's the status here? It would nice to get this resolved.

Hoikas:
I'm converting this to a draft because there is outstanding feedback and to clarify from the main PR page what the status is.

colincornaby:
Following up on this one...

I'd still like to get a 'beginActivityWithOptions:reason:' alternative in because there are other flags that probably need to be reset such as background, user interactive, etc. Plasma sometimes acts funny when it's not in the foreground, so flagging that we do things in the background might be helpful. (Also, we should declare that anyway because we stay talking to a server in the background.)

This IOKit call also doesn't exist on iOS, where 'beginActivityWithOptions:reason:'  does.

I don't mind the IOKit call being there if the version of OS X is too old. But I don't think we can rely on IOKit alone being used.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -50,6 +50,10 @@
 #endif
 #import <QuartzCore/QuartzCore.h>
 
+#if MAC_OS_X_VERSION_MIN_REQUIRED < 1090
```
dgelessus:
```suggestion
#if defined(HS_BUILD_FOR_MACOS) && MAC_OS_X_VERSION_MIN_REQUIRED < 1090
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -532,6 +537,21 @@ - (void)startClient
     ((PLSView*)self.window.contentView).inputManager = gClient->GetInputManager();
     [self.window makeFirstResponder:self.window.contentView];
 
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1090
```
dgelessus:
```suggestion
#if !defined(HS_BUILD_FOR_MACOS) || MAC_OS_X_VERSION_MIN_REQUIRED >= 1090
```

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -243,6 +243,7 @@ target_link_libraries(
         CURL::libcurl
         "$<$<PLATFORM_ID:Darwin>:-framework Cocoa>"
         "$<$<PLATFORM_ID:Darwin>:-framework QuartzCore>"
+        "$<$<PLATFORM_ID:Darwin>:-framework IOKit>"
```
dgelessus: This line is already correct for non-macOS, I think. According to [the CMake docs for `CMAKE_SYSTEM_NAME`](https://cmake.org/cmake/help/v4.1/variable/CMAKE_SYSTEM_NAME.html#system-names-known-to-cmake) (which is what `PLATFORM_ID` corresponds to), `Darwin` actually means only macOS. For the other Apple OSes, `CMAKE_SYSTEM_NAME` is `iOS`, `tvOS`, etc.

If we want to be *really* correct, this generator expression would also have to check if we're targeting OS X older than 10.9, but that may be pointless complexity. (At least until Apple decides that linking against IOKit is verboten starting with macOS 29 or whatever.)

dgelessus (commented):
> There's probably something to be done to only do the version check on macOS and not iOS/visionOS/etc.

Would this work?

Sources/Plasma/Apps/plClient/CMakeLists.txt
dpogue: There's probably some `-weak_framework` or something we could use to make it only loaded if needed (but I don't know how far back that flag goes, it might not work on OSX 10.5 - 10.9)

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
dpogue: I think Apple defined `TARGET_OS_MAC` as well... feels a bit weird to mix and match Plasma and Apple macros, but that should work

Sources/Plasma/Apps/plClient/CMakeLists.txt
dgelessus: If we really care, I think this would work (untested). There's no generator expression for `CMAKE_OSX_DEPLOYMENT_TARGET`, so it has to be substituted in as a variable. This should be safe, because the value can't change depending on the build type or anything - only if the user re-runs CMake.

```suggestion
        "$<$<AND:$<PLATFORM_ID:Darwin>,$<VERSION_LESS:${CMAKE_OSX_DEPLOYMENT_TARGET},10.9>>:-framework IOKit>"
```

dpogue:
Pretty sure everyone is going to hate this latest revision, but it should support everything properly... 😛

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -99,6 +104,80 @@
...
+#endif
+#if defined(HS_BUILD_FOR_MACOS) && MAC_OS_X_VERSION_MIN_REQUIRED < 1090
+        fAssertionID = 0;
+#endif
+    }
+
+    void startUserActivity() {
+#if !defined(HS_BUILD_FOR_MACOS) || MAC_OS_X_VERSION_MIN_REQUIRED >= 1090
+        fActivity = [[NSProcessInfo processInfo]
+                        beginActivityWithOptions:(NSActivityUserInitiated | NSActivityIdleDisplaySleepDisabled)
```
colincornaby: This looks good, but could we also get a NSActivityBackground while we're here? I think that's necessary to ensure our background networking isn't messed with.

PR #1604, opened 2024-07-15
Fixing banner pngs being included in root folder of Mac client

colincornaby:
There's an extra `install` function call that seems to be causing the pngs to be copied to the client directory. This history of this seems to go back to 6fa6565 which was a commit about fixing Metal shaders. Metal shaders do not seem to be affected by this change now. The other possibility is that the Mac build didn't use to call plasma_executable (which I believe is where install dirs are supposed to be set) so this might have been patching over that missing bit.

### DIFF
--- a/Sources/Plasma/Apps/plClient/CMakeLists.txt
+++ b/Sources/Plasma/Apps/plClient/CMakeLists.txt
@@ -165,10 +165,6 @@ if(APPLE)
     set_source_files_properties(Mac-Cocoa/Assets.xcassets ${RESOURCES} PROPERTIES
          MACOSX_PACKAGE_LOCATION Resources
     )
-    install(
-        TARGETS plClient
-        DESTINATION client
-    )
     
     if(${CMAKE_VERSION} VERSION_LESS 3.28)
         message(FATAL_ERROR "Cannot build Mac client without CMake 3.28")


colincornaby:
Fun FYI - Xcode seems to re-encode the two banner pngs to a single tiff when it copies to the resource directory. So the two bare pngs are never actually included in the Mac client application - but they were getting copied to the client directory.

Hoikas (approved):

dpogue:
> Fun FYI - Xcode seems to re-encode the two banner pngs to a single tiff when it copies to the resource directory. So the two bare pngs are never actually included in the Mac client application - but they were getting copied to the client directory.

How will this behave for non-Xcode builds on Mac?

colincornaby:
> > Fun FYI - Xcode seems to re-encode the two banner pngs to a single tiff when it copies to the resource directory. So the two bare pngs are never actually included in the Mac client application - but they were getting copied to the client directory.
> 
> How will this behave for non-Xcode builds on Mac?

I'm not completely sure how this will behave under different tooling. I think copying the pngs directly will still work - what Xcode is doing is probably an optimization. CMake is responsible for emulating Xcode's behavior outside of Xcode - and I would assume the CMake side is doing something rational.

That's probably ok unless our intention is to drop support for xcodebuild. However - my personal thinking for official builds is just as we use msbuild on the Windows side we should probably stick to xcodebuild for CI on the Mac side.

dpogue (approved):

PR #1605, opened 2024-07-20
Use system wide file locks to synchronize log files.

Hoikas:
This is an alternative to #1603, inspired by https://github.com/H-uru/Plasma/pull/1603#issuecomment-2235700787.

The global semaphore has been observed to not be released on unix like platforms when the process exits. So, we convert to file locks, which will be released on process exit on unix like platforms. Unfortunately, Win32 file locks and POSIX file locks have slightly different semantics in that Win32 locks are mandatory and POSIX locks are advisory, which can result in slightly different behavior. OTOH, this should fix the client deadlocking until the system is restarted, so this should be an overall win.

### DIFF
--- a/Sources/Plasma/CoreLib/CMakeLists.txt
+++ b/Sources/Plasma/CoreLib/CMakeLists.txt
@@ -6,6 +6,7 @@ set(CoreLib_SOURCES
     hsExceptions.cpp
     hsExceptionStack.cpp
     hsFastMath.cpp
+    hsFILELock.cpp
     hsGeometry3.cpp
     hsMatrix33.cpp
     hsMatrix44.cpp
@@ -43,6 +44,7 @@ set(CoreLib_HEADERS
     hsExceptions.h
     hsExceptionStack.h
     hsFastMath.h
+    hsFILELock.h
     hsGeometry3.h
     hsLockGuard.h
     hsMatrix44.h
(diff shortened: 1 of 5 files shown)


dpogue (approved):
This looks reasonable to me.

I don't like the `hsFILELock` name (would prefer `hsFileLock`), but it locks a `FILE*` so I guess it's accurate. We prefer this in CoreLib rather than PubUtilLib/plFile?

colincornaby:
Has this been tested on macOS yet? It's implied that file locks should automatically release on macOS - but I didn't check myself.

Hoikas:
> Has this been tested on macOS yet? It's implied that file locks should automatically release on macOS - but I didn't check myself.

I tested it on my mac mini insofar as making sure that running multiple clients did not break in surprising ways. I haven't done any testing where the client crashes while the lock is being held, yet.

colincornaby:
Not a huge surprise - but I validated that macOS will release the file locks if the process is unexpectedly terminated.

Hoikas:
> Not a huge surprise - but I validated that macOS will release the file locks if the process is unexpectedly terminated.

Thanks! I've been forgetting to fire up the Mac and verify that. Did you also perchance check to see if the client deadlocks without this?

colincornaby:
> > Not a huge surprise - but I validated that macOS will release the file locks if the process is unexpectedly terminated.
> 
> Thanks! I've been forgetting to fire up the Mac and verify that. Did you also perchance check to see if the client deadlocks without this?

I've had it deadlock before and have had to manually delete the lock file that was created. My assumption had been the Windows client had the same issue - until I read the source.

Hoikas:
Oh, this is still open. Any objections to this being merged?

colincornaby:
Nothing from me

PR #1609, opened 2024-08-06
Update building/running instructions for macOS

### LINKED ISSUE #1608: macOS build instructions need updating for full client build
The ["Building on macOS & Linux" instructions](https://h-uru.github.io/Plasma/building.html#building-nix) currently say:

> Linux and macOS are **not officially supported targets**, and the game client does not compile. However, several of the tools can be built and run.

This should be updated to mention that macOS has a full client now.

While trying to guide someone through the macOS build process, I also noticed some other things that aren't mentioned in the instructions, and a few small errors:

* There seems to be a mistake in the CMake options: for the configure step, it should be `-B` instead of `--builddir`, and for the build step, it should be `--build` instead of `--builddir`. (This applies to Linux as well. I can't find a `--builddir` option documented anywhere.)
* The CMake command for the configure step needs `-G Xcode` to support building the Mac client. (This means that macOS and Linux need different CMake configure commands now.)
* The instructions don't mention what the minimum required/supported macOS version is. According to various places in the CMake build scripts, macOS 10.14 seems to be the minimum for building the client. The tools can probably built with much older versions.
* `-DUSE_VCPKG=ON` should probably be recommended as the default for macOS builds. At least PhysX is otherwise difficult/impossible to install via the usual macOS package managers (Homebrew, MacPorts).
* Some of the vcpkg dependencies require extra commands that are not part of the standard macOS developer tools. It should be mentioned somewhere that these need to be manually installed via a different package manager before trying to use vcpkg.
  * `pkg-config` is required by `zlib` (and probably others)
  * `python3` is required by `vcpkg-tool-meson` (recent Xcode versions include some version of Python 3 I believe, but older ones don't)
  * `nasm` is required by `libjpeg-turbo` and `libvpx` (and possibly others)

dpogue:
Hopefully fixes #1608

### DIFF
--- a/building.html
+++ b/building.html
@@ -27,8 +27,8 @@
 
       <h2>Building</h2>
 
-      <p>Currently, the complete game client can only be built for Windows, using Visual Studio.</p>
-      <p>By default these steps will compile a client targeting your machine architecture. It is possible to build a 64-bit game client, but only the 32-bit (x86) client is officially supported.</p>
+      <p>Currently, the complete game client can only be built for Windows (using Visual Studio) and macOS (using Xcode). Several tools can be built for Linux, but not a complete game client.</p>
+      <p>By default these steps will compile a client targeting your machine architecture. It is possible to build a 64-bit game client, but only the 32-bit (x86) Windows client is officially supported.</p>
 
 
       <h3 id="dependencies">Library Dependencies</h3>
@@ -64,7 +64,7 @@
           <li><a href="https://www.webmproject.org/">VPX and WebM</a> (for video)</li>
           <li><a href="https://www.opus-codec.org/">Opus</a> (for video and voice chat)</li>
           <li><a href="https://www.speex.org/downloads/">Speex</a> (for legacy voice chat)</li>
-          <li>Autodesk 3ds Max 7 (non-free SDK; for 3ds Max plugin)</li>
+          <li>Autodesk 3ds Max 7 or newer (non-free SDK; for 3ds Max plugin)</li>
         </ul>
       </details>
 
@@ -151,11 +151,13 @@
       </ol>
 
 
-      <h3 id="building-nix">Building on macOS &amp; Linux</h3>
+      <h3 id="building-mac">Building on macOS</h3>
 
-      <p>Linux and macOS are <strong>not officially supported targets</strong>, and the game client does not compile. However, several of the tools can be built and run.</p>
+      <p>Compiling on macOS requires a minimum of macOS 11 and the Xcode build system. Xcode is available as a free download from the <a href="https://apps.apple.com/us/app/xcode/id497799835?mt=12/">App Store</a>.</p>
 
-      <p>You will need to ensure all the required dependencies are available. On macOS, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code>. You can also use vcpkg on these platforms for automatically managing dependencies.</p>
+      <p>We strongly encourage using vcpkg to automatically build dependencies; however, this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config. They can be installed using common package managers such as Homebrew or MacPorts.</p>
+
+      <p>For Homebrew users who wish to manage all dependencies, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code> after cloning the project.</p>
 
       <ol>
         <li>
@@ -165,19 +167,55 @@
 
         <li>
           <p>In the folder where you cloned the repository, run CMake to create a build folder:<br>
-            <code>cmake --builddir build -S .</code></p>
-          <p>If you want to use vcpkg to automatically build dependencies, enable the <tt>USE_VCPKG</tt> option:<br>
-            <code>cmake -D USE_VCPKG=ON --builddir build -S .</code></p>
+            <code>cmake -G Xcode -DUSE_VCPKG=ON -B build -S .</code></p>
+          <p>This assumes you wish to use vcpkg to build dependencies. Remove <tt>-DUSE_VCPKG=ON</tt> to manage dependencies yourself.</p>
         </li>
 
         <li>
           <p>Compile the project with the following command:<br>
-            <code>cmake --builddir build --config RelWithDebInfo</code></p>
+            <code>cmake --build build --config RelWithDebInfo</code></p>
         </li>
 
         <li>
           <p>Install the resulting tools with the following command:<br>
-            <code>cmake --install build --config RelWithDebInfo</code></p>
+            <code>cmake --build build --target install --config RelWithDebInfo</code></p>
+        </li>
+
+        <li>
+          <p>The client will be built and installed into the <tt>build/install/client</tt> subfolder of where you cloned the repo.</p>
+          <p>The client output folder will be referred to as your “MOUL-OS” folder in the instructions on <a href="running.html">Running Plasma</a>.</p>
+        </li>
+      </ol>
+
+      <p>You can also open the project in the Xcode IDE by opening the <tt>build/Plasma.xcodeproj</tt> file.</p>
+
+      <h3 id="building-nix">Building on Linux</h3>
+
+      <p>Linux and other UNIX-like operating systems are <strong>not officially supported targets</strong>, and the game client does not compile. However, several of the tools can be built and run.</p>
+
+      <p>You will need to ensure all the required dependencies are available. You can also use vcpkg on these platforms for automatically managing dependencies.</p>
+
+      <ol>
+        <li>
+          <p>Clone the repository from GitHub, including all submodules, using a git client. On the command line, this can be done by executing the following:<br>
+            <code>git clone --recurse-submodules https://github.com/H-uru/Plasma.git</code></p>
+        </li>
+
+        <li>
+          <p>In the folder where you cloned the repository, run CMake to create a build folder:<br>
+            <code>cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -B build -S .</code></p>
+          <p>If you want to use vcpkg to automatically build dependencies, enable the <tt>USE_VCPKG</tt> option:<br>
+            <code>cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_VCPKG=ON -B build -S .</code></p>
+        </li>
+
+        <li>
+          <p>Compile the project with the following command:<br>
+            <code>cmake --build build</code></p>
+        </li>
+
+        <li>
+          <p>Install the resulting tools with the following command:<br>
+            <code>cmake --build build --target install</code></p>
         </li>
 
         <li>
--- a/running.html
+++ b/running.html
@@ -35,14 +35,17 @@
           <p>You should <strong>never</strong> use a custom-built client with servers that are not explicitly intended for testing. Newer features in the client code might be incompatible with those servers and could break the game for other players or result in permanent player data corruption.</p>
       </div>
 
-
-      <h3 id="running-windows">Running on Windows</h3>
+      <p>Before running the client, you will need to do the following:</p>
 
       <ol>
         <li><p>Copy the folders <tt>avi</tt>, <tt>dat</tt>, and <tt>sfx</tt> from your existing MOULa installation to the MOUL-OS folder.</p></li>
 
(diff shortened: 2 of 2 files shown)


running.html
```diff
@@ -72,14 +71,19 @@ <h3 id="running-vs">Running in Visual Studio (for debugging)</h3>
...
-      <h3 id="running-linux">Running on Linux</h3>
+      <h3 id="running-mac">Running on macOS</h3>
 
-      <ol>
-        <li><p>Copy the folders <tt>avi</tt>, <tt>dat</tt>, and <tt>sfx</tt> from your existing MOULa installation to the MOUL-OS folder.</p></li>
+      <ol start="3">
+        <li><p>Open a terminal window in your MOUL-OS folder.</p></li>
 
-        <li><p>Get the appropriate <tt>server.ini</tt> file for the server you will be connecting to.</p></li>
+        <li><p>Run <code>./plClient.app/Contents/MacOS/plClient --LocalData</code> to start the client.</p></li>
```
dpogue, line 79: @colincornaby Is this the best option, or does something like `open plClient.app --LocalData` work?

Hoikas: On my mini, I've been using `open -n ./plClient.app --args -LocalData`.

colincornaby: I've been using `NoSelfPatch` so that I download all the game data without looking for a client patch.

(Curious if that's the intended behavior or if maybe there is a difference between Mac/Windows.)

dpogue: `--LocalData` will skip all patching and use local game data and python files.
`--NoSelfPatch` will (on macOS) skip client patching, but depends on the server having the data files for all the game content (which is not guaranteed in the case of things like local Dockersand shards)

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
       </ol>
 
 
-      <h3 id="building-nix">Building on macOS &amp; Linux</h3>
+      <h3 id="building-mac">Building on macOS</h3>
 
-      <p>Linux and macOS are <strong>not officially supported targets</strong>, and the game client does not compile. However, several of the tools can be built and run.</p>
+      <p>Compiling on macOS requires macOS version 10.14 or newer, and the Xcode build system. Xcode is available as a free download from the <a href="https://apps.apple.com/us/app/xcode/id497799835?mt=12/">App Store</a>.</p>
```
Hoikas, line 156:
```suggestion
      <p>Compiling on macOS requires macOS version 10.14 or newer and the Xcode build system. Xcode is available as a free download from the <a href="https://apps.apple.com/us/app/xcode/id497799835?mt=12/">App Store</a>.</p>
```

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
 
-      <p>You will need to ensure all the required dependencies are available. On macOS, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code>. You can also use vcpkg on these platforms for automatically managing dependencies.</p>
+      <p>We strongly encourage using vcpkg to automatically build dependencies, however this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config, and can be installed using common package managers such as Homebrew or MacPorts.</p>
```
Hoikas, line 158:
```suggestion
      <p>We strongly encourage using vcpkg to automatically build dependencies; however, this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config, and can be installed using common package managers such as Homebrew or MacPorts.</p>
```

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
+      <ol>
+        <li>
+          <p>Clone the repository from GitHub, including all submodules, using a git client. On the command line, this can be done by executing the following:<br>
+            <code>git clone --recurse-submodules https://github.com/H-uru/Plasma.git</code></p>
+        </li>
+
+        <li>
+          <p>In the folder where you cloned the repository, run CMake to create a build folder:<br>
+            <code>cmake -G Xcode -DUSE_VCPKG=ON -B build -S .</code></p>
+          <p>This assumes you wish to use vcpkg to build dependencies. Leave off the <tt>USE_VCPKG</tt> flag to manage dependencies yourself.</p>
```
Hoikas, line 171:
```suggestion
          <p>This assumes you wish to use vcpkg to build dependencies. Remove <tt>-DUSEVCPKG=ON</tt> to manage dependencies yourself.</p>
```

building.html
colincornaby, line 156: Compiling requires Xcode 12.2. In turn - Xcode 12.2 requires macOS 10.15.4.

The resulting output will run on 10.14 - but it would be inaccurate to say you can compile from 10.14.

If we made MSL 2.3 an optional part of the build - we could get this lower. But for now Xcode 12.2 and the macOS 11 SDK are required.

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
 
-      <p>Linux and macOS are <strong>not officially supported targets</strong>, and the game client does not compile. However, several of the tools can be built and run.</p>
+      <p>Compiling on macOS requires macOS version 10.14 or newer, and the Xcode build system. Xcode is available as a free download from the <a href="https://apps.apple.com/us/app/xcode/id497799835?mt=12/">App Store</a>.</p>
 
-      <p>You will need to ensure all the required dependencies are available. On macOS, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code>. You can also use vcpkg on these platforms for automatically managing dependencies.</p>
+      <p>We strongly encourage using vcpkg to automatically build dependencies, however this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config, and can be installed using common package managers such as Homebrew or MacPorts.</p>
+
+      <p>For Homebrew users who wish to manage all dependencies, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code> after cloning the project.</p>
```
colincornaby, line 160: I think the strongly recommend above covers things - but I'm not sure one could build a real release on Homebrew. I'm not sure whats included are universal binaries - and bundling Python as anything other than the static lib we get from vcpkg is not at all working. So the resulting output is really only going to be usable from the Mac it was built on.

Hoikas:
I'm converting this to a draft because there is outstanding feedback and to clarify from the main PR page what the status is.

dpogue:
Since the build requires the macOS 11 SDK, I've opted for simplicity to just say the minimum required version for building is macOS 11.

dgelessus:
I haven't tested it, but I assume macOS 10.15 will work as well, assuming you have a recent enough Xcode version. Generally, for any given macOS version, the newest supported Xcode version includes the macOS SDK for the *next* macOS version, so e. g. Xcode 12.4 runs on macOS 10.15 and includes the macOS 11 SDK.

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
+
+        <li>
+          <p>In the folder where you cloned the repository, run CMake to create a build folder:<br>
+            <code>cmake -G Xcode -DUSE_VCPKG=ON -B build -S .</code></p>
+          <p>This assumes you wish to use vcpkg to build dependencies. Remove <tt>-DUSE_VCPKG=ON</tt> to manage dependencies yourself.</p>
+        </li>
+
+        <li>
+          <p>Compile the project with the following command:<br>
+            <code>cmake -B build --config RelWithDebInfo</code></p>
```
dgelessus, line 176: Is `-B` the correct option here? The option for *running* the build is `--build`. `-B` is for setting the build directory when *configuring* the build.

```suggestion
            <code>cmake --build build --config RelWithDebInfo</code></p>
```

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
+        </li>
+
+        <li>
+          <p>Install the resulting tools with the following command:<br>
+            <code>cmake -B build --target install --config RelWithDebInfo</code></p>
```
dgelessus, line 181:
```suggestion
            <code>cmake --build build --target install --config RelWithDebInfo</code></p>
```

dpogue:
Yeah, per Colin's comments:

> Compiling requires Xcode 12.2. In turn - Xcode 12.2 requires macOS 10.15.4.
> The resulting output will run on 10.14 - but it would be inaccurate to say you can compile from 10.14.

(and in theory even lower if you disable the Metal pipeline, but the OpenGL pipeline isn't merged yet so it's not useful to do that)

I think for simplicity it might be worth just documenting it as macOS 11 for now. Some software (not Plasma) also has issues with Xcode versions whose SDK doesn't match the system version, so people are hesitant to update to newer versions on older systems.

dgelessus:
> Some software (not Plasma) also has issues with Xcode versions whose SDK doesn't match the system version, so people are hesitant to update to newer versions on older systems.

FWIW, I see people warn about this a lot, but in my experience, I've never had issues due to the SDK being one version newer than the running system - except in one case where I had to build a kernel extension from source, so I think that case isn't representative for other macOS software 😄

building.html
```diff
@@ -165,19 +203,19 @@ <h3 id="building-nix">Building on macOS &amp; Linux</h3>
...
+            <code>cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -B build -S .</code></p>
           <p>If you want to use vcpkg to automatically build dependencies, enable the <tt>USE_VCPKG</tt> option:<br>
-            <code>cmake -D USE_VCPKG=ON --builddir build -S .</code></p>
+            <code>cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_VCPKG=ON -B build -S .</code></p>
         </li>
 
         <li>
           <p>Compile the project with the following command:<br>
-            <code>cmake --builddir build --config RelWithDebInfo</code></p>
+            <code>cmake -B build</code></p>
```
dgelessus, line 213:
```suggestion
            <code>cmake --build build</code></p>
```

building.html
```diff
@@ -165,19 +203,19 @@ <h3 id="building-nix">Building on macOS &amp; Linux</h3>
...
         </li>
 
         <li>
           <p>Install the resulting tools with the following command:<br>
-            <code>cmake --install build --config RelWithDebInfo</code></p>
+            <code>cmake -B build --target install</code></p>
```
dgelessus, line 218:
```suggestion
            <code>cmake --build build --target install</code></p>
```

dpogue:
> FWIW, I see people warn about this a lot, but in my experience, I've never had issues due to the SDK being one version newer than the running system

I know for sure this is problem that affects WebKit, and I seem to run into it every year. You cannot build a working WebKit for OS using an SDK for OS+1, and their stance is "That's not a supported configuration, your tools need to match your system"

building.html
```diff
@@ -151,11 +151,49 @@ <h3 id="building-windows">Building on Windows</h3>
...
 
 
-      <h3 id="building-nix">Building on macOS &amp; Linux</h3>
+      <h3 id="building-mac">Building on macOS</h3>
 
-      <p>Linux and macOS are <strong>not officially supported targets</strong>, and the game client does not compile. However, several of the tools can be built and run.</p>
+      <p>Compiling on macOS requires a minimum of macOS 11 and the Xcode build system. Xcode is available as a free download from the <a href="https://apps.apple.com/us/app/xcode/id497799835?mt=12/">App Store</a>.</p>
 
-      <p>You will need to ensure all the required dependencies are available. On macOS, a Homebrew bundle file is included in the repository, which can be installed by running <code>brew bundle</code>. You can also use vcpkg on these platforms for automatically managing dependencies.</p>
+      <p>We strongly encourage using vcpkg to automatically build dependencies; however, this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config, and can be installed using common package managers such as Homebrew or MacPorts.</p>
```
Hoikas, line 158:
```suggestion
      <p>We strongly encourage using vcpkg to automatically build dependencies; however, this requires some tools be manually installed in advance. The required tools are CMake, nasm, and pkg-config. They can be installed using common package managers such as Homebrew or MacPorts.</p>
```

The comma before the second `and` is not quite correct, and that structure is a little awkward, IMO. It would be better to separate that into two sentences.

Hoikas (commented):
Just one nitpick, and I think I will be OK with this.

Hoikas (approved):

Deledrius (approved):

PR #1612, opened 2024-08-10
Create folder hierarchy in IDE project views

dpogue:
Makes a folder hierarchy in IDEs that roughly matches the folder layout of the Plasma projects.

VS 2022 project Solution Explorer:
Visual Studio 2022 Project

VS 2022 open folder CMake Targets:
Visual Studio 2022 CMake Folder View

Xcode project file sidebar:
Xcode Project


Hoikas (commented):
Does `cmake/PlasmaTargets.cmake` use the wrong line endings or something? The diff is basically the whole file, which makes it very hard to read.

dpogue:
> Does `cmake/PlasmaTargets.cmake` use the wrong line endings or something?

I think the line endings might have been wrong on master, and it got swapped to LF here 😕

Hoikas:
Can we defer the line ending conversion or another PR?

dpogue:
Should be better now

Hoikas (approved):

Deledrius (approved):

PR #1619, opened 2024-10-12
Fix #1617.

### LINKED ISSUE #1617: Changing the layer dropdown in a Clothing Material does not re-render the texture selection
Screenshot From 2024-10-07 15-34-21

If I change the layer dropdown from "Tint 1" to "Base" in the above image, I should see the Texture buttons update to reflect the assigned textures for the base layer. In Max 2022, that is not happening.

If I change the dropdown, click on to a different material and then back, it does render the selection correctly, so this is clearly a case of not re-rendering when the dropdown changes.

Hoikas:
Fixes #1617.

This invalidates the UI when the clothing material combo box selections are changed. By invalidating the UI, the texture pickers update appropriately. I also added some extra safety valves to reduce spurious updates that the debugger showed coming through.

### DIFF
--- a/Sources/MaxPlugin/MaxPlasmaMtls/Materials/plClothingMtlPBDec.h
+++ b/Sources/MaxPlugin/MaxPlasmaMtls/Materials/plClothingMtlPBDec.h
@@ -255,15 +255,27 @@ public:
         case WM_COMMAND:
             if (id == IDC_CLOTHING_TILESET)
             {
+                if (HIWORD(wParam) != CBN_SELCHANGE)
+                    return FALSE;
+
                 int setIdx = (int)SendMessage(GetDlgItem(hWnd, id), CB_GETCURSEL, 0, 0);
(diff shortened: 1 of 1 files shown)


dpogue (approved):
The code looks good to me, I can see about trying to build and test it in Max

Hoikas:
It seems to fix the problem for me in Max 2022. I'm a little unsure as to *why* this is needed though - it feels like the code should have simply worked as is. I wonder if this was a problem in Max 7?

Deledrius:
That makes me wonder about regressions on Max 7.  It doesn't look like it should be a problem, but it wouldn't be the first time that a seemingly-innocuous change around the WinAPI breaks something in older Max versions.

Hoikas:
This isn't actually a WinAPI regression. What's happening here is that, in Max 2022, the `Update` method on `ClothingBasicDlgProc` isn't being called when values on the param bock are being changed during the `DlgProc()` method.

dpogue:
Built and tested this and it's working in Max 2022 :+1:

PR #1623, opened 2024-11-01
Resolve Max texture paths before exporting

dpogue:
Max seems to save the original full file path to the texture image in its data, but happily resolves it against the current working directory when a file is opened in the editor. Except that it doesn't ever update the stored full file path, so when the PlasmaMax plugin asks for files it can get paths that don't exist. This is a problem because it tries to detect changes to textures based on modified times, and nonexistent files have nonexistent modified times, which can lead to the same image getting exported multiple times because PlasmaMax tries to treat each instance of it as a new texture.

The Max SDK includes a function to resolve texture file paths, so we'll call that in a bunch of spots before we use the texture file path for anything.

### DIFF
--- a/Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
+++ b/Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
@@ -4793,6 +4793,7 @@ bool plGUISkinComp::Convert(plMaxNode *node, plErrorMsg *pErrMsg)
         PBBitmap *texture = layer->GetPBBitmap();
         if (texture != nullptr)
         {
+            BMMGetFullFilename(&texture->bi);
             plBitmap *bMap = plLayerConverter::Instance().CreateSimpleTexture( M2ST(texture->bi.Name()), fConvertedSkin->GetKey()->GetUoid().GetLocation(), 0, plMipmap::kForceNonCompressed | plMipmap::kAlphaChannelFlag | plMipmap::kNoMaxSize );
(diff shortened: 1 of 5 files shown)


Hoikas (commented):
I'm not sure I'm comfortable with the idea of running this on the BitmapInfo itself because it may cause the max file to change on export. I'm a little wary of creating additional deltas. I think I would prefer to see the fix limited to places where the plugin code is trying to access the file system, which is really just in `plBitmapCreator::ICreateTexture()` AFAICT. There is an overload of `BMMGetFullFilename()` would be helpful for this. I left it in a suggestion - not for it to be applied there but as an example.

dpogue:
That is a good point about this causing the paths to change in the actual Max file and leading to unexpected git diffs!

I'll try to get the string-based on hooked up in plBitmapConverter. IIRC the one thing we need to watch out for there is that it isn't available in older APIs (i.e., Max 7) and changed to take a length parameter in recent versions.

Hoikas:
> the one thing we need to watch out for there is that it isn't available in older APIs (i.e., Max 7) and changed to take a length parameter in recent versions.

😞

dpogue:
While we have the quick-fix hack around this, it's probably still best to try to fix this properly.

Do we want to implement the new-style `BMMGetFullFilename` fix only on versions that support it, and fall back to the workaround mod-time fix for older Max versions? I think most people are using newer Max versions, so maybe that's the best compromise?

PR #1626, opened 2024-11-12
Fix #1620

### LINKED ISSUE #1620: Names of the people are far away
Hello again.
I've seen something like this several times but didn't think it was that important, but I think it should be fixed.
If you move the cursor over someone, the person's name appears.

For me, the name is very far away, see picture.

2024-10-15 14_34_23-The Open Cave

I started the game once in full screen and window mode but there was no difference.

For your information:
I have 3 monitors. 2x 1920x1080 and 1x 4k = 3840x2160 (main monitor) resolution.

Thanks for your work and if I knew more about C++, I would fix this myself ;)

Hoikas:
Fixes #1620 by avoiding a double-scale of mouse coordinates. I also included a bonus multi-monitor fix.

I hardwired the cursor to always think it's hovering over a player with the name `It's a Spooky Ghost!`...

On master:
image

With this PR:
image

For reference:
image

### DIFF
--- a/Sources/Plasma/PubUtilLib/plInputCore/plInputDevice.cpp
+++ b/Sources/Plasma/PubUtilLib/plInputCore/plInputDevice.cpp
@@ -492,7 +492,7 @@ bool plMouseDevice::MsgReceive(plMessage* msg)
             fXPos = pXMsg->fX;
 
         SetCursorX(fXPos);
-        fWXPos = pXMsg->fWx * fScale;
+        fWXPos = pXMsg->fWx;
         return true;
     }
 
@@ -529,7 +529,7 @@ bool plMouseDevice::MsgReceive(plMessage* msg)
         else
             fYPos = pYMsg->fY;
 
-        fWYPos = pYMsg->fWy * fScale;
+        fWYPos = pYMsg->fWy;
         SetCursorY(fYPos);
         
         return true;
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
@@ -191,18 +191,19 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
     CGPoint viewLocation = [self convertPoint:windowLocation fromView:nil];
 
     NSRect windowViewBounds = self.bounds;
-    CGFloat deltaX = (windowLocation.x) / windowViewBounds.size.width;
-    CGFloat deltaY =
+    CGFloat xNormal = (windowLocation.x) / windowViewBounds.size.width;
+    CGFloat yNormal =
         (windowViewBounds.size.height - windowLocation.y) / windowViewBounds.size.height;
 
     plIMouseXEventMsg* pXMsg = new plIMouseXEventMsg;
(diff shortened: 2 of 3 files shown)


dpogue (commented):
I suspect this will need testing on macOS retina screens /cc @colincornaby

Sources/Plasma/PubUtilLib/plInputCore/plInputDevice.cpp
```diff
@@ -492,7 +492,7 @@ bool plMouseDevice::MsgReceive(plMessage* msg)
             fXPos = pXMsg->fX;
 
         SetCursorX(fXPos);
-        fWXPos = pXMsg->fWx * fScale;
```
colincornaby, line 495: This change broke macOS. I'll dig into why. Just for background - why didn't this work on Windows?

Hoikas: WM_MOUSEMOVE and friends send coordinates that are already DPI aware, therefore there is no need to scale them. They are already scaled.

colincornaby: That explains it. The input system on macOS does not scale with DPI. Let me consider what to do (including possibly scaling the macOS input by hand) and I'll get back to you shortly.

colincornaby:
Suggested change for macOS, in Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm

```
    NSRect windowViewBounds = self.bounds;
    CGFloat xNormal = (windowLocation.x) / windowViewBounds.size.width;
    CGFloat yNormal =
        (windowViewBounds.size.height - windowLocation.y) / windowViewBounds.size.height;

    plIMouseXEventMsg* pXMsg = new plIMouseXEventMsg;
    plIMouseYEventMsg* pYMsg = new plIMouseYEventMsg;
    
    pXMsg->fX = xNormal;
    pYMsg->fY = yNormal;
    
    // Plasma internally uses input coords as display coords
    pXMsg->fWx = viewLocation.x * self.window.screen.backingScaleFactor;
    pYMsg->fWy = (windowViewBounds.size.height - windowLocation.y) * self.window.screen.backingScaleFactor;

    @synchronized(self.layer) {
        if (self.inputManager) {
            self.inputManager->MsgReceive(pXMsg);
            self.inputManager->MsgReceive(pYMsg);
        }
    }
```

Hoikas:
@colincornaby I've pushed your change and an additional two to ensure that scaled coordinates are going down when the mouse is being recentered. I've verified that the code compiles, but the client login window doesn't actually seem to appear on my system. Maybe this is just a problem with my setup or an unrelated regression, but I'd appreciate your taking a glance at things.

colincornaby:
I am seeing the login window. Weird.

I've seen the login window not appear when the client can't connect to the game server (it just gets stuck in a retry loop, might be its own bug.) When the login window doesn't appear you could try pausing in the debugger to see where it is stuck.

Otherwise, this change looks good on my machine.

Hoikas:
Great! I'm going to wait for a green checkmark before merging, though ;)

colincornaby (approved):

PR #1632, opened 2024-11-16
Allow `ptLayer.setTexture()` for unloaded images.

Hoikas:
This is useful for setting textures that are in global PRPs.


dpogue (approved):
LGTM

PR #1633, opened 2024-11-16
Dynamic Nexus

Hoikas:
Requires #1632.
Requires corresponding assets PR H-uru/moul-assets#268.

This will allow us to more easily ingest fan Ages in the game without requiring their information to be mainlined.

### DIFF
--- a/Scripts/Python/nxusBookMachine.py
+++ b/Scripts/Python/nxusBookMachine.py
@@ -40,13 +40,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
       Mead, WA   99021
 
  *==LICENSE==* """
-"""
-Module: nxusBookMachine
-Age: nexus
-Date: September, 2002
-Author: Bill Slease
-Handler for the nexus book machine.
-"""
+
+from __future__ import annotations
 
 from Plasma import *
 from PlasmaKITypes import *
@@ -58,8 +53,11 @@ from xPsnlVaultSDL import *
 import xLocTools
 
 import PlasmaControlKeys
+
 import datetime
 import random
+import re
+from typing import *
 
 # define the attributes that will be entered in max
 NexusGUI = ptAttribGUIDialog(1, "The Nexus GUI")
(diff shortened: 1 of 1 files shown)


Hoikas:
I'm marking this as a draft because this change to the Nexus Python will "break" linking panels without the corresponding assets PR. That PR is currently a draft because some work is pending on it.

Hazado:
Would it be possible to load pngs from the file directory for the linking panels? It would solve issues with having to have all the linking panels be in the global prps or nexus...

Hoikas:
It's possible, but then we run into the problem of not knowing which files to distribute. Currently, UruManifest (and other tools) crawl the PRPs to determine which files to include in the manifests. If we allow fetching arbitrary images from the data directory as part of the expected game functionality, we lose the ability to know everything by crawling the PRPs. So that's probably not going to be a workable solution. It might be better for us to have a script that can package the BkBookImages PRP from the images in moul-assets.

dpogue:
> Would it be possible to load pngs from the file directory for the linking panels? It would solve issues with having to have all the linking panels be in the global prps or nexus...

This might be something useful to have for Internal clients for easy testing of linking panels during development.

However, in addition to the distribution concerns Hoikas noted, the ability to load arbitrary images from disk poses a security risk, particularly given recent history of image loader libraries being targeted for buffer overflows and remote code execution.

Hazado:
> ability to load arbitrary images from disk

This ability already exists in the code, its used to load the custom avatar images...

Hoikas:
We are aware of that, yes. A long-standing goal of ours has been to remove those in favor of a better preview mechanism. Hopefully this will appear, in time.

dpogue:
> This ability already exists in the code, its used to load the custom avatar images...

Yes, but that is limited to the startup screen

dpogue:
(Sorry, tried to cancel a comment I was typing and hit the wrong button 🤦🏼‍♂️)

Hoikas:
I've updated this to fix a problem with the GuildPub linking panels.

dpogue (approved):
Missing panels for Kalamee, Tiam, and Elonin, but I think that's expected with the removal of fan Age content from moul-assets.

Everything else I checked looked good (although that check did not include Descent and K'veer).

PR #1645, opened 2024-12-24
Unhooking Metal shadows from the lighting system

colincornaby:
The D3D9 shadow system would manipulate lights to properly cast shadows. This change decouples Metal shadows from the lighting system and instead passes the required variables directly to the shadow renderer. This allows Metal to avoid expensive light changes to render shadows.

Also includes a fix for high shadow strength causing shadows to be too dark. The Metal pipeline was not clamping the output diffuse color into a valid range.

These changes are to support #1551 by preventing shadows from needing to touch the cache states for the lighting system. But these changes can be merged and reviewed independently.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
@@ -3202,9 +3202,6 @@ bool plMetalPipeline::IPushShadowCastState(plShadowSlave* slave)
     if (!slave->SetupViewTransform(this))
         return false;
 
-    // Set texture to U_LUT
-    fCurrentRenderPassUniforms->specularSrc = 0.0;
-
     // if( !ref->fTexture )
     //{
     //     if( ref->fData )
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -3759,6 +3756,7 @@ void plMetalPipeline::IRenderShadowsOntoSpan(const plRenderPrimFunc& render, con
             // The shadow light isn't used in generating the shadow map, it's used
             // in projecting the shadow map onto the scene.
             plShadowState shadowState;
+            shadowState.opacity = first? mat->GetLayer(0)->GetOpacity() : 1.f;
```
Hoikas, line 3759:
```suggestion
            shadowState.opacity = first ? mat->GetLayer(0)->GetOpacity() : 1.f;
```

Hoikas (approved):
Looks fine to me

colincornaby:
@dpogue - It looks like Linux builds are failing. Is that expected right now?

dpogue:
> @dpogue - It looks like Linux builds are failing. Is that expected right now?

Ugh, yeah... The newer Ubuntu 24.04 runners don't have mono or NuGet installed 😒

Hoikas:
The linux CI is not a blocker here, IMO

PR #1647, opened 2024-12-30
Remove copy from hsMatrix2SIMD in Metal renderer

colincornaby:
hsMatrix2SIMD used to do a transpose of the input hsMatrix - but shaders and code were rewritten to make this unnecessary. hsMatrix2SIMD still does a vestigial memory copy. This commit removes that memory copy.

This changes the meaning of hsMatrix2SIMD - so functions that need to make a copy now do so manually. Otherwise a direct cast is made. hsMatrix2SIMD has also been made constexpr.

This should result in a small (~1%) performance improvement in the Metal renderer main loop.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
@@ -65,8 +65,11 @@ class plCubicEnvironmap;
 class plLayerInterface;
 class plMetalPipelineState;
 
-// NOTE: Results of this will be row major
-matrix_float4x4* hsMatrix2SIMD(const hsMatrix44& src, matrix_float4x4* dst);
+inline const matrix_float4x4& hsMatrix2SIMD(const hsMatrix44& src)
+{
+    // reinterpret_cast not allowed in constexpr
+    return *reinterpret_cast<const simd_float4x4* >(src.fMap);
+}
 
 class plMetalDevice
 {
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
@@ -1147,9 +1147,8 @@ void plMetalPipeline::ISetupTransforms(plDrawableSpans* drawable, const plSpan& 
     }
 
     if (span.fNumMatrices == 2) {
-        matrix_float4x4 mat;
-        hsMatrix2SIMD(drawable->GetPaletteMatrix(span.fBaseMatrix + 1), &mat);
-        fDevice.CurrentRenderCommandEncoder()->setVertexBytes(&mat, sizeof(matrix_float4x4), VertexShaderArgumentBlendMatrix1);
+        const matrix_float4x4* mat = &hsMatrix2SIMD(drawable->GetPaletteMatrix(span.fBaseMatrix + 1));
+        fDevice.CurrentRenderCommandEncoder()->setVertexBytes(mat, sizeof(matrix_float4x4), VertexShaderArgumentBlendMatrix1);
     }
 
     fCurrentRenderPassUniforms->projectionMatrix = fDevice.fMatrixProj;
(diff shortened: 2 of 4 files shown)


dpogue (commented):
I'm wondering if it makes sense to return a `matrix_float4x4&` rather than a pointer, since we always seem to be immediately dereferencing it

colincornaby:
I've rebased this branch to include the Linux runner changes.

@dpogue - I'm still considering your question. No real thoughts yet - I think the underlying fMap in the matrix is a pointer type which is why it was a pointer here. Shame the matrix type includes extra data beyond the matrix, but using the fMap as the raw underlying matrix data is a small inconvenience.

colincornaby:
I've edited this commit to use reinterpret_cast and to return a reference instead of a pointer.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -65,8 +65,11 @@ class plCubicEnvironmap;
 class plLayerInterface;
 class plMetalPipelineState;
 
-// NOTE: Results of this will be row major
-matrix_float4x4* hsMatrix2SIMD(const hsMatrix44& src, matrix_float4x4* dst);
+inline const matrix_float4x4& hsMatrix2SIMD(const hsMatrix44& src)
+{
+    //reinterperate_cast not allowed in constexpr
```
Hoikas, line 70:
```suggestion
    // reinterpret_cast not allowed in constexpr
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -65,8 +65,11 @@ class plCubicEnvironmap;
...
+    return *reinterpret_cast<const simd_float4x4 *>(src.fMap);
```
Hoikas, line 71:
```suggestion
    return *reinterpret_cast<const simd_float4x4*>(src.fMap);
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1147,9 +1147,8 @@ void plMetalPipeline::ISetupTransforms(plDrawableSpans* drawable, const plSpan&
     }
 
     if (span.fNumMatrices == 2) {
-        matrix_float4x4 mat;
-        hsMatrix2SIMD(drawable->GetPaletteMatrix(span.fBaseMatrix + 1), &mat);
-        fDevice.CurrentRenderCommandEncoder()->setVertexBytes(&mat, sizeof(matrix_float4x4), VertexShaderArgumentBlendMatrix1);
+        const matrix_float4x4 *mat = &hsMatrix2SIMD(drawable->GetPaletteMatrix(span.fBaseMatrix + 1));
```
Hoikas, line 1150:
```suggestion
        const matrix_float4x4* mat = &hsMatrix2SIMD(drawable->GetPaletteMatrix(span.fBaseMatrix + 1));
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1315,8 +1314,7 @@ void plMetalPipeline::IRenderProjection(const plRenderPrimFunc& render, plLightI
     fCurrentRenderPassUniforms->fogColor = {0.f, 0.f, 0.f};
     fCurrentRenderPassUniforms->diffuseCol = {1.f, 1.f, 1.f, 1.f};
 
-    matrix_float4x4 tXfm;
-    hsMatrix2SIMD(proj->GetTransform(), &tXfm);
+    const matrix_float4x4 &tXfm = hsMatrix2SIMD(proj->GetTransform());
```
Hoikas, line 1317:
```suggestion
    const matrix_float4x4& tXfm = hsMatrix2SIMD(proj->GetTransform());
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -4198,15 +4188,14 @@ void plMetalPipeline::IBlendVertBuffer(plSpan* span, hsMatrix44* matrixPalette,
...
 
-        simd_float4x4 simdMatrix;
-
         // Blend
         for (uint32_t j = 0; j < numWeights + 1; ++j) {
-            hsMatrix2SIMD(matrixPalette[indices & 0xFF], &simdMatrix);
-            if (weights[j]) {
+            float weight = weights[j];
+            if (weight) {
+                const simd_float4x4 simdMatrix = hsMatrix2SIMD(matrixPalette[indices & 0xFF]);
```
Hoikas, line 4195:
```suggestion
                const simd_float4x4& simdMatrix = hsMatrix2SIMD(matrixPalette[indices & 0xFF]);
```

colincornaby:
Sorry for the delay in fixes. I've fixed the issues you found @Hoikas.

Hoikas (approved):

PR #1651, opened 2025-01-20
Add some RAII helpers for profiling.

Hoikas:
This adds some RAII helpers for guarded versions of the common `plProfile` operations. A few uses have been ported over to demonstrate the improvement.


dpogue (approved):
Looks good to me

PR #1653, opened 2025-01-25
Support Mouse Scrolling in GUIs

dpogue:
This is a most of the work to get scrolling with the mouse wheel working in GUIs. This should address concerns about not being able to scroll in KI Mail items after the change to being readonly due to clickable links. This also allows mouse scrolling in the KI chat display, which feels about 20 years overdue.

- [x] Plumb scroll events through pfGameGUIMgr and pfGUIControlMod
- [x] Support wheel-based scrolling in pfGUIMultiLineEditCtrl
- [x] Support wheel-based scrolling in pfGUIListBoxMod
- [x] Figure out a way to stop the camera zooming on scroll when we're on a GUI

That last point involves some indirection by using the Avatar or Scene input interface to grab the mouse wheel events and forward them as control events to the camera brain. Using the input interface to do this means that the GUI input interface consumes those events before they pass down to the more general interfaces.

Note that the big KI view with folders for each Age is not actually a list box, and all the scrolling there is faked in Python, so mouse scrolling does not work to scroll the list there.


dpogue:
This should be good to go. It works for scrolling in the KI chat area, the KI Age player list, and within KI journal notes.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1046,6 +1046,28 @@ void    pfGUIMultiLineEditCtrl::HandleMouseHover(hsPoint3& mousePt, uint8_t modi
...
+    if (IHandleMouse(mousePt)) {
+        deltaAccum += wheelAmount;
+        if (fabs(deltaAccum) >= 120.f) {
+            int32_t delta = -1 * int32_t(deltaAccum / 120.f);
+            SetScrollPosition(GetScrollPosition() + delta);
+
+            deltaAccum += (delta * 120.f);
+        }
+    }
+}
```
Hoikas, line 1069: It feels like the wheel delta accumulation should be handled before we get to the individual GUI controls. AFAIK, we don't really support partial scrolling anywhere, so it might be worth accumulating and sending out whole lines either in the input manager or the GUI manager... or perhaps having wheel accumulation being something in the base `pfGUIControlMod` class. Otherwise, we get duplicated accumulation code and interesting semantics where the different controls in the same type knowing about each others' wheel accumulation but different control types do not. That might make for some interesting behavior.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIListBoxMod.cpp
```diff
@@ -775,6 +775,25 @@ void    pfGUIListBoxMod::HandleMouseDblClick( hsPoint3 &mousePt, uint8_t modifie
...
+//// HandleMouseWheel /////////////////////////////////////////////////////////
+
+void pfGUIListBoxMod::HandleMouseWheel(hsPoint3& mousePt, uint8_t modifiers)
+{
+    static float deltaAccum = 0.f;
+
+    deltaAccum += mousePt.fZ;
+
+    // Set the Z-component to 0 for the bounds test
+    mousePt.fZ = 0;
```
Hoikas, line 787: I don't see any bounds testing here...

Hoikas (commented):
Great job getting this figured out!

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
dpogue: Putting this in the GameGUIMgr might make sense.

Do you think the mousePt Z should contain the pre-divided delta (i.e., how many lines to scroll) or just the accumulated delta and let each control figure that out?

Hoikas: That's a good question. I think both of those would be valid approaches, but I have a slight preference for sending the accumulated delta.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1046,6 +1046,21 @@ void    pfGUIMultiLineEditCtrl::HandleMouseHover(hsPoint3& mousePt, uint8_t modi
...
 }
 
+//// HandleMouseWheel /////////////////////////////////////////////////////////
+
+void pfGUIMultiLineEditCtrl::HandleMouseWheel(hsPoint3& mousePt, uint8_t modifiers)
+{
+    float wheelAmount = mousePt.fZ;
+
+    // Set the Z-component to 0 for the bounds test
+    mousePt.fZ = 0;
```
Hoikas, line 1056:
```suggestion
    mousePt.fZ = 0.f;
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGameGUIMgr.cpp
```diff
@@ -738,34 +740,39 @@ bool    pfGameUIInputInterface::InterpretInputEvent( plInputEventMsg *pMsg )
...
             if ((handled) || (fButtonState & kLeftButtonDown)) // even if we didn't handle the mouse up, if we think the button is still down, we should clear our flag
                 fButtonState &= ~kLeftButtonDown;
         }
-        else if( pMouseMsg->GetButton() == kLeftButtonDblClk )
-            handled = fGUIManager->IHandleMouse( pfGameGUIMgr::kMouseDblClick, pMouseMsg->GetXPos(), pMouseMsg->GetYPos(), fModifiers, &fCurrentCursor );
-        else if( fButtonState & kLeftButtonDown )
-            handled = fGUIManager->IHandleMouse( pfGameGUIMgr::kMouseDrag, pMouseMsg->GetXPos(), pMouseMsg->GetYPos(), fModifiers, &fCurrentCursor );
+        else if (pMouseMsg->GetButton() == kLeftButtonDblClk)
+            handled = fGUIManager->IHandleMouse(pfGameGUIMgr::kMouseDblClick, pMouseMsg->GetXPos(), pMouseMsg->GetYPos(), 0.f, fModifiers, &fCurrentCursor);
+        else if (pMouseMsg->GetButton() == kWheelPos || pMouseMsg->GetButton() == kWheelNeg) {
```
Hoikas, line 757: Since this branch has a curly brace, we should give each branch of this if/else a curly brace.

Hoikas (commented):
Just two minor nitpicks

Hoikas:
This looks good from a CR standpoint. However, in my play testing, the mouse wheel is now very reluctant to zoom the camera compared to master 😢

dpogue:
> However, in my play testing, the mouse wheel is now very reluctant to zoom the camera compared to master 😢

It should zoom as it did previously, provided your cursor is not overtop of a scrollable GUI element (and unfortunately the chat element spans most of the upper portion of the screen).

Are you seeing dead spots like that, or a change in how the scrolling is behaving with the cursor in the middle of the screen?

Hoikas:
With the mini KI closed, moving the wheel is barely causing any zoom at all. I have to go around multiple times to get anything to happen. If I revert back to master, it's very smooth.

dpogue:
hmm, it seems a bit slow to zoom on the first scroll for me, but fine after that.
I think what might be happening is that we consume the first scroll wheel event to set the B_CAMERA_ADJUST_OFFSET control event flag, and then the camera doesn't start handling wheel events until after that. I had hoped to avoid needing to do bigger changes and actually read the percentage out of the control event, but maybe that's necessary 😕

Hoikas:
> Or maybe we're being too aggressive in clearing the scrolling event flag, so it's going on and off and causing events to be missed by the camera? The issue is that we don't really have a "button up" event for scrolling, which means there's no obvious signal for when to clear the flag

That seems like the problem. I deleted the two lines that clear the scrolling flag, and the problem went away.

Sources/Plasma/PubUtilLib/plInputCore/plAvatarInputInterface.cpp
```diff
@@ -916,6 +929,8 @@ bool plAvatarInputInterface::InterpretInputEvent( plInputEventMsg *pMsg )
                         continue;
                     if (mouseInfo->fControlFlags & kControlFlagMiddleButtonUp && !(pMouseMsg->GetButton() == kMiddleButtonUp))
                         continue;
+                    if (mouseInfo->fControlFlags & kControlFlagWheel && !(pMouseMsg->GetButton() == kWheelPos | pMouseMsg->GetButton() == kWheelNeg))
```
Hoikas, line 932: MSVC is warning about this. Did you mean:
```suggestion
                    if (mouseInfo->fControlFlags & kControlFlagWheel && !(pMouseMsg->GetButton() == kWheelPos || pMouseMsg->GetButton() == kWheelNeg))
```

dpogue:
> I deleted the two lines that clear the scrolling flag, and the problem went away.

hmmm, presumably we do need to clear these at some point though, and it's not clear where to do that :\

dpogue:
Current thought is maybe to store the timestamp of the last mouse msg with a wheel type, and on the next mouse msg without a wheel type after [TBD] period of time, clear the wheel flags.

Hoikas:
More experimenting... Removing the `HasMovementFlag(B_CAMERA_ADJUST_OFFSET)` check from `plCameraBrain1_Avatar` also fixes the problem.

dpogue:
The `HasMovementFlag(B_CAMERA_ADJUST_OFFSET)` bit was added because otherwise your camera goes zooming in and out while scrolling in the KI chat, because it subscribes directly to the mouse messages rather than going through the input interfaces where the GUI can claim and stop those messages from further processing by later interfaces.

Hoikas:
Yeah, I noticed that. It's looking to me like the camera brain is going to
need to act on the control message instead of the mouse message.

dpogue:
I hooked it up to the ControlEventMsg and it seems to work better in my testing, but your scrolling seemed more sensitive so it would be good for you to confirm that it's working smoothly

TheScarFr:
Would it be possible to use the mouse wheel to turn the book pages?
It would be great for the Relto book...

Hoikas (approved):
This works a lot better in game, and it looks like control event messages will pass through to the base class `plCameraBrain1` correctly, so I would say this is good to go.

dpogue:
> Would it be possible to use the mouse wheel to turn the book pages?

This PR doesn't support that, because the books are like the big KI folder lists where all the scrolling is faked with Python. To support mouse scrolling there, we would need to expose the wheel events to Python and update the Python scripts to handle them.

PR #1654, opened 2025-01-25
SetActivePlayer when double clicking on the player name.

Hoikas:
This allows setting the active player from the StartUp Age by simply double clicking on the button corresponding to the player you want to activate.

### DIFF
--- a/Scripts/Python/xDialogStartUp.py
+++ b/Scripts/Python/xDialogStartUp.py
@@ -301,6 +301,17 @@ class xDialogStartUp(ptResponder):
             elif event == kInterestingEvent: ## RollOver Event ##
                 self.ToggleColor(GUIDiag4b, tagID)
 
+            elif event == kSpecialAction:
+                if tagID in gExp_HotSpot or tagID in gVis_HotSpot:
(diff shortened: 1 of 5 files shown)


Scripts/Python/xDialogStartUp.py
```diff
@@ -301,6 +301,17 @@ def OnGUINotify(self,id,control,event):
             elif event == kInterestingEvent: ## RollOver Event ##
                 self.ToggleColor(GUIDiag4b, tagID)
 
+            elif event == kSpecialAction:
+                if tagID in gExp_HotSpot or tagID in gVis_HotSpot:
+                    if gSelectedSlot and gPlayerList[gSelectedSlot-gMinusExplorer]:
+                        PtShowDialog("GUIDialog06a")
```
dpogue, line 307: GUIDialog06a is the "Linking..." popup?

dpogue (approved):
I don't love the vagueness of "special action", but that's probably fine for now, and otherwise this looks good to me

Scripts/Python/xDialogStartUp.py
Hoikas: Yes, this is just copy pasta'd from the code executed by the regular play button.

PR #1656, opened 2025-01-26
Fix Key Map options not saving secondary key bindings

dgelessus:
The second column in the Key Map options wasn't working - trying to change any of the secondary key bindings caused a Python exception and the new binding wasn't saved/applied. This PR fixes that. Apparently, this was a regression introduced in #1573.


Scripts/Python/xOptionsMenu.py
```diff
@@ -699,11 +696,22 @@ def OnGUINotify(self,id,control,event):
...
+                        PtDebugPrint(f"Missing control code definition for ID {kmID}")
+                        return
+
                     self.ISetKeyMapping(
-                        kmID - 300,
+                        keyLine.controlCode,
                         control.getLastKeyCaptured(),
                         control.getLastModifiersCaptured(),
-                        kmID in gKM1ControlCodesRow1
+                        isPrimary,
```
Hoikas, line 714:
```suggestion
                        isPrimary
```

Hoikas (approved):

Hoikas (approved):

PR #1657, opened 2025-01-27
Remove last uses of `hsStrcpy` and `hsStrncpy`

dgelessus:
Most calls to these functions have already been removed in previous `ST::string`/Unicode conversion work. This PR takes care of the last few remaining uses and finally drops these custom functions (in favor of `ST::string` and standard C APIs).

Important difference: `hsStrncpy` always ensured that the destination string is 0-terminated, whereas the standard C `strncpy` will *not* add a 0 terminator if the destination buffer is too small to hold the entire source string. So we do lose this convenience/safety feature by switching to standard functions. But considering that (after this PR) the codebase only contains 12 `strncpy` calls in total, I don't think `hsStrncpy` is worth keeping around (especially not in the HeadSpin.h header that's included almost everywhere).

### DIFF
--- a/Sources/Plasma/PubUtilLib/plSDL/plStateVariable.cpp
+++ b/Sources/Plasma/PubUtilLib/plSDL/plStateVariable.cpp
@@ -379,7 +379,10 @@ bool plSimpleStateVariable::SetFromString(const ST::string& value, int idx, bool
             int i=idx*fVar.GetAtomicCount();
             for (std::vector<ST::string>::iterator ptr = bits.begin(); ptr != bits.end(); ++ptr)
             {
-                hsStrncpy(fS32[i++], ptr->c_str(), 32);
+                plVarDescriptor::String32& dest = fS32[i];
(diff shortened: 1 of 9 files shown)


Sources/Plasma/PubUtilLib/plSDL/plStateVariable.cpp
```diff
@@ -379,7 +379,9 @@ bool plSimpleStateVariable::SetFromString(const ST::string& value, int idx, bool
             int i=idx*fVar.GetAtomicCount();
             for (std::vector<ST::string>::iterator ptr = bits.begin(); ptr != bits.end(); ++ptr)
             {
-                hsStrncpy(fS32[i++], ptr->c_str(), 32);
+                strncpy(fS32[i], ptr->c_str(), 31);
+                fS32[i][31] = 0;
```
Hoikas, line 383: Could we replace the magic numbers with a call to `std::size(fS32[i]) - 1`?

Hoikas (approved):

dpogue (approved):

PR #1658, opened 2025-01-27
Optimize some poorly performing parts of ASYS.

Hoikas:
This optimizes the "Soft Update" portion of the audio system update. According to my profiling, with 101 male avatars in an Age (resulting in approximately 31,000 sounds loaded), this function would take 7ms on my Ryzen 9 7900X3D. I have something of an overkill CPU, so this kind of drag in the update loop is unacceptable. The problem is that the sounds are all being stored in an intrusive doubly linked list, which is being traversed and partially bubble sorted each update. This pull request changes the container from the intrusive doubly-linked list to an `std::vector` and uses `std::sort()` to sort the active sounds. This change results in the "Soft Update" portion of the update loop only taking about 3ms on my Ryzen 9 7900X3D.

As a bonus, I have improved the semantics around accessing the `plSoundBuffer*` from `plSound` objects. There was already a `plSoundBuffer*` member variable, but it was only ever set by the exporter. The client code was grabbing an active ref and continually grabbing the object pointer from the `plKey`. I have changed this to store the `plSoundBuffer*` in the field that already exists, removing the pointless indirection.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
+++ b/Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
@@ -88,76 +88,14 @@ plProfile_CreateTimer("Soft Update", "Sound", SoundSoftUpdate);
 plProfile_CreateCounter("Max Sounds", "Sound", SoundMaxNum);
 plProfile_CreateTimer("AudioUpdate", "RenderSetup", AudioUpdate);
 
-//// Internal plSoftSoundNode Class Definition ///////////////////////////////
-class plSoftSoundNode
+//// Internal plSoftSound Methods Definition //////////////////////////////////////////////
+
+void plSoftSound::BootSourceOff() const
 {
-    public:
-        const plKey fSoundKey;
-        float fRank;
-
-        plSoftSoundNode* fNext;
-        plSoftSoundNode** fPrev;
-
-        plSoftSoundNode* fSortNext;
-        plSoftSoundNode** fSortPrev;
-
-        plSoftSoundNode(const plKey& s)
-            : fSoundKey(s), fNext(), fPrev(), fSortNext(), fSortPrev()
-        { }
-        ~plSoftSoundNode() { Unlink(); }
-
-        void Link(plSoftSoundNode** prev)
-        {
-            fNext = *prev;
-            fPrev = prev;
-            if (fNext)
-                fNext->fPrev = &fNext;
-            *prev = this;
-        }
-
-        void Unlink()
-        {
-            if (fPrev) {
-                *fPrev = fNext;
-                if (fNext)
-                    fNext->fPrev = fPrev;
-                fPrev = nullptr;
-                fNext = nullptr;
-            }
-        }
-
-        void SortedLink(plSoftSoundNode** prev, float rank)
-        {
-            fRank = rank;
-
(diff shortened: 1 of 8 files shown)


Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
```diff
@@ -670,13 +598,14 @@ void    plAudioSystem::IUpdateSoftSounds(const hsPoint3 &newPosition)
     fDebugActiveSoundDisplay->AddLine(0xff00ffff, "Incidentals");
     fDebugActiveSoundDisplay->AddLine("--------------------");
 
-    for (i = 0; sortedList && i < fMaxNumSounds; sortedList = sortedList->fSortNext) {
-        plSound* sound = plSound::ConvertNoRef(sortedList->fSoundKey->GetObjectPtr());
+    auto soundIt = fActiveSofts.begin();
+    for (i = 0; soundIt != fActiveSofts.end() && i < fMaxNumSounds; soundIt++) {
+        plSound* sound = (plSound*)soundIt->fSoundKey->GetObjectPtr();
```
dpogue, line 603: Safe to cast directly instead of ConvertNoRef?

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
```diff
@@ -693,36 +622,37 @@ void    plAudioSystem::IUpdateSoftSounds(const hsPoint3 &newPosition)
...
         }
         i++;
     }
 
-    for (; sortedList; sortedList = sortedList->fSortNext, i++)
-    {
-        plSound* sound = plSound::ConvertNoRef(sortedList->fSoundKey->GetObjectPtr());
-        if(!sound)
+    for (; soundIt != fActiveSofts.end(); ++soundIt) {
+        plSound* sound = (plSound*)soundIt->fSoundKey->GetObjectPtr();
```
dpogue, line 654: Also here

Sources/Plasma/PubUtilLib/plAudio/plSound.cpp
```diff
@@ -757,34 +774,32 @@ bool plSound::ILoadDataBuffer()
 {
     if(!fDataBufferLoaded)
     {
-        plSoundBuffer *buffer = (plSoundBuffer *)fDataBufferKey->RefObject();
-        if(!buffer)
+        fDataBuffer = (plSoundBuffer*)fDataBufferKey->RefObject();
```
dpogue, line 777: I guess this was already a cast, but... ConvertNoRef? or actually, Convert (with ref) in this case?

Sources/Plasma/PubUtilLib/plAudio/plAudioSystem.cpp
Hoikas: Yes, only `plSound`s ever get added to this collection. The loop is tight enough that I feel it's worth cutting out as much code as possible.

Sources/Plasma/PubUtilLib/plAudio/plSound.cpp
Hoikas: `ConvertNoRef` would be more appropriate here.

dpogue (approved):

PR #1659, opened 2025-01-27
Rework server.ini parsing to not depend on the real console

dgelessus:
Reworks how server.ini files are parsed, based on my console parsing refactoring in #1459. Advantages of this reworked code include:

* Visible error messages (instead of silent failure) when something in the server.ini is obviously invalid: unknown option names, invalid integers, incorrectly formatted base-64 keys.
* The server.ini handling is now decoupled from the console engine - only the parser is shared. This means you can no longer put console commands into the server.ini or set server.ini options from the in-game console.
* The parsed server settings are stored in a struct and not directly into global variables. This is useful groundwork for possible future projects that need to parse more than one server.ini, such as a shard selector.

Also includes some bonus refactoring of how the Diffie-Hellman keys are stored and passed around.

Example of the new error messages:

Windows error dialog saying: "Error in server.ini file. Please check your URU installation. Line 5: Invalid key: Invalid base-64 data (did you forget to put quotes around the value?)"


Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.h
```diff
@@ -0,0 +1,91 @@
...
+    {}
+
+private:
+    ST::string IParseOption(const std::vector<ST::string>& name, const ST::string& value);
+
+public:
+    ST::string Parse(const plFileName& fileName);
+    void Apply();
+
+    static ST::string Load(const plFileName& fileName);
```
Hoikas, line 88: I have a preference for returning a boolean status value and having the error string fetch being a separate call.

Sources/Plasma/NucleusLib/pnNetBase/pnNbConst.h
```diff
@@ -43,6 +43,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #ifndef pnNbConst_inc
 #define pnNbConst_inc
 
+#include "HeadSpin.h"
```
Hoikas, line 46: This is a very lightweight constant only file. I think it would be better to avoid adding this and just using `unsigned` like the rest of the file.

Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.h
dgelessus: Would a boolean return and an out parameter for the error message be okay too? `pfServerIni::Load` is a static method and I would rather not introduce a new global "last error" variable just for this one call...

Hoikas: That's a good point. I was thinking that the error should be a class member that can be fetched with a getter, and the client code would do something like the body of `Load()`. I'm not a big fan of out parameters in C++ - I know there are some in the engine already, but they are a code smell IMO. I would consider throwing an exception acceptable as well. I generally don't care much for those either, but this isn't performance sensitive code, and it feels like a more natural way of receiving an error as opposed to an error message return.

dgelessus: Hmm. Genuine question: why exactly do you prefer a `GetErrorMsg` function/method? To me, it seems like it just has lots of disadvantages (forced shared state for no good reason, every class needs its own error variable/getter, it composes badly, and it's easy for callers to miss that there is an error message at all). All of the alternatives (error message in return value, as an out parameter, or an exception) have less of these issues and just seem better to me, even if they're a little awkward in other ways. I think the ideal solution would be [`std::expected`](https://en.cppreference.com/w/cpp/utility/expected), but that only exists since C++23...

Exceptions would probably work well enough though. I guess I should define a custom exception specifically for this API, similar to e. g. `plPXCookingException`?

Hoikas: To me, returning an error message and testing its emptiness isn't a great way to communicate success. There are plenty of libraries out there that use error codes and have strings for "Success," and it's not uncommon to see amusing situations where users get "Error: Success". I know this isn't an error code, but who knows who will come behind us and not realize an empty string is success. I would prefer an explicit, easily understandable success condition. Out parameters do that, but are C style. This is C++, where we have more idiomatic ways of doing things. Those can be members with a getter, exceptions, tuples (and, with C++17, structured bindings to unpack the return value), or, as you mention, `std::expected` in C++23.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -319,8 +322,12 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification
     FILE* serverIniFile = plFileSystem::Open(serverIni, "rb");
     if (serverIniFile) {
         fclose(serverIniFile);
-        pfConsoleEngine tempConsole;
-        tempConsole.ExecuteFile(serverIni);
+        try {
+            pfServerIni::Load(serverIni);
+        } catch (const std::exception& exc) {
```
Hoikas, line 327: I wonder if we should catch only `pfServerIniParseException`?

Sources/Plasma/Apps/plUruLauncher/winmain.cpp
```diff
@@ -408,15 +410,17 @@ int __stdcall WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdL
         return PLASMA_OK; // see you on the other side...
 
     // Load the doggone server.ini
-    if (!launcher.LoadServerIni()) {
-        IShowErrorDialog(L"No server.ini file found.  Please check your URU installation.");
+    ST::string errorMsg;
+    try {
+        launcher.LoadServerIni();
+    } catch (const std::exception& exc) {
```
Hoikas, line 416: I wonder if we should catch only `pfServerIniParseException`?

Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.cpp
```diff
@@ -0,0 +1,184 @@
...
+        if (name[2].compare_i("Host") == 0) {
+            fAuthHostname = value;
+        } else if (name[2].compare_i("N") == 0) {
+            ParseBase64KeyInto(value, fAuthDhConstants.n);
+        } else if (name[2].compare_i("X") == 0) {
+            ParseBase64KeyInto(value, fAuthDhConstants.x);
+        } else if (name[2].compare_i("G") == 0) {
+            ParseIntegerInto(value, fAuthDhConstants.g);
+        } else {
+            throw pfServerIniParseException("Unknown option name");
```
Hoikas, line 104: It might be nice to have a more specific error message to indicate which command group we encountered the error in.

Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.h
```diff
@@ -0,0 +1,127 @@
...
+#include "pnNetBase/pnNbConst.h"
+#include "pnNetBase/pnNbKeys.h"
+
+class plFileName;
+
+class pfServerIniParseException : public std::runtime_error
+{
+public:
+    explicit pfServerIniParseException(const char* message) :
+        std::runtime_error(message)
```
Hoikas, line 59:
```suggestion
    explicit pfServerIniParseException(const char* message)
        : std::runtime_error(message)
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.h
```diff
@@ -0,0 +1,127 @@
...
+    {}
+
+    explicit pfServerIniParseException(const ST::string& message) :
+        pfServerIniParseException(message.c_str())
```
Hoikas, line 63:
```suggestion
    explicit pfServerIniParseException(const ST::string& message)
        : pfServerIniParseException(message.c_str())
```

Sources/Plasma/FeatureLib/pfConsoleCore/pfServerIni.h
```diff
@@ -0,0 +1,127 @@
...
+        fStatusUrl(),
+        fSignupUrl(),
+        fDisplayName(),
+        fPort(kDefaultClientPort),
+        fFileHostname(),
+        fAuthHostname(),
+        fAuthDhConstants {kNetAuthDefaultDhGValue},
+        fGameDhConstants {kNetGameDefaultDhGValue},
+        fGateKeeperHostname(),
+        fGateKeeperDhConstants {kNetGateKeeperDefaultDhGValue}
```
Hoikas, line 91:
```suggestion
    pfServerIni()
        : fStatusUrl(),
          fSignupUrl(),
          fDisplayName(),
          fPort(kDefaultClientPort),
          fFileHostname(),
          fAuthHostname(),
          fAuthDhConstants {kNetAuthDefaultDhGValue},
          fGameDhConstants {kNetGameDefaultDhGValue},
          fGateKeeperHostname(),
          fGateKeeperDhConstants {kNetGateKeeperDefaultDhGValue}
```

Hoikas (commented):
Getting really close :+1:

Hoikas (approved):

PR #1661, opened 2025-02-01
Generate the Python stubs automatically

dgelessus:
Closes #1660. Adds a script to automatically generate the Python "stub" files under Scripts/Python/plasma. The script must be run from the in-game Python console:

```python
>>> __import__("generate_stubs").run()
```

This will generate stubs for all Plasma built-in modules into a subfolder "plasma_stubs_generated" under the game folder.

Everything happens fully automatically using runtime introspection. Only one part is "semi-manual": because Python cannot introspect the signatures of C-defined functions, all function parameters must be specified manually in the C++-defined function docstrings, on a line starting with `Params: `. This convention was apparently used by Cyan originally, so most function docstrings already follow this format.

The generator itself is almost completely finished, but this PR still needs a bit of work:

* [x] The generator currently removes all type annotations.
  * [x] This will be mostly fixed by copying the parameter types from the existing stubs into the `Params: ` lines in the C++-defined docstrings.
  * [x] There's no way to specify type annotations for function return types and attribute/property types. Most likely, I'll implement this using another docstring line prefix, such as `Returns: `.
* [x] Some C++-defined docstrings are outdated, incorrect, slightly misformatted, or don't match the stubs, leading to spurious and incorrect changes.
* [x] The existing stubs are sometimes not sorted alphabetically. I'll fix this in a separate PR (#1662), so that the diff of this PR will become more readable/meaningful.

With those things in mind, early feedback is welcome. I'm particularly interested in the opinion of @Hoikas, who has recently hand-written some stubs that will be overwritten by this script.


Scripts/Python/plasma/PlasmaGame.py
```diff
@@ -41,79 +41,150 @@
...
 from __future__ import annotations
 from typing import *
 
 class ptGameCli:
     """Abstract bass class for legacy game clients."""
 
-    gameID: int = ...
-    """The ID of the game instance on the server."""
+    gameID: Any
+    """Gets the ID of the game instance on the server."""
```
dpogue, line 59: For things that are implemented as properties, it would be nice if the comments didn't start with "Gets" (but I assume that's something to fix on the C++ docstring side?)

dgelessus: Yes, this comes from the docstrings defined in C++. I'll adjust that soon (along with some other issues in the C++ docstrings).

dpogue:
The generated signature for `PtLocalizedYesNoDialog` is resulting in Python syntax errors (it has a `/` as an argument, which I think is supposed to have special meaning and not show up as an argument itself?)

dgelessus:
I think `/` in parameter lists is allowed in real Python code since... some recent-ish version of Python 3. The actual syntax error is that the `/` appears after a `*args` (`/` must come before `*`). This should get fixed along the way once I get type annotations working, because the manually type-annotated version of `PtLocalizedYesNoDialog` in the stub doesn't have this syntax error.

dgelessus:
Type annotations are now supported by the stub generator. I've introduced a new `Type: ` docstring prefix, which works similarly to the existing `Params: ` prefix, but allows specifying a full function signature including a return type annotation (example: `Type: (x: int, y: int) -> int`) or the type of a property getter (example: `Type: Tuple[str, int]`).

I've also brought all of the C++-defined docstrings up-to-date with the manual changes and added type annotations in the stubs. This should resolve all of the spurious changes.

This PR is basically finished now. I'll un-draft it once #1662 is merged and I've rebased this PR onto that (so that this PR's diff is more readable for reviewing).

Scripts/Python/plasma/PlasmaGameConstants.py
```diff
@@ -54,11 +63,11 @@ class PtGameJoinError:
     kGameJoinErrAlreadyJoined = 6
     kGameJoinErrNoInvite = 7
     kNumGameJoinErrors = 8
-
+    kGameJoinPending = 4294967295
```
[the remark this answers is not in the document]
dpogue: I guess a good question is whether a comparison against -1 would ever work in Python, maybe this MAX_UINT32 is the only correct option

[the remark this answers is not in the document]
dgelessus: Yeah, it's declared in C++ as `kGameJoinPending = (uint32_t)-1`, so it gets output as unsigned.

Also worth noting that I used a 64-bit client to generate the stubs. Plasma's custom `EnumValue`s store the numeric value as `Py_ssize_t`, so it's possible that a 32-bit client *would* see this constant as `-1` on the Python side. That seems like something we should fix...

Scripts/Python/plasma/PlasmaGame.py
```diff
@@ -41,79 +41,150 @@
...
+    ownerID: int
     """The ID of the player who owns this game instance."""
 
+    def __init__(self):
+        """None"""
+        pass
+
+    def hitCloth(self):
+        """Request for the server to hit a specific cloth index and validate the correct sequence of cloth inputs."""
+        pass
```
Hoikas, line 99: The binding itself must be missing the `Params: ...` line

Scripts/Python/plasma/PlasmaGameConstants.py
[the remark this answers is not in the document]
Hoikas: I wonder if would be clearer to use `0xFFFFFFFF`?

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+        params = params_line.removeprefix(docstring_params_prefix)
+        return f"({params})", doc_body
+    else:
+        return "", doc
+
+def format_qualified_name(cls: type, context_module_name: str) -> str:
+    if cls.__module__ in {"builtins", context_module_name}:
+        return cls.__qualname__
+    else:
+        return cls.__module__ + "." + cls.__qualname__
```
Hoikas, line 164:
```suggestion
        return f"{cls.__module__}.{cls.__qualname__}"
```

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+        # Assume that functions without a "Type: " or "Params: " line have no parameters.
+        # Safer would be to default to "*args, **kwargs"
+        # and require an explicit "Type: ()" for no parameters,
+        # but this would require extensive manual updating of the docstrings.
+        signature = "()"
+
+    if self_param is not None:
+        # Insert the self parameter into the method signature.
+        if signature.startswith("()"):
+            signature = "(" + self_param + ")" + signature.removeprefix("()")
```
Hoikas, line 205:
```suggestion
            signature = f"{self_param}){signature.removeprefix('()')}"
```

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+        elif signature.startswith("("):
+            # No space after the comma for now, to match the existing stubs.
+            signature = "(" + self_param + "," + signature.removeprefix("(")
```
Hoikas, line 208:
```suggestion
            signature = f"({self_param },{signature.removeprefix('(')}"
```

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+
+def generate_class_stub(name: str, cls: type) -> Iterable[str]:
+    if tuple(cls.__bases__) == (object,):
+        class_parens = ""
+    else:
+        base_names = []
+        for base in cls.__bases__:
+            base_names.append(format_qualified_name(base, cls.__module__))
+
+        class_parens = "(" + ", ".join(base_names) + ")"
```
Hoikas, line 238:
```suggestion
        class_parens = f"("{', '.join(base_names)})"
```

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+    # Note: Keep this list in sync with the IInitBuiltinModule calls
+    # in PythonInterface::initPython in pfPython/cyPythonInterface.cpp!
+    for module in all_plasma_modules:
+        # Generate all the lines first and only write the file afterwards
+        # to avoid leaving behind a half-written file if generate_module_stub throws an exception.
+        lines = list(generate_module_stub(module))
+        stub_file = os.path.join(dest_dir, module.__name__ + ".py")
+        with open(stub_file, "w", encoding="utf-8", newline="\n") as f:
+            for line in lines:
+                f.write(line + "\n")
```
Hoikas, line 382: Using the `pathlib` module would make all of this quite a bit less tedious, and is highly preferable IMO to all this manual path string fiddling.

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+    else:
+        return cls.__module__ + "." + cls.__qualname__
+
+def add_indents(indent: str, lines: Iterable[str]) -> Iterable[str]:
+    for line in lines:
+        if line:
+            yield indent + line
+        else:
+            # Don't add extra whitespace on blank lines.
+            yield ""
```
Hoikas, line 172: This could be replaced with [`textwrap.indent()`](https://docs.python.org/3/library/textwrap.html#module-textwrap).

Scripts/Python/plasma/PlasmaGameConstants.py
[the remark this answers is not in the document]
dgelessus: Hmm, how would we detect this in the generator though? Add a special case for `0xffffffff`, or just use hex for all values larger than some number?

In any case, I've now regenerated the stubs using a 32-bit client, so we can ignore this until the underlying problem with `kGameJoinPending` is fixed.

Scripts/Python/plasma/generate_stubs.py
dgelessus: Unfortunately, this module/script needs to be run from inside the game, and `pathlib` cannot be used there, because it unconditionally imports `urllib`:

```python-traceback
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../python/system\pathlib.py", line 14, in <module>
    from urllib.parse import quote_from_bytes as urlquote_from_bytes
ModuleNotFoundError: No module named 'urllib'
```

Scripts/Python/plasma/generate_stubs.py
```diff
@@ -0,0 +1,382 @@
...
+    except FileExistsError:
+        pass
+
+    # Note: Keep this list in sync with the IInitBuiltinModule calls
+    # in PythonInterface::initPython in pfPython/cyPythonInterface.cpp!
+    for module in all_plasma_modules:
+        # Generate all the lines first and only write the file afterwards
+        # to avoid leaving behind a half-written file if generate_module_stub throws an exception.
+        lines = list(generate_module_stub(module))
+        stub_file = os.path.join(dest_dir, module.__name__ + ".py")
```
Hoikas, line 379:
```suggestion
        stub_file = os.path.join(dest_dir, f"{module.__name__}.py")
```

Scripts/Python/plasma/generate_stubs.py
Hoikas, line 382:
```suggestion
                f.write(f"{line}\n")
```

Hoikas (commented):
(Can you tell that I'm not a fan of string concats?)

Scripts/Python/plasma/generate_stubs.py
dgelessus: This is a good idea, but I would prefer to do this in a later PR, because it changes the formatting of docstrings that contain newlines. It turns out we have quite a few of those and it would clutter up the diff with even more whitespace changes.

Hoikas: Interesting. I thought based on the documentation that it would give the same result, especially if we provided a predicate, eg

```python
add_indents = functools.partial(textwrap.indent, predicate=lambda line: bool(line))
```

(The call to `bool()` is for the sake of clarity and could be omitted.)

Hoikas: With that being said, with the need for `functools.partial()`, we might be reaching too far into the bag of tricks. I think I'd be okay with keeping `add_indents()` if we remove the string concat in favor of an f-string 😉

dgelessus: The "problem" is that `textwrap.indent` will indent every line, whereas my `add_indents` lets you insert newlines that will not cause an indent, e. g. `add_indents("    ", ["indented\nnot indented", "indented again"])`.

This behavior was by accident, but it turned out to be necessary for matching Cyan's original formatting, where multiline docstrings only have their first line indented and not the following lines.

Hoikas: Makes sense. Thanks for the clarification

Scripts/Python/plasma/PlasmaGame.py
dgelessus: I've added type annotations and parameter names for all the `ptGmBlueSpiral` methods now.

Hoikas (approved):
This will make maintaining the stubs significantly easier. Thanks!

PR #1664, opened 2025-02-20
Remove xLinkingBookDefs import from psnlYeeshaPageChanges.py

Hazado:
All the comments was doing here was list all the Relto pages you had if you had not gotten the 5th page
Removing that section allows removing the xLinkingBookDefs import

### DIFF
--- a/Scripts/Python/Personal.py
+++ b/Scripts/Python/Personal.py
@@ -50,7 +50,7 @@ event manager hooks for the personal age
 from Plasma import *
 from PlasmaTypes import *
 from PlasmaKITypes import *
-
+from xLinkingBookDefs import xYeeshaPages
 
 kEmptyGuid = '0000000000000000'
 kIntroPlayedChronicle = "IntroPlayed"
@@ -202,6 +202,7 @@ class Personal(ptResponder):
         self._AddClothingFromAgeSDL("SweatshirtPod", "psnlBahroWedge07", "psnlBahroWedge08", "psnlBahroWedge09", "psnlBahroWedge10")
         self._AddClothingFromAgeSDL("LJacketMinkata", "psnlBahroWedge11")
         self._AddClothingFromAgeSDL("VestShell", "psnlBahroWedge12", "psnlBahroWedge13")
+        self.ReportFoundPages()
 
 
     def Load(self):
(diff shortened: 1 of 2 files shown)


Hoikas (commented):
I'm not really sure why we want to remove the `xLinkingBookDefs` import. But, we could probably salvage the log message with something like this:

```python
# Top of file
import itertools

# Here:
PtDebugPrint("psnlYeeshaPageChanges: You've found the following Yeesha Pages:")
for thispage in itertools.count(1):
    varName = f"YeeshaPage{thispage}"
    if FoundValue := ageSDL.findVar(varName):
        PtDebugPrint(f"\t The previous value of the SDL variable '{varName}' is {FoundValue.getInt()}")
        if FoundValue.getInt() != 0: 
            PtDebugPrint(f"psnlYeeshaPageChanges: You have found Yeesha Page #{thispage}')
    else:
        # The first not found page variable is the end of the list of Yeesha Pages.
        break
```

Although, tbh, this should probably be in somewhere like Personal.py so we don't have to have the obnoxious guard hack.

Hazado:
> Although, tbh, this should probably be in somewhere like Personal.py so we don't have to have the obnoxious guard hack.

Moved it to Personal.py instead

Hoikas:
Did the snippet I post not work? What you've submitted still imports `xLinkingBookDefs`.

Scripts/Python/Personal.py
```diff
@@ -211,5 +212,16 @@ def Load(self):
...
-
+    def ReportFoundPages(self):
+        if psnlSDL := ptAgeVault().getAgeSDL():
+            PtDebugPrint("psnlYeeshaPageChanges: You've found the following Yeesha Pages:")
+            for thispage in itertools.count(1):
+                varName = f"YeeshaPage{thispage}"
+                if FoundValue := psnlSDL.findVar(varName):
+                    PtDebugPrint(f"\t The previous value of the SDL variable '{varName}' is {FoundValue.getInt()}")
+                    if FoundValue.getInt() != 0: 
+                        PtDebugPrint(f"psnlYeeshaPageChanges: You have found Yeesha Page #{thispage}")
```
Hoikas, line 223:
```suggestion
                        PtDebugPrint(f"Personal: You have found Yeesha Page #{thispage}")
```

Hoikas (commented):
Just one minor nitpick

Scripts/Python/Personal.py
Hazado: Should be fixed.
I used the force

Hoikas (approved):

PR #1666, opened 2025-02-23
Fix incorrect parsing of some KeyMap chronicles written by old code

dgelessus:
The old code sometimes wrote KeyMap chronicles containing *two* spaces between key bindings. The new code incorrectly assumed that there is always exactly *one* space between key bindings, causing all of the key bindings to be shifted around if the KeyMap chronicle contained an extra space.

Detailed repro steps:

1. Log in using a client that has the "old" key map code (before #1573).
2. Change a key binding for an action that can have two keys (e. g. add "J" as the secondary key for "Jump").
3. Close the game.
4. Log in with the same avatar using a client that has the "new" key map code (after #1573).
5. Open the Key Map options and observe that most key bindings are mixed up.

If you repeat steps 4 and 5 with a client that has this PR applied, the key bindings should show up correctly again.

(If you played with a client that has #1573, but not this PR, and you changed any key bindings after encountering this bug, your key map is now persistently mixed up in the chronicle - you need to remap/reset them all manually.)

### DIFF
--- a/Scripts/Python/xOptionsMenu.py
+++ b/Scripts/Python/xOptionsMenu.py
@@ -1778,7 +1778,7 @@ class xOptionsMenu(ptModifier):
             return
 
         # set the key binds back to the saved
-        for controlCode, mappedKey in zip(kDefaultControlCodeBinds, KeyMapString.split(" ")):
+        for controlCode, mappedKey in zip(kDefaultControlCodeBinds, KeyMapString.split()):
             if isinstance(controlCode, str):
                 PtDebugPrint(f"xOptionsMenu.LoadKeyMap(): Binding {mappedKey=} to {controlCode=}", level=kWarningLevel)
                 km.bindKeyToConsoleCommand(mappedKey, controlCode)


Hoikas (approved):

Hoikas:
I hit the approve button, then something started to bother me. Do we know under what circumstance extra spaces would be inserted by the old code? I think I changed the split to only one space for a reason, but I'm struggling to remember why exactly.

dgelessus:
It's this part of the old code:

https://github.com/H-uru/Plasma/blob/8cc47caa5359e571579dd3a67cfbdcb60718f212/Scripts/Python/xOptionsMenu.py#L727-L731

As far as I can tell, this happens every time you rebind an action that can have two keys. Line 728 adds one space after the keys for the newly rebound action, then line 730 in the loop adds one space after every element of the array, which in combination produces *two* spaces after the newly rebound action.

I assumed you changed the `.split()` to `.split(" ")` just to make it clear that the bindings are *supposed* to be separated by exactly one space. If there's actually a deeper reason, we should look into that of course. But the old code also did `.split()`, so I assume it should be safe...

Hoikas:
That makes sense. I would say let's go with this, and if anything bad happens, we can fix that later. It's better to fix the problem we know about than not in favor of some poorly defined concern.

PR #1670, opened 2025-03-04
Update plPickMaterialMap.cpp fix for german

Filtik:
little fix for german


Hoikas:
This only fixes German and breaks the build for 32-bit versions of 3ds Max. I'm working on a more comprehensive fix that I hope to submit soon.

PR #1672, opened 2025-03-04
Fix #1622

### LINKED ISSUE #1622: 3dsMax 2020 freezes after selecting dynamic texture in Python settings
Hi everyone,
I'm having a problem with 3dsMax 2020. When I select a dynamic texture in the Python settings or close the material window without selecting one, 3ds freezes, and even after a long wait, it stays frozen.
The same problem occurs with xSimpleImager python file.
2024-10-21 22_02_37-cod_GameWorld max - Autodesk 3ds Max 2020

I also noticed that the descriptions are cut off. Can this be fixed?

Hoikas:
Fixes #1622 by changing the infinite loop to an "until we quit" loop. This will prevent 3ds Max from hanging when requesting the hack thread to go away. As a bonus, don't burn up the CPU.

Closes #1670. That PR tries to fix the issue by allowing the German window name to be found. Unfortunately, its implementation is not correct for 32-bit max builds. Also, this weird manual dialog fiddling doesn't have any effect in modern maxes. The items that we are trying to hide seem to simply not exist any more due to the UI changing so much since Max 7. IMO, it's better to just fix the hang and leave the UI changes alone.

### DIFF
--- a/Sources/MaxPlugin/MaxComponent/plPickMaterialMap.cpp
+++ b/Sources/MaxPlugin/MaxComponent/plPickMaterialMap.cpp
@@ -70,7 +70,7 @@ public:
     void Run() override
     {
         SetThisThreadName(ST_LITERAL("hsHackWinFindThread"));
-        while (1)
+        while (!GetQuit())
         {
             HWND hMtlDlg = FindWindow(nullptr, _T("Material/Map Browser"));
             if (hMtlDlg && IsWindowVisible(GetDlgItem(hMtlDlg, kOK)))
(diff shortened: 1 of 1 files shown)


dpogue:
Just checking, "until we quit" here means "until we quit Max"?

If we know for sure that certain Max versions don't have whatever problem this is trying to work around, gating the thread initialization on Max version would also be nice.

Hoikas:
"until we quit" means "until hsThread::Stop() is called" - at which point we no longer care about any of this code, discontinue the infinite loop.

dpogue (approved):
This is a clear improvement over the existing code, although my long-term preference would still be that this code doesn't run at all when it's not needed.

PR #1676, opened 2025-03-12
Negilahn and Payiferen creature spawn time edge case

Hazado:
If the spawn times array size exceeded 20 it would fail to write to the age SDL preventing that entire days creature appearances

This is pretty unlikely to happen due to the randomness and large timeframe the creatures can spawn
I only encountered it because I shortened the spawn windows to a few minutes.


Hoikas (approved):
Yikes, this existing code is all really questionable. But, let's go ahead and get this improvement in.

PR #1679, opened 2025-03-21
Adds returning to last read page for journals

Hazado:
https://youtu.be/A0GeF4ayUkA
This saves the last 68 journals read pages to a player chronicle and will open to that page when you open the journal again.

68 journals is how many alder32 hashed strings you can fit in a 1024 character json string
1024 is character limit decided on due to this [issue](https://github.com/H-uru/dirtsand/issues/174)
2048 is probably the string limit for other vault implementations

https://youtu.be/A0GeF4ayUkA?t=27
The only issue I found during testing has to do with the cover and first page of the journal.
There is currently no notify message when you turn from the cover to the first page. 
Due to this I coded after the book kNotifyShow is received, it sets the page to current page (zero).
If you open the book but dont turn any pages and leave, the next time you open the book it will open the book to page 0 instead of the cover.

Operations tested with new code:
I tested these also closing client and entering age again as well
Also tested journals that don't have a cover (Serene Journal)

Cover -> First Page -> Close = First Page
Cover -> First Page -> Cover = Cover
Cover -> Close = First Page (Only issue found so far)
Any Page -> Close = Last used page
Any Page -> Cover -> Close = Cover

### DIFF
--- a/Scripts/Python/xJournalBookGUIPopup.py
+++ b/Scripts/Python/xJournalBookGUIPopup.py
@@ -80,7 +80,11 @@ GUIType             = ptAttribString(14,"Book GUI Type",default="bkBook")
 # globals
 LocalAvatar = None
 JournalBook = None
-CurrentPage = 2
+CurrentPage = -1
+
+
+# Timer variable
+CoverOrPage0 = 1
 
 class xJournalBookGUIPopup(ptModifier):
     "The Journal Book GUI Popup python code"
@@ -133,30 +137,35 @@ class xJournalBookGUIPopup(ptModifier):
                         # disable the KI
                         PtSendKIMessage(kDisableKIandBB,0)
                         if CurrentPage > -1:
-                            JournalBook.open(CurrentPage)
-                            JournalBook.goToPage(CurrentPage)
+                            JournalBook.open(CurrentPage) #Opens to saved page
+                            JournalBook.goToPage(CurrentPage) #shows page tabs at bottom when used after open
+                        CurrentPage = JournalBook.getCurrentPage() #Save current page
                     if event[1] == PtBookEventTypes.kNotifyHide:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyHide")
+                        PtClearTimerCallbacks(self.key)
                         # re-enable KI
                         PtSendKIMessage(kEnableKIandBB,0)
                         # re-enable our avatar
                         PtToggleAvatarClickability(True)
-                        if CurrentPage == 0:
-                            CurrentPage = -1
-                        elif CurrentPage > 0:
+                        if CurrentPage > -1:
                             CurrentPage = JournalBook.getCurrentPage()
-                            PtDebugPrint(JournalBook.getCurrentPage())
                     elif event[1] == PtBookEventTypes.kNotifyClose:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyClose")
+                        PtAtTimeCallback(self.key, 1.01, CoverOrPage0)
                     elif event[1] == PtBookEventTypes.kNotifyNextPage:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyNextPage")
-                        CurrentPage = JournalBook.getCurrentPage()
+                        #CurrentPage = JournalBook.getCurrentPage()
                     elif event[1] == PtBookEventTypes.kNotifyPreviousPage:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyPreviousPage")
-                        CurrentPage = JournalBook.getCurrentPage()
+                        #CurrentPage = JournalBook.getCurrentPage()
                     elif event[1] == PtBookEventTypes.kNotifyCheckUnchecked:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyCheckUncheck",level=kDebugDumpLevel)
                         pass
+
+    def OnTimer(self, id):
+        global CurrentPage
+        if id == CoverOrPage0:
+            CurrentPage = -1
 
     def IShowBook(self):
         global JournalBook
--- a/Scripts/Python/xJournalBookGUIPopup.py
+++ b/Scripts/Python/xJournalBookGUIPopup.py
@@ -79,6 +79,8 @@ GUIType             = ptAttribString(14,"Book GUI Type",default="bkBook")
 
 # globals
 LocalAvatar = None
+JournalBook = None
+CurrentPage = 2
 
 class xJournalBookGUIPopup(ptModifier):
     "The Journal Book GUI Popup python code"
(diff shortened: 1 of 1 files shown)


Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -212,3 +242,32 @@ def IsThereACover(self, bookHtml):
...
+        else:
+            BookJson.pop(self.EncodeLocPath(), None)
+        temp = json.dumps(BookJson)
+        while len(temp) > 1024:
+            BookJson.pop(next(iter(BookJson)))
+            temp = json.dumps(BookJson)
+        ptVault().addChronicleEntry(kBookChronicle, 0, temp)
+        
+    def EncodeLocPath(self):
+        temp = zlib.adler32(LocPath.value.encode('utf-8')) & 0xffffffff
```
Hoikas, line 271:
```suggestion
        temp = zlib.adler32(LocPath.value.encode('utf-8'))
```

`zlib.adler32` is guaranteed to return a 32-bit integer, so we shouldn't need to mask to 32-bits

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -212,3 +242,32 @@ def IsThereACover(self, bookHtml):
...
+            BookJson[self.EncodeLocPath()] = JournalBook.getCurrentPage()
+        else:
+            BookJson.pop(self.EncodeLocPath(), None)
+        temp = json.dumps(BookJson)
+        while len(temp) > 1024:
+            BookJson.pop(next(iter(BookJson)))
+            temp = json.dumps(BookJson)
+        ptVault().addChronicleEntry(kBookChronicle, 0, temp)
+        
+    def EncodeLocPath(self):
```
Hoikas, line 270: This would be a good place to use a `@property` instead of a function, IMO.

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -212,3 +242,32 @@ def IsThereACover(self, bookHtml):
...
+        
+    def WriteCurrentPage(self):
+        PtDebugPrint("xJournalBookGUIPopup WriteCurrentPage",level=kDebugDumpLevel)
+        global BookJson
+        if CurrentPage > -1:
+            BookJson[self.EncodeLocPath()] = JournalBook.getCurrentPage()
+        else:
+            BookJson.pop(self.EncodeLocPath(), None)
+        temp = json.dumps(BookJson)
+        while len(temp) > 1024:
```
Hoikas, line 265: A comment for posterity about why this limitation is needed would be nice.

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -212,3 +242,32 @@ def IsThereACover(self, bookHtml):
...
+        CurrentPage = BookJson.get(self.EncodeLocPath(), -1)
+        
+    def WriteCurrentPage(self):
+        PtDebugPrint("xJournalBookGUIPopup WriteCurrentPage",level=kDebugDumpLevel)
+        global BookJson
+        if CurrentPage > -1:
+            BookJson[self.EncodeLocPath()] = JournalBook.getCurrentPage()
+        else:
+            BookJson.pop(self.EncodeLocPath(), None)
+        temp = json.dumps(BookJson)
```
Hoikas, line 264:
```suggestion
        temp = json.dumps(BookJson, separators=(",", ":")))
```

This will compact the json more by eliminating whitespace.

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -79,6 +81,14 @@
 
 # globals
 LocalAvatar = None
+JournalBook = None
+BookJson = None
+CurrentPage = -1
```
Hoikas, line 86: Could we do this without adding more globals? Instead, prefer instance members.

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -128,12 +140,21 @@ def OnNotify(self, state, id, events):
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyShow",level=kDebugDumpLevel)
                         # disable the KI
                         PtSendKIMessage(kDisableKIandBB,0)
+                        if CurrentPage > -1:
+                            JournalBook.open(CurrentPage) #Opens to saved page
+                            JournalBook.goToPage(CurrentPage) #shows page tabs at bottom when used after open
```
Hoikas, line 145: Looks like we may need to fix a bug in `pfJournalBook`. Is it that the page turn buttons are missing?

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -212,3 +242,32 @@ def IsThereACover(self, bookHtml):
...
+
+    def ReadCurrentPage(self):
+        global BookJson
+        global CurrentPage
+        PtDebugPrint("xJournalBookGUIPopup ReadCurrentPage",level=kDebugDumpLevel)
+        BookJson = ptVault().findChronicleEntry(kBookChronicle)
+        if BookJson:
+            BookJson = json.loads(BookJson.getValue())
+        else:
+            BookJson = {}
```
Hoikas, line 254: We probably shouldn't be hanging onto the value of the chronicle, in case some rogue vault manager comes along and changes the value of the chronicle while we're reading the journal. This would cause the vault manager's change to be discarded.

Hoikas:
A few general thoughts about this approach....

JSON is a bit much for this, I think, but OTOH it's definitely got less boilerplate required for getting started. We might consider just making a chronicle for each loc path under a parent journal chronicle, that way we don't have to worry about the character limits and can forget about the whole JSON thing.

I am a little concerned about how this might work for journals in Ages that are being updated. Cyan's journals are unlikely to change, so that's not a big deal. But, it might be a little surprising if you're in a certain fan Age on page 100 of the journal, and the author suddenly rewrites a significant part of the first 100 pages, meaning you open the journal up to something very different. Also, we'll want to test what happens if we try to open to page 100 if there are only 98 pages (pages get deleted).

> There is currently no notify message when you turn from the cover to the first page.

That sounds like a bug we need to fix, IMO.

Scripts/Python/xJournalBookGUIPopup.py
Hazado: Yes, during my testing both commands were needed to consistently open the book to the correct page and show the tabs.

Hazado:
I dont know if its possible but checking the max pages and if the current page is greater than max, go back to zero or cover

dpogue:
> JSON is a bit much for this, I think, but OTOH it's definitely got less boilerplate required for getting started. We might consider just making a chronicle for each loc path under a parent journal chronicle, that way we don't have to worry about the character limits and can forget about the whole JSON thing.

FWIW, JSON was suggested for this to avoid making a bunch of Chronicles that all need to be fetched at login as part of the player Vault tree. The concern was that it could potentially lead to SetActivePlayer timeouts (although the Chronicles would be small so it's probably not likely, but a single JSON blob is less data to look up and transfer than a hundred single-int Chronicles).

PR #1680, opened 2025-03-25
Fix mouse cursor unpredictably not fading out

### LINKED ISSUE #1677: Cursor Behavior Issues
Got a strange one here - not sure whether this is directly related, but after a recent upgrade to Windows 11, recent builds of the Huru client now show some odd cursor side effects: while several seconds of inactivity usually cause the cursor to fade out, this now appears to be inconsistent - sometimes moving around a little and jiggling the cursor will cause it to fade normally, but sometimes it just sits on the screen and never fades out like it's supposed to. There are a few other odd behaviors too, where essentially things seem to "re-activate" the cursor that normally wouldn't - I've seen it fade out but then get re-triggered by switching between first and third person, and most noticeably when moving around (ie with the arrow keys). **Not seeing any of this behavior on the MOULa client.**

For a second it looked like purging AppData/Local/Uru Live made a difference, but after playing for a few minutes it was back to the behavior I noticed before. In hindsight, it's hard to tell whether there was any actual difference here or this was just a coincidence.

Not sure whether this is _directly_ related to Win11 (it could be I simply didn't notice before?) and have been having some trouble getting people to reproduce, but the fact that I cannot reproduce it in the MOULa client tells me there must be _some_ key behavioral difference here. FWIW @Hoikas also reports he is able to reproduce some of the behavior I'm describing.

dgelessus:
Fixes #1677.

The field `plMouseEventMsg::fWheelDelta` was uninitialized by default and thus had an unpredictable value (which may or may not be 0) for most mouse events. This made the code for fading out the cursor sometimes wrongly think that the player is constantly scrolling, which prevented the cursor from ever fading out.

The underlying bug (the uninitialized field) has always been there since the original open-source release, but went unnoticed, because apparently nothing used the field. It only became noticeable recently with #1653, which added code to reset the cursor fade out timer upon scrolling.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plMessage/plInputEventMsg.cpp
+++ b/Sources/Plasma/PubUtilLib/plMessage/plInputEventMsg.cpp
@@ -321,7 +321,7 @@ void plIMouseBEventMsg::Write(hsStream* stream, hsResMgr* mgr)
     stream->WriteLE16(fButton);
 }
 
-plMouseEventMsg::plMouseEventMsg() : fXPos(0.0f), fYPos(0.0f), fDX(0.0f), fDY(0.0f), fWheelDelta(), fButton(0)
+plMouseEventMsg::plMouseEventMsg() : fXPos(), fYPos(), fDX(), fDY(), fWheelDelta(), fButton()
 {
 }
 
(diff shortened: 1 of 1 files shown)


dpogue (approved):

patmauro:
Tested using the artifacts auto-built by GitHub and can confirm this resolves the issue I reported in #1677.

PR #1682, opened 2025-03-30
Fixing Python path sensitivity

colincornaby:
These paths do not work on case sensitive file systems (i.e. iOS APFS.) Fixing to match actual case.

Assuming the same problem could happen on case sensitive macOS or Linux (although case sensitive macOS installs are rare.)

I'm open to alternatives - but I'm not personally aware of a way to do a case incentive file load on a case sensitive system. Maybe someone else has ideas. I'm assuming the capital-P-Plasma is intentional in the installs.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
@@ -912,10 +912,10 @@ void PythonInterface::initPython()
 
     // Allow importing from the local python directory if and only if this is an internal client.
 #ifndef PLASMA_EXTERNAL_RELEASE
-    PyWideStringList_Append(&config.module_search_paths, L"./python");
-    PyWideStringList_Append(&config.module_search_paths, L"./python/plasma");
-    PyWideStringList_Append(&config.module_search_paths, L"./python/system");
-    PyWideStringList_Append(&config.module_search_paths, L"./python/system/lib-dynload");
+    PyWideStringList_Append(&config.module_search_paths, L"./Python");
+    PyWideStringList_Append(&config.module_search_paths, L"./Python/plasma");
+    PyWideStringList_Append(&config.module_search_paths, L"./Python/system");
+    PyWideStringList_Append(&config.module_search_paths, L"./Python/system/lib-dynload");
     config.module_search_paths_set = 1;
 #endif
 
--- a/Sources/Plasma/FeatureLib/pfPython/plPythonFileMod.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/plPythonFileMod.cpp
@@ -372,8 +372,8 @@ bool plPythonFileMod::ILoadPythonCode()
 #ifndef PLASMA_EXTERNAL_RELEASE
     // get code from file and execute in module
     // see if the file exists first before trying to import it
-    plFileName pyfile = plFileName::Join(".", "python", ST::format("{}.py", fPythonFile));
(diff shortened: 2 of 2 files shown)


dpogue:
It looks like I'm getting `build/install/client/python` on Linux when I run `make install`, so I suspect the CMake install step for Scripts might need updating too: https://github.com/H-uru/Plasma/blob/bcbb906ad96fb49f477d0d435b77096a948060c1/Scripts/CMakeLists.txt#L16

dgelessus (changes requested):
This is not the correct fix. I'm not sure what issue you're running into, but the "python" folder created in an internal client build is lowercase on all systems. You can confirm this by looking at the CI build logs for example ([Windows](https://github.com/H-uru/Plasma/actions/runs/14152317702/job/39647166902?pr=1682#step:9:135), [Mac](https://github.com/H-uru/Plasma/actions/runs/14152317702/job/39647162801?pr=1682#step:9:1286), [Linux](https://github.com/H-uru/Plasma/actions/runs/14152317702/job/39647162790?pr=1682#step:9:12)) - they all show lines like this, with `python` in lowercase:

```
-- Installing: D:/a/Plasma/Plasma/build/install/client/python
```

```
-- Installing: /Users/runner/work/Plasma/Plasma/build/install/client/python
```

```
-- Installing: /home/runner/work/Plasma/Plasma/build/install/client/python
```

If your build is creating a "Python" folder with an uppercase P *in the install directory*, then something is wrong with the build.

colincornaby:
> This is not the correct fix. I'm not sure what issue you're running into, but the "python" folder created in an internal client build is lowercase on all systems.

I'm a little concerned there might be multiple cases being used. I'll check my Windows machine (away from it for now.) I have an install that I think came from OpenUru that has an uppercase P. I also have Dirtsand output with an uppercase P - I'll check on why that occurred.

dpogue:
In older versions of Uru (like the offline game) and other engine versions where Python is loaded from .pak files on the filesystem, the standard seems to be "Python".

It feels to me like consistently capitalizing it probably makes the most sense

dgelessus:
As far as I can tell, the folder containing .pak files is called "Python" with a capital P, and the folder containing .py files is called "python" in all lowercase. The difference between the two is odd, but each usage seems to be consistent *within itself*. So I would be careful with changing the capitalization, to not introduce new inconsistencies in places where it actually matters.

I don't think the capitalization difference between the .pak and .py folders would cause any problems even on a case-sensitive system, because a single client should use either *only* .pak files or *only* .py files, not both at once.

If there's no deeper reason behind the difference, then making them consistent would be a good idea regardless. In that case, "Python" with a capital P would be the better option, because that capitalization seems to be used in more places (including some that are risky to change, like the manifest names expected by clients, as already mentioned). The lowercase "python" seems to be used only for paths to local .py files, so there should be no external dependencies on that capitalization.

colincornaby:
Checking back in on this PR. I'm not sure we arrived at agreement.

Hoikas:
I believe the current consensus is that this fix is incorrect. Loose Python files live in the lowercase `python` directory.

Hoikas:
Given that this is invalid, I propose closing this PR.

Hoikas:
- Python.pak should be in the lowercase `python` directory, per [plPythonPak](https://github.com/H-uru/Plasma/blob/52e36f22c7dbdbb74a35a98fcc4685878fa71212/Sources/Plasma/FeatureLib/pfPython/plPythonPack.cpp?plain=1#L121C1-L121C100). The case restriction only applies to internal clients using local python.pak.
- Individual scripts should live in the lowercase `python` directory.

PR #1683, opened 2025-03-30
Support for Metal renderer on iOS

colincornaby:
Only using storage mode managed on Mac - shared elsewhere. See: https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/MTLBestPracticesGuide/ResourceOptions.html

Fixes for a few other compile issues when building for iPad OS.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
@@ -190,6 +190,15 @@ public:
     MTL::PixelFormat GetFramebufferFormat() const { return fFramebufferFormat; };
     
     MTL::Library* GetShaderLibrary() const { return fShaderLibrary; }
+    
+    static constexpr MTL::StorageMode GetDefaultStorageMode()
+    {
+#if HS_BUILD_FOR_MACOS
+        return MTL::StorageModeManaged;
+#else
+        return MTL::StorageModeShared;
+#endif
+    }
 
 private:
     struct plMetalPipelineRecord
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.cpp
@@ -595,7 +595,7 @@ void plMetalDevice::FillVertexBufferRef(VertexBufferRef* ref, plGBufferGroup* gr
         return;
     }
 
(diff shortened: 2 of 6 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -190,6 +190,15 @@ class plMetalDevice
     MTL::PixelFormat GetFramebufferFormat() const { return fFramebufferFormat; };
     
     MTL::Library* GetShaderLibrary() const { return fShaderLibrary; }
+    
+    static constexpr MTL::StorageMode GetDefaultStorageMode()
+    {
+#if TARGET_OS_MAC
```
dpogue, line 196: For consistency with the rest of the engine, we might want to use `HS_BUILD_FOR_MACOS` here (but we should double check that's actually only getting set on macOS through CMake).

IIRC we have two defines that have mostly been conflated thus far:
* `HS_BUILD_FOR_APPLE`
* `HS_BUILD_FOR_MACOS`

colincornaby: I'll change to 'HS_BUILD_FOR_MACOS'. We don't have an 'HS_BUILD_FOR_IOS' but we also don't require that here.

colincornaby:
To be clear on the intention of this PR: MTL::StorageModeManaged is not supported on iOS. :(

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
dpogue: Nobody has been... courageous... enough to make `HS_BUILD_FOR_IOS` exist yet 😂

dpogue (approved):
LGTM

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -190,6 +190,15 @@ class plMetalDevice
     MTL::PixelFormat GetFramebufferFormat() const { return fFramebufferFormat; };
     
     MTL::Library* GetShaderLibrary() const { return fShaderLibrary; }
+    
+    static constexpr MTL::StorageMode GetDefaultStorageMode()
+    {
+#if HS_BUILD_FOR_MACOS
```
Hoikas, line 196:
```suggestion
#ifdef HS_BUILD_FOR_MACOS
```

PR #1684, opened 2025-03-30
Hopefully better cross-compiling support with Python

dpogue:
CMake historically has had issues with finding dependencies for the target system and the host system in the same project. vcpkg has its own concept of host dependencies, but that doesn't generalize to the rest of CMake.

In Plasma's case, the only host dependency that is causing is pain is the Python interpreter, to run the resources.dat building step on the host at compile time.

CMake 4 introduces the idea of Python artifact prefixes, to allow finding Python multiple times. They give cross-compiling and needing the host interpreter as an example of where this is useful (although it's unclear to me how exactly it's told to find the host interpreter vs the target interpreter).

This change should allow us to opt-in to the prefixed behaviour on CMake 4, while retaining the existing behaviour on CMake 3 with a conditional assignment to the prefixed executable variable. In my testing with cross-compiling for Windows, it does seem to actually find the right values when using CMake 4.

Note: We can't adopt CMake 4 in CI until vcpkg properly supports it. They seem to be doing a bunch of work around that right now.


dgelessus:
I assume this would also need some vcpkg adjustments (once that supports CMake 4.0) to differentiate between host and target dependencies? As I understand it, part of the issue is that some of the dependencies (specifically `python3-cairosvg`) are used at build time and so need to be built for the host system instead of the target system.

dpogue:
yeah, vcpkg.json does currently specify that it wants `python3-cairosvg` as a host dependency, but then it only looks at the target `python3` dependency. So cross-compiling with vcpkg is currently broken (this is most noticeable on macOS where you can't make an x86_64 build from an Apple Silicon machine, and vice versa, even though all the tools support that).

But there are definitely some vcpkg/CMake 4 things to sort out before this is ready to go. I'm not happy with the CMake 4 version checks I had to add here, but vcpkg was doing weird stuff when asked to find the interpreter 😕

PR #1685, opened 2025-03-30
Remove special case for plain IP addresses when connecting to servers

dgelessus:
`AsyncAddressLookupName` can handle plain IP addresses just as well as host names, so there's no need for this special case.

This fixes the client being unable to connect to server addresses that contain a plain IP address with a port number (either in server.ini or when returned by the gatekeeper server). Cyan's original code supported this, but it was apparently broken by accident way back as part of some `plNetAddress` cleanup in #182 (3ea3473d1397d54322bbff177ca8ae6a36e46856).

This is relevant at least for MOSS, where the gatekeeper server returns addresses with an explicit port number if the server is running on a port other than the default 14617.

Server addresses consisting of a host name with a port number have always worked fine. Only the plain IP address case was broken.


dpogue (approved):
LGTM

PR #1689, opened 2025-04-04
Add iOS iconset

patmauro:
As requested by @colincornaby, adds an Xcode-readable iconset that can be used for a hypothetical iOS build - in addition to the standard icon, also includes everything needed for dark and tinted variants as currently required by Apple.

I considered _merging_ this into the existing `AppIcon.appiconset` asset and modifying the json file accordingly, but the current build of Xcode seems to prefer that you keep each purpose-bulit set separate, so I let Xcode generate a new directory here instead. If this should be moved elsewhere let me know.


colincornaby:
Thank you - really above and beyond. I was not expecting to receive an icon for dark mode and tint. Looks good to me!

IMG_0134
IMG_0135

Issue #1649, opened 2025-01-05
_PyModule_Clear missing in Python 3.13


dpogue:
This was a private method that was removed from API visibility in Python 3.13.

This is currently used in cyPythonInterface.cpp to clear an existing module before re-registering it. It seems like an edge case (and we have an assert to catch that case when running in debug mode), but there doesn't seem to be any alternative in the public Python API.

Hoikas:
Broken by the misguided python/cpython#106320. We basically need to reimplement [`_PyModule_ClearDict`](https://github.com/python/cpython/blob/3b231be8f000ae59faa04d5a2f1af11beafee866/Objects/moduleobject.c#L669-L737).

PR #1692, opened 2025-04-26
iOS/iPad Support

colincornaby:
_This PR is very much a draft - but is at that point I should probably start getting feedback. The Python integration is a pain point._

This PR enables Plasma to run on iPhone and iPad. This version supports mouse, keyboard, and stylus input.

The oldest device I have run this on is an A12X. This PR includes changes to manually decompress DX compressed textures on devices that don't support that in hardware.

Changes:
- Adds a new SwiftUI front end. SwiftUI is the modern multiplatform framework that supports all of Apple's platforms. This frontend runs on iOS and sets things up for visionOS
   - I'm not moving the Mac client to the new frontend - there are some missing features in the framework and higher platform requirements. But it _could_ in the future.
- Touch enabled view that includes Apple Pencil support
- Support for the Game Controller framework for keyboards.  Provides HID key codes.
   - Could also be used for macOS in the future
   - Does not provide characters so KI chat does not work
- Adds support for the Python 3.13 iOS build.

This version will not build entirely with vcpkg. The iOS build of Python does not align with how vcpkg includes files. Vcpkg integration would take some thought. iOS 15 is required - I've only tested on iOS 18.

There is no self patching available in this version.

To build:

1.  Download and build the iOS version of Python as detailed here: https://docs.python.org/3/using/ios.html
2. Copy the resulting arm64-iphoneos folder and dylib-Info-template.plist file to the root of the Plasma folder.
3. Invoke CMake - here's the build command I use: `cmake -G Xcode -DPLASMA_APPLE_DEVELOPMENT_TEAM_ID={Your team ID here} -DUSE_VCPKG=1 -DCMAKE_SYSTEM_NAME=iOS -DPLASMA_BUILD_TOOLS=OFF -DPLASMA_PIPELINE_GL=OFF -DVCPKG_CMAKE_SYSTEM_NAME=iOS -DVCPKG_DEBUG=1 -DVCPKG_TARGET_TRIPLET="arm64-ios-plasma" ../`
4. Build/install/run Plasma on your iOS device. This run will fail because Uru data has to be loaded on the device.
5. Using the iOS Files app (or other method of choice) load an Uru data install into the documents folder of the plClient application. This should include a server.ini.
_6. In the device's Uru data, also copy the Python System folder into the Python directory, like you would on Windows and Linux. _You must use the iOS version of System._


Sources/Plasma/Apps/plClient/Apple-SwiftUI/PLSPatcher.mm
colincornaby, line 1: Looks like this was redundantly copied - I'll fix.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginWindowController.mm
```diff
@@ -238,8 +111,8 @@ - (IBAction)loginButtonHit:(id)sender
 
     [self.loginParameters makeCurrent];
 
-    [PLSLoginController attemptLogin:^(ENetError result) {
-        [self loginAttemptEndedWithResult:result];
+    [PLSLoginController attemptLogin:^(int result) {
```
colincornaby, line 114: This is a funny looking change - but Swift doesn't support C++ by default. There is new C++ support, but it can be finicky and I'm not eager to start exposing Plasma headers.  So for now I'm anonymizing ENetError as an int.

LoginView.swift
```diff
@@ -0,0 +1,71 @@
+//
```
dpogue, line 1: This doesn't belong at the top repo level?

NSString+StringTheory.h
```diff
@@ -0,0 +1,56 @@
+/*==LICENSE==*
```
dpogue, line 1: This doesn't belong at the top repo level?

colincornaby:
The Python integration is kind of tortured because of how we need to integrate:

- We _must_ integrate as an Apple framework bundle instead of a library and headers.
- That also means we have to copy it into the bundle resources and sign it.
- We need to also convert all of Python's dynamic library modules into frameworks and then bundle those frameworks (which Python provides us a script for that I've integrated.)

I started working on a find package path for integration - but it was unclear how to handle the bundling of the Python framework.

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -241,7 +306,12 @@ target_link_libraries(
         CURL::libcurl
         "$<$<PLATFORM_ID:Darwin>:-framework Cocoa>"
         "$<$<PLATFORM_ID:Darwin>:-framework QuartzCore>"
+        "$<$<PLATFORM_ID:iOS>:-framework Accelerate>"
```
dpogue, line 309: On macOS, Accelerate is pulled in via CoreLib (due to its use in hsMatrix44)

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -479,19 +483,21 @@ plFileName plFileSystem::GetUserDataPath()
 #if defined(HAVE_BUILTIN_AVAILABLE) && defined(HAVE_SYSDIR)
         if (__builtin_available(macOS 10.12, *)) {
             sysdir_search_path_enumeration_state state;
-            state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_APPLICATION_SUPPORT, SYSDIR_DOMAIN_MASK_USER);
+            state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_DOCUMENT, SYSDIR_DOMAIN_MASK_USER);
```
dpogue, line 486: This might not be right for macOS?

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -241,7 +306,12 @@ target_link_libraries(
         CURL::libcurl
         "$<$<PLATFORM_ID:Darwin>:-framework Cocoa>"
         "$<$<PLATFORM_ID:Darwin>:-framework QuartzCore>"
```
dpogue, line 308: You can provide a list of platforms to PLATFORM_ID to cover both macOS and iOS:
`$<$<PLATFORM_ID:Darwin,iOS>:-framework QuartzCore>`

Sources/Plasma/Apps/plClient/CMakeLists.txt
```diff
@@ -241,7 +306,12 @@ target_link_libraries(
...
+        "$<$<PLATFORM_ID:iOS>:-framework Accelerate>"
+        "$<$<PLATFORM_ID:iOS>:-framework GameController>"
+        "$<$<PLATFORM_ID:iOS>:-framework QuartzCore>"
+        "$<$<PLATFORM_ID:iOS>:-framework UIKit>"
         $<$<PLATFORM_ID:Darwin>:${LibArchive_LIBRARIES}>
+        $<$<PLATFORM_ID:iOS>:${LibArchive_LIBRARIES}>
```
dpogue, line 314: Probably better to unify these with the multiplatform syntax mentioned above

Sources/Plasma/Apps/plClient/Mac-Cocoa/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,61 +1,91 @@
...
+          "value" : "tinted"
+        }
+      ],
+      "filename" : "Icon-iOS-1024x1024-Tinted.png",
+      "idiom" : "universal",
+      "platform" : "ios",
+      "size" : "1024x1024"
+    },
+    {
+      "filename" : "Icon-MacOS-16x16@1x 1.png",
```
dpogue, line 34: Can we do the macOS icons without a ` 1` suffix?

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSView.mm
```diff
@@ -26,6 +26,7 @@
 containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
 PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
 JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
```
dpogue, line 29: This line got duplicated

Sources/Plasma/FeatureLib/pfPython/cyPythonInterface.cpp
```diff
@@ -1447,7 +1447,7 @@ PyObject* PythonInterface::CreateModule(const char* module)
     {
         // clear it
         hsAssert(false, ST::format("ERROR! Creating a python module of the same name - {}", module).c_str());
-        _PyModule_Clear(m);
+        //_PyModule_Clear(m);
```
dpogue, line 1450: At some point we need to actually address this (see https://github.com/H-uru/Plasma/issues/1649)

Sources/Plasma/PubUtilLib/plPipeline/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ target_link_libraries(plPipeline
         $<$<PLATFORM_ID:Windows>:plWinDpi>
         freetype
         "$<$<PLATFORM_ID:Darwin>:-framework CoreText>"
```
dpogue, line 77: Probably need iOS here too?

CMakeLists.txt
```diff
@@ -22,7 +28,11 @@ include(FeatureSummary)
 include(MetalShaderSupport)
 include(VcpkgToolchain)
 
-project(Plasma)
+if(APPLE)
+    project(Plasma LANGUAGES C CXX Swift)
```
dpogue, line 32: Would be nice to do a check for Swift support before enabling this (selfishly saying so because this will blow up spectacularly on OS X Leopard)

dpogue: You can also do this by calling `enable_language(Swift)` at a later point in the file too.

```cmake
include(CheckLanguage)

if(APPLE)
    check_language(Swift)
    if(CMAKE_Swift_COMPILER)
        enable_language(Swift)
    else()
        message(WARNING "Swift is not supported")
    endif()
endif()

Sources/Plasma/CoreLib/plFileSystem.cpp
colincornaby: I did this for diagnostic reasons (takes a while to pull down App Support.) I'll undo.

dpogue:
My suggestion would be to split some of the more self-contained parts of this into a few PRs for easier review/merging head of the iOS client itself:

1. The fallback DXT handling in pfMetalPipeline
2. The bits of condition handling for iOS support across the rest of the core modules
3. Whatever Python stuff is not https://github.com/H-uru/Plasma/pull/1752 and a revised version of https://github.com/H-uru/Plasma/pull/1682

colincornaby:
I'll start breaking this up. I'm going to leave this draft PR open (with apologies to @Hoikas) so I can track what's been merged in.

Python for iOS links in a completely different manner than it does on any other platform. I'm not sure vcpkg supports it (or how it would support the very custom linking) but I'll look into that. It looks like the iOS version requires 3.13 so we may not be far enough ahead to merge in real support.

PR #1699, opened 2025-05-22
Fix CMake Metal support for CMake 4.x

dpogue:
Pulling in updates and fixes from https://github.com/dpogue/CMake-MetalShaderSupport

### DIFF
--- a/cmake/CMakeDetermineMetalCompiler.cmake
+++ b/cmake/CMakeDetermineMetalCompiler.cmake
@@ -9,7 +9,7 @@ if(NOT CMAKE_Metal_COMPILER_NAMES)
     set(CMAKE_Metal_COMPILER_NAMES metal)
 endif()
 
-if("${CMAKE_GENERATOR}" STREQUAL "Xcode")
+if("${CMAKE_GENERATOR}" STREQUAL "Xcode" OR (APPLE AND CMAKE_VERSION VERSION_GREATER_EQUAL "4.0"))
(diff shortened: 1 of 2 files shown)


colincornaby:
It looks like the changes disable the manual shader compilation for CMake 4.0. If I'm understanding that right - does that mean CMake added support for Metal shaders?

dpogue:
No, CMake 4 [includes a change](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9977/diffs#4c0f105291a4108fd44f3771e47c9ccfc934e7c9) to no longer automatically look for compilers in Xcode's SDK root, preferring ones that exist in `/usr/bin`. That is apparently fine for almost every tool that Xcode includes, with the exception of `metal` 🙃

So now, if we're on an Apple system and using CMake >=4.0 we always use `xcrun --find metal` to try to find the Metal compiler rather than asking CMake to find it the way it finds other compilers in `/usr/bin`

colincornaby (approved):

Hoikas:
The logic seems to be getting really fiddly here. This is probably fine for a stop-gap, but what we should probably be doing IMO, instead of "if the generator is this or the version is some magic number":
1. Look for the metal compiler unconditionally. If it's found, stop doing all this.
2. `find_program(xcrun)` if notfound, stop with error, allowing metal or xcrun path to be set manually.
3. Run `xcrun --find metal`, if error, stop with error, allowing metal path to be set manually.
4. Set path to metal compiler, profit.

Issue #1701, opened 2025-05-28
Exclude Region issue


TheScarFr:
If there is no "Exclude Region Message" node in the node tree, the "Exclude Region Logic" node seems to not work with the "xStandardDoor.py and "xHighLevelStarTrekDoor.py" files.
This archive contains 2 BLEND files, 1 with not working Exclude Region (without any "Exclude Region Message" node), and 1 with working Exclude Region (with 2 "Exclude Region Message" nodes): [exclude_region_issue-TheScarFr.zip](https://github.com/user-attachments/files/20471266/exclude_region_issue-TheScarFr.zip)

Hoikas:
Just to summarize.... My initial diagnosis in discord was incorrect. Apparently the xStandardDoor and xHighLevelStarTrekDoor scripts do require the clear and release commands to be part of the responder. This is a questionable design decision that we may want to revisit at some point. Anyway, point is that the working files in your archive are correct.

I did find and fix a few minor issues with the door scripts in the process, so this was actually a helpful diversion. Thanks!

PR #1702, opened 2025-05-28
Fix some minor issues uncovered by #1701.

Hoikas:


Hazado:
This may be a different commit but this wont release the exclude region during normal operations.
Just during the initial link in.

PR #1709, opened 2025-06-18
Fix #1708.

### LINKED ISSUE #1708: Exclude Regions can sometimes take a LOOONG time to clear in HURU vs. standard client
---
ORIGINAL MESSAGE FROM DISCORD:

idle observation on HURU vs OU client:
exclude regions seem to take ABSURDLY long to clear on HURU.
i was noticing a slight lag on kirel's private room and classroom doors while i was testing and thought id compare to see if i was imagining things. a brief pause on destiny/huru preventing me from getting through the door isnt there on OU. so i checked gahreesen to see if it was just kirel and...

OU client: if i run against them constantly as they open, Gahreesen's first floor auto doors have a 1-second pause before letting players run through the gap.
HURU client: if i run against them constantly as they open, it was upwards of half a minute for some of the doors to properly let my player char through.
---

https://www.youtube.com/watch?v=67JsMgs92i8

Here is a video link URL showcasing the issue I brought up in discord last night. Gahreesen had minimal problems today compared to last night, but Kirel's doors were exceedingly inconsistent at times today. Please watch through the entire Kirel section from about 5min in through to the end atleast, especially the classroom door specifically. I haven't encountered it being that buggy in a while. 

In this test run, I was able to hold an exclude region in its 'blocking' state twice for atleast a full 60 count, if not longer in reality. I believe that @Hoikas is right in his brief assessment on Discord that this is probably physx taking a while to process an avatar's positioning and interaction with the Exclude Region, as I observed that after avatar directional changes, intentional on my part or not, I was soon let through the regions in most cases except for the Classroom Door. 

In that case, when the region blocked me out, i could run against it, stop, and run again without being let through. sometimes i was able to even readjust my position during that section of the test.  I thought at first maybe I screwed up the node tree, but the issue DID eventually clear itself and after the issue cleared i was able to pass through it fine on subsequent so I don't think it's a problem on the node tree end for that door. (Even if I will be poking and prodding it to make sure everything is working right again.)

Frankly, something weird is happening here, but IDK what. *shrugs*

Hoikas:
Exclude regions were in practice not clearing because the PhysX CCT was holding onto the collision so as long as the "details" of the collision are largely the same. According to empirical evidence, once the preFilter stage is passed, PhysX never questions an actor again unless there's a position change or something of that nature. So, changing from an exclude region group to a detector volume isn't enough. We actually have to invalidate the character controller's cache to force re-evaluation of the collision. This allows us to move through a released exclude region immediately.

I suppose we could destroy and recreate the actor, but that feels like a more expensive option than just telling the character controllers to reevaluate their lives. Ideally, we would only do this for characters in contact with an exclude region, but that starts to get complicated. It's probably better to go with this simple fix.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysicalControllerCore.h
+++ b/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysicalControllerCore.h
@@ -202,6 +202,9 @@ public:
     // Update controllers when not performing a physics step
     static void UpdateNonPhysical(float alpha);
 
+    /** Invalidates the collision cache for the specified subworld. */
+    static void InvalidateCache(plKey world = {});
+
 #ifndef PLASMA_EXTERNAL_RELEASE
     static bool fDebugDisplay;
 #endif
--- a/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysical.cpp
+++ b/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysical.cpp
@@ -123,6 +123,20 @@ bool plPXPhysical::IsTrigger() const
 
 // ==========================================================================
 
+void plPXPhysical::SetGroup(int group)
+{
+    fGroup = (plSimDefs::Group)group;
+    IUpdateShapeFlags();
+    ISyncFilterData();
+    if (fActor) {
+        // Invalidate character collision cache for exclude regions changing groups.
+        plPXPhysicalControllerCore::InvalidateCache(fWorldKey);
+        InitProxy();
+    }
+}
+
+// ==========================================================================
+
 void plPXPhysical::IUpdateShapeFlags()
 {
     physx::PxShape* shape;
--- a/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysical.h
+++ b/Sources/Plasma/PubUtilLib/plPhysX/plPXPhysical.h
@@ -213,14 +213,7 @@ public:
     void GetTransform(hsMatrix44& l2w, hsMatrix44& w2l) override;
 
     int GetGroup() const override { return fGroup; }
-    void SetGroup(int group) override
-    {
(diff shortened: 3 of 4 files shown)


Sources/Plasma/PubUtilLib/plPhysX/plPXPhysicalControllerCore.h
```diff
@@ -202,6 +202,9 @@ class plPXPhysicalControllerCore: public plPhysicalControllerCore
     // Update controllers when not performing a physics step
     static void UpdateNonPhysical(float alpha);
 
+    /** Invalidates the collision cache for the specified subworld. */
+    static void InvalidateCache(plKey world = {});
```
dgelessus, line 206: Is the default value intentional here? A nil subworld key seems to stand for the main world (no subworld) - seems like this could be a risky default.

```suggestion
    static void InvalidateCache(plKey world);
```

dgelessus (commented):
I think I can reproduce the original issue. For me, it's most clearly visible with the Neighborhood private room doors: after pressing a door button, walking forward gets my avatar stuck on the already open door until I turn/strafe. Enabling `Physics.ShowControllerDebugDisplay` also helps to see if the avatar is colliding with an XRegion.

Can confirm that with this PR, my avatar no longer gets stuck there, and the XRegions disappear more quickly (i. e. when you'd expect them to) from the debug display.

Sources/Plasma/PubUtilLib/plPhysX/plPXPhysicalControllerCore.h
Hoikas: Yes, I was going for consistency with most of the other functions that accept subworld arguments, such as `plPXSimulation::CreateCharacterController()`, which gives the same default option.

dpogue (approved):

PR #1710, opened 2025-06-18
Fix #1705.

### LINKED ISSUE #1705: Nexus - personal hood link issue
In the Nexus, after having clicked the personal hood link the first time, if you exit the Nexus terminal and enter it again, the personal hood link stays highlighted and isn't clickable anymore.
It's fixed by leaving the Nexus and linking back to it, but if you redo the same steps, the issue comes back.

Image

Image

Hoikas:
The Nexus dialog never marks that it is closed. This causes the open-dialog init to only run once. The dialog is fairly resilient and is able to cope with that not happening, with the exception of the Neighborhood button not being cleared out properly on exit. This fixes the state tracking so that the button is reinitialized correctly.

### DIFF
--- a/Scripts/Python/nxusBookMachine.py
+++ b/Scripts/Python/nxusBookMachine.py
@@ -750,6 +750,7 @@ class nxusBookMachine(ptModifier):
         respKISlotReturn.run(self.key)
         self.controlsEnabled = True
         self.animCount = 0
+        self.dialogVisible = False
 
         if self.presentedBookAls is not None:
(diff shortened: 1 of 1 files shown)


dgelessus (approved):
Reproduced the original issue and can confirm that this PR fixes it.

Can't offer much feedback on the code, as I don't have much experience with writing PythonFileMods. But based on the existing implementation, this looks like the obvious way to fix the state tracking issue.

dpogue (approved):

PR #1717, opened 2025-06-19
Fix #1714.

### LINKED ISSUE #1714: Avatar creation name field scrolls oddly if the name is longer than the text field
After typing an avatar name that is longer than the text field, moving the cursor back towards the beginning (using the arrow keys) makes the text start somewhere in the middle of the text field, rather than at the left edge like it's supposed to. Pressing Home makes the text disappear completely until you move the cursor towards the right again.

Haven't tested yet if this also affects other text fields in the game.

Avatar creation dialog with a long name ending in "name be of great length"

The same avatar creation dialog, now showing the start of the name ("long avata"), but the text starts in the middle of the text field instead of at the left edge

Hoikas:
I'm not really sure what's happened - the code is basically identical between the source drop and here, but, for some reason, we are getting a negative scrolling position, which throws everything off. It looks like Cyan anticipated that in one code path, but not the other. Debounce it everywhere.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
+++ b/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
@@ -139,10 +139,9 @@ void    pfGUIEditBoxMod::IUpdate()
         else if( 4 + cursorPos < 4 )
         {
             fScrollPos -= 4 - ( 4 + cursorPos );
-            if( fScrollPos < 0 )
-                fScrollPos = 0;
         }
 
+        fScrollPos = std::max(0, fScrollPos);
         cursorPos = (int16_t)(oldCursorPos - fScrollPos);
     }
 


dgelessus:
Can confirm that this solves the original issue.

Though with this fix, I'm seeing other slightly odd behavior with the cursor movement. When the cursor hits the right edge, one character from beyond the right edge gets scrolled into view - as expected. When the cursor hits the left edge, sometimes much more than one character gets scrolled in from the left. I feel like perhaps the calculation for scrolling to the left is wrong somehow?

dgelessus:
\*sigh\* It's a classic C bug - in the *other* branch. `plDynamicTextMap::GetVisibleWidth` returns an unsigned integer, so the entire condition `4 + cursorPos > fDynTextMap->GetVisibleWidth() - 4 - 2` gets evaluated as unsigned. So if the signed `cursorPos` is less than -4, the unsigned calculation overflows and is always true.

Casting `fDynTextMap->GetVisibleWidth()` to signed seems to fix it. Then the clipping of `fScrollPos` is also not needed in that branch, because the condition (if evaluated correctly) ensures that that branch can only *increase* the scroll position and not decrease it.

Hoikas:
Would you like to submit a PR to supersede this one, @dgelessus, or should I do it?

dgelessus:
Done: #1720

PR #1719, opened 2025-06-19
Fix #1716.

### LINKED ISSUE #1716: After deleting an avatar, another avatar slot incorrectly displays as selected
For example, after deleting the avatar in slot 3, I see this:

Avatar selection screen with 2 avatars, looking as if both are selected

Slot 1 is the one that is actually selected. Slot 2 looks selected, but isn't. Hovering over slot 2 temporarily makes it look un-selected. Actually selecting slot 2 makes both selection rectangles disappear.

Similarly, after deleting the avatar in slot 2, I see this:

Avatar selection screen with only one avatar, but the empty slot 3 also looks like it's selected

Once again, hovering over slot 3 makes it look un-selected.

This seems to be a purely visual issue - the avatar selection otherwise still works properly (unlike previously in #1169).

Tested using 18ca8452f335b9101886766644581b38dac14d52 and H-uru/moul-assets@49da52338e47968e060740ce8747f743ed5d8987.

Hoikas:
Fixes #1716 by making the implicit logic around focused explorers explicit. Gone is the idea of "toggling" - now the state must be explicitly requested. Also, apparently, there was some problem with a default state returning that Cyan was hacking around. Instead, when the GUI dialog is shown, we simply fixup everything.

### DIFF
--- a/Scripts/Python/xDialogStartUp.py
+++ b/Scripts/Python/xDialogStartUp.py
@@ -260,7 +260,7 @@ class xDialogStartUp(ptResponder):
                     PtShowDialog("GUIDialog05")
 
             elif event == kInterestingEvent:
-                self.ToggleColor(GUIDiag4a, tagID)
+                self.ToggleColor(GUIDiag4a, tagID, on=control.isInteresting())
 
         #################################
         ##    Explorer Player List     ##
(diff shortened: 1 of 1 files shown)


dpogue (approved):
LGTM

PR #1720, opened 2025-06-19
Fix accidental conversion to unsigned in `pfGUIEditBoxMod`

### LINKED ISSUE #1714: Avatar creation name field scrolls oddly if the name is longer than the text field

dgelessus:
Fixes #1714, supersedes #1717.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
+++ b/Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
@@ -132,9 +132,14 @@ void    pfGUIEditBoxMod::IUpdate()
         oldCursorPos = cursorPos;
         cursorPos -= (int16_t)fScrollPos;
 
-        if( 4 + cursorPos > fDynTextMap->GetVisibleWidth() - 4 - 2 )
+        // plDynamicTextMap::GetVisibleWidth returns an unsigned integer.
+        // We have to explicitly convert it to signed,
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -132,9 +132,14 @@ void    pfGUIEditBoxMod::IUpdate()
         oldCursorPos = cursorPos;
         cursorPos -= (int16_t)fScrollPos;
 
-        if( 4 + cursorPos > fDynTextMap->GetVisibleWidth() - 4 - 2 )
+        // plDynamicTextMap::GetVisibleWidth returns an unsigned integer.
+        // We have to explicitly convert it to signed,
+        // otherwise the entire comparison will become unsigned and behave incorrectly for negative numbers.
+        ASSERT(fDynTextMap->GetVisibleWidth() <= INT16_MAX);
```
Hoikas: We should probably use `hsAssert` outside of the networking code.

Hoikas (approved):

PR #1723, opened 2025-06-27
Replace hack `std::map` with `std::find`.

Hoikas:
I found this absolute gem while working on bringing back VarSync quab games.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCritter.cpp
+++ b/Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCritter.cpp
@@ -44,6 +44,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plgDispatch.h"
 
+#include <algorithm>
 #include <string_theory/formatter>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


Sources/Plasma/PubUtilLib/plAvatar/plAvBrainCritter.cpp
```diff
@@ -651,21 +652,16 @@ std::vector<unsigned long> plAvBrainCritter::IGetAgePlayerIDList() const
...
     for (size_t i = 0; i < nc->TransportMgr().GetNumMembers(); ++i)
     {
         plNetTransportMember* mbr = nc->TransportMgr().GetMember(i);
         unsigned long id = mbr->GetPlayerID();
-        if (tempMap.find(id) == tempMap.end())
-        {
-            playerIDs.push_back(id);
-            tempMap[id] = true;
-        }
+        playerIDs.push_back(id);
```
zrax, line 660: This won't remove any duplicates from the TransportMgr's list, which the old code also did...  I suppose you could also just use a set and then convert it back to a vector (if necessary), assuming preserving the original order is not important.

Hoikas: That's true, but the comment below here indicates that the purpose of the map is to ensure that the local player specifically is not added as a duplicate. I think that we can assume that there are no duplicated transport members since the rest of the codebase tends to return the transport list without doing any kind of de-duplication.

zrax (approved):

dgelessus (approved):

PR #1724, opened 2025-06-27
Ahnoying GameMgr Redux

Hoikas:
This continues the game manager restore compatibility work by bringing back the VarSync quab game. The non-game manager version has been rewritten simultaneously to improve ownership and synchronization. This changeset has been in progress for over a year, so there are some remnants of ideas about how to improve synchronizing critter AI in the code that has been somewhat left by the wayside, specifically around the whole GoToGoal aspect. A lot of the commit messages are aspirational in that regard.

Notes:
- Fixed the client crash on Cyan's server when the owner of the VarSync game links out AND retains the quabs in their previous location.
- When the game owner is running this implementation, the "goal position" of the quabs will be synchronized to other clients, improving the behavior of the quabs on remote clients (e.g. they should run in the same general direction now). Older clients will safely ignore this information.
- This will not work 100% correctly on MOSS because MOSS "knows" that the quab game is broken and refuses to correctly reassign ownership when the owner leaves. This could probably be mitigated with a special case, but I have elected to not do this in the hope that MOSS will be fixed instead of me needing to add yet another hack.

### DIFF
--- a/Scripts/Python/ahnyQuabs.py
+++ b/Scripts/Python/ahnyQuabs.py
@@ -51,7 +51,6 @@ from dataclasses import dataclass, field
 import enum
 import itertools
 import random
-import re
 from typing import *
 import weakref
 
@@ -71,10 +70,6 @@ kQuabAvatarName = "Quab"
 # Silly behavior name constants
 kQuabIdleBehNames = ("Idle02", "Idle03",)
 kQuabRunBehNames  = ("Run02", "Run03",)
-
-# Regexes for variable notifications
-goalRegex = re.compile(r"^QuabGP(?P<axis>[XYZ])(?P<quabNum>\d+)$")
-runRegex = re.compile(r"^QuabRun(?P<quabNum>\d+)$")
 
 @dataclass
 class _QuabVar:
--- a/Scripts/Python/ahnyQuabs.py
+++ b/Scripts/Python/ahnyQuabs.py
@@ -70,32 +70,36 @@ class ahnyQuabs(ptModifier, object):
         self.version = 2
         self.brains = []
         random.seed()
-        PtDebugPrint("__init__ahnyQuabs v%d " % (self.version))
+        PtDebugPrint(f"__init__ahnyQuabs v{self.version} ")
 
-    def _last_update_get(self):
+    @property
+    def last_update(self) -> int:
+        """Gets the last time a quab was killed/born"""
         ageSDL = PtGetAgeSDL()
         return ageSDL[kLastQuabUpdate][0]
-    last_update = property(_last_update_get, doc="Gets the last time a quab was killed/born")
 
-    def _quabs_get(self):
+    @property
+    def quabs(self) -> int:
+        """Gets the number of quabs alive"""
         ageSDL = PtGetAgeSDL()
         return ageSDL[SDLQuabs.value][0]
-    def _quabs_set(self, value):
+
+    @quabs.setter
+    def quabs(self, value: int) -> None:
         ageSDL = PtGetAgeSDL()
         ageSDL[SDLQuabs.value] = (value,)
         ageSDL[kLastQuabUpdate] = (PtGetServerTime(),)
-    quabs = property(_quabs_get, _quabs_set, doc="Gets the number of quabs alive")
 
     def OnServerInitComplete(self):
         PtDebugPrint("ahnyQuabs.OnServerInitComplete():\tWhen I got here...", level=kWarningLevel)
-        PtDebugPrint("ahnyQuabs.OnServerInitComplete():\t... there were already %i quabs" % self.quabs, level=kWarningLevel)
+        PtDebugPrint(f"ahnyQuabs.OnServerInitComplete():\t... there were already {self.quabs} quabs", level=kWarningLevel)
         self.brains = PtGetAIAvatarsByModelName(kQuabAvatarName)
 
         # Sanity Check: Before we think about doing any processing, make sure there are no quabs
         #               already loaded. We may have arrived after the last man left but before the
         #               server shut down. Therefore, we will already have quabs... So we don't want
         #               to spawn another 20 or so dupe avatar clones.
-        if len(self.brains) != 0:
+        if self.brains:
             PtDebugPrint("ahnyQuabs.OnServerInitComplete():\t... and they were already spawned!", level=kWarningLevel)
             for brain in self.brains:
                 self._PrepCritterBrain(brain[0])
@@ -103,12 +107,12 @@ class ahnyQuabs(ptModifier, object):
 
         if self.sceneobject.isLocallyOwned():
             delta = PtGetServerTime() - self.last_update
-            toSpawn = int(math.floor(delta / kQuabGestationTime))
+            toSpawn = delta // kQuabGestationTime
             if toSpawn:
-                PtDebugPrint("ahnyQuabs.OnServerInitComplete():\t... and I need to spawn %i more" % toSpawn, level=kWarningLevel)
+                PtDebugPrint(f"ahnyQuabs.OnServerInitComplete():\t... and I need {toSpawn=} more", level=kWarningLevel)
                 self.quabs += toSpawn
             if self.quabs > kMaxNumQuabs:
-                PtDebugPrint("ahnyQuabs.OnServerInitComplete():\t... woah, %i quabs?!" % self.quabs, level=kWarningLevel)
+                PtDebugPrint(f"ahnyQuabs.OnServerInitComplete():\t... woah,  {self.quabs=}?!", level=kWarningLevel)
                 self.quabs = kMaxNumQuabs
 
             # Shuffle the spawn points around so we don't get the same quabs appearing
@@ -120,19 +124,19 @@ class ahnyQuabs(ptModifier, object):
             # We will load the avatar clones manually if we are the first one in.
             # We will obtain the ptCritterBrains in an OnAIMsg callback.
             for i in range(self.quabs):
-                PtLoadAvatarModel(kQuabAvatarName, qSpawns[i].getKey(), "Quab %i" % i)
+                PtLoadAvatarModel(kQuabAvatarName, qSpawns[i].getKey(), f"Quab {i}")
 
     def OnAIMsg(self, brain, msgType, userStr, args):
         if msgType == PtAIMsgType.kBrainCreated:
             # Init the brain and push it into our collection
-            PtDebugPrint("ahnyQuabs.OnAIMsg():\t%s created" % userStr, level=kDebugDumpLevel)
+            PtDebugPrint(f"ahnyQuabs.OnAIMsg():\t{userStr} created", level=kDebugDumpLevel)
             self._PrepCritterBrain(brain)
             self.brains.append((brain, userStr,))
             return
 
         if msgType == PtAIMsgType.kArrivedAtGoal:
             # Not really important, but useful for debugging
-            PtDebugPrint("ahnyQuabs.OnAIMsg():\t%s arrived at goal" % userStr, level=kDebugDumpLevel)
+            PtDebugPrint(f"ahnyQuabs.OnAIMsg():\t{userStr} arrived at goal", level=kDebugDumpLevel)
             return
 
     def OnNotify(self, state, id, events):
@@ -143,7 +147,7 @@ class ahnyQuabs(ptModifier, object):
             colso = PtFindAvatar(events)
             if colso.isAvatar() and not colso.isHuman():
                 self.quabs -= 1
-                PtDebugPrint("ahnyQuabs.OnNotify():\tQuabs remaining: %i" % self.quabs, level=kWarningLevel)
+                PtDebugPrint(f"ahnyQuabs.OnNotify():\tQuabs remaining: {self.quabs}", level=kWarningLevel)
                 return
 
     def OnUpdate(self, seconds, delta):
@@ -163,9 +167,9 @@ class ahnyQuabs(ptModifier, object):
         # ability to turn quickly (which the dog does not actually do)...
         # This evolutionary biology lesson is thanks to Hoikas, the Mammalogy drop-out
         monsters = brain.playersICanHear()
-        if len(monsters) == 0:
+        if not monsters:
             if running:
-                PtDebugPrint("ahnyQuabs._Think():\t%s is now safe." % name, level=kDebugDumpLevel)
+                PtDebugPrint(f"ahnyQuabs._Think():\t{name} is now safe.", level=kDebugDumpLevel)
                 self._RunAway(brain, False)
             return
         runaway  = None
@@ -184,7 +188,7 @@ class ahnyQuabs(ptModifier, object):
         if not running:
             # Note: low level brain will make the quab play the run behavior
             #       no need to court a race condition by playing it here
-            PtDebugPrint("ahnyQuabs._Think():\tTime for %s to run away!" % name, level=kDebugDumpLevel)
+            PtDebugPrint(f"ahnyQuabs._Think():\tTime for {name} to run away!", level=kDebugDumpLevel)
         brain.goToGoal(endPos)
 
     def _PrepCritterBrain(self, brain):
--- a/Scripts/Python/ahnyQuabs.py
+++ b/Scripts/Python/ahnyQuabs.py
@@ -47,7 +47,7 @@ from PlasmaGame import *
 from PlasmaTypes import *
 
 import abc
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 import enum
 import itertools
 import random
@@ -80,31 +80,23 @@ runRegex = re.compile(r"^QuabRun(?P<quabNum>\d+)$")
 class _QuabVar:
     name: str
     varID: Optional[int] = None
-    value: float = 0.0
-    dirty: bool = False
-    lastUpdate: float = 0.0
-
-    def __hash__(self):
-        return hash((self.name, self.varID))
-
-    def __eq__(self, other):
-        if not isinstance(other, _QuabVar):
-            return False
-        return (self.name, self.varID) == (other.name, other.varID)
+    value: float = field(default=0.0, compare=False)
+    dirty: bool = field(default=False, compare=False)
+    lastUpdate: float = field(default=0.0, compare=False)
 
 
-@dataclass
+@dataclass(unsafe_hash=True)
 class _QuabState:
     name: str
-    runningVar: _QuabVar
-    goalVarX: _QuabVar
-    goalVarY: _QuabVar
-    goalVarZ: _QuabVar
-    brain: Optional[ptCritterBrain]
-    lastXform: ptMatrix44
-    lastW2L: ptMatrix44
-    respawn: bool
-    pending: bool
+    runningVar: _QuabVar = field(compare=False)
+    goalVarX: _QuabVar = field(compare=False)
+    goalVarY: _QuabVar = field(compare=False)
+    goalVarZ: _QuabVar = field(compare=False)
+    brain: Optional[ptCritterBrain] = field(compare=False)
+    lastXform: ptMatrix44 = field(compare=False)
+    lastW2L: ptMatrix44 = field(compare=False)
+    respawn: bool = field(compare=False)
+    pending: bool = field(compare=False)
 
     def __init__(self, num: int):
         # Argh, Cyan's code offsets the quab name and variable indices!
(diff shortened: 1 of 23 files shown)


Sources/Plasma/FeatureLib/pfPython/pyGmVarSyncGlue.cpp
```diff
@@ -0,0 +1,140 @@
...
+PYTHON_METHOD_DEFINITION_STATIC_NOARGS(ptGmVarSync, isSupported)
+{
+    return plPython::ConvertFrom(self->fThis->IsSupported());
+}
+
+PYTHON_START_METHODS_TABLE(ptGmVarSync)
+    PYTHON_METHOD_STATIC(ptGmVarSync, join, "Join the common var sync game in the current Age."),
+    PYTHON_METHOD_STATIC_NOARGS(ptGmVarSync, isSupported, "Checks for the presence of a server-side var sync game manager."),
+    PYTHON_METHOD(ptGmVarSync, setVariable, "Change the value of a variable on the server."),
+    PYTHON_METHOD(ptGmVarSync, createVariable, "Create a new variable on the server."),
```
dgelessus: can haz type signatures plz?

Sources/Plasma/FeatureLib/pfPython/pyGmVarSync.h
```diff
@@ -0,0 +1,80 @@
...
+*==LICENSE==*/
+
+#ifndef _pyGmVarSync_h_
+#define _pyGmVarSync_h_
+
+#include "pyGameCli.h"
+
+#include <variant>
+
+using pfGmVarSyncValue = std::variant<ST::string, double>;
```
dgelessus, line 50: I assume this duplicate declaration is intentional to avoid including pfGameMgr/pfGmVarSync.h?

Sources/Plasma/PubUtilLib/plMessage/plAIMsg.h
```diff
@@ -113,4 +115,55 @@ class plAIArrivedAtGoalMsg : public plAIMsg
     hsPoint3 fGoal;
 };
 
+/**
+ * Message spammed to anyone listening so they can discard the brain's key
+ * does NOT get net-propped
```
dgelessus:
```suggestion
 * Message spammed to anyone listening so they can discard the brain's key.
 * Does NOT get net-propped.
```

Else the text looks odd when auto-wrapped in my IDE's documentation popup.

Sources/Plasma/PubUtilLib/plMessage/plAIMsg.h
```diff
@@ -113,4 +115,55 @@ class plAIArrivedAtGoalMsg : public plAIMsg
...
+    CLASSNAME_REGISTER(plAIBrainDestroyedMsg);
+    GETINTERFACE_ANY(plAIBrainDestroyedMsg, plAIMsg);
+
+    void Read(hsStream* stream, hsResMgr* mgr) override { plAIMsg::Read(stream, mgr); }
+    void Write(hsStream* stream, hsResMgr* mgr) override { plAIMsg::Write(stream, mgr); }
+};
+
+/**
+ * Message sent to the AI brain to force it to go to a specific goal
+ * might get net-proppped
```
dgelessus:
```suggestion
 * Message sent to the AI brain to force it to go to a specific goal.
 * Might get net-propped.
```

Sources/Plasma/FeatureLib/pfPython/pyCritterBrainGlue.cpp
```diff
@@ -388,7 +408,11 @@ PYTHON_METHOD_DEFINITION(ptCritterBrain, vectorToPlayer, args)
 PYTHON_START_METHODS_TABLE(ptCritterBrain)
     PYTHON_METHOD(ptCritterBrain, addReceiver, "Params: key\nTells the brain that the specified key wants AI messages"),
     PYTHON_METHOD(ptCritterBrain, removeReceiver, "Params: key\nTells the brain that the specified key no longer wants AI messages"),
+    PYTHON_METHOD(ptCritterBrain, setLocallyControlled, "Params: local\n"
```
dgelessus:
```suggestion
    PYTHON_METHOD(ptCritterBrain, setLocallyControlled, "Type: (local: bool) -> None\n"
```

Sources/Plasma/FeatureLib/pfPython/pyCritterBrainGlue.cpp
```diff
@@ -388,7 +408,11 @@ PYTHON_METHOD_DEFINITION(ptCritterBrain, vectorToPlayer, args)
...
+        "Tells the brain that we are the ones making all the AI decisions, and to prop location and other information to the server."),
     PYTHON_METHOD(ptCritterBrain, getSceneObject, "Returns the ptSceneObject this brain controls."),
+    PYTHON_METHOD_NOARGS(ptCritterBrain, getLocallyControlled, "Are we the one making AI decisions? NOTE: Not set automatically, some python script needs to "
+        "tell the brain this using setLocallyControlled()."),
```
dgelessus: Return type:

```suggestion
    PYTHON_METHOD_NOARGS(ptCritterBrain, getLocallyControlled, "Type: () -> bool\n"
        "Are we the one making AI decisions? NOTE: Not set automatically, some python script needs to "
        "tell the brain this using setLocallyControlled()."),
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+    dirty: bool = False
+    lastUpdate: float = 0.0
+
+    def __hash__(self):
+        return hash((self.name, self.varID))
+
+    def __eq__(self, other):
+        if not isinstance(other, _QuabVar):
+            return False
+        return (self.name, self.varID) == (other.name, other.varID)
```
dgelessus: Haven't tested, but according to the docs, `field(compare=False)` can be used to exclude specific fields from equality and hashing without having to manually override `__eq__` and `__hash__`.

```suggestion
    name: str
    varID: Optional[int] = None
    value: float = field(default=0.0, compare=False)
    dirty: bool = field(default=False, compare=False)
    lastUpdate: float = field(default=0.0, compare=False)
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -40,10 +40,20 @@
...
 
+from __future__ import annotations
+
 from Plasma import *
+from PlasmaGame import *
 from PlasmaTypes import *
-import math
+
+import abc
+from dataclasses import dataclass
```
dgelessus:
```suggestion
from dataclasses import dataclass, field
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+    name: str
+    runningVar: _QuabVar
+    goalVarX: _QuabVar
+    goalVarY: _QuabVar
+    goalVarZ: _QuabVar
+    brain: Optional[ptCritterBrain]
+    lastXform: ptMatrix44
+    lastW2L: ptMatrix44
+    respawn: bool
+    pending: bool
```
dgelessus: Same here, but it's debatable whether this is more readable...

```suggestion
    runningVar: _QuabVar = field(compare=False)
    goalVarX: _QuabVar = field(compare=False)
    goalVarY: _QuabVar = field(compare=False)
    goalVarZ: _QuabVar = field(compare=False)
    brain: Optional[ptCritterBrain] = field(compare=False)
    lastXform: ptMatrix44 = field(compare=False)
    lastW2L: ptMatrix44 = field(compare=False)
    respawn: bool = field(compare=False)
    pending: bool = field(compare=False)
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        self.pending = False
+
+    def __hash__(self):
+        return hash(self.name)
+
+    def __eq__(self, other: _QuabState):
+        if not isinstance(other, _QuabState):
+            return False
+        return self.name == other.name
+
```
dgelessus:
```suggestion
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+    @property
+    def isRunning(self) -> bool:
+        return self.runningVar.value
+
+    @property
+    def isRunningDirty(self) -> bool:
+        return self.runningVar.dirty
+
+    def itervars(self) -> Generator[_QuabVar]:
```
dgelessus: As it's not relevant to callers that this is implemented as a generator:

```suggestion
    def itervars(self) -> Iterable[_QuabVar]:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        yield self.runningVar
+        yield self.goalVarX
+        yield self.goalVarY
+        yield self.goalVarZ
+
+    @property
+    def lastGoalUpdate(self) -> float:
+        return max((i.lastUpdate for i in (self.goalVarX, self.goalVarY, self.goalVarZ)))
```
dgelessus:
```suggestion
        return max(i.lastUpdate for i in (self.goalVarX, self.goalVarY, self.goalVarZ))
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        quab.runningVar.dirty = True
+
+        if sync:
+            self.ISendVars(quab.runningVar)
+
+    def changeQuabGoal(self, quab: _QuabState, goal: ptPoint3, *, sync: bool = False):
+        # Make sure that the adjustment is actually enough to trigger some movement.
+        if not quab.isRunning:
+            distSq = goal.distanceSq(quab.goal)
+            if distSq < pow(quab.brain.getStopDistance(), 2):
```
dgelessus:
```suggestion
            if distSq < quab.brain.getStopDistance()**2:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        self.changeQuabState(quab, True)
+        if sync:
+            self.sync(quab.itervars())
+
+    @property
+    @abc.abstractmethod
+    def currentTime(self) -> int:
+        ...
+
+    def iterAllVars(self) -> Generator[Tuple[_QuabState, _QuabVar]]:
```
dgelessus:
```suggestion
    def iterAllVars(self) -> Iterable[Tuple[_QuabState, _QuabVar]]:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+                if (isinstance(varIdent, str) and var.name == varIdent) or (isinstance(varIdent, int) and var.varID == varIdent):
+                    return quab, var
+        raise LookupError
+
+    @abc.abstractmethod
+    def OnNotify(self, events):
+        ...
+
+    @property
+    def quabs(self) -> Generator[_QuabState]:
```
dgelessus:
```suggestion
    def quabs(self) -> Iterable[_QuabState]:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+
+    def findQuab(self, name: str) -> Optional[_QuabState]:
+        return next((i for i in self.quabState if name == i.name), None)
+
+    def findQuabVar(self, varIdent: Union[int, str]) -> Tuple[_QuabState, _QuabVar]:
+        for quab in self.quabState:
+            for var in quab.itervars():
+                if (isinstance(varIdent, str) and var.name == varIdent) or (isinstance(varIdent, int) and var.varID == varIdent):
+                    return quab, var
+        raise LookupError
```
dgelessus:
```suggestion
        raise LookupError(varIdent)
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+    def numQuabs(self) -> int:
+        return len([i for i in self.quabState if i.brain is not None or i.pending])
+
+    @property
+    def parent(self) -> ahnyQuabs:
+        return self._parent()
+
+    @property
+    def ready(self) -> bool:
+        return self.gameState & GameState.kReady == GameState.kReady
```
dgelessus:
```suggestion
        return GameState.kReady in self.gameState
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        amOwner = self.gameCli.isLocallyOwned
+        PtDebugPrint(f"_QuabVarSyncBrain.OnOwnerChanged(): Ownership of VarSync has changed {amOwner=} {newOwnerID=}", level=kWarningLevel)
+        if amOwner:
+            self.parent.HandleOwnershipChange()
+
+    def OnNotify(self, events):
+        PtDebugPrint(f"_QuabVarSyncBrain.OnNotify(): Unhandled {events=}")
+
+    def ICheckGameReady(self) -> None:
+        if not self.gameState & GameState.kInitialSyncComplete:
```
dgelessus:
```suggestion
        if GameState.kInitialSyncComplete not in self.gameState:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+            PtDebugPrint("_QuabVarSyncBrain.ICheckGameReady(): The initial var sync is NOT complete, we are clearly not ready", level=kDebugDumpLevel)
+            return
+
+        if not self.ready and all((i.varsReady for i in self.quabState)):
```
dgelessus:
```suggestion
        if not self.ready and all(i.varsReady for i in self.quabState):
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+            PtDebugPrint("_QuabVarSyncBrain.ICheckGameReady(): All variables created!", level=kWarningLevel)
+            self.gameState |= GameState.kAllVarsCreated
+
+        if self.ready:
+            self.parent.SpawnQuabs()
+            return
+
+        # The initial state is available, but not all variables are known. Therefore, we must create them.
+        if self.gameState & GameState.kVarCreateRequested:
```
dgelessus:
```suggestion
        if GameState.kVarCreateRequested in self.gameState:
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+                    PtDebugPrint(f"_QuabPyNotifyBrain.OnNotify(): Invalid data type {event[1]=} {eventDataType=}")
+                    continue
+
+                var.value = event[3]
+                var.lastUpdate = secs
+                updatedQuabs.add(quab)
+
+                # It's probably best to just assume that receiving a valid variable notification
+                # is enough to be a state init.
+                self.gameState |= (GameState.kInitialSyncComplete | GameState.kAllVarsCreated)
```
dgelessus:
```suggestion
                self.gameState |= GameState.kReady
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+        ).scale(100)
+        curPos = quab.brain.getSceneObject().position()
+        return ptPoint3(
+            curPos.getX() + vec.getX(),
+            curPos.getY() + vec.getY(),
+            curPos.getZ() + vec.getZ()
+        )
+
+    def _UpdateQuabGoal(self, quab: _QuabState, *, force: bool = False, sync: bool = False):
+        # We only want to update the goal every 3 seconds while a qub is running.
```
dgelessus:
```suggestion
        # We only want to update the goal every 3 seconds while a quab is running.
```

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +72,737 @@
...
+            self.brain.startBehavior(beh)
+            if run and goal:
+                self.brain.goToGoal(goal, True)
+
+    @property
+    def varsReady(self) -> bool:
+        return all(i.varID is not None for i in self.itervars())
+
+
+class GameState(enum.IntFlag):
```
dgelessus: If this is never de-/serialized anywhere and only used for internal state tracking, we don't need the implicit `int` conversion behavior of `IntFlag`:

```suggestion
class GameState(enum.Flag):
```

Sources/Plasma/FeatureLib/pfPython/pyGmVarSyncGlue.cpp
Hoikas: Thanks for the reminder. This commit is from before your generate_stubs.py PR ;)

Sources/Plasma/FeatureLib/pfPython/pyGmVarSync.h
Hoikas: Indeed

Hoikas:
I've updated this to address feedback and tested that everything works as expected.

Scripts/Python/ahnyQuabs.py
```diff
@@ -63,148 +71,717 @@
...
+@dataclass
+class _QuabVar:
+    name: str
+    varID: Optional[int] = None
+    value: float = field(default=0.0, compare=False)
+    dirty: bool = field(default=False, compare=False)
+    lastUpdate: float = field(default=0.0, compare=False)
+
+
+@dataclass(unsafe_hash=True)
```
dgelessus, line 83: Yeah, I remembered too late that mutable dataclasses aren't hashable by default...

Sources/Plasma/FeatureLib/pfPython/pyGmVarSyncGlue.cpp
```diff
@@ -0,0 +1,152 @@
...
+    PYTHON_METHOD_STATIC(ptGmVarSync, join,
+        "Type: (handler: Any) -> None\n"
+        "Join the common var sync game in the current Age."
+    ),
+    PYTHON_METHOD_STATIC_NOARGS(ptGmVarSync, isSupported,
+        "Type: () -> bool\n"
+        "Checks for the presence of a server-side var sync game manager."
+    ),
+    PYTHON_METHOD(ptGmVarSync, setVariable,
+        "Type: (varID: int, value: Union[float, str])\n"
```
dgelessus:
```suggestion
        "Type: (varID: int, value: Union[float, str]) -> None\n"
```

Sources/Plasma/FeatureLib/pfPython/pyGmVarSyncGlue.cpp
```diff
@@ -0,0 +1,152 @@
...
+        "Change the value of a variable on the server."
+    ),
+    PYTHON_METHOD(ptGmVarSync, createVariable,
+        "Type: (varName: str, value: Union[float, str])\n"
```
dgelessus:
```suggestion
        "Type: (varName: str, value: Union[float, str]) -> None\n"
```

dgelessus (commented):
I may try to do a less surface-level review later - but as I'm not familiar with the inner workings of the GameMgr stuff or any version of the quab AI, I might not have much useful further feedback.

Scripts/Python/ahnyQuabs.py
Hoikas: No worries, I was *extremely* pleased to be able to remove the custom `__eq__()` and `__hash__()` implementations.

Hoikas:
> There are a couple of asserts guarding against plMessages with an unknown class index, but they should only affect debug builds, so in standard gameplay it should be fine.

Yeah. And, at this time, I'm not letting any of the Python code send the new messages over the network, so the new messages shouldn't pose an issue. The tricky part was making sure that we don't clobber the legacy client's ahnyQuabs.py state. It doesn't really examine the variable names at all - it just slices the variable name and uses the quab number to index into some internal state. Yikes.

> So I take it that this can currently only be tested on Cyan's servers?

You can *test* it on MOSS, but, if the owner leaves the Age, the quabs will all disappear and not respawn. The new Python code is expecting that the server will assign a new VarSync game owner correctly when the current owner leaves. I could mitigate that in code by adding a special case to revert to "who owns the scene object" ownership model as the PyNotify/DirtSand uses when the VarSync game starts reporting MOSS's special invalid owner ID (ownerID == 1):

```C++
void VarSyncGameMgr::player_left(kinum_t player, GameServer *server) {
  // Because of the buggy Python for quab games, owner handoff is a
  // nightmare: exception-fest galore. Since the game is broken anyway,
  // just break it preemptively by not transferring the owner.

  // We can't leave the owner as the player who left, because they may return
  // to the age. We can't set the owner to zero because that means the next
  // to link in will be assigned the owner. Setting it to 1 (which is not a
  // valid KI number) works around these problems. Using -1 feels like it
  // could trip a special case in client code.
  m_owner = 1;
  GameMgr::player_left(player, server);
  if (m_players.size() == 0) {
    // we can revive the game next time someone arrives, so fix up the owner
    // back to "no owner"
    m_owner = 0;
    // age is now empty -- throw out the variables
    m_vars.clear();
  }
}

```

but I'm not a huge fan of adding such a hack when MOSS itself could be fixed instead.

PR #1726, opened 2025-07-02
Remove `PtGetPublicAgeList` `cbObject` parameter that had no effect

dgelessus:
The `gotPublicAgeList` callback method is called on any PythonFileMod that has one, not on the particular object passed to `PtGetPublicAgeList`.

Alternatively, I could try to change the implementation so that it actually uses the `cbObject` argument for calling the callback method. It depends on which style of callback API we would prefer going forward.

(For reference, a while back in 9b8523cdf7493a2889e81f26abb884cb6a4d98c0 in #1486, I made a similar change for another Python function.)

### DIFF
--- a/Scripts/Python/Personal.py
+++ b/Scripts/Python/Personal.py
@@ -119,7 +119,7 @@ class Personal(ptResponder):
             if cityInfo:
                 if cityInfo.getAgeInstanceGuid()==kEmptyGuid:
                     # we don't have it yet, so make it! (the callback will make it for us)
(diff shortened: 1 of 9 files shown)


Hoikas (approved):
> Alternatively, I could try to change the implementation so that it actually uses the cbObject argument for calling the callback method. It depends on which style of callback API we would prefer going forward.

That might be nice at some point, or we should pick back up Python awaitables as I was proposing in #1149. For now, I think this is a good idea as-is.

PR #1727, opened 2025-07-04
WIP: Finish the engine side of GameMgr.

Hoikas:
This changeset readds all of the GameMgr games. Unfortunately, the work is not complete, but I figure this needs to be out here so folks can start poking.

One fun thing I added is `xGameCliDbgHandler.py`, which will let you play with the GameMgr in the console. Get started by:
```python
from PlasmaGame import *
from PlasmaGameConstants import *
import xGameCliDbgHandler

handler = xGameCliDbgHandler.new()
```

Then you can join a game by doing something like `ptGmClimbingWall.join(handler, 0)` and watch the results in the Python log. Fire off GameMgr messages using e.g. `handler.gameCli.changeNumBlockers(20)`.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnGameMgr/pnGmClimbingWall.h
+++ b/Sources/Plasma/NucleusLib/pnGameMgr/pnGmClimbingWall.h
@@ -1,0 +1,191 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011 Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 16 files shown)


Sources/Plasma/NucleusLib/pnGameMgr/pnGmClimbingWall.h
```diff
@@ -0,0 +1,191 @@
...
+    kSrv2Cli_ClimbingWall_Ready,
+    kSrv2Cli_ClimbingWall_BlockersChanged,
+    kSrv2Cli_ClimbingWall_PlayerEntered,
+    kSrv2Cli_ClimbingWall_SuitMachineLocked,
+    kSrv2Cli_ClimbingWall_GameOver,
+};
+
+//============================================================================
+// Begin networked data structures
+#include <PshPack1.h>
```
dpogue: This isn't cross-platform, right?

Hoikas: Probably not. It looks like the other files are doing

```C++
// Begin networked data scructures
#pragma pack(push, 1)
```

PR #1729, opened 2025-07-06
Add state for smashed Bevin heek table

dpogue:
Depends on https://github.com/H-uru/moul-assets/pull/286 for content


dpogue:
The content side of this is proving difficult, so I'll close this for now and reopen when I've got working content.

PR #1730, opened 2025-07-08
Adding dedicated hsDisplayHndl type

colincornaby:
This PR is in preparation for further macOS resolution work. I thought I'd ask for review for just this chunk to lighten the load in the next PR.

Plasma uses hsWindowHndl for both the window and display types. This worked under Windows because both displays and windows are HWND types. (That's technically not true, under Windows the display should be an HMONITOR type. Was this refactored at some point?)

I'd added the dedicated type and made it a CGDirectDisplayID on macOS.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -199,7 +199,7 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
     [window setDelegate:self];
     
     gClient.SetClientWindow((__bridge void *)view.layer);
-    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] intValue]);
+    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
 
     self = [super initWithWindow:window];
     self.window.acceptsMouseMovedEvents = YES;
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -199,7 +199,7 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
     [window setDelegate:self];
     
     gClient.SetClientWindow((__bridge void *)view.layer);
-    gClient.SetClientDisplay((hsWindowHndl)NULL);
+    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] intValue]);
 
     self = [super initWithWindow:window];
     self.window.acceptsMouseMovedEvents = YES;
--- a/Sources/Plasma/Apps/plClient/plClientLoader.h
+++ b/Sources/Plasma/Apps/plClient/plClientLoader.h
@@ -48,7 +48,7 @@ class plClientLoader : private hsThread
 {
     class plClient* fClient;
     hsWindowHndl fWindow;
-    hsWindowHndl fDisplay;
+    hsDisplayHndl fDisplay;
     uint32_t fDevType;
 
     void OnQuit() override
@@ -85,7 +85,7 @@ public:
     /**
      * Sets the client display handle.
      */
-    void SetClientDisplay(hsWindowHndl hDC) { fDisplay = hDC; }
+    void SetClientDisplay(hsDisplayHndl hDC) { fDisplay = hDC; }
 
     /**
      * Sets the preferred rendering backend for the client pipeline.
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -69,17 +69,29 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
     struct HINSTANCE__; typedef struct HINSTANCE__ *HINSTANCE;
 
     typedef HWND hsWindowHndl;
+    typedef HWND hsDisplayHndl;
     typedef HINSTANCE hsWindowInst;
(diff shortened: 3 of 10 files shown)


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -69,17 +69,29 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
     struct HINSTANCE__; typedef struct HINSTANCE__ *HINSTANCE;
 
     typedef HWND hsWindowHndl;
+    typedef HWND hsDisplayHndl;
     typedef HINSTANCE hsWindowInst;
     typedef HINSTANCE HMODULE;
     typedef HMODULE hsLibraryHndl;
     typedef long HRESULT;
     typedef void* HANDLE;
 #elif HS_BUILD_FOR_MACOS
```
dpogue: I think this should now be `#elif defined(HS_BUILD_FOR_APPLE)` since it's not macOS-specific?

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -69,17 +69,29 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    // Same note as Windows above - would rather not forward declare but I don't want to
+    // import Foundation or CoreGraphics
+#if HS_BUILD_FOR_IOS
+    // Exception - iOS doesn't support CGDirectDisplayID.
+    // It has UIScreen but that's a Cocoa type.
+    typedef void* CGDirectDisplayID;
+#else
```
dpogue, line 85: I think this should be `#endif` not else? (otherwise hsDisplayHndl never actually gets defined for iOS)

or else CGDirectDisplayID should be hsDisplayHndl

dpogue (commented):
I could have sworn I'd done something like this already for the OpenGL stuff, but I don't see it on any of my branches 🤷🏼‍♂️ 

The approach makes sense to me, but a few comments on the implementation here

Sources/Plasma/Apps/plClient/plClientLoader.h
```diff
@@ -85,7 +85,7 @@ class plClientLoader : private hsThread
     /**
      * Sets the client display handle.
      */
-    void SetClientDisplay(hsWindowHndl hDC) { fDisplay = hDC; }
+    void SetClientDisplay(hsDisplayHndl hDC) { fDisplay = hDC; }
```
dpogue, line 88: You mentioned `HMONITOR` for Windows, but in the way the OpenGL code uses (/will use) this, it's expected to be the `HDC` device context

colincornaby:
@dpogue - You're right on the preprocessor bits - I'm amending the commit and force pushing.

I... don't know what we'll do with iOS here. iOS requires the use of UIDisplay. But I guess we can cram that through a void *.

Sources/Plasma/Apps/plClient/plClientLoader.h
colincornaby: Got it.

On the Mac side - I've written the second part of this that also updates the OpenGL code to make use of the CoreGraphics display id.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -199,7 +199,7 @@ - (id)init
     [window setDelegate:self];
     
     gClient.SetClientWindow((__bridge void *)view.layer);
-    gClient.SetClientDisplay((hsWindowHndl)NULL);
+    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] intValue]);
```
dgelessus: Because `CGDirectDisplayID` is `uint32_t`, this would be slightly more correct, right?

```suggestion
    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
```

dgelessus (commented):
If `hsDisplayHndl` is only used in graphics-related code, would plPipeline for example be a better place for this declaration? As far as I can tell, there's no need to declare it in HeadSpin.h, which gets included by almost everything.

I realize that the existing code declares `hsWindowHndl` and others in HeadSpin.h, but at some point I'd like to try to move those out of HeadSpin.h to more appropriate places. (I suspect that `hsWindowHndl` is only in HeadSpin.h because it was used by `hsMessageBoxWithOwner`, but that went away with 7feef4212477dd366d43169b8a01e51535a3f9f7 in #1466.) So it would be nice if we didn't add more graphics-specific definitions to the "absolute core header".

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -69,17 +69,29 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
     typedef HINSTANCE hsWindowInst;
     typedef HINSTANCE HMODULE;
     typedef HMODULE hsLibraryHndl;
     typedef long HRESULT;
     typedef void* HANDLE;
-#elif HS_BUILD_FOR_MACOS
+#elif HS_BUILD_FOR_APPLE
+    // Same note as Windows above - would rather not forward declare but I don't want to
+    // import Foundation or CoreGraphics
+#if HS_BUILD_FOR_IOS
```
Hoikas:
```suggestion
#ifdef HS_BUILD_FOR_IOS
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
colincornaby: Yes. Technically correct - the best kind of correct! 😆

Sources/Plasma/CoreLib/HeadSpin.h
colincornaby: Thanks - I force pushed this change in.

PR #1731, opened 2025-07-12
Mac resolution changing support

colincornaby:
- Adding display helper to help rendering backends query Mac displays
- Filtering to make sure resolutions don’t include notch area (which we don’t draw into)
- Adding proper hooks into Metal device and renderer
- Mac frontend now responds to resolution switching
- Resizing in windowed mode now is constrained to the aspect mode of the resolution
- Adding a ruleset for picking a default resolution for Metal
- Aligning the OpenGL renderer with these changes

This changes a few things in the Mac client behavior:
- The user can no longer freely resize the resolution in windowed mode. The user can resize an aspect locked window, but it will not change the resolution. Just like in Windows.
- The full screen resolution will follow the set resolution, just like Windows
- Multiple displays are now broken - just like Windows.

This is the PR that I've been bouncing Claude Code against. But if you've been following my blog posts you know that went... not great. The code here was written by me.


Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -261,6 +265,16 @@ void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
         hsG3DDeviceRecord devRec;
         devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
         devRec.SetDriverName("OpenGL.framework");
+        
+        plDisplayHelper* displayHelper = plDisplayHelper::CurrentDisplayHelper();
```
colincornaby: @dpogue - FYI, I've upgraded the GL backend to hook into the new code.

Display helpers are a platform specific object that can vend display information to multiple backends. I kept this concept from the previous draft PR. So all rendering backends on the Mac can sit on the Mac display helper.

This lets OpenGL inherit all the behaviors I baked in for the Mac client like handling displays with a notch.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,153 @@
...
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plPipeline.h"
+#include "plMacDisplayHelper.h"
+
+// CGDirectDisplayCopyCurrentMetalDevice only declared in Metal.h?
+#include <Metal/Metal.h>
```
dpogue: Can this go in plMetalEnumerate.mm? We don't seem to use `CGDirectDisplayCopyCurrentMetalDevice` in here...

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,153 @@
...
+    // Calculate the region actually available for full screen
+    NSRect currentResolution = [screen frame];
+    if (@available(macOS 12.0, *)) {
+        NSEdgeInsets currentSafeAreaInsets = [screen safeAreaInsets];
+        // Sigh... Origin doesn't matter but lets do it for inspectability
+        currentResolution.origin.x += currentSafeAreaInsets.left;
+        currentResolution.origin.y += currentSafeAreaInsets.top;
+        currentResolution.size.width -= currentSafeAreaInsets.left + currentSafeAreaInsets.right;
+        currentResolution.size.height -= currentSafeAreaInsets.top + currentSafeAreaInsets.bottom;
+    }
```
colincornaby, line 95: This is an ugly bit here. On MacBook Pros, we can only draw below the notch. But they'll supply us both notch-y and notch-less resolutions. We need to filter out the resolutions that are for an area we can't draw into.

Unfortunately this code also filters out any aspect ratio that doesn't match the panel aspect ratio. I'm considering how to be less aggressive with the filtering.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,76 @@
...
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#ifndef plMacDisplayHelper_hpp
+#define plMacDisplayHelper_hpp
+
+#include <stdio.h>
```
dpogue: Prefer `<cstdio>` (if it's actually needed here?)

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -580,6 +601,11 @@ - (NSApplicationPresentationOptions)window:(NSWindow*)window
            NSApplicationPresentationAutoHideMenuBar;
 }
 
+- (void)windowDidChangeScreen:(NSNotification *)notification
+{
+    _displayHelper->SetCurrentScreen(self.window.screen);
```
colincornaby: This should be removed - I'll make sure it gets deleted. It's leftover from an attempt to make this support multiple displays.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby, line 1: Sorry @dpogue - this had to become an .mm. CGDirectDisplayCopyCurrentMetalDevice returns a Cocoa type. I embraced the Cocoa in the rest of the file.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -41,26 +41,24 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 #include "plMetalPipeline.h"
 
 void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
 {
-    // For now - just use the default device. If there is a high power discrete device - this will
-    // spin it up. This will also automatically pin us to an eGPU if present and the user has
-    // configured us to use it.
-    MTL::Device* device = MTL::CreateSystemDefaultDevice();
+    CGDirectDisplayID mainDisplay = CGMainDisplayID();
+    id<MTLDevice> device = CGDirectDisplayCopyCurrentMetalDevice(mainDisplay);
```
colincornaby: I'm actually suspicious of what this might do on a dual GPU MacBook Pro. This may not kick on the discrete GPU.

Plasma's original architecture definitely never had to contend with GPU switching. If this doesn't force a GPU switch I could force the GPU a different way.

I'll need to get out a dual GPU MacBook Pro to test this.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
dpogue: Never did I expect to see safe area inset code in Plasma, after spending multiple years of my life dealing with safe area insets on the web 🙃

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,153 @@
...
+void plMacDisplayHelper::SetCurrentScreen(hsDisplayHndl display)
+{
+    NSScreen *nsScreen;
+    for (NSScreen *screen in [NSScreen screens]) {
+        NSDictionary *deviceDescription = [screen deviceDescription];
+        NSNumber *screenID = deviceDescription[@"NSScreenNumber"];
+        if ([screenID unsignedIntValue] == display) {
+            nsScreen = screen;
+        }
+    }
```
dpogue: Are you sure this can't be done with `CGGetActiveDisplayList`? 😝

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,153 @@
...
+        if(modeAspectRatio < safeAspectRatio)
+        {
+            continue;
+        }
+
+        // aspect ratio is good - add the mode to the list
+        plDisplayMode plasmaMode;
+        plasmaMode.ColorDepth = 32;
+        plasmaMode.Width = int(CGDisplayModeGetWidth(mode));
+        plasmaMode.Height = int(CGDisplayModeGetHeight(mode));
```
dpogue: I know we're handling scaling elsewhere, but would it make sense for plDisplayMode to contain a scaling factor?

@Hoikas Is that something that would make any sense with Windows DPI stuff, or is that too weird?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
dpogue: Compiler won't let you cast it to `MTL::Device*`? Maybe you should embrace `objc_msgSend` :trollface:

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,55 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+            devMode.SetWidth(mode.Width);
+            devMode.SetHeight(mode.Height);
+            devMode.SetColorDepth(mode.ColorDepth);
+            devRec.GetModes().emplace_back(devMode);
+        }
+
+        // now inspect the GPU and figure out a good default resolution
+        // This code is in Metal (for now) - but it should also work
+        // under OpenGL/Mac as long as the GPU also supports Metal.
+        if (@available(macOS 10.15, *)) {
```
dpogue, line 92: We have a `HAVE_BUILTIN_AVAILABLE` macro that's set for when `@available` is defined (or `__builtin_available` in C++)

I doubt this is ever getting built on compilers without that, but it might be good to wrap it for consistency

Sources/Plasma/NucleusLib/inc/pnSingletons.cpp
```diff
@@ -43,7 +43,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plgDispatch.h"
 #include "hsResMgr.h"
 #include "plPipeResReq.h"
+#include "plPipeline.h"
 
+plDisplayHelper* plDisplayHelper::_currentDisplayHelper = nullptr;
```
dpogue: Feels really weird sticking this in NucleusLib, but that's where plPipeline lives, so maybe it's okay?

pl3DPipeline in PubUtilLib might be a better spot

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,76 @@
...
+    void SetCurrentScreen(NSScreen *screen);
+
+    plDisplayMode DesktopDisplayMode() override { return fDesktopDisplayMode; };
+    std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) override;
+    
+    plMacDisplayHelper();
+    ~plMacDisplayHelper();
+private:
+    CGDirectDisplayID fCurrentDisplay = -1;
+    MTL::Device* fRenderDevice = nullptr;
```
dpogue: This doesn't ever appear to get initialized or assigned a value, and my preference would be to avoid introducing Metal-isms into shared code as much as possible

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -357,6 +373,9 @@ class plPipeline : public plCreatable
     
     float fBackingScale = 1.0f;
     void SetBackingScale(float scale) { fBackingScale = scale; };
+    
+    std::shared_ptr<plDisplayHelper> fDisplayHelper;
+    void SetDisplayHelper(std::shared_ptr<plDisplayHelper> helper) { fDisplayHelper = helper; };
```
dpogue: Do we need both the pipeline holding this, and an internal singleton in plDisplayHelper? Is this member variable used anywhere in the pipeline or does it always use `plDisplayHelper::CurrentDisplayHelper()`?

I'm undecided which my preference would be... I guess we need the display helper before we have a pipeline?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
dpogue: I guess the issue is just importing the ObjC header in general?
Can you declare it as `extern "C" MTL::Device* CGDirectDisplayCopyCurrentMetalDevice(CGDirectDisplayID display)`? >_>

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: I wish, actually. I have feedback in with Apple around this whole topic and I suggested they implement their own filter under the CoreGraphics functions. But right now NSScreen is the only way.

Sources/Plasma/NucleusLib/inc/plPipeline.h
colincornaby: I think this is leftover from the previous draft. IIRC I didn't want a singleton but it was forced by resolutions needing to be queried during the pipelines constructor.

colincornaby:
I've done some force pushing to clean up the detritus from earlier iterations.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
colincornaby: plMacDisplayHelper used to decide the default resolution - but I farmed that back out to the Metal enumerator. I've removed the remnants of this.

Do you have any opinions here? That would leave the GL renderer also responsible for deciding the default resolution. But that might be where that responsibility belongs.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby: I'll go ahead and wrap it. If this were code out in some generic or OpenGL part of the client I'd be more worried. But this being Metal code keeps it in a certain compiler range.

colincornaby:
Doing another pass through - the Metal enumerate will need to get the device back out. Right now the pipeline just picks its own device. That will happen to work on a lot of Macs but isn't going to be correct everywhere.

I'll have to look again at how D3D does it.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+        if ([screenID unsignedIntValue] == display) {
+            nsScreen = screen;
+        }
+    }
+    SetCurrentScreen(nsScreen);
+}
+
+void plMacDisplayHelper::SetCurrentScreen(NSScreen *screen)
+{
+    CGDirectDisplayID displayID = (CGDirectDisplayID)[screen.deviceDescription[@"NSScreenNumber"] intValue];
```
dpogue: `unsignedIntValue`?

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+    fDesktopDisplayMode.Height = windowArea.size.height;
+    fDesktopDisplayMode.ColorDepth = 32;
+    
+    // visibleFrame is in points - put into pixels
+    fDesktopDisplayMode.Width *= [screen backingScaleFactor];
+    fDesktopDisplayMode.Height *= [screen backingScaleFactor];
+    
+    // Calculate the region actually available for full screen
+    NSRect currentResolution = [screen frame];
+    if (@available(macOS 12.0, *)) {
```
dpogue, line 88: This should also be wrapped in `HAVE_BUILTIN_AVAILABLE`

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+        currentResolution.size.height -= currentSafeAreaInsets.top + currentSafeAreaInsets.bottom;
+    }
+
+    float safeAspectRatio = currentResolution.size.width / currentResolution.size.height;
+
+    fDisplayModes.clear();
+
+    CFArrayRef displayModes = CGDisplayCopyAllDisplayModes(fCurrentDisplay, nullptr);
+    for(int i=0; i< CFArrayGetCount(displayModes); i++)
+    {
```
dpogue: nit: code formatting (spacing and brace)

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+        // We could force that with kCGDisplayShowDuplicateLowResolutionModes but the aspect ratio
+        // filter would still drop some or all of those.
+        //
+        // Asked for better guidance here from Apple - FB13375033
+        
+        CGDisplayModeRef mode = (CGDisplayModeRef)CFArrayGetValueAtIndex(displayModes, i);
+
+        float modeAspectRatio = float(CGDisplayModeGetWidth(mode)) / float(CGDisplayModeGetHeight(mode));
+        if(modeAspectRatio < safeAspectRatio)
+        {
```
dpogue: nit: brace format and spacing around condition

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+    SetCurrentScreen(nsScreen);
+}
+
+void plMacDisplayHelper::SetCurrentScreen(NSScreen *screen)
+{
+    CGDirectDisplayID displayID = (CGDirectDisplayID)[screen.deviceDescription[@"NSScreenNumber"] intValue];
+    if (fCurrentDisplay == displayID)
+    {
+        return;
+    }
```
dpogue: nit: no need for braces around a 1-line unambiguous `if` statement

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+        NSDictionary *deviceDescription = [screen deviceDescription];
+        NSNumber *screenID = deviceDescription[@"NSScreenNumber"];
+        if ([screenID unsignedIntValue] == display) {
+            nsScreen = screen;
+        }
+    }
+    SetCurrentScreen(nsScreen);
+}
+
+void plMacDisplayHelper::SetCurrentScreen(NSScreen *screen)
```
dpogue: Would it make sense to refactor the looping over `[NSScreen screens]` into its own static function that takes in an `hsDisplayHndl` and returns an `NSScreen*`, and then have this method just take the `hsDisplayHndl`?

I'm wondering if there's a way to avoid needing to have `NSScreen` defined in the header file, because that precludes it from being included in any non-ObjC files.

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
```diff
@@ -240,6 +241,9 @@ class hsG3DDeviceRecord
     bool    IsInvalid() const { return 0 != ( fFlags & kInvalid ); }
 
     std::vector<hsG3DDeviceMode>& GetModes() { return fModes; }
+    
+    hsG3DDeviceMode   GetDefaultMode() const { return fDefaultMode; }
+    void    SetDefaultMode( hsG3DDeviceMode mode ) { fDefaultMode = mode; }
```
dpogue: Where is `GetDefaultMode()` used? I see this being set, but never read

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // if it's a Metal 3 GPU - it should be good
+                // Pick the native display resolution
+                // (Re-picking the first one here for clarity)
+                defaultMode = devRec.GetModes().front();
+            }
+        }
+#endif
+
+        if (defaultMode.GetWidth() == 0)
+        {
```
dpogue: nit: formatting on this and other if statements in the same block

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -580,6 +601,10 @@ - (NSApplicationPresentationOptions)window:(NSWindow*)window
            NSApplicationPresentationAutoHideMenuBar;
 }
 
+- (void)windowDidChangeScreen:(NSNotification *)notification
+{
+}
+
```
dpogue: needed?

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -232,20 +232,24 @@ void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
 {
     IGNORE_WARNINGS_BEGIN("deprecated-declarations")
     CGLPixelFormatObj pix = nullptr;
     CGLContextObj ctx = nullptr;
+    
+    CGDirectDisplayID mainDisplay = CGMainDisplayID();
 
     do {
-        CGLPixelFormatAttribute attribs[6] = {
+        CGLPixelFormatAttribute attribs[8] = {
```
dgelessus: No need to manually count array elements:

```suggestion
        CGLPixelFormatAttribute attribs[] = {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -41,26 +41,24 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
-    MTL::Device* device = MTL::CreateSystemDefaultDevice();
+    CGDirectDisplayID mainDisplay = CGMainDisplayID();
+    id<MTLDevice> device = CGDirectDisplayCopyCurrentMetalDevice(mainDisplay);
 
     if (device) {
         hsG3DDeviceRecord devRec;
         devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeMetal);
         devRec.SetDriverName("Metal");
-        devRec.SetDeviceDesc(device->name()->utf8String());
+        devRec.SetDeviceDesc(static_cast<const char*>([device.name UTF8String]));
```
dgelessus: What's the cast needed for? [`NSString.UTF8String`](https://developer.apple.com/documentation/foundation/nsstring/utf8string?language=objc) is already a `const char *`. Plus, if this is Objective-C code now, we can use our custom `NSString` category for the conversion.

```suggestion
        devRec.SetDeviceDesc([device.name STString]);
```

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -138,6 +138,22 @@ class plDisplayMode
...
+{
+public:
+    virtual plDisplayMode DesktopDisplayMode() = 0;
+    virtual std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) = 0;
+    
+    plDisplayHelper() = default;
+    virtual ~plDisplayHelper() = default;
+    
+    static plDisplayHelper* CurrentDisplayHelper() { return _currentDisplayHelper; }
+    void MakeCurrentDisplayHelper() { _currentDisplayHelper = this; }
```
Hoikas: The rest of the engine uses `SetInstance()` and `GetInstance()` for singletons. It would be nice to maintain consistency with the already existing patterns.

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -138,6 +138,22 @@ class plDisplayMode
     int ColorDepth;
 };
 
+class plDisplayHelper
+{
+public:
+    virtual plDisplayMode DesktopDisplayMode() = 0;
+    virtual std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) = 0;
+    
+    plDisplayHelper() = default;
```
Hoikas:
```suggestion
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,74 @@
...
+// different for older GPUs.
+#include <AppKit/AppKit.h>
+#include <QuartzCore/QuartzCore.h>
+
+#include "plPipeline/hsG3DDeviceSelector.h"
+
+class plMacDisplayHelper: public plDisplayHelper
+{
+public:
+    CGDirectDisplayID CurrentDisplay() { return fCurrentDisplay; }
```
Hoikas:
```suggestion
    CGDirectDisplayID CurrentDisplay() const { return fCurrentDisplay; }
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,74 @@
...
+
+    plDisplayMode DesktopDisplayMode() override { return fDesktopDisplayMode; };
+    std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) override;
+    
+    plMacDisplayHelper();
+    ~plMacDisplayHelper();
+private:
+    CGDirectDisplayID fCurrentDisplay = -1;
```
Hoikas: Needs to be initialized in the constructor

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,74 @@
...
+*==LICENSE==*/
+
+#ifndef plMacDisplayHelper_hpp
+#define plMacDisplayHelper_hpp
+
+// Currently requires Metal to query attached GPU capabilities
+// Capability check will also work for GL - but will need something
+// different for older GPUs.
+#include <AppKit/AppKit.h>
+#include <QuartzCore/QuartzCore.h>
```
Hoikas, line 50: Prefer to not include platform specific stuff in headers if possible.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -107,6 +108,7 @@ @interface AppDelegate : NSWindowController <NSApplicationDelegate,
    @public
     plClientLoader gClient;
     dispatch_source_t _displaySource;
+    std::shared_ptr<plMacDisplayHelper> _displayHelper;
```
Hoikas: Admittedly, I have an unnatural distaste for `std::shared_ptr`, but the display helper doesn't actually seem to be shared. This should probably be at most `std::unique_ptr`  or not a pointer at all, like `gClient`.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -580,6 +601,10 @@ - (NSApplicationPresentationOptions)window:(NSWindow*)window
            NSApplicationPresentationAutoHideMenuBar;
 }
 
+- (void)windowDidChangeScreen:(NSNotification *)notification
+{
+}
```
Hoikas:
```suggestion
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+
+#include "plPipeline.h"
+#include "plMacDisplayHelper.h"
+
+void plMacDisplayHelper::SetCurrentScreen(hsDisplayHndl display)
+{
+    NSScreen *nsScreen;
+    for (NSScreen *screen in [NSScreen screens]) {
+        NSDictionary *deviceDescription = [screen deviceDescription];
+        NSNumber *screenID = deviceDescription[@"NSScreenNumber"];
```
Hoikas: Pointers bind to type

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
Hoikas: Pointer binds to type

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+    SetCurrentScreen(nsScreen);
+}
+
+void plMacDisplayHelper::SetCurrentScreen(NSScreen *screen)
+{
+    CGDirectDisplayID displayID = (CGDirectDisplayID)[screen.deviceDescription[@"NSScreenNumber"] intValue];
+    if (fCurrentDisplay == displayID)
+    {
```
Hoikas: style

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
Hoikas: `if` is not a function. Brackets on same line.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
Hoikas: `if` is not a function. Brackets on same line.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+            continue;
+        }
+
+        // aspect ratio is good - add the mode to the list
+        plDisplayMode plasmaMode;
+        plasmaMode.ColorDepth = 32;
+        plasmaMode.Width = int(CGDisplayModeGetWidth(mode));
+        plasmaMode.Height = int(CGDisplayModeGetHeight(mode));
+
+        fDisplayModes.push_back(plasmaMode);
```
Hoikas: Prefer to construct in place with `emplace_back()`.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
Hoikas: `plWinDpi` largely handles all of this right now, so I don't know what the value of adding a scale factor here is, but if we wanted to, `plWinDpi::GetDpi()` shows how to extract the DPI for an arbitrary `HMONITOR` on Windows 8.1, which can be converted to a scale factor by dividing by `96.f`.

```C++
    UINT dpiX, dpiY;
    HRESULT hr = GetDpiForMonitor(someHMonitorHere, MDT_EFFECTIVE_DPI, &dpiX, &dpiY);
    if (SUCCEEDED(hr)) {
        hsAssert(dpiX == dpiY, "X and Y dpi values should be the same on Windows apps");
        float scale  = (float)dpiY / 96.f; 
        return scale;
    }
    LogRed("Per-Monitor DPI failed: {}", hsCOMError(hr.value()));
    return 0.f;
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,147 @@
...
+    return fDisplayModes;
+}
+
+plMacDisplayHelper::plMacDisplayHelper()
+{
+}
+
+plMacDisplayHelper::~plMacDisplayHelper()
+{
+}
```
Hoikas: Should probably just be defaulted/omitted.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -261,6 +265,16 @@ void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
         devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
         devRec.SetDriverName("OpenGL.framework");
+        
+        plDisplayHelper* displayHelper = plDisplayHelper::CurrentDisplayHelper();
+        for (const auto& mode : displayHelper->GetSupportedDisplayModes(mainDisplay)) {
+            hsG3DDeviceMode devMode;
+            devMode.SetWidth(mode.Width);
+            devMode.SetHeight(mode.Height);
+            devMode.SetColorDepth(mode.ColorDepth);
+            devRec.GetModes().emplace_back(devMode);
```
Hoikas: This is still making a copy. You'll need to `std::move(devMode)` or supply the arguments to `emplace_back()` to construct in place.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
-        devMode.SetColorDepth(hsG3DDeviceSelector::kDefaultDepth);
-        devRec.GetModes().emplace_back(devMode);
+        plDisplayHelper* displayHelper = plDisplayHelper::CurrentDisplayHelper();
+        hsG3DDeviceMode defaultMode;
+        for (const auto& mode : displayHelper->GetSupportedDisplayModes(mainDisplay)) {
+            hsG3DDeviceMode devMode;
+            devMode.SetWidth(mode.Width);
+            devMode.SetHeight(mode.Height);
+            devMode.SetColorDepth(mode.ColorDepth);
+            devRec.GetModes().emplace_back(devMode);
```
Hoikas: Would be nice to avoid the copy

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+        }
+
+        // now inspect the GPU and figure out a good default resolution
+        // This code is in Metal (for now) - but it should also work
+        // under OpenGL/Mac as long as the GPU also supports Metal.
+#if defined(HAVE_BUILTIN_AVAILABLE)
```
Hoikas:
```suggestion
#ifdef HAVE_BUILTIN_AVAILABLE
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+        if (@available(macOS 10.15, *)) {
+            // The available check needs to be on it's own line or
+            // we'll get a compiler warning
+            if([device supportsFamily:MTLGPUFamilyMac2])
+            {
```
Hoikas: If is not a function. Brackets.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+            }
+        }
+#endif
+
+        if (defaultMode.GetWidth() == 0)
+        {
+            // time to go down the rabit hole
+            int maxWidth = INT_MAX;
+            if([device isLowPower])
+            {
```
Hoikas: Brackets

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // (Re-picking the first one here for clarity)
+                defaultMode = devRec.GetModes().front();
+            }
+        }
+#endif
+
+        if (defaultMode.GetWidth() == 0)
+        {
+            // time to go down the rabit hole
+            int maxWidth = INT_MAX;
```
Hoikas:
```suggestion
            int maxWidth = std::numeric_limits::<int>::max();
```

This is C++

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+            if([device isLowPower])
+            {
+                // integrated - not Apple Silicon, we know it's not very good
+                maxWidth = 1080;
+            }
+            else if([device recommendedMaxWorkingSetSize] < 4000000000)
+            {
```
Hoikas: If is not a function. Brackets.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // if it has less than around 4 gigs of VRAM it might still be performance
+                // limited
+                maxWidth = 1400;
+            }
+            
+            for (auto & mode : devRec.GetModes()) {
```
Hoikas: Reference binds to the type

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                if(mode.GetWidth() <= maxWidth) {
```
Hoikas: If is not a function

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,12 +74,57 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                    defaultMode = mode;
+                    abort();
```
Hoikas: Let's not just throw the whole app away?

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -138,6 +138,22 @@ class plDisplayMode
...
+    virtual ~plDisplayHelper() = default;
+    
+    static plDisplayHelper* CurrentDisplayHelper() { return _currentDisplayHelper; }
+    void MakeCurrentDisplayHelper() { _currentDisplayHelper = this; }
+    
+private:
+    static plDisplayHelper* _currentDisplayHelper;
```
Hoikas:
```suggestion
    static plDisplayHelper* fCurrentDisplayHelper;
```

consistency with field naming

Sources/Plasma/NucleusLib/inc/pnSingletons.cpp
Hoikas: Agreed - the null pipeline probably doesn't care about displays

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
Hoikas: This is making an unnecessary copy.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,74 @@
...
+class plMacDisplayHelper: public plDisplayHelper
+{
+public:
+    CGDirectDisplayID CurrentDisplay() { return fCurrentDisplay; }
+
+    plDisplayMode DesktopDisplayMode() override { return fDesktopDisplayMode; };
+    std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) override;
+    
+    plMacDisplayHelper();
+    ~plMacDisplayHelper();
```
Hoikas: We prefer to see these at the top of the public list.

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -138,6 +138,22 @@ class plDisplayMode
     int ColorDepth;
 };
 
+class plDisplayHelper
+{
+public:
+    virtual plDisplayMode DesktopDisplayMode() = 0;
+    virtual std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) = 0;
```
Hoikas: These feel like the should be `const` methods.

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -138,6 +138,22 @@ class plDisplayMode
...
+    
+    plDisplayHelper() = default;
+    virtual ~plDisplayHelper() = default;
```
Hoikas: We prefer to see these at the top of the public list.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
colincornaby: I'm leaving this in place because it will be. There's going to have to be code to wired in to fix multiple displays - and this will be an important bit.

But thats a deeper set of changes that's going to affect both Mac and Windows.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
colincornaby: Understandable. But - I think it's going to be necessary here. And the only thing that should ever see this header is the Mac client. plDisplayHelper is what the world should see, which is clean.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: I should probably get Clang format going again.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: Let me think about it. I mentioned above to @Hoikas that this header should never be included by anyone but the client. plDisplayHelper is the clean version.

What these bits are circling around is the client needs to be able to tell the display helper when the game window changes displays, and the display helper needs to react to that. Thats why this interface exposes things like NSScreen (and the client has a window did change screen delegate message.)

I should probably decide what a likely path forward looks like for that, and then change the code to match.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: plDisplayMode doesn't have a constructor, do you want me to add one?

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
colincornaby: hsG3DDeviceMode doesn't have a constructor. I'll std::move it for now.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby: I'm not sure it's worth the C++ shenanigans to get this file away from Obj-C++.

One thing to note is that Metal-cpp is not catching API use that's inconsistent without our minimum deployment target. One nice thing about the Obj-C++ use here is it immediately caught that our family query wasn't valid API on 10.14.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby: The category that adds STString is in the client. It's a pain to pull in - and is probably also the unclean choice.

I'm going to remove the cast but continue using UTF8String.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
dgelessus: `plDisplayMode` is basically a struct with no customization, so it should support "aggregate initialization" or whatever the default is called. Does `fDisplayModes.emplace_back(width, height, depth);` not work?

colincornaby: It does not:
`/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.5.sdk/usr/include/c++/v1/__memory/allocator.h:165:24: error: no matching constructor for initialization of 'plDisplayMode'
  165 |     ::new ((void*)__p) _Up(std::forward<_Args>(__args)...);`
  I was going to remark that plDisplayMode is a class and not a structure - but then I read C++17 should potentially support aggregate initialization from POD type classes. So now I'm not sure why this isn't working.

Hoikas: IME you need to have a constructor for construct in place to work, so we should definitely add one. What does work with no extra work is something like `fContainer.emplace_back({ member1, member2, member3 })`, but that may cause copying on these structures with only trivially copyable members.

colincornaby: @Hoikas - It feels like that should work (the approach with {}) - but that also doesn't. Not sure why.

Regardless, it sounds like we might want to add a constructor. But I also think it's a little weird it's not a struct? This type originated from Cyan, right?

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
colincornaby: Looking at this. This never got wired in and needs some massaging. We're using the constant defaults right now to find a default resolution and there needs to be a path to query the default resolution that comes back here.

I'm going to do it in a way where we don't have to mess with Windows right now. Windows can keep using the 800x600 constants.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
Hoikas: Oh, probably because `plDisplayMode` is a class with every member public. I think that could be safely changed to a struct with a constructor.

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
colincornaby: @dpogue - there were some changes here. Please review b9ca050cb94836043c295f3f0de74bacde5434cd.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: I got it figured out. The syntax with emplace is just a little twitchy - different than directly constructing.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
colincornaby: I'll go ahead and remove this. But it will return when I get multiple displays figured out.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
dgelessus: By the way, a `struct` in C++ behaves exactly like a `class`, except that `struct` defaults to `public` access instead of `private`. So whether something is a `class` with everything public or just a `struct` should never make a difference.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -125,3 +125,9 @@
         records.emplace_back(devRec);
     }
 }
+
+MTL::Device* plMetalEnumerate::DeviceForDisplay(hsDisplayHndl display)
+{
+    id<MTLDevice> device = CGDirectDisplayCopyCurrentMetalDevice(display);
+    return static_cast<MTL::Device*>(device);
+}
```
colincornaby, line 133: This is required because there isn't a good way to pass the Metal device from the enumeration types to the pipeline. The display id gets passed, so this is a convenience function to get the Metal device from the display passed to the pipeline.

I'm aware this is muddying the waters a bit with the Obj-C use in this file (this is case back to MTL::Device for the pipeline.) In since the file still has to be Obj-C++ to support CGDirectDisplayCopyCurrentMetalDevice without starting to hack things I'm not sure yet what I'll do here.

dpogue: I really feel like it should be possible to use `extern MTL::Device* CGDirectDisplayCopyCurrentMetalDevice(CGDirectDisplayId)` to get around the Obj-C-ness...

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby: This does cause the discrete GPU to spin up on a dual GPU MacBook Pro.

Sources/Plasma/Apps/plClient/plClient.cpp
```diff
@@ -2061,8 +2061,8 @@ void plClient::IDetectAudioVideoSettings()
     else
 #endif
     {
-        plPipeline::fDefaultPipeParams.Width = hsG3DDeviceSelector::kDefaultWidth;
-        plPipeline::fDefaultPipeParams.Height = hsG3DDeviceSelector::kDefaultHeight;
```
colincornaby, line 2065: Not sure what was going on here. 'mode' is the default mode - and is already matched against kDefaultWidth and kDefaultHeight within 'GetDefault'

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: I'm going to try to put a pin in this one. @dpogue, @Hoikas - this feedback was given in multiple comments so I'm going to try to pull the discussion together here.

The line I'm going to draw is that plMacDisplayHelper is part of the Mac client (which is accurate to where this lives) and thus should never be touched by any other module. Display helpers should inherit from plDisplayHelper, but be owned by the client. It's a way for the client to inject display logic into the engine.

With that said - _my opinion is_ removing Obj-C from the header is unnecessary because this is part of the Obj-C client. And this header should not be imported in any other module.

colincornaby:
Note for myself: Different Macs are returning different resolutions in different orders. I'll probably need to sort if I want to guarantee anything about resolution order.

- [ ] Make sure resolution order is the same across Macs.

colincornaby:
It actually was the SIMD change a few PRs back that broke older Macs. I'll have to dig in more.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,75 @@
...
+#include <QuartzCore/QuartzCore.h>
+
+#include "plPipeline/hsG3DDeviceSelector.h"
+#include "plPipeline/pl3DPipeline.h"
+
+class plMacDisplayHelper: public plDisplayHelper
+{
+public:
+    plMacDisplayHelper();
+    ~plMacDisplayHelper() = default;
```
Hoikas:
```suggestion
```

Hoikas (commented):
It appears that we still have some code style issues, unfortunately.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,150 @@
...
+plMacDisplayHelper::plMacDisplayHelper() :
+    fCurrentDisplay(-1)
+{
+}
+
+void plMacDisplayHelper::SetCurrentScreen(hsDisplayHndl display) const
+{
+    NSScreen* nsScreen;
+    for (NSScreen* screen in [NSScreen screens]) {
+        NSDictionary *deviceDescription = [screen deviceDescription];
```
Hoikas:
```suggestion
        NSDictionary* deviceDescription = [screen deviceDescription];
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,150 @@
...
+        currentResolution.size.height -= currentSafeAreaInsets.top + currentSafeAreaInsets.bottom;
+    }
+#endif
+
+    float safeAspectRatio = currentResolution.size.width / currentResolution.size.height;
+
+    fDisplayModes.clear();
+
+    CFArrayRef displayModes = CGDisplayCopyAllDisplayModes(fCurrentDisplay, nullptr);
+    for(int i=0; i< CFArrayGetCount(displayModes); i++) {
```
Hoikas:
```suggestion
    for (int i = 0; i < CFArrayGetCount(displayModes); i++) {
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,150 @@
...
+        // like 800x600, so don't expect those here. My Macbook Pro gives me back 1080p as the lowest.
+        // We could force that with kCGDisplayShowDuplicateLowResolutionModes but the aspect ratio
+        // filter would still drop some or all of those.
+        //
+        // Asked for better guidance here from Apple - FB13375033
+        
+        CGDisplayModeRef mode = (CGDisplayModeRef)CFArrayGetValueAtIndex(displayModes, i);
+
+        float modeAspectRatio = float(CGDisplayModeGetWidth(mode)) / float(CGDisplayModeGetHeight(mode));
+        if(modeAspectRatio < safeAspectRatio) {
```
Hoikas:
```suggestion
        if (modeAspectRatio < safeAspectRatio) {
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,150 @@
...
+            continue;
+        }
+
+        // aspect ratio is good - add the mode to the list
+
+        // Plasma likes to handle modes from largest to smallest,
+        // CG likes to go from smallest to largest. Insert modes
+        // at the front.
+        fDisplayModes.emplace_back(plDisplayMode{int(CGDisplayModeGetWidth(mode)), int(CGDisplayModeGetHeight(mode)), 32});
```
Hoikas:
```suggestion
        fDisplayModes.emplace_back(plDisplayMode{ int(CGDisplayModeGetWidth(mode)), int(CGDisplayModeGetHeight(mode)), 32 });
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,13 +74,60 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+        }
+
+        // now inspect the GPU and figure out a good default resolution
+        // This code is in Metal (for now) - but it should also work
+        // under OpenGL/Mac as long as the GPU also supports Metal.
+#ifdef HAVE_BUILTIN_AVAILABLE
+        if (@available(macOS 10.15, *)) {
+            // The available check needs to be on it's own line or
+            // we'll get a compiler warning
+            if([device supportsFamily:MTLGPUFamilyMac2]) {
```
Hoikas:
```suggestion
            if ([device supportsFamily:MTLGPUFamilyMac2]) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,13 +74,60 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // (Re-picking the first one here for clarity)
+                defaultMode = &devRec.GetModes().front();
+            }
+        }
+#endif
+
+        if (defaultMode == nullptr) {
+            // time to go down the rabit hole
+            int maxWidth = std::numeric_limits<int>::max();
+            if([device isLowPower]) {
```
Hoikas:
```suggestion
            if ([device isLowPower]) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,13 +74,60 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // integrated - not Apple Silicon, we know it's not very good
+                maxWidth = 1080;
+            }
+            else if([device recommendedMaxWorkingSetSize] < 4000000000) {
```
Hoikas:
```suggestion
            } else if ([device recommendedMaxWorkingSetSize] < 4000000000) {
```

That weird bracing is one of the symptoms I see in VS.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,13 +74,60 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
+                // if it has less than around 4 gigs of VRAM it might still be performance
+                // limited
+                maxWidth = 1400;
+            }
+            
+            for (auto& mode : devRec.GetModes()) {
+                if(mode.GetWidth() <= maxWidth) {
```
Hoikas:
```suggestion
                if (mode.GetWidth() <= maxWidth) {
```

2x

Sources/Plasma/PubUtilLib/plPipeline/pl3DPipeline.h
```diff
@@ -107,6 +107,19 @@ static const float kPerspLayerTrans  = 0.00002f;
...
 static const float kAvTexPoolShrinkThresh = 30.f; // seconds
 
+class plDisplayHelper
+{
+public:
+    virtual plDisplayMode DesktopDisplayMode() = 0;
+    virtual std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) const = 0;
+    
+    static plDisplayHelper* CurrentDisplayHelper() { return fCurrentDisplayHelper; }
+    void MakeCurrentDisplayHelper() { fCurrentDisplayHelper = this; }
```
Hoikas: Should probably be `GetInstance()`/`SetInstance()` to match with the rest of the codebase.

Hoikas (commented):
These are the code style issues that I see. clang-format is being weird for some reason, and it's causing Visual Studio to completely freak out. I'm tempted to say we should just toss .clang-format.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: Wow, this seems like an obvious one for Clang-format to get. Even though I think our relationship with Clang-format is ending, I am morbidly curious why this happened...

colincornaby: I was using git-clang-format before to try to target only changes. It seems to have missed this one.

If I was running it on every commit maybe it would have done better. But I gave it every commit in the branch and it missed this. Could be a weakness in the tool that it gets selective about which commits it applies. Especially if the commits maybe conflict with each other.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -108,7 +108,7 @@ @interface AppDelegate : NSWindowController <NSApplicationDelegate,
    @public
     plClientLoader gClient;
     dispatch_source_t _displaySource;
-    plMacDisplayHelper _displayHelper;
+    plMacDisplayHelper *_displayHelper;
```
Hoikas:
```suggestion
    plMacDisplayHelper* _displayHelper;
```

What is happening?

colincornaby: Clang-format didn't get this one either (possibly because I gave it multiple commits to work through.)

I had it do a full pass on this file - but it's giving me a lot of changes. I'll trim that change set down.

Hoikas (approved):
This LGTM. I'd like to make sure that @dpogue and @dgelessus are :+1: before we merge, though.

dpogue (approved):
This seems good to me now

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.h
```diff
@@ -0,0 +1,76 @@
...
+// Currently requires Metal to query attached GPU capabilities
+// Capability check will also work for GL - but will need something
+// different for older GPUs.
+#include <AppKit/AppKit.h>
+#include <QuartzCore/QuartzCore.h>
+
+#include "plPipeline/hsG3DDeviceSelector.h"
+#include "plPipeline/pl3DPipeline.h"
+
+class plMacDisplayHelper: public plDisplayHelper
```
dgelessus:
```suggestion
class plMacDisplayHelper : public plDisplayHelper
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+        if ([screenID unsignedIntValue] == display) {
+            nsScreen = screen;
+        }
+    }
+    SetCurrentScreen(nsScreen);
+}
+
+void plMacDisplayHelper::SetCurrentScreen(NSScreen* screen) const
+{
+    CGDirectDisplayID displayID = (CGDirectDisplayID)[screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue];
```
dgelessus, line 65: Named casts are generally better/safer:

```suggestion
    auto displayID = static_cast<CGDirectDisplayID>([screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
```

but I know the existing Mac code tends to use C-style casts, so it may be cleaner to stick with that style...

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+    fDesktopDisplayMode.Height = windowArea.size.height;
+    fDesktopDisplayMode.ColorDepth = 32;
+
+    // visibleFrame is in points - put into pixels
+    fDesktopDisplayMode.Width *= [screen backingScaleFactor];
+    fDesktopDisplayMode.Height *= [screen backingScaleFactor];
+
+    // Calculate the region actually available for full screen
+    NSRect currentResolution = [screen frame];
+#if defined(HAVE_BUILTIN_AVAILABLE)
```
dgelessus:
```suggestion
#ifdef HAVE_BUILTIN_AVAILABLE
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+        NSEdgeInsets currentSafeAreaInsets = [screen safeAreaInsets];
+        // Sigh... Origin doesn't matter but lets do it for inspectability
+        currentResolution.origin.x += currentSafeAreaInsets.left;
+        currentResolution.origin.y += currentSafeAreaInsets.top;
+        currentResolution.size.width -= currentSafeAreaInsets.left + currentSafeAreaInsets.right;
+        currentResolution.size.height -= currentSafeAreaInsets.top + currentSafeAreaInsets.bottom;
+    }
+#endif
+
+    float safeAspectRatio = currentResolution.size.width / currentResolution.size.height;
```
dgelessus: For consistency with the types of `NSSize.width` and `.height`:

```suggestion
    CGFloat safeAspectRatio = currentResolution.size.width / currentResolution.size.height;
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+
+    fDisplayModes.clear();
+
+    CFArrayRef displayModes = CGDisplayCopyAllDisplayModes(fCurrentDisplay, nullptr);
+    for (int i = 0; i < CFArrayGetCount(displayModes); i++) {
```
dgelessus:
```suggestion
    for (CFIndex i = 0; i < CFArrayGetCount(displayModes); i++) {
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+        // Apple also doesn't include some traditional resolutions on some of their hardware
+        // like 800x600, so don't expect those here. My Macbook Pro gives me back 1080p as the lowest.
+        // We could force that with kCGDisplayShowDuplicateLowResolutionModes but the aspect ratio
+        // filter would still drop some or all of those.
+        //
+        // Asked for better guidance here from Apple - FB13375033
+
+        CGDisplayModeRef mode = (CGDisplayModeRef)CFArrayGetValueAtIndex(displayModes, i);
+
+        float modeAspectRatio = float(CGDisplayModeGetWidth(mode)) / float(CGDisplayModeGetHeight(mode));
```
dgelessus:
```suggestion
        CGFloat modeAspectRatio = static_cast<CGFloat>(CGDisplayModeGetWidth(mode)) / static_cast<CGFloat>(CGDisplayModeGetHeight(mode));
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
```diff
@@ -0,0 +1,149 @@
...
+        if (modeAspectRatio < safeAspectRatio) {
+            continue;
+        }
+
+        // aspect ratio is good - add the mode to the list
+
+        // Plasma likes to handle modes from largest to smallest,
+        // CG likes to go from smallest to largest. Insert modes
+        // at the front.
+        fDisplayModes.emplace_back(plDisplayMode{ int(CGDisplayModeGetWidth(mode)), int(CGDisplayModeGetHeight(mode)), 32 });
```
dgelessus:
```suggestion
        fDisplayModes.emplace_back(plDisplayMode{ static_cast<int>(CGDisplayModeGetWidth(mode)), static_cast<int>(CGDisplayModeGetHeight(mode)), 32 });
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -228,6 +228,8 @@ class plMetalDevice
     
     void SetOutputLayer(CA::MetalLayer* layer) { fLayer = layer; }
     CA::MetalLayer* GetOutputLayer() const { return fLayer; };
+
```
dgelessus: Whitespace on empty line

```suggestion

```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -76,13 +74,59 @@ void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
...
         devRec.SetLayersAtOnce(8);
 
-        // Just make a fake mode so the device selector will let it through
-        hsG3DDeviceMode devMode;
-        devMode.SetWidth(hsG3DDeviceSelector::kDefaultWidth);
-        devMode.SetHeight(hsG3DDeviceSelector::kDefaultHeight);
-        devMode.SetColorDepth(hsG3DDeviceSelector::kDefaultDepth);
-        devRec.GetModes().emplace_back(devMode);
+        plDisplayHelper* displayHelper = plDisplayHelper::GetInstance();
+        hsG3DDeviceMode* defaultMode = nullptr;
```
dgelessus: `defaultMode` should go after the `for` loop, because it's only used later.

dgelessus: Also, I wonder if `defaultMode` should be an index right away, instead of a pointer that gets converted to an index a few lines later. I think all of the pointer arithmetic here is safe, but it would be nice to avoid it entirely...

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.cpp
```diff
@@ -95,12 +95,18 @@ void hsG3DDeviceMode::Clear()
...
+      fLayersAtOnce(0),
+      fMemoryBytes(0),
+      fG3DHALorHEL(hsG3DDeviceSelector::kHHTypeUnknown),
+      fZBiasRating(0),
+      fLODBiasRating(0),
+      fFogExpApproxStart(0.0),
+      fFogExp2ApproxStart(0.0),
+      fFogEndBias(0.0),
+      fMaxAnisotropicSamples(1),
+      fDefaultModeIndex(-1)
```
dgelessus, line 109: -1 is not a `size_t`, at least not really. If it's important to have a distinct "no default mode" value, you should change the type of `fDefaultModeIndex` to `hsSsize_t`.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -232,22 +232,25 @@ void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
+        CGLPixelFormatAttribute attribs[] = {
             kCGLPFAAccelerated,
             kCGLPFANoRecovery,
             kCGLPFADoubleBuffer,
-            kCGLPFAOpenGLProfile, (CGLPixelFormatAttribute) kCGLOGLPVersion_3_2_Core,
-            (CGLPixelFormatAttribute) 0
-        };
+            kCGLPFAOpenGLProfile, (CGLPixelFormatAttribute)kCGLOGLPVersion_3_2_Core,
+            kCGLPFADisplayMask, (CGLPixelFormatAttribute)CGDisplayIDToOpenGLDisplayMask(mainDisplay),
+            (CGLPixelFormatAttribute)0};
```
dgelessus:
```suggestion
            (CGLPixelFormatAttribute)0,
        };
```

dgelessus (commented):
Just a couple more style nits (sorry) and some type pedantry. The actual logic looks reasonable to me, as far as I can follow it (I'm not familiar with Mac graphics stuff).

Aside from the `hsSsize_t` suggestion, none of this is super critical.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
dgelessus: If we're already adjusting the style here: these casts could also be `static_cast`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
colincornaby: I'm pondering this.

I've also been thinking about spans, but we don't have those yet, and those seem really meant for multiple items.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm
colincornaby: I'll go ahead and static_cast it, but agree generally C casts are used in Obj-C.

Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
```diff
@@ -241,6 +242,9 @@ class hsG3DDeviceRecord
 
     std::vector<hsG3DDeviceMode>& GetModes() { return fModes; }
 
+    const hsG3DDeviceMode* GetDefaultMode() const { return fDefaultModeIndex == -1 ? nullptr : &fModes[fDefaultModeIndex]; }
+    void                   SetDefaultModeIndex(size_t defaultModeIndex) { fDefaultModeIndex = defaultModeIndex; }
```
dgelessus:
```suggestion
    void                   SetDefaultModeIndex(hsSsize_t defaultModeIndex) { fDefaultModeIndex = defaultModeIndex; }
```

PR #1733, opened 2025-07-16
Lambda-ify the high-level vault API callbacks

dgelessus:
Follow-up to #1703. As expected, this is another big diff because of lots of helper structs and functions being inlined (especially commit f5759534475496170b45ae9a83c1f9ab0f04ee68). But now you can read the async vault code internals without your brain catching fire immediately.


Hoikas:
> But now you can read the async vault code internals without your brain catching fire immediately.

This made me chuckle a few days ago because the diff made my brain catch fire. 🤯

Hoikas (approved):

PR #1739, opened 2025-07-21
Add clickable links to journal books.

Hoikas:
This adds clickable links to journals. Links already somewhat existed in journals in that you could have an image or movie with a `link=[event ID]` attribute on the tag that fired a notification to the host Python script. This pull request extends that in a number of ways, rather creatively in some respects by:
- adding named anchors to journals
- adding the ability to link to named anchors using text, images, or movies
- adding the ability to link to URLs using text, images, or movies

Text based links can be created using `<a link=[event ID]>`. So, if you want some text to fire event 1 in the host script, use `<a link=1>`. EsHTML does not use closing tags, so to stop making text a link, use `<a link=clear>`. This is mostly consistent with `` and `<movie link=1>`. By default, links will be colored solid blue, but this can be controlled using the `color` attribute, which accepts a hex color.

To add an anchor named `foo` into a journal, use `<a name=foo>`. You can then link to that anchor by using `<a href=#foo>`  This also works with images and movies, eg `` and `<movie href=#foo>`. Alternatively, any images or movies added to a journal while a link is "active" will inherit that link unless they define their own links.

URL support was a bit more tricky. The existing KI URL support uses the Python `webbrowser` module. This module does a good job of opening URLs in browsers in a cross platform way. Unfortunately, it weighs in at approximately 24 KB. I have no interest in reimplementing that in C++, so that means Python needs to open URLs. Even more unfortunately, the book notification message only sends out an integer event ID. So, I decided to have the journal book Python script inspect all `href` attributes for URLs. If one is found, it translates that to an integer ID, and dispatches any events that come back from the book.

As part of this pull request, I identified some already existing issues with how journals work. Specifically, if using center or right alignment, any attempt to continue an already in progress paragraph will cause the paragraph text to overlap. A trivial way to do this would be to change the font color or style without starting a new paragraph, eg:
```
<p align=center>This is an example <font color=FF00FF0000>of the brokenness!
```
This means that using links in a center or right aligned paragraph may yield disappointing results. The entire paragraph will need to be a link in those cases because any uses of the `<a>` tag cause a new continuation paragraph to start internally.

For reference, [check out this video](https://www.youtube.com/watch?v=evyGFLz3GmE) or try the [sample Age](https://github.com/user-attachments/files/21338748/KormanGUI.zip) attached to this pull request.

### DIFF
--- a/Scripts/Python/xJournalBookGUIPopup.py
+++ b/Scripts/Python/xJournalBookGUIPopup.py
@@ -87,7 +87,7 @@ LocalAvatar = None
 # a url by doing <a href="https://foo.com">, and this regex will
 # convert that to an integer id. It will do the same for any tag, but
 # only <img> and <movie> are useful.
-_URL_REGEX = re.compile(r"(?P<href>href\s*=\s*(?P<quote>[\"\']?)\s*?(?P<url>https?:\/\/[^\s>]+)\s*?(?P=quote))(?=[^<]*>)")
+_URL_REGEX = re.compile(r"(?P<href>href\s*=\s*(?P<quote>[\"\']?)\s*?(?P<url>https?:\/\/[^\s>]+)\s*?(?P=quote))(?=[^<]*>)", re.IGNORECASE)
 
 class xJournalBookGUIPopup(ptModifier):
     "The Journal Book GUI Popup python code"
--- a/Scripts/Python/xJournalBookGUIPopup.py
+++ b/Scripts/Python/xJournalBookGUIPopup.py
@@ -61,6 +61,9 @@ from Plasma import *
 from PlasmaTypes import *
 from PlasmaKITypes import *
 
+import re
+import webbrowser
+
 import xJournalBookDefs
 
 
@@ -78,6 +81,12 @@ GUIType             = ptAttribString(14,"Book GUI Type",default="bkBook")
 
 # globals
 LocalAvatar = None
+
+# pfJournalBook's EsHTML will only allow links with integer events.
+# This isn't very friendly for age creators, so we'll let them specify
+# a url by doing <link href="https://foo.com">, and this regex will
+# convert that to an integer id.
+_URL_REGEX = re.compile(r"<link\s+href\s*=\s*(?P<quote>[\'\"]?)\s*(?P<url>https?:\/\/[\S]+|www\.[^\s.]+\.[^\s]+)\s*(?P=quote)\s*>")
 
 class xJournalBookGUIPopup(ptModifier):
     "The Journal Book GUI Popup python code"
@@ -123,11 +132,20 @@ class xJournalBookGUIPopup(ptModifier):
                 # is it from the OpenBook? (we only have one book to worry about)
                 if event[0] == PtEventType.kBook:
                     PtDebugPrint("xJournalBookGUIPopup: BookNotify  event=%d, id=%d" % (event[1],event[2]),level=kDebugDumpLevel)
-                    if event[1] == PtBookEventTypes.kNotifyShow:
+                    if event[1] == PtBookEventTypes.kNotifyImageLink:
+                        PtDebugPrint("xJournalBookGUIPopup:Book: NotifyImageLink",level=kDebugDumpLevel)
+                        try:
+                            url = self.links[event[2]]
+                        except IndexError:
+                            pass
+                        else:
+                            PtDebugPrint(f"xJournalBookGUIPopup: NotifyImageLink: opening {url=} {event[2]=}", level=kWarningLevel)
+                            webbrowser.open_new_tab(url)
+                    elif event[1] == PtBookEventTypes.kNotifyShow:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyShow",level=kDebugDumpLevel)
                         # disable the KI
                         PtSendKIMessage(kDisableKIandBB,0)
-                    if event[1] == PtBookEventTypes.kNotifyHide:
+                    elif event[1] == PtBookEventTypes.kNotifyHide:
                         PtDebugPrint("xJournalBookGUIPopup:Book: NotifyHide",level=kDebugDumpLevel)
                         # re-enable KI
                         PtSendKIMessage(kEnableKIandBB,0)
@@ -194,7 +212,7 @@ class xJournalBookGUIPopup(ptModifier):
         PtSendKIMessage(kDisableKIandBB, 0)
 
         # now build the book
-        self.JournalBook = ptBook(journalContents, self.key)
+        self.JournalBook = ptBook(self.IPreprocessJournalContents(journalContents), self.key)
         self.JournalBook.setSize(BookWidth.value, BookHeight.value)
         self.JournalBook.setGUI(GUIType.value)
 
@@ -204,6 +222,17 @@ class xJournalBookGUIPopup(ptModifier):
         else:
             self.JournalBook.show(StartOpen.value)
 
+    def IPreprocessJournalContents(self, journalContents: str):
+        self.links = []
+        def replace_url(match: re.Match) -> str:
+            url = match.group("url")
+            event = len(self.links)
+
+            PtDebugPrint(f"xJournalBookGUIPopup.IPreprocessJournalContents(): Found {url=} {event=}")
+            self.links.append(url)
+            return f"<link event={event}>"
+
+        return _URL_REGEX.sub(replace_url, journalContents)
 
     def IsThereACover(self, bookHtml):
         # search the bookhtml string looking for a cover
--- a/Scripts/Python/xJournalBookGUIPopup.py
+++ b/Scripts/Python/xJournalBookGUIPopup.py
@@ -85,7 +85,8 @@ LocalAvatar = None
 # pfJournalBook's EsHTML will only allow links with integer events.
 # This isn't very friendly for age creators, so we'll let them specify
 # a url by doing <a href="https://foo.com">, and this regex will
-# convert that to an integer id.
+# convert that to an integer id. It will do the same for any tag, but
+# only <img> and <movie> are useful.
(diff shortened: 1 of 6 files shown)


Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -123,11 +134,25 @@ def OnNotify(self, state, id, events):
...
+                            url = self.links[event[2]]
+                        except IndexError:
+                            pass
+                        else:
+                            def open_url(result):
+                                if result == PtConfirmationResult.Yes:
+                                    webbrowser.open_new_tab(url)
+
+                            hostname = _HOST_REGEX.sub(R"\1", url)
+                            PtDebugPrint(f"xJournalBookGUIPopup: NotifyImageLink: Promping user to open {hostname=} {url=} {event[2]=}", level=kWarningLevel)
```
Deledrius:
```suggestion
                            PtDebugPrint(f"xJournalBookGUIPopup: NotifyImageLink: Prompting user to open {hostname=} {url=} {event[2]=}", level=kWarningLevel)
```

Scripts/Python/xJournalBookGUIPopup.py
```diff
@@ -79,6 +82,14 @@
 # globals
 LocalAvatar = None
 
+# pfJournalBook's EsHTML will only allow links with integer events.
+# This isn't very friendly for age creators, so we'll let them specify
+# a url by doing <a href="https://foo.com">, and this regex will
+# convert that to an integer id. It will do the same for any tag, but
+# only <img> and <movie> are useful.
+_URL_REGEX = re.compile(r"(?P<href>href\s*=\s*(?P<quote>[\"\']?)\s*?(?P<url>https?:\/\/[^\s>]+)\s*?(?P=quote))(?=[^<]*>)", re.IGNORECASE)
+_HOST_REGEX = re.compile(R"(?:https?:\/\/)?(?:www\.)?([^\/?#]+).*", re.IGNORECASE)
```
dgelessus, line 91: Would be really cool if we could make `urllib.parse` usable inside Plasma at some point...

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -103,16 +109,32 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+            kImage,
+            kPageBreak,
+            kFontChange,
+            kMargin,
+            kCover, // Just a placeholder, never actually used after compile time
+            kBook,  // another placeholder
+            kDecal,
+            kMovie,
+            kEditable, // placeholder, ver 3.0
+            kTextLink
```
dgelessus: To reduce future diff churn:

```suggestion
            kTextLink,
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -154,22 +176,8 @@ class pfEsHTMLChunk
...
-            kPageBreak,
-            kFontChange,
-            kMargin,
-            kCover,         // Just a placeholder, never actually used after compile time
-            kBook,          // another placeholder
-            kDecal,
-            kMovie,
-            kEditable       // placeholder, ver 3.0
+            kTranslucent= 0x00000100,   // is the image translucent? if so, use fCurrOpacity
+            kValidEvent = 0x00000200
```
dgelessus:
```suggestion
            kValidEvent = 0x00000200,
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -103,16 +109,32 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 class pfEsHTMLChunk
 {
     public:
+        struct TextLink_Type {};
+        static constexpr TextLink_Type TextLink{};
+
+        enum Types : uint8_t
```
dgelessus:
```suggestion
        enum Type : uint8_t
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -103,16 +109,32 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+            kEditable, // placeholder, ver 3.0
+            kTextLink
+        };
 
         ST::string fText; // Paragraph text, or face name
         plKey   fImageKey;  // Key of image
         uint8_t   fFontSize;
         uint32_t  fFlags;
-        uint8_t   fType;
+        Types   fType;
```
dgelessus:
```suggestion
        Type    fType;
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -1550,7 +1637,7 @@ hsSsize_t pfJournalBook::IFindCurrVisibleLink(bool rightNotLeft, bool hoverNotUp
     // Search through the list of visible hotspots
     for (size_t i = 0; i < fVisibleLinks.size(); i++)
     {
-        if( fVisibleLinks[ i ]->fLinkRect.Contains( (int16_t)pt.fX, (int16_t)pt.fY ) )
+        if( fVisibleLinks[ i ].Contains( (int16_t)pt.fX, (int16_t)pt.fY ) )
```
dgelessus:
```suggestion
        if (fVisibleLinks[i].Contains((int16_t)pt.fX, (int16_t)pt.fY))
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -1672,6 +1876,7 @@ bool    pfJournalBook::ICompileSource(const ST::string& source, const plLocation
     IFreeSource();
 
     pfEsHTMLChunk *chunk, *lastParChunk = new pfEsHTMLChunk(ST::string());
+    pfEsHTMLChunk *lastLinkChunk = nullptr;
```
dgelessus, line 1878:
```suggestion
    pfEsHTMLChunk* lastLinkChunk = nullptr;
```

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -2578,7 +2911,7 @@ void    pfJournalBook::IMoveMovies( hsGMaterial *source, hsGMaterial *dest )
 
 void    pfJournalBook::IDrawMipmap( pfEsHTMLChunk *chunk, uint16_t x, uint16_t y, plMipmap *mip, plDynamicTextMap *dtMap, uint32_t whichDTMap, bool dontRender )
 {
-    plMipmap *copy = new plMipmap();
+    auto copy = std::make_unique<plMipmap>();
```
dgelessus: `hsRef` would be better here, because `plMipmap` extends `hsRefCnt`. (I know the previous code didn't care and `delete`d it directly, but that seems like a risky thing that we should stop doing.)

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
```diff
@@ -3100,6 +3437,25 @@ void    pfJournalBook::IFindFontProps( uint32_t chunkIdx, ST::string &face, uint
...
 }
 
+//// IFindTextLink ///////////////////////////////////////////////////////////
+// Starting at the given chunk, works backward to determine the current text link ID
+pfEsHTMLChunk* pfJournalBook::IFindTextLink(uint32_t chunkIdx) const
+{
+    if (fHTMLSource.empty())
+        return nullptr;
+
+    for (size_t i = chunkIdx; i > 0; --i) {
```
dgelessus: `i > 0` as the loop condition means that the loop body will never run for index 0 - is that intentional?

If the condition is supposed to be `i >= 0`, you'd also have to change the type of `i` to `hsSsize_t` (like the other loops above) to prevent an infinite loop.

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.h
```diff
@@ -513,8 +533,9 @@ class pfJournalBook : public hsKeyedObject
         // Some animation keys we use
         plKey   fPageTurnAnimKey;
 
-        // Current list of linkable image chunks we have visible on the screen, for quick hit testing
-        std::vector<pfEsHTMLChunk *> fVisibleLinks;
+        // Current list of visible link hotspots for quick hit testing.
+        // Can be images or lines of text.
+        std::vector<class pfJournalVisibleLink> fVisibleLinks;
```
dgelessus: Inline type declarations are eeeehhhhh if you ask me... could we declare `pfJournalVisibleLink` together with all the other classes, at the top of the file?

Scripts/Python/xJournalBookGUIPopup.py
Hoikas: That would be some good future work. It looks like pulling in just `urllib.parse` would be safe. It's `urllib.request` that we especially don't want due to the dependency on `socket`. `urllib.parse` doesn't seem to depend on any other `urllib` submodules.

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
Hoikas: That's what our code style says, strictly speaking, but we're also supposed to match the style of the immediately surrounding code. It's a tough call sometimes :(

Hoikas:
Would be nice to get a :+1: or 👎 here.

Sources/Plasma/FeatureLib/pfJournalBook/pfJournalBook.cpp
Deledrius: I run into that conflict a lot.

PR #1740, opened 2025-07-26
Clean up some unused or redundant debug macros

dgelessus:
See commit messages for explanations.

Also replaces many (but not all) uses of the global `DebugMsg` function with other logging functions. The output of `DebugMsg` only goes to stderr (and the Windows debugger output) and so cannot be seen by the average tester. So it seems best to use `DebugMsg` only for strictly developer-oriented things.


Hoikas (commented):
I wonder if it would make more sense for the coop coordinator logs to go into the avatar log? Otherwise, LGTM

dgelessus:
Seems reasonable, considering that the coop stuff is under plAvatar. I didn't think much about what the new destination should be 😄

Hoikas (approved):

PR #1742, opened 2025-07-28
Rename global `DebugMsg` function (and others) to avoid name collisions/confusion

dgelessus:
Renames the following global functions in HeadSpin.h that were missing the usual `hs` prefix, and also adjusts some names for clarity (bikeshedding welcome 😉):

| previously | with this PR |
|---|---|
| `ErrorEnableGui` | `hsDebugEnableGuiAsserts` |
| `ErrorAssert` | `hsDebugAssertionFailed` |
| `DebugIsDebuggerPresent` | `hsDebugIsDebuggerPresent` |
| `DebugBreakIfDebuggerPresent` | `hsDebugBreakIfDebuggerPresent` |
| `DebugBreakAlways` | `hsDebugBreakAlways` |
| `DebugMsg` | `hsDebugPrintToTerminal` |

`DebugMsg` is the main motivation for this change, because there are multiple other functions/methods named `DebugMsg` elsewhere in the codebase, e. g. in `plLoggable`, `plSDLParser`, and `plClient`. This made it very difficult to tell if a call to `DebugMsg` went to the global function or one of the local variants.

The other renames are to avoid possible confusion about whether a function is from Plasma or from Win32. For example, the Win32 functions [`DebugBreak`](https://learn.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-debugbreak) and [`IsDebuggerPresent`](https://learn.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-isdebuggerpresent) look dangerously similar to some of the previous HeadSpin.h functions.

### DIFF
--- a/Sources/MaxPlugin/MaxExport/plExportErrorMsg.cpp
+++ b/Sources/MaxPlugin/MaxExport/plExportErrorMsg.cpp
@@ -129,7 +129,7 @@ void plExportErrorMsg::Quit()
 void plExportErrorMsg::IDebugThrow()
 {
     try {
-        DebugBreakIfDebuggerPresent();
+        hsDebugBreakIfDebuggerPresent();
     }
     catch(...)
     {
--- a/Sources/Plasma/Apps/plClient/plClient.cpp
+++ b/Sources/Plasma/Apps/plClient/plClient.cpp
@@ -1569,7 +1569,7 @@ bool plClient::MainLoop()
 #endif
 
 #ifdef PLASMA_EXTERNAL_RELEASE
-    if (DebugIsDebuggerPresent())
+    if (hsDebugIsDebuggerPresent())
     {
         NetCliAuthLogClientDebuggerConnect();
         SetDone(true);
--- a/Sources/Plasma/CoreLib/HeadSpin.cpp
+++ b/Sources/Plasma/CoreLib/HeadSpin.cpp
@@ -140,8 +140,7 @@ void ErrorAssert(int line, const char* file, const char* fmt, ...)
     } else
 #endif // _MSC_VER
     {
-        DebugMsg("-------\nASSERTION FAILED:\nFile: %s   Line: %i\nMessage: %s\n-------",
-                 file, line, msg);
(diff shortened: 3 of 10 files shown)


Hoikas (approved):
This all makes sense to me and seems like a clear improvement. I was going to suggest, instead of `hsDebugPrintToTerminal()` to make the name `hsDebugPrint()`, but the doxygen comment has a good rationale for having such a verbose name. I'd like to leave this open for awhile to give other stakeholders an opportunity to chime in, though.

Deledrius (approved):

dgelessus:
Yeah, I am not fully happy with `hsDebugPrintToTerminal` either. It's kind of the only thing I could think of after deciding against all the other obvious options:

* `hsDebugMsg` would have been the obvious name, but that sounds way too similar to the existing different function named `hsDebugMessage`.
* `hsDebugPrint` might have crossed my mind too, but that's still easy to confuse with `hsDebugMessage`. Plus, as explained, I want to discourage using this function for general logging.
* I considered `hs[Debug]PrintToConsole`, but one could mis-interpret that as "print to the in-game Plasma console".

`hsDebugPrintToStderr` could also be a good option. I went with "terminal" to hint at how it prints not *just* to stderr but also to the Windows debugger output, but that doesn't communicate it very well either tbh.

Aside from the points above, I don't have super strong opinions on the name. So if we agree that `hsDebugPrintToStderr` is possibly clearer than `hsDebugPrintToTerminal`, I could change it.

dgelessus:
I was considering something similar as well (`hsPrintToDebugger`), but then decided against it, because it sounds like the output goes *only* to the debugger, which isn't the case. (Especially on non-Windows, this function doesn't interact with a debugger in any special way.) It would communicate the "for developer output only" aspect well though.

Hazado:
hsPrintToConsole?

dpogue:
> hsPrintToConsole?

It doesn't print to the Plasma console though (or to any of the log files you can bring up with ctrl+L), nor does it print to the macOS Console.app (which is a system log viewer with a terribly confusing name).

`hsPrintToStderrOrDebugOutputWindow()`? That would be one way to discourage its continued use 😝

dpogue (approved):

PR #1744, opened 2025-08-12
[Hacky tool] hsG3DDeviceDumper

dpogue:
Prints out the available hsG3DDeviceRecord records along with their hsG3DDeviceMode resolutions for debugging purposes.

Incredibly hacky code 😛

### DIFF
--- a/Sources/Plasma/CoreLib/hsOptionalCall.h
+++ b/Sources/Plasma/CoreLib/hsOptionalCall.h
@@ -207,6 +207,44 @@ public:
     }
 };
 
+#if defined(HS_BUILD_FOR_WIN32)
+template<class, class...>
+class plOptionalWinCall;
+
+template<class _ReturnT, class... _ArgsT>
+class plOptionalWinCall<_ReturnT(_ArgsT...)> : public hsOptionalCallBase
+{
+    using _FuncPtrT = _ReturnT(WINAPI *)(_ArgsT...);
+    using _ResultT = typename std::conditional<std::is_void<_ReturnT>::value, std::nullptr_t, _ReturnT>::type;
+
+public:
+    template<class _StrT>
+    plOptionalWinCall(const _StrT* lib, const char* func)
+        : hsOptionalCallBase(lib, func)
+    { }
+
+    std::optional<_ResultT> operator()(_ArgsT... args) const
+    {
+        _FuncPtrT proc = Get();
+        hsAssert(proc, "Trying to invoke undefined function");
+
+        if (proc) {
+            if constexpr (std::is_void<_ReturnT>::value) {
+                return proc(std::forward<_ArgsT>(args)...), nullptr;
+            } else {
+                return (_ResultT)proc(std::forward<_ArgsT>(args)...);
+            }
+        }
+        return std::nullopt;
+    }
+
+    _FuncPtrT Get() const
+    {
+        return reinterpret_cast<_FuncPtrT>(fValue);
+    }
+};
+#endif
+
 
 #if defined(HS_BUILD_FOR_WIN32)
 #   define HS_OPTIONAL_LIB_SUFFIX ""
--- a/Sources/Plasma/CoreLib/hsThread_Win.cpp
+++ b/Sources/Plasma/CoreLib/hsThread_Win.cpp
@@ -40,7 +40,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
-#include "hsWindows.h"
+#include "hsOptionalCall.h"
 #include <process.h>
 
 #include "hsThread.h"
--- a/Sources/Plasma/CoreLib/hsWindows.h
+++ b/Sources/Plasma/CoreLib/hsWindows.h
@@ -142,55 +142,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
             ST::format_type(format, output, hr.fResult);
         }
     }
-
-    template<class, class...>
-    class plOptionalWinCall;
-
-    template<
-        class _ReturnT,
-        class... _ArgsT
-    >
-    class plOptionalWinCall<_ReturnT(_ArgsT...)>
-    {
-        using _FuncPtrT = _ReturnT(WINAPI*)(_ArgsT...);
-
-        HMODULE fHModule;
-        _FuncPtrT fProc;
-
-    public:
-        plOptionalWinCall(const wchar_t* mod, const char* func)
-            : fHModule(), fProc()
-        {
-            fHModule = LoadLibraryW(mod);
-            if (fHModule) {
-                fProc = (_FuncPtrT)GetProcAddress(fHModule, func);
-                if (!fProc) {
-                    FreeLibrary(fHModule);
-                    fHModule = nullptr;
-                }
-            }
-        }
-
-        ~plOptionalWinCall()
-        {
-            if (fHModule)
-                FreeLibrary(fHModule);
-        }
-
-        std::optional<_ReturnT> operator()(_ArgsT... args) const
-        {
-            if (fProc)
-                return (_ReturnT)fProc(std::forward<_ArgsT>(args)...);
-            return std::nullopt;
-        }
-
-        operator bool() const
-        {
-            return fProc != nullptr;
-        }
-
-        _FuncPtrT Get() const { return fProc; }
-    };
 #endif // HS_BUILD_FOR_WIN32
 
 #endif // _hsWindows_inc_
--- a/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
+++ b/Sources/Plasma/PubUtilLib/plWinDpi/plWinDpi.h
@@ -46,7 +46,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <optional>
 #include <variant>
 
-#include "hsWindows.h"
+#include "hsOptionalCall.h"
 
 #include "plMessage/plDisplayScaleChangedMsg.h"
 


Sources/Plasma/PubUtilLib/plPipeline/hsG3DDeviceSelector.h
```diff
@@ -241,6 +241,7 @@ class hsG3DDeviceRecord
     bool    IsInvalid() const { return 0 != ( fFlags & kInvalid ); }
 
     std::vector<hsG3DDeviceMode>& GetModes() { return fModes; }
+    const std::vector<hsG3DDeviceMode>& GetConstModes() const { return fModes; }
```
Hoikas:
```suggestion
    const std::vector<hsG3DDeviceMode>& GetModes() const { return fModes; }
```

Sources/Tools/hsG3DDeviceDumper/main.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+    devSel.Enumerate(nullptr);
+    devSel.RemoveUnusableDevModes(true);
+
+    std::cout << devSel.GetDeviceRecords().size() << " found." << std::endl << std::endl;
+
+    for (const hsG3DDeviceRecord& rec : devSel.GetDeviceRecords()) {
+        std::cout << rec.GetDriverName() << " (" << rec.GetDriverVersion() << ") - " << rec.GetDeviceDesc() << std::endl;
+
+        for (const hsG3DDeviceMode& mode : rec.GetConstModes()) {
```
Hoikas:
```suggestion
        for (const hsG3DDeviceMode& mode : rec.GetModes()) {
```

Sources/Tools/hsG3DDeviceDumper/main.cpp
```diff
@@ -0,0 +1,90 @@
...
+#ifdef HS_BUILD_FOR_MACOS
+    plDisplayHelper::SetInstance(new plMacDisplayHelper());
+#endif
+
+    std::cout << "Checking device records... ";
+
+    devSel.Enumerate(nullptr);
+    devSel.RemoveUnusableDevModes(true);
+
+    std::cout << devSel.GetDeviceRecords().size() << " found." << std::endl << std::endl;
```
Hoikas: IIRC string theory has a nice print-format-to-stdout facility.

dpogue:
I'm still unclear whether the tool itself has any long-term value and should be merged or not, but this seems like a good starting point for other Linux and OpenGL work

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
```diff
@@ -49,10 +49,9 @@
 
 #include "plMetalPipeline.h"
 
-void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records)
+void plMetalEnumerate::Enumerate(std::vector<hsG3DDeviceRecord>& records, hsDisplayHndl mainDisplay)
```
colincornaby: Is this actually mainDisplay, or is this meant to be any display? If it's not going to always be the mainDisplay we should rename that.

dpogue: Renamed to `displayHndl` because it's not specifically the main display (although currently that's what it always ends up being)

Sources/Tools/hsG3DDeviceDumper/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+)
+
+plasma_executable(hsG3DDeviceDumper TOOL
+    FOLDER Tools
+    SOURCES
+        ${hsG3DDeviceDumper_HEADERS} ${hsG3DDeviceDumper_SOURCES}
+
+        # Hack hack hack
+        $<$<PLATFORM_ID:Darwin>:../../Plasma/Apps/plClient/Mac-Cocoa/plMacDisplayHelper.mm>
+        $<$<PLATFORM_ID:Windows>:../../Plasma/Apps/plClient/win32/plWinDisplayHelper.cpp>
```
Hoikas: I wonder if these should be split out into a `pfDisplayHelper` project. That would also mean the wayland and x11 code don't need to live in this tool and we can remove these hacky references.

Sources/Tools/hsG3DDeviceDumper/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+)
+
+target_include_directories(
+    hsG3DDeviceDumper
+    PRIVATE
+    $<$<AND:$<BOOL:${USE_WAYLAND}>,$<TARGET_EXISTS:Wayland::Client>>:$<TARGET_PROPERTY:Wayland::Client,INCLUDE_DIRECTORIES>>
```
Hoikas: We should make `USE_WAYLAND` and `USE_X11` be a `cmake_dependent_option` on whether or not the all of the required components were found. That will allow you to simplify these generator expressions to just the boolean portion and omit the `AND`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalEnumerate.mm
Hoikas: This still seems to be `mainDisplay`?

Sources/Plasma/Apps/plClient/win32/plWinDisplayHelper.cpp
```diff
@@ -0,0 +1,90 @@
...
+void plWinDisplayHelper::SetCurrentScreen(hsDisplayHndl display) const
+{
+    if (fCurrentDisplay == display)
+        return;
+
+    fCurrentDisplay = display;
+
+    fDisplayModes.clear();
+
+    DEVMODE dm = {};
```
Hoikas:
```suggestion
    DEVMODE dm{};
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/plMacDisplayHelper.h
```diff
@@ -43,25 +43,31 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#endif
 
 #include "plPipeline/hsG3DDeviceSelector.h"
 #include "plPipeline/pl3DPipeline.h"
 
+#ifdef __OBJC__
+@class NSScreen;
+#else
+class NSScreen;
+#endif
```
colincornaby, line 59: I think this should be ok for now - but this situation will get worse. iOS doesn't support CGDirectDisplayID - it only has Cocoa types. I'll continue thinking about it.

dpogue: Maybe a `plIOSDisplayHelper` would make sense for that case if it's different enough?

Sources/Plasma/FeatureLib/pfDisplayHelpers/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+        "$<$<AND:$<PLATFORM_ID:Darwin>,$<VERSION_LESS:${CMAKE_OSX_DEPLOYMENT_TARGET},10.9>>:-framework ApplicationServices>"
+        "$<$<AND:$<PLATFORM_ID:Darwin>,$<VERSION_GREATER_EQUAL:${CMAKE_OSX_DEPLOYMENT_TARGET},10.9>>:-framework CoreGraphics>"
+)
+
+# We're not linking against X11 or Wayland because we want to load them
+# dynamically at runtime, but we do need their include paths to be added here
+# so the header definitions can be loaded
+target_include_directories(
+    pfDisplayHelpers
+    INTERFACE
```
Hoikas:
```suggestion
    PUBLIC
```

It looks like cpp files in this project are consuming the Plasma headers that depend on these headers, so we should use `PUBLIC` instead of `INTERFACE`.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "plWaylandFunctions.h"
+
+template <typename... Args> void ignore(Args...) {}
+
+void plWaylandDisplayHelper::output_handle_mode(void* data, wl_output* wl_output, uint32_t flags, int32_t width, int32_t height, int32_t refresh)
+{
+    plWaylandDisplayHelper* self = reinterpret_cast<plWaylandDisplayHelper*>(data);
+
+    if (flags & WL_OUTPUT_MODE_CURRENT) {
+        self->fDisplayModes.emplace_back(plDisplayMode {
```
Hoikas:
```suggestion
        self->fDisplayModes.emplace_back(plDisplayMode{
```

Also, does `plDisplayMode` have a constructor? If so, we could simply construct in place instead of construct and move (meaning, pass the constructor arguments as the arguments to `emplace_back`).

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,152 @@
...
+            &ignore,
+            &ignore
+        };
+
+        self->fCurrentOutput = reinterpret_cast<wl_output*>(wl_registry_bind(registry, id, &__wl_output_interface, std::min<uint32_t>((*__wl_output_interface).version, version)));
+        wl_output_add_listener(self->fCurrentOutput, &output_listener, data);
+    }
+}
+
+plWaylandDisplayHelper::plWaylandDisplayHelper() : fCurrentDisplay(), fCurrentRegistry(), fCurrentOutput()
```
Hoikas: We usually like to see these initializers on a new line.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWinDisplayHelper.cpp
```diff
@@ -0,0 +1,89 @@
...
+    fDisplayModes.clear();
+
+    DEVMODE dm{};
+    dm.dmSize = sizeof(DEVMODE);
+
+    for (int i = 0;; i++) {
+        if (!EnumDisplaySettings(nullptr, i, &dm))
+            break;
+
+        fDisplayModes.emplace_back(plDisplayMode {
```
Hoikas:
```suggestion
        fDisplayModes.emplace_back(plDisplayMode{
```

Same comment about constructing in place

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWinDisplayHelper.cpp
```diff
@@ -0,0 +1,89 @@
...
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plWinDisplayHelper.h"
+#include "hsWindows.h"
+
+plWinDisplayHelper::plWinDisplayHelper() : fCurrentDisplay(INVALID_HANDLE_VALUE)
```
Hoikas:
```suggestion
plWinDisplayHelper::plWinDisplayHelper()
     : fCurrentDisplay(INVALID_HANDLE_VALUE)
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.cpp
```diff
@@ -0,0 +1,123 @@
...
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plX11DisplayHelper.h"
+
+// X11 includes MUST be after everything else to avoid contamination!
+#include "plX11Functions.h"
+
+plX11DisplayHelper::plX11DisplayHelper() : fCurrentDisplay()
```
Hoikas:
```suggestion
plX11DisplayHelper::plX11DisplayHelper()
    : fCurrentDisplay()
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.cpp
```diff
@@ -0,0 +1,123 @@
...
+        int num_sizes;
+        XRRScreenSize* xrrs;
+
+        xrrs = *__XRRSizes(fCurrentDisplay, 0, &num_sizes);
+        for (int i = 0; i < num_sizes; i++) {
+            // Don't include unreasonably small sizes
+            if (xrrs[i].width < 800 || xrrs[i].height < 600)
+                continue;
+
+            fDisplayModes.emplace_back(plDisplayMode {
```
Hoikas:
```suggestion
            fDisplayModes.emplace_back(plDisplayMode{
```

Constructing in place, if ppossible.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.cpp
```diff
@@ -0,0 +1,123 @@
...
+                xrrs[i].width,
+                xrrs[i].height,
+                DefaultDepth(fCurrentDisplay, screen)
+            });
+        }
+    }
+    else
+#endif
+    {
+        fDisplayModes.emplace_back(plDisplayMode {
```
Hoikas:
```suggestion
        fDisplayModes.emplace_back(plDisplayMode{
```

Constructing in place, if ppossible.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -47,11 +47,14 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plGLPipeline.h"
 
-bool fillDeviceRecord(hsG3DDeviceRecord& devRec)
+bool fillDeviceRecord(hsG3DDeviceRecord& devRec, const ST::string& driverName, hsDisplayHndl display)
```
Hoikas: While we're here, this should probably be `static`

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -68,43 +71,73 @@ bool fillDeviceRecord(hsG3DDeviceRecord& devRec)
...
-    devMode.SetColorDepth(hsG3DDeviceSelector::kDefaultDepth);
-    devRec.GetModes().emplace_back(devMode);
+    plDisplayHelper* displayHelper = plDisplayHelper::GetInstance();
+    if (displayHelper) {
+        for (const auto& mode : displayHelper->GetSupportedDisplayModes(display)) {
+            hsG3DDeviceMode devMode;
+            devMode.SetWidth(mode.Width);
+            devMode.SetHeight(mode.Height);
+            devMode.SetColorDepth(mode.ColorDepth);
+            devRec.GetModes().emplace_back(std::move(devMode));
```
Hoikas: Would be nice if we constructed these in place.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -68,43 +71,73 @@ bool fillDeviceRecord(hsG3DDeviceRecord& devRec)
...
+        }
+        devRec.SetDefaultModeIndex(0);
+    } else {
+        // Just make a fake mode so the device selector will let it through
+        hsG3DDeviceMode devMode;
+        devMode.SetWidth(hsG3DDeviceSelector::kDefaultWidth);
+        devMode.SetHeight(hsG3DDeviceSelector::kDefaultHeight);
+        devMode.SetColorDepth(hsG3DDeviceSelector::kDefaultDepth);
+        devRec.GetModes().emplace_back(devMode);
```
Hoikas: Would be nice if we constructed these in place.

Sources/Plasma/PubUtilLib/plPipeline/pl3DPipeline.h
```diff
@@ -110,8 +110,11 @@ static const float kAvTexPoolShrinkThresh = 30.f; // seconds
 class plDisplayHelper
 {
 public:
+    virtual ~plDisplayHelper() {}
+
     virtual plDisplayMode              DesktopDisplayMode() = 0;
```
Hoikas: I was noticing this in the derived classes you added - this will make a copy of the display mode when called. I wonder if we should - potentially in another PR - change this to return a const reference to avoid that.

Sources/Tools/hsG3DDeviceDumper/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+#   include "pfDisplayHelpers/plX11DisplayHelper.h"
+#endif
+
+#ifdef USE_WAYLAND
+#   include "pfDisplayHelpers/plWaylandDisplayHelper.h"
+#endif
+
+#ifdef PLASMA_PIPELINE_GL
+#   include "pfGLPipeline/plGLPipeline.h"
+plGLEnumerate glEnum;
```
Hoikas:
```suggestion
static plGLEnumerate glEnum;
```

Sources/Tools/hsG3DDeviceDumper/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+#endif
+
+#ifdef PLASMA_PIPELINE_METAL
+#   include "pfMetalPipeline/plMetalPipeline.h"
+plMetalEnumerate mtlEnum;
```
Hoikas:
```suggestion
static plMetalEnumerate mtlEnum;
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/CMakeLists.txt
colincornaby, line 1: If you're pulling plMacDisplayHelper from the Mac client you're going to need to make sure to compile with ARC.

dpogue: What specifically needs ARC in here? It looks like we're properly doing a `CFRelease` of the only owned object...

Sources/Plasma/FeatureLib/pfDisplayHelpers/plMacDisplayHelper.h
colincornaby, line 1: I'm still not a huge fan of this code being pulled from the client. But for this tool to exist that might have to be how things are. I've done my best to keep Obj-C out of the rest of the code base except for one place in the Metal renderer. So if we're doing this - just keep in mind that quarantine will be broken. More Obj-C code, and likely eventually some Swift code will land here.

Sources/Plasma/FeatureLib/pfDisplayHelpers/CMakeLists.txt
colincornaby: I think you're correct? This code was originally written in an ARC environment - but I think you're correct it does not seem to rely on it.

Just keep in mind the iOS version of this class will be very Obj-C and so we will need to revisit the ARC question.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plMacDisplayHelper.h
Hoikas: I don't think we should try to adhere to a policy of limiting Objective-C or Swift code to plClient if that means making reusable code more difficult. That doesn't mean we should simply go nuts writing Objective C or Swift everywhere, but rather treat it in the same way that we do as Win32. Use standard solutions where available; otherwise, dip into whatever the platform requires in as constrained manner as feasible.

Sources/Plasma/CoreLib/hsOptionalCall.h
```diff
@@ -0,0 +1,258 @@
...
+        }
+#endif
+    }
+
+    ~hsOptionalCall()
+    {
+        fLibrary = nullptr;
+    }
+
+    _ValueT operator *() const {
```
Hoikas:
```suggestion
    _ValueT operator *() const
    {
```

Sources/Plasma/CoreLib/hsOptionalCall.h
```diff
@@ -0,0 +1,258 @@
...
+        return *fValue;
+    }
+
+    _ValueT* operator &() const {
```
Hoikas:
```suggestion
    _ValueT* operator &() const
    {
```

Sources/Plasma/CoreLib/hsOptionalCall.h
```diff
@@ -0,0 +1,258 @@
...
+        fLibrary = dlopen(lib, RTLD_LAZY | RTLD_LOCAL);
+        if (fLibrary) {
+            fProc = (_FuncPtrT)dlsym(fLibrary, func);
+            if (!fProc) {
+                dlclose(fLibrary);
+                fLibrary = nullptr;
+            }
+        }
+#endif
+    }
```
Hoikas: This API will be inconsistent from Windows to other platforms. If we every want to optionally load in a cross platform library, we would have to do something like

```C++
#ifdef HS_BUILD_FOR_WINDOWS
hsOptionalCall<void(void)> s_foo(L"libfoo", "foo");
#else
hsOptionalCall<void(void)> s_foo(L"libfoo", "foo");
#endif
```

I can't say I'm an fan of exposing that kind of API. I would be tempted to do something like:


```suggestion
    template<_class StrT>
    hsOptionalCall(const _StrT* lib, const char* func)
        : fLibrary(), fProc()
    {
        static_assert(std::is_same_v<_StrT, char> || std::is_same_v<_StrT, wchar_t>);

#ifdef HS_BUILD_FOR_WIN32
        if constexpr (std::is_same_v<_StrT, char>)
            fLibrary = LoadLibraryA(lib);
        else
            fLibrary = LoadLibraryW(lib);

        if (fLibrary) {
            fProc = (_FuncPtrT)GetProcAddress(fLibrary, func);
            if (!fProc) {
                FreeLibrary(fLibrary);
                fLibrary = nullptr;
            }
        }
#else
        static_assert(std::is_same_v<_StrT, char>)
        fLibrary = dlopen(lib, RTLD_LAZY | RTLD_LOCAL);
        if (fLibrary) {
            fProc = (_FuncPtrT)dlsym(fLibrary, func);
            if (!fProc) {
                dlclose(fLibrary);
                fLibrary = nullptr;
            }
        }
#endif
    }
```

Sources/Plasma/CoreLib/hsOptionalCall.h
```diff
@@ -0,0 +1,258 @@
...
+        return fValue;
+    }
+
+    operator bool() const
+    {
+        return fValue != nullptr;
+    }
+
+    _ValueT Get() const { return *fValue; }
+};
```
Hoikas, line 245: Why is `hsOptionalCall` triplecated?

Sources/Plasma/CoreLib/hsOptionalCall.h
```diff
@@ -0,0 +1,258 @@
...
+
+
+#ifdef HS_BUILD_FOR_APPLE
+#   define HS_OPTIONAL_LIB_SUFFIX ".dylib"
+#else
+#   define HS_OPTIONAL_LIB_SUFFIX ".so"
+#endif
```
Hoikas: No suffix on Windows.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plWaylandDisplayHelper.h"
+
+#include <cstring>
+#include "plWaylandFunctions.h"
+
+template <typename... Args> void ignore(Args...) {}
```
Hoikas:
```suggestion
template <typename... Args>
void ignore(Args&&...) {}
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+void plWaylandDisplayHelper::output_handle_mode(void* data, wl_output* wl_output, uint32_t flags, int32_t width, int32_t height, int32_t refresh)
```
Hoikas: This line wraps in split diff mode

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+    plWaylandDisplayHelper* self = reinterpret_cast<plWaylandDisplayHelper*>(data);
+
+    if (flags & WL_OUTPUT_MODE_CURRENT) {
+        self->fDisplayModes.emplace_back(width, height, 32);
+    }
+}
+
+void plWaylandDisplayHelper::global_handler(void* data, wl_registry* registry, uint32_t id, const char* interface, uint32_t version)
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+        static const wl_output_listener output_listener = {
+            &ignore,
+            &plWaylandDisplayHelper::output_handle_mode,
+            &ignore,
+            &ignore,
+            &ignore,
+            &ignore
+        };
+
+        self->fCurrentOutput = reinterpret_cast<wl_output*>(wl_registry_bind(registry, id, &__wl_output_interface, std::min<uint32_t>((*__wl_output_interface).version, version)));
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    if (fCurrentDisplay)
+        __wl_display_disconnect(fCurrentDisplay);
+}
+
+void plWaylandDisplayHelper::SetCurrentScreen(wl_display* display) const
+{
+    if (fCurrentDisplay == display)
+        return;
+    else if (fCurrentDisplay) {
```
Hoikas:
```suggestion
    if (fCurrentDisplay == display) {
        return;
    } else if (fCurrentDisplay) {
```

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.cpp
```diff
@@ -0,0 +1,149 @@
...
+        wl_output_destroy(fCurrentOutput);
+        fCurrentOutput = nullptr;
+    }
+
+    std::sort(fDisplayModes.begin(), fDisplayModes.end(), std::greater());
+    auto last = std::unique(fDisplayModes.begin(), fDisplayModes.end());
+    fDisplayModes.erase(last, fDisplayModes.end());
+}
+
+std::vector<plDisplayMode> plWaylandDisplayHelper::GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth) const
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.h
```diff
@@ -0,0 +1,76 @@
...
+struct wl_registry;
+
+class plWaylandDisplayHelper : public plDisplayHelper
+{
+public:
+    plWaylandDisplayHelper();
+    ~plWaylandDisplayHelper() override;
+
+    const plDisplayMode& DesktopDisplayMode() const override { return fDesktopDisplayMode; };
+    std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) const override;
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWaylandDisplayHelper.h
```diff
@@ -0,0 +1,76 @@
...
+    mutable wl_display*                 fCurrentDisplay;
+    mutable wl_registry*                fCurrentRegistry;
+    mutable wl_output*                  fCurrentOutput;
+    mutable plDisplayMode               fDesktopDisplayMode;
+    mutable std::vector<plDisplayMode>  fDisplayModes;
+
+    void SetCurrentScreen(wl_display* screen) const;
+
+    static void global_handler(void* data, wl_registry* registry, uint32_t id, const char* interface, uint32_t version);
+    static void output_handle_mode(void* data, wl_output* wl_output, uint32_t flags, int32_t width, int32_t height, int32_t refresh);
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWinDisplayHelper.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+        fDisplayModes.emplace_back(static_cast<int>(dm.dmPelsWidth), static_cast<int>(dm.dmPelsHeight), 32);
+    }
+
+    std::sort(fDisplayModes.begin(), fDisplayModes.end(), std::greater());
+    auto last = std::unique(fDisplayModes.begin(), fDisplayModes.end());
+    fDisplayModes.erase(last, fDisplayModes.end());
+}
+
+std::vector<plDisplayMode> plWinDisplayHelper::GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth) const
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.cpp
```diff
@@ -0,0 +1,116 @@
...
+            if (xrrs[i].width < 800 || xrrs[i].height < 600)
+                continue;
+
+            fDisplayModes.emplace_back(xrrs[i].width, xrrs[i].height, DefaultDepth(fCurrentDisplay, screen));
+        }
+    }
+    else
+#endif
+    {
+        fDisplayModes.emplace_back(DisplayWidth(fCurrentDisplay, screen), DisplayHeight(fCurrentDisplay, screen), DefaultDepth(fCurrentDisplay, screen));
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.cpp
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    std::sort(fDisplayModes.begin(), fDisplayModes.end(), std::greater());
+    auto last = std::unique(fDisplayModes.begin(), fDisplayModes.end());
+    fDisplayModes.erase(last, fDisplayModes.end());
+}
+
+std::vector<plDisplayMode> plX11DisplayHelper::GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth) const
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.h
```diff
@@ -0,0 +1,69 @@
...
+struct _XDisplay;
+
+class plX11DisplayHelper : public plDisplayHelper
+{
+public:
+    plX11DisplayHelper();
+    ~plX11DisplayHelper() override;
+
+    const plDisplayMode& DesktopDisplayMode() const override { return fDesktopDisplayMode; };
+    std::vector<plDisplayMode> GetSupportedDisplayModes(hsDisplayHndl display, int ColorDepth = 32) const override;
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plX11DisplayHelper.h
```diff
@@ -0,0 +1,69 @@
...
+#include "plPipeline/hsG3DDeviceSelector.h"
+#include "plPipeline/pl3DPipeline.h"
+
+struct _XDisplay;
+
+class plX11DisplayHelper : public plDisplayHelper
+{
+public:
+    plX11DisplayHelper();
+    ~plX11DisplayHelper() override;
```
Hoikas:
```suggestion
    ~plX11DisplayHelper();
```

We omit `override` on destructors.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.cpp
```diff
@@ -610,30 +596,107 @@ bool    hsG3DDeviceSelector::IGetD3DCardInfo( hsG3DDeviceRecord &record,
...
+    total = width * height;
+    if( bpp > 8 )
+        total *= ( bpp >> 3 );
+
+    return cardMem + total;
+}
+
+//// IFudgeDirectXDevice //////////////////////////////////////////////////////
+
+void hsGDirect3DTnLEnumerate::IFudgeDirectXDevice(hsG3DDeviceRecord& record, D3DEnum_DisplayInfo* driverInfo, D3DEnum_RendererInfo* deviceInfo)
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.cpp
```diff
@@ -610,30 +596,107 @@ bool    hsG3DDeviceSelector::IGetD3DCardInfo( hsG3DDeviceRecord &record,
...
+        record.SetFudgeFactors(hsG3DDeviceRecord::kDefaultChipset);
     }
 }
 
 //// ITryDirect3DDriver ///////////////////////////////////////////////////////
 //
 //  New DirectX Way
 
-void hsG3DDeviceSelector::ITryDirect3DTnLDriver(D3DEnum_DisplayInfo* drivInfo)
+void hsGDirect3DTnLEnumerate::TryDirect3DTnLDriver(std::vector<hsG3DDeviceRecord>& records, D3DEnum_DisplayInfo* drivInfo)
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
```diff
@@ -152,6 +150,30 @@ class hsGDirect3DTnLEnumerate
...
     HRESULT IConfirmDevice( D3DCAPS9* pCaps, DWORD dwBehavior, D3DFORMAT format );
 
+    void ITryDirect3DTnLDevice(hsG3DDeviceRecord& srcDevRec, D3DEnum_RendererInfo* devInfo);
+
+    /**
+     * Given a driver enum struct, strips out and produces the vendor ID,
+     * device ID and the driver name. Returns true if processed, false
+     * otherwise.
+     */
+    bool IGetD3DCardInfo(hsG3DDeviceRecord& record, D3DEnum_DisplayInfo* driverInfo, uint32_t* vendorID, uint32_t* deviceID, ST::string& driverString, ST::string& descString);
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfDXPipeline/plDXEnumerate.h
```diff
@@ -152,6 +150,30 @@ class hsGDirect3DTnLEnumerate
...
+     * According to Microsoft, the best way to do this is to add in the memory
+     * necessary for the entire desktop. Okay, whatever...
+     */
+    uint32_t IAdjustDirectXMemory(uint32_t cardMem);
+
+    /**
+     * Checks this DirectX device against all our known types and fudges our
+     * caps flags and bias values, etc, accordingly
+     */
+    void IFudgeDirectXDevice(hsG3DDeviceRecord& record, D3DEnum_DisplayInfo* driverInfo, D3DEnum_RendererInfo* deviceInfo);
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -47,11 +47,15 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "plGLPipeline.h"
 
-bool fillDeviceRecord(hsG3DDeviceRecord& devRec)
+static bool fillDeviceRecord(hsG3DDeviceRecord& devRec, const ST::string& driverName, uint32_t provider, hsDisplayHndl display)
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -113,22 +145,25 @@ void plEGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
 
         if (!eglMakeCurrent(display, surface, surface, context))
             break;
 
         hsG3DDeviceRecord devRec;
-        devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
-        devRec.SetDriverName("EGL");
-
-        if (fillDeviceRecord(devRec))
+        if (fillDeviceRecord(devRec, ST_LITERAL("EGL OpenGL API"), hsG3DDeviceSelector::kGLProviderEGL, displayHndl))
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -206,10 +242,7 @@ void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
             break;
 
         hsG3DDeviceRecord devRec;
-        devRec.SetG3DDeviceType(hsG3DDeviceSelector::kDevTypeOpenGL);
-        devRec.SetDriverName("opengl32.dll");
-
-        if (fillDeviceRecord(devRec))
+        if (fillDeviceRecord(devRec, ST_LITERAL("opengl32.dll"), hsG3DDeviceSelector::kGLProviderWGL, displayHndl))
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -225,58 +258,69 @@ void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
+#include <AvailabilityMacros.h>
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 1090
+#   include <ApplicationServices/ApplicationServices.h>
+#else
+#   include <CoreGraphics/CoreGraphics.h>
+#endif
 #include <OpenGL/OpenGL.h>
 
-void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
+void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records, CGDirectDisplayID displayHndl = kCGNullDirectDisplay)
```
Hoikas, line 281: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -225,58 +258,69 @@ void plWGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
-            hsG3DDeviceMode devMode;
-            devMode.SetWidth(mode.Width);
-            devMode.SetHeight(mode.Height);
-            devMode.SetColorDepth(mode.ColorDepth);
-            devRec.GetModes().emplace_back(std::move(devMode));
-        }
-        devRec.SetDefaultModeIndex(0);
-
-        if (fillDeviceRecord(devRec))
+        if (fillDeviceRecord(devRec, ST_LITERAL("OpenGL.framework"), hsG3DDeviceSelector::kGLProviderCGL, displayHndl))
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -291,8 +335,77 @@ void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
+
+        context = glXCreateContext(display, visinfo, False, True);
+        if (!context)
+            break;
+
+        if (glXMakeCurrent(display, None, context) != True)
+            break;
+
+        hsG3DDeviceRecord devRec;
+        if (fillDeviceRecord(devRec, ST_LITERAL("libGL (GLX)"), hsG3DDeviceSelector::kGLProviderGLX, displayHndl))
```
Hoikas: This line wraps in split diff mode.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLEnumerate.cpp
```diff
@@ -291,8 +335,77 @@ void plCGLEnumerate(std::vector<hsG3DDeviceRecord>& records)
...
+            records.emplace_back(std::move(devRec));
+    } while(0);
```
Hoikas:
```suggestion
    } while (0);
```

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -134,6 +134,19 @@ class plDisplayMode
     int Width;
     int Height;
     int ColorDepth;
+
+    plDisplayMode() = default;
+
+    plDisplayMode(int width, int height, int depth)
+        : Width(width), Height(height), ColorDepth(depth) {}
```
Hoikas:
```suggestion
    plDisplayMode(int width, int height, int depth)
        : Width(width), Height(height), ColorDepth(depth)
    { }
```

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -134,6 +134,19 @@ class plDisplayMode
...
+
+    bool operator>(const plDisplayMode& dm) const {
+        return (Width * Height) > (dm.Width * dm.Height);
+    }
```
Hoikas:
```suggestion
    bool operator>(const plDisplayMode& dm) const
    {
        return (Width * Height) > (dm.Width * dm.Height);
    }
```

Sources/Plasma/NucleusLib/inc/plPipeline.h
```diff
@@ -134,6 +134,19 @@ class plDisplayMode
...
+
+    bool operator==(const plDisplayMode& dm) const {
+        return (Width == dm.Width && Height == dm.Height && ColorDepth == dm.ColorDepth);
+    }
```
Hoikas:
```suggestion
    bool operator==(const plDisplayMode& dm) const
    {
        return (Width == dm.Width && Height == dm.Height && ColorDepth == dm.ColorDepth);
    }
```

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -136,12 +137,35 @@ uint16_t  *plTextFont::IInitFontTexture()
...
-    
-    CFURLRef fontURL = (CFURLRef) CTFontDescriptorCopyAttribute(fulfilledFontDescriptor, kCTFontURLAttribute);
-    CFStringRef fileSystemPath = CFURLCopyFileSystemPath(fontURL, kCFURLPOSIXPathStyle);
+
     char cPath[PATH_MAX];
+    CFURLRef fontURL = nullptr;
+
+#if !defined(HS_BUILD_FOR_MACOS) || (defined(HS_BUILD_FOR_MACOS) && MAC_OS_X_VERSION_MAX_ALLOWED >= 1060)
+    // On Mac, kCTFontURLAttribute is only supported in 10.6+
+    fontURL = static_cast<CFURLRef>(CTFontDescriptorCopyAttribute(fulfilledFontDescriptor, kCTFontURLAttribute));
```
Hoikas: This line wraps in split diff mode.

dpogue:
> Given the comments I had about the hsOptionalCall stuff, it feels like it might be a good idea to remove `hsOptionalWinCall` and replace it with the new `hsOptionalCall`. potentially in a separate PR so there's less moving parts here.

I would love to unify `hsOptionalCall` and `hsOptionalWinCall`, the issue was the `WINAPI` (`__stdcall`) vs `__cdecl` handling

Sources/Plasma/CoreLib/hsOptionalCall.h
dpogue: One is for accessing normal functions, one is for accessing exported variables, and one is for accessing variadic functions

Sources/Plasma/CoreLib/hsOptionalCall.h
dgelessus: Why not use `plFileName` and convert to the preferred OS string format internally? That would completely avoid exposing this implementation detail.

Though I'd guess that there won't be many optional calls that have identical function *and* library names on Windows and Unix, so you'll end up with `#ifdef`s no matter what.

Sources/Plasma/CoreLib/hsOptionalCall.h
Hoikas: That makes sense. Could we pull the shared code that does the LoadLibrary/dlopen into a base class to avoid all the repetition?

Sources/Plasma/CoreLib/hsOptionalCall.h
Hoikas: I would prefer to use literals to avoid the string copy, tbh

Hoikas:
This is looking pretty good to me. @colincornaby did you have any additional items that needed to be addressed?

PR #1745, opened 2025-08-12
Remove incorrect unreachable declaration in `DEFAULT_FATAL`

dgelessus:
In debug builds, the code after `FATAL` is in fact reachable, because the failed assertion dialog offers the user the option to ignore the failure.

`hsDebugAssertionFailed` is already declared as `[[noreturn]]` where appropriate (i. e. in non-debug builds). This should be enough for compilers to figure out when the code after the call is unreachable.

Alternatively, if we want `DEFAULT_FATAL` to truly never continue execution, even in debug builds if the user asks for it, we could replace the unreachable declarations with an `abort()` call or similar.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -338,7 +338,7 @@ void hsDebugPrintToTerminal(const char* fmt, ...);
 
 #endif  // HS_DEBUGGING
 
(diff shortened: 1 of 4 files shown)


Hoikas (approved):
I think I submitted something similar to this for the MOUL server code to prevent it from blowing up on bad data from clients <_<

> Alternatively, if we want DEFAULT_FATAL to truly never continue execution, even in debug builds if the user asks for it, we could replace the unreachable declarations with an abort() call or similar.

Definitely not, IMO.

PR #1746, opened 2025-08-12
Don't define `MAX_PATH` on non-Windows

dgelessus:
The current Plasma code uses `MAX_PATH` only in Windows-specific code. All Unix-specific code already uses the native spelling `PATH_MAX`. None of our cross-platform code needs `MAX_PATH`, and we should keep it that way to avoid arbitrary platform-specific path length limits.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -286,10 +286,6 @@ template <> inline double hsToLE(double value) { return hsToLEDouble(value); }
 #   ifndef fileno
 #       define fileno(__F) _fileno(__F)
 #   endif
-#else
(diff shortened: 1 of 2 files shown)


Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -286,10 +286,6 @@ template <> inline double hsToLE(double value) { return hsToLEDouble(value); }
 #   ifndef fileno
 #       define fileno(__F) _fileno(__F)
 #   endif
```
dgelessus, line 288: btw, I am tempted to remove this one as well, in the interest of getting random rarely-used platform handling out of the core header that's included everywhere. But this macro is actually used in two places and thus not completely pointless (it would be *very* easy to replace though).

PR #1753, opened 2025-08-20
vcpkg 2025.07.25 & hopeful epoxy static linking fix

dpogue:
I'm not sure whether the libepoxy static linking on Windows actually works (which is part of what led me to build hsG3DDeviceDumper to try to test it).

I can pull that commit into a difference branch if we want.


Hoikas:
@dpogue Any reason why this is still a draft? Once this is merged, I will rebase #1752 to see if the autotools errors are fixed.

PR #1754, opened 2025-08-20
Setting LocalData on Mac client when NoSelfPatch is set

colincornaby:
I think what happened here is that Windows has the separate patcher and client. I would guess servers like Destiny don't even bother including the launcher.

I'm hard coding this inference on the Mac client.

(While I'm here - I'm happy with how the Mac client was setup but I'm not sure if there is any reason to have a separate Mac launcher. Always open to suggestions - but it's at least aligned with how an iOS client could work.)

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -378,7 +378,7 @@ dispatch_queue_t loadingQueue = dispatch_queue_create("", DISPATCH_QUEUE_SERIAL)
     NetCommConnect();
     [[PLSServerStatus sharedStatus] loadServerStatus];
 
-    BOOL skipPatch = cmdParser.IsSpecified(kArgNoSelfPatch);
+    BOOL skipPatch = cmdParser.IsSpecified(kArgNoSelfPatch) ||  cmdParser.IsSpecified(kArgLocalData);
     if (gDataServerLocal || skipPatch) {
         [self initializeClient];
     } else {


Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -378,7 +378,7 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification
     NetCommConnect();
     [[PLSServerStatus sharedStatus] loadServerStatus];
 
-    BOOL skipPatch = cmdParser.IsSpecified(kArgNoSelfPatch);
+    BOOL skipPatch = cmdParser.IsSpecified(kArgNoSelfPatch) ||  cmdParser.IsSpecified(kArgLocalData);
     if (gDataServerLocal || skipPatch) {
```
Hoikas, line 382: This is not needed because `gDataServerLocal` is true when the `kArgLocalData` is specified.

colincornaby: Huh. I've had to specify NoSelfPatch but now I'm just wondering if that's muscle memory. I will retest and close this PR if it's irrelevant.

Hoikas: I only ever specify `--LocalData`  when I'm launching the mac client ;)

colincornaby: Yep, this is indeed working as expected, my muscle memory was wrong. Closing this PR.

PR #1757, opened 2025-08-21
fix Plugin Translate

Filtik:


Sources/MaxPlugin/MaxConvert/hsControlConverter.cpp
```diff
@@ -633,7 +633,10 @@ plCompoundController *hsControlConverter::MakeTransformController(Control *contr
             Control* sub = (Control*)control->SubAnim(i);
             if (sub)
             {
-                IConvertSubTransform(sub, control->SubAnimName(i), node, tmc, start, end);
+                if (MAX_VERSION_MAJOR >= 24)
+                    IConvertSubTransform(sub, control->SubAnimName(i, false), node, tmc, start, end);
+                else
+                    IConvertSubTransform(sub, control->SubAnimName(i), node, tmc, start, end);
```
Hoikas, line 639: This will unfortunately cause the build to fail for 3ds Max versions prior to 2022. It looks like there are a few other cases in the plugin that are affected as well. I'll get these fixed in a few minutes for you.

PR #1760, opened 2025-08-21
Add tool: plGeneratePythonStubs.

Hoikas:
This is a convenience tool that allows running `generate_stubs.py` outside of the game client. We may want to consider adding this to CI.

### DIFF
--- a/Sources/Plasma/NucleusLib/pnGameMgr/pnGameMgrConst.h
+++ b/Sources/Plasma/NucleusLib/pnGameMgr/pnGameMgrConst.h
@@ -48,7 +48,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 //============================================================================
 // EGameJoinError
 //============================================================================
-enum class EGameJoinError : uint32_t
+enum class EGameJoinError : int32_t
 {
     kGameJoinSuccess,
     kGameJoinErrNotExist,
@@ -59,7 +59,7 @@ enum class EGameJoinError : uint32_t
     kGameJoinErrAlreadyJoined,
     kGameJoinErrNoInvite,
     kNumGameJoinErrors,
-    kGameJoinPending = (uint32_t)-1,
+    kGameJoinPending = -1,
 };
 
 //============================================================================


Sources/Tools/plGeneratePythonStubs/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+        ${plGeneratePythonStubs_SOURCES}
+        ${plGeneratePythonStubs_HEADERS}
+)
+
+target_link_libraries(
+    plGeneratePythonStubs
+    PRIVATE
+        CoreLib
+        plPubUtilInc
+        pfFeatureInc
```
dgelessus, line 23: Doesn't this need `pnNucleusInc` as well?

```suggestion
        pnNucleusInc
        plPubUtilInc
        pfFeatureInc
```

Hoikas:
TODO:
- Update to use hsMain()
- Add to CI

.github/workflows/ci.yml
```diff
@@ -346,3 +346,42 @@ jobs:
...
+    runs-on: ${{ matrix.platform.runner }}
+    name: stubs-${{ matrix.platform.str }}
+    strategy:
+      matrix:
+        platform:
+          - { str: macos-arm64, runner: "macos-14" }
+          - { str: macos-x64, runner: "macos-13" }
+          - { str: linux-x64, runner: "ubuntu-22.04" }
+          - { str: windows-x64, runner: "windows-2022" }
+          - { str: windows-x86, runner: "windows-2022" }
```
dpogue, line 361: Is there actual value in running this 5 times in CI? Do we have Python APIs that are different between the platforms?

It's a good test that the tool works, but seems excessive for validating the stub files

dgelessus: We already had a case where an enum constant accidentally had a different value in 32-bit and 64-bit builds, which was uncovered by the stub generator producing different output when run as 32-bit vs. 64-bit. That's what @Hoikas fixed in 254dcb1f07bc9d79d7abcc4fa16745fc1b80852c in this PR and what I discussed in #1781. For cases like that, it's useful to verify the stubs in multiple different environments - though hopefully such discrepancies won't happen often.

Hoikas: Basically what @dgelessus says. I know it makes the matrix more complicated, but it's pretty important (IMO) that the stubs be consistent on all of our supported platforms. It would be a bummer for CI to reject a set of stubs that someone literally just generated and committed on their own machine. Each stub job only takes like 30 seconds, so I felt like it's worth the trouble to get that extra bit of assurance.

dpogue (approved):

PR #1764, opened 2025-08-23
Clean up assorted oddities/bugs around assert macros and `hsStatusMessage`

dgelessus:
While preparing to `ST::string`-ify these APIs, I came across a bunch of oddities and small bugs. The main changes are:

* The assert macros no longer accept `printf`-style varargs. (This was very rarely used and caused issues for literal `%` characters in the message.)
* The assert macros now reliably support literal `%` characters in the message. (There was another format string confusion bug here, unrelated to the previous point!)
* `hsStatusMessage` now always adds a newline after the message. (Previously, there was a special case for messages that already ended in a newline.)

The other things are boring minor fixes - see commit messages for details.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plModifier/plSimpleModifier.cpp
+++ b/Sources/Plasma/PubUtilLib/plModifier/plSimpleModifier.cpp
@@ -167,16 +167,16 @@ bool plSimpleModifier::IHandleCmd(plAnimCmdMsg* modMsg)
 
 #if 0   // debug
     char str[256];
-        sprintf(str, "ModHandleCmd: time=%f, ts=%f FWD=%d, BWD=%d, SpeedChange=%d sp=%f, CONT=%d, STOP=%d",
-            hsTimer::GetSysSeconds(),
-            modMsg->GetTimeStamp(),
-            modMsg->Cmd(plAnimCmdMsg::kSetForewards),
-            modMsg->Cmd(plAnimCmdMsg::kSetBackwards),
-            modMsg->Cmd(plAnimCmdMsg::kSetSpeed),
-            modMsg->fSpeed,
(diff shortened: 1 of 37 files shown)


Sources/Plasma/PubUtilLib/plModifier/plSimpleModifier.cpp
```diff
@@ -167,16 +167,16 @@ bool plSimpleModifier::IHandleCmd(plAnimCmdMsg* modMsg)
...
-            hsTimer::GetSysSeconds(),
-            modMsg->GetTimeStamp(),
-            modMsg->Cmd(plAnimCmdMsg::kSetForewards),
-            modMsg->Cmd(plAnimCmdMsg::kSetBackwards),
-            modMsg->Cmd(plAnimCmdMsg::kSetSpeed),
-            modMsg->fSpeed,
-            modMsg->Cmd(plAnimCmdMsg::kContinue),
-            modMsg->Cmd(plAnimCmdMsg::kStop));
-        hsStatusMessage(str);
+    sprintf(str, "ModHandleCmd: time=%f, ts=%f FWD=%d, BWD=%d, SpeedChange=%d sp=%f, CONT=%d, STOP=%d",
```
dpogue, line 170: Would this be better as an `ST::format` call and then `.c_str()` it into `hsStatusMessage`?

dgelessus: Yes. I'm doing that later, because there are quite a few cases of `sprintf` + `hsStatusMessage` and some of them are a bit more complex to translate. I wanted to get all of these simple fixes out of the way first.

Hoikas (approved):

PR #1766, opened 2025-08-26
Improve Mac compatibility for hsMessageBox

dpogue:
@colincornaby Curious for your thoughts here...

The big change is moving away from blocks (which are not GCC-compatible and don't work on Mac OS X <10.6) to `performSelectorOnMainThread:`

The part I am least certain about is (as always) the autorelease stuff. Ugh.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
+++ b/Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
@@ -46,72 +46,111 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #import <Cocoa/Cocoa.h>
 
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 1090
+typedef NSInteger NSModalResponse;
+#endif
+
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 101200
+#   define NSAlertStyleCritical         NSCriticalAlertStyle
+#   define NSAlertStyleInformational    NSInformationalAlertStyle
+#   define NSAlertStyleWarning          NSWarningAlertStyle
+#endif
+
+#pragma mark - hsMessageBoxPresenter
+__attribute__((visibility("hidden")))
+@interface hsMessageBoxPresenter : NSObject {
+@public
+    hsMessageBoxResult result;
+}
+
+- (void)showAlert:(NSAlert*)alert;
+@end
+
+@implementation hsMessageBoxPresenter
+- (id)init
+{
+    self = [super init];
+    if (self) {
+        result = hsMBoxCancel;
+    }
+    return self;
+}
+
+- (void)showAlert:(NSAlert*)alert
+{
+    NSModalResponse response = [alert runModal];
+
+    // This is only really safe for OK and Cancel
+    result = static_cast<hsMessageBoxResult>(response);
+}
+@end
+
+
+#pragma mark - hsMessageBox
 hsMessageBoxResult hsMessageBox(const ST::string& message, const ST::string& caption, hsMessageBoxKind kind, hsMessageBoxIcon icon)
 {
     if (hsMessageBox_SuppressPrompts)
         return hsMBoxOk;
 
-    @autoreleasepool {
-        NSString* nsMessage = NSStringCreateWithSTString(message);
-        NSString* nsCaption = NSStringCreateWithSTString(caption);
+    hsMessageBoxResult result = hsMBoxCancel;
 
-        __block NSModalResponse response;
-        NSCondition* lock = [NSCondition new];
-        dispatch_block_t alertBlock = ^{
-            NSAlert* alert = [NSAlert new];
-            alert.messageText = nsCaption;
-            alert.informativeText = nsMessage;
+    hsAutoreleasingScope;
 
-            if (icon == hsMessageBoxIconError)
-                alert.alertStyle = NSAlertStyleCritical;
-            else if (icon == hsMessageBoxIconQuestion)
-                alert.alertStyle = NSAlertStyleInformational;
-            else if (icon == hsMessageBoxIconExclamation)
-                alert.alertStyle = NSAlertStyleWarning;
-            else if (icon == hsMessageBoxIconAsterisk)
-                alert.alertStyle = NSAlertStyleWarning;
-            else
-                alert.alertStyle = NSAlertStyleCritical;
+    NSString* nsMessage = NSStringCreateWithSTString(message);
+    NSString* nsCaption = NSStringCreateWithSTString(caption);
 
-            if (kind == hsMessageBoxNormal)
-                [alert addButtonWithTitle:@"OK"];
-            else if (kind == hsMessageBoxAbortRetyIgnore) {
-                [alert addButtonWithTitle:@"Retry"];
-                [alert addButtonWithTitle:@"Ignore"];
-            } else if (kind == hsMessageBoxOkCancel) {
-                [alert addButtonWithTitle:@"OK"];
-                [alert addButtonWithTitle:@"Cancel"];
-            } else if (kind == hsMessageBoxRetryCancel) {
-                [alert addButtonWithTitle:@"Retry"];
-                [alert addButtonWithTitle:@"Cancel"];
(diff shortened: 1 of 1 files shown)


dgelessus:
I wonder if the autoreleasepool is needed at all, if the code is all linear/blocking? I think you could just do a regular `release` of every object at the end of the block where it's declared. (With ARC, nothing would need to be done at all.)

colincornaby:
> I wonder if the autoreleasepool is needed at all, if the code is all linear/blocking? I think you could just do a regular `release` of every object at the end of the block where it's declared. (With ARC, nothing would need to be done at all.)

An autorelease pool would be needed because Cocoa may internally autorelease things. Whenever one interacts with Cocoa there should be an active autorelease pool unless the documentation supports it being unnecessary.

But what I don't know is if this code inherits the auto release pool from the main thread. But I'm ok with adding one just in case it doesn't.

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,121 @@
...
+    [nsCaption autorelease];
+    [nsMessage autorelease];
+    [alert autorelease];
+#endif
+
+    if ([NSThread isMainThread]) {
+        NSModalResponse response = [alert runModal];
+        result = static_cast<hsMessageBoxResult>(response);
+    } else {
+        HSMBPresenter* presenter = [[HSMBPresenter alloc] init];
```
colincornaby: Nit: Third party Cocoa prefixes should be three letters according to the language guidelines.

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,121 @@
...
+
+#if !__has_feature(objc_arc)
+        [presenter autorelease];
```
colincornaby: I need to pull down the commit and take a look - so don't change anything yet... But in theory ARC would need an autorelease pool too. ARC doesn't automatically create autorelease pools. But unfortunately autorelease pool creation is different under ARC.

dgelessus: In its current state, this PR uses an `@autoreleasepool` block in ARC mode and a traditional `NSAutoreleasePool` in non-ARC mode. That should be enough, right?

dpogue:
Now I'm wondering, is there any time this file would be compiled with ARC and need to use `@autoreleasepool` rather than `NSAutoreleasePool`? I believe plClient is the only module that enables ARC

colincornaby:
We should check on that. It may not matter after the changes - but this code looks like it assumed ARC so it may have been leaky if ARC wasn't on.

dpogue:
I've rebased this and gotten rid of the autorelease usage... but I am definitely not certain that I've got the retain/release stuff right

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
-        // This is only really safe for OK and Cancel
-        return (hsMessageBoxResult)response;
+    hsMessageBoxResult result = hsMBoxCancel;
+
+    hsAutoreleasingScope;
+
+    NSString* nsMessage = NSStringCreateWithSTString(message);
+    NSString* nsCaption = NSStringCreateWithSTString(caption);
+
+    NSAlert* alert = [[NSAlert new] retain];
```
dgelessus: That's one `retain` too much I think. `[NSAlert new]` is the same as `[[NSAlert alloc] init]`, so the newly created object is already retained.

```suggestion
    NSAlert* alert = [NSAlert new];
```

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
+    alert.messageText = nsCaption;
+    alert.informativeText = nsMessage;
```
dgelessus, line 104: `nsCaption` and `nsMessage` could be released here already. The `messageText` and `informativeText` properties are declared as `copy`, so there should be no need to keep the original strings around after the assignment.

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
+        [alert addButtonWithTitle:@"Cancel"];
+    } else {
+        [alert addButtonWithTitle:@"OK"];
     }
+
+    if ([NSThread isMainThread]) {
+        NSModalResponse response = [alert runModal];
+        result = static_cast<hsMessageBoxResult>(response);
+    } else {
+        hsMessageBoxPresenter* presenter = [[hsMessageBoxPresenter new] retain];
```
dgelessus: As above, one `retain` too many:

```suggestion
        hsMessageBoxPresenter* presenter = [hsMessageBoxPresenter new];
```

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
+        [alert addButtonWithTitle:@"Yes"];
+        [alert addButtonWithTitle:@"No"];
+        [alert addButtonWithTitle:@"Cancel"];
+    } else {
+        [alert addButtonWithTitle:@"OK"];
     }
+
+    if ([NSThread isMainThread]) {
+        NSModalResponse response = [alert runModal];
+        result = static_cast<hsMessageBoxResult>(response);
```
dgelessus: To make it clear that both branches do the same thing except for the thread handling, perhaps `hsMessageBoxPresenter` should be used in both cases. Then the only conditional code would be:

```objc
if ([NSThread isMainThread]) {
    [presenter showWithAlert:alert];
} else {
    [presenter performSelectorOnMainThread:@selector(showAlert:) withObject:alert waitUntilDone:YES];
}
```

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
+        result = presenter->result;
+
+        [presenter release];
+    }
+
+    [alert release];
+    [nsMessage release];
+    [nsCaption release];
+
+    // This is only really safe for OK and Cancel
```
dgelessus: I think this comment was meant to go with the cast to `hsMessageBoxResult`, which has now moved to `-[hsMessageBoxPresenter showAlert:]`

dpogue:
@colincornaby This should be ready for review now

I'm semi-deliberately not following Apple's 3-letter-prefix-based naming convention for an ObjC class that is only ever used internally and explicitly not exposed outside of the module. `hsMessageBox` should work well enough as a unique prefix compared to classes in Apple frameworks.

Sources/Plasma/PubUtilLib/plMessageBox/hsMessageBox_Mac.mm
```diff
@@ -46,72 +46,111 @@
...
+
+#if MAC_OS_X_VERSION_MAX_ALLOWED < 101200
+#   define NSAlertStyleCritical         NSCriticalAlertStyle
+#   define NSAlertStyleInformational    NSInformationalAlertStyle
+#   define NSAlertStyleWarning          NSWarningAlertStyle
+#endif
+
+#pragma mark - hsMessageBoxPresenter
+__attribute__((visibility("hidden")))
+@interface hsMessageBoxPresenter : NSObject {
```
colincornaby, line 61: Not blocking on this - just responding to the comment on this in since it's an Obj-C technicality. I'm not sure trying to keep this symbol unlinkable will work in since it still gets dynamically inserted into the Obj-C global namespace at runtime.

If Apple shipped an hsMessageBoxPresenter class we'd be in trouble. But I'm not blocking here because a) I don't think they would and b) They wouldn't use a lowercase hs namespace anyway.

colincornaby (approved):

PR #1768, opened 2025-08-27
Namespace some system-conflicting enum values

dpogue:
Some of these enum key names in the global scope were conflicting with enums and other definitions in system header files, and it's easier to address it in our code than it is to change system files.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAudio/plSoundEvent.cpp
+++ b/Sources/Plasma/PubUtilLib/plAudio/plSoundEvent.cpp
@@ -151,10 +151,10 @@ plSoundEvent::Types plSoundEvent::GetTypeFromCallbackMsg( plEventCallbackMsg *ms
 {
     switch( msg->fEvent )
     {
-        case ::kStart: return kStart;
-        case ::kTime:  return kTime;
-        case ::kStop:  return kStop;
-        case ::kLoop:  return kLoop;
+        case plEventCallbackMsg::kStart: return kStart;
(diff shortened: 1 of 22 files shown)


Sources/Plasma/PubUtilLib/plAudio/plSoundEvent.cpp
```diff
@@ -151,10 +151,10 @@ plSoundEvent::Types plSoundEvent::GetTypeFromCallbackMsg( plEventCallbackMsg *ms
...
     {
-        case ::kStart: return kStart;
-        case ::kTime:  return kTime;
-        case ::kStop:  return kStop;
-        case ::kLoop:  return kLoop;
+        case plEventCallbackMsg::kStart: return kStart;
+        case plEventCallbackMsg::kTime:  return kTime;
+        case plEventCallbackMsg::kStop:  return kStop;
+        case plEventCallbackMsg::kLoop:  return kLoop;
         default:       return kStop;
```
dgelessus, line 158: Here we even had name conflicts within Plasma itself! I remember my IDE complaining about this (or another similar conflict). Thanks!

dgelessus (approved):

Sources/Plasma/PubUtilLib/plAudio/plSoundEvent.cpp
Hoikas: Yoooooooooo WTF

PR #1769, opened 2025-08-28
Fix #1761 stragglers.

### LINKED ISSUE #1767: URULauncher no longer starts
Hi,
I was just about to create a new client and noticed that the URULauncher no longer starts.
So I went through the latest commits and found that it no longer works since commit https://github.com/H-uru/Plasma/commit/79f2db35a2bd67845cf11135d8cfef99638cddf7.

Hoikas:
Fixes #1767 and fixes the plClient entrypoint on Windows to use UCS-2 as well.


dpogue (commented):
I'm :+1: on this if someone tests it and confirms it solves the problem

dpogue (approved):

PR #1770, opened 2025-08-29
Check for (and polyfill) strnlen API

dpogue:
`strnlen` isn't available on Mac OS X pre-10.7, as well as some Solaris systems, so we'll check if it exists and do the counting loop ourselves otherwise.

If we ever convert the whole console over to ST::string, this problem likely goes away

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -80,6 +80,14 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "pfConsoleCore/pfConsoleEngine.h"
 #include "pfPython/cyPythonInterface.h"
 
+#ifdef HAVE_STRNLEN
+#   define pfStrnlen strnlen
+#else
+    static size_t pfStrnlen(const char* str, size_t max)
+    {
+        return static_cast<const char*>(memchr(str, '\0', max)) - str;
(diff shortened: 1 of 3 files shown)


Deledrius (approved):

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -80,6 +80,18 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#ifdef HAVE_STRNLEN
+#   define pfStrnlen strnlen
+#else
+    static size_t pfStrnlen(const char* str, size_t max)
+    {
+        size_t i = 0;
+        while (i < max && str[i])
+            i++;
+
+        return i;
```
dgelessus: Would this work?

```suggestion
        return memchr(str, '\0', max) - str;
```

dpogue:
This file does appear to be the only use of `strnlen` in Plasma

I have some concerns about bringing in MacPorts legacy-support due to the amount of other stuff it's trying to work around, and not wanting to tie us to a specific package manager's meta-package

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
dpogue: This should work, I'll give it a shot

Hoikas (approved):

PR #1771, opened 2025-08-31
Minor improvements to Darwin string tests

dpogue:
It's apparently ill-advised to use non-ASCII characters in NSString `@`-literals or CFString `CFSTR` literals, so test those with pure ASCII strings and add separate tests with manually constructed strings for UTF8 testing.

Also wrap the NSString tests in autorelease pools to make sure nothing leaks.

### DIFF
--- a/Sources/Tests/CoreTests/test_hsDarwin_NS.mm
+++ b/Sources/Tests/CoreTests/test_hsDarwin_NS.mm
@@ -47,21 +47,55 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 TEST(hsDarwin_Foundation, converts_to_ST_string)
 {
-    NSString* str = @"Test ö";
+    NSAutoreleasePool* pool = [[NSAutoreleasePool alloc] init];
+
+    NSString* str = @"Test 123";
     ST::string st = STStringFromNSString(str);
-    EXPECT_STREQ(st.c_str(), "Test \xc3\xb6");
+    EXPECT_STREQ(st.c_str(), "Test 123");
+
+    [pool drain];
 }
 
 TEST(hsDarwin_Foundation, converts_to_NSString)
 {
+    NSAutoreleasePool* pool = [[NSAutoreleasePool alloc] init];
+
+    ST::string st = ST_LITERAL("Test 123");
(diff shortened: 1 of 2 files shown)


dpogue:
The tests were definitely failing on Leopard, although that could have either been a 10.5 bug or a compiler bug. I tried adding `-finput-encoding` and `-fexec-encoding` flags to GCC and it didn't make a difference 😞 

On a [StackOverflow post](https://stackoverflow.com/questions/5690172/xcode-utf-8-literals) about putting UTF-8 literals in NSString someone cites that Apple's old documentation seemed to recommend using always escaping Unicode characters: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Strings/Articles/FormatStrings.html#//apple_ref/doc/uid/20000943-SW2

colincornaby:
Interesting. That doc isn't market as out of date so I'll assume it's still relevant today as well. But it does look like this was working on modern macOS so maybe it was fixed.

Sources/Tests/CoreTests/test_hsDarwin_NS.mm
```diff
@@ -47,21 +47,55 @@
...
 
 TEST(hsDarwin_Foundation, returns_retained_NSString)
 {
+    NSAutoreleasePool* pool = [[NSAutoreleasePool alloc] init];
+
     ST::string st = ST::format("{} Test {}", 12345, "Hello");
     NSString* nstr = NSStringCreateWithSTString(st);
     EXPECT_EQ(1, [nstr retainCount]);
+
+    [pool drain];
```
colincornaby, line 100: I'm assuming this was not built with ARC on modern macOS either making this a leak?

dpogue: Unless CMake is automagically turning on ARC for us, I think the only project that opts in to ARC is plClient

I saw some warnings about leaks when running tests that disappeared after adding the NSAutoreleasePools. It would be nice if there was a good way to polyfill `@autoreleasepool` for both ARC and non-ARC while supporting non-Clang compilers with proprietary syntax 😕

colincornaby (approved):

PR #1773, opened 2025-09-03
Migrate `hsStatusMessage` and `hsDebugPrintToTerminal` to `ST::string`

dgelessus:
Haven't touched `hsAssert` yet, because there are just *so many* of them. If there are no complaints here, I'll repeat the same procedure for `hsAssert` later.

### DIFF
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -348,6 +348,9 @@ hsStatusMessageProc hsSetStatusMessageProc(hsStatusMessageProc newProc);
     void hsStatusMessage(const char* message);
 #endif // PLASMA_EXTERNAL_RELEASE
 
+// Defined as a macro instead of a template function to avoid globally including <string_theory/format>.
+#define hsStatusMessageF(message, ...) (hsStatusMessage(ST::format((message), __VA_ARGS__).c_str()))
+
 #if defined(__clang__) || defined(__GNUC__)
 #   define _COMPILER_WARNING_NAME(warning) "-W" warning
 
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -344,10 +344,8 @@ hsStatusMessageProc hsSetStatusMessageProc(hsStatusMessageProc newProc);
 
 #ifdef PLASMA_EXTERNAL_RELEASE
 #   define hsStatusMessage(x) ((void)0)
-#   define hsStatusMessageF(x, ...) ((void)0)
 #else
     void hsStatusMessage(const char* message);
-    void hsStatusMessageF(const char* fmt, ...);
 #endif // PLASMA_EXTERNAL_RELEASE
 
 #if defined(__clang__) || defined(__GNUC__)
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -319,7 +319,7 @@ void hsDebugBreakAlways();
  *
  * @param msg message to print
  */
-void hsDebugPrintToTerminal(const char* msg);
+void hsDebugPrintToTerminal(const ST::string& msg);
 
 #ifdef HS_DEBUGGING
     
--- a/Sources/Plasma/CoreLib/HeadSpin.h
+++ b/Sources/Plasma/CoreLib/HeadSpin.h
@@ -337,7 +337,7 @@ void hsDebugPrintToTerminal(const char* msg);
 
 #define DEFAULT_FATAL(var) default: FATAL("No valid case for switch variable '" #var "'"); break
 
-typedef void (*hsStatusMessageProc)(const char message[]);
+typedef void (*hsStatusMessageProc)(const ST::string& message);
 
 extern hsStatusMessageProc gHSStatusProc;
 hsStatusMessageProc hsSetStatusMessageProc(hsStatusMessageProc newProc);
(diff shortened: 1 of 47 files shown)


dpogue (approved):
LGTM

Sources/Plasma/CoreLib/HeadSpin.h
```diff
@@ -337,19 +337,20 @@ void hsDebugPrintToTerminal(const char* msg);
...
 #ifdef PLASMA_EXTERNAL_RELEASE
 #   define hsStatusMessage(x) ((void)0)
-#   define hsStatusMessageF(x, ...) ((void)0)
 #else
-    void hsStatusMessage(const char* message);
-    void hsStatusMessageF(const char* fmt, ...);
+    void hsStatusMessage(const ST::string& message);
 #endif // PLASMA_EXTERNAL_RELEASE
 
+// Defined as a macro instead of a template function to avoid globally including <string_theory/format>.
```
Hoikas, line 351: Good. A global dependency on the string_theory headers has been a concern of mine for quite some time. The indirect dependency on type_traits everywhere is a killer for compile time. Not that Plasma compiles particularly quickly now.

Sources/Plasma/NucleusLib/pnKeyedObject/plKey.cpp
```diff
@@ -243,16 +236,17 @@ void plKey::IIncRef()
...
-        char msg[ 512 ];
+        ST::string msg;
         plConst(int) kMaxCnt(30);
+        // FIXME These ifs seem to do nothing - what is the point of this?
         if( fKeyData->fRefCount > kMaxCnt )
-            *msg = 0;
+            msg.clear();
         if( lastData && (fKeyData != lastData) )
-            *msg = 0;
+            msg.clear();
```
Hoikas: Yeah, this is weird. I wonder if, at some point long ago, this was a static error buffer? Anyway, I would be OK with you deleting this whole block.

Sources/Plasma/NucleusLib/pnKeyedObject/plKey.cpp
```diff
@@ -271,18 +265,19 @@ void plKey::IDecRef()
...
-        char msg[ 512 ];
+        ST::string msg;
         plConst(int) kMinCnt(2);
+        // FIXME These ifs seem to do nothing - what is the point of this?
         if( fKeyData->fRefCount < kMinCnt )
-            *msg = 0;
+            msg.clear();
         if( lastData && (fKeyData != lastData) )
-            *msg = 0;
+            msg.clear();
```
Hoikas: Same as above

Sources/Plasma/NucleusLib/pnKeyedObject/plKey.cpp
```diff
@@ -271,18 +265,19 @@ void plKey::IDecRef()
...
         lastData = fKeyData;
-        sprintf( msg, "-: Key %s %s is being de-reffed! Refcount: %d", keyNameToLookFor, CloneString(fKeyData), fKeyData->fRefCount );
-        //hsAssert( false, msg );
-        hsStatusMessageF(msg);
+        msg = ST::format("-: Key {} {} is being de-reffed! Refcount: {}", keyNameToLookFor, CloneString(fKeyData), fKeyData->fRefCount);
+        //hsAssert(false, msg.c_str());
+        hsStatusMessage(msg);
         if( fKeyData->fRefCount == 0 )
-            *msg = 0;
+            msg.clear();
```
Hoikas: Another one?

Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
```diff
@@ -439,7 +441,7 @@ void    plDynamicTextMap::SetFont( const ST::string &face, uint16_t size, uint8_
     if (fCurrFont == nullptr)
     {
         if (!fCurrFont)
-            hsStatusMessageF("Font missing - %s. Using Arial", fFontFace.c_str("nil"));
+            hsStatusMessageF("Font missing - {}. Using Arial", fFontFace);
```
Hoikas, line 444: Can `fFontFace` ever be empty? The explicit replacement value makes me suspicious.

```suggestion
            hsStatusMessageF("Font missing - {}. Using Arial", fFontFace.empty() ? ST_LITERAL("nil") : fFontFace);
```

Sources/Plasma/CoreLib/HeadSpin.h
dgelessus: I mean, HeadSpin.h already does `#include <type_traits>`, and it's been that way for years now. I have some ideas how that could be avoided though - will be interesting to see how much that affects the overall compile time.

Sources/Plasma/PubUtilLib/plGImage/plDynamicTextMap.cpp
dgelessus: I think `face` isn't *supposed* to be empty. But even if it is, printing an empty string into a log message doesn't hurt.

In Cyan's original code, `face` was a `const char*` and "nil" was printed only for actual null pointers. During the move to `ST::string` (well, `plString` back then), this was changed to print "nil" for empty strings. I don't think there's a deeper reason behind this special case, so I simply removed it.

Sources/Plasma/NucleusLib/pnKeyedObject/plKey.cpp
dgelessus: That's my guess as well. I was unsure about the part with `kMaxCnt` - that feels like it's meant to limit how many messages are printed for objects with a lot of refs, but the logic doesn't make sense to me.

Probably best to just remove it, yeah. The next person who actually uses this `TRACK_REFS` feature can improve it if they like :P

Hoikas (approved):

PR #1774, opened 2025-09-03
Remove last non-Asio socket/internet API calls

dgelessus:
The code behind `plNetAddress` still used `inet_ntop`, `inet_pton`, and `getaddrinfo` instead of the corresponding Asio APIs. `getaddrinfo` is especially problematic, because it can block for an unpredictable amount of time because of DNS queries.

Thankfully, most of this code was already unused, except for the IPv4-to-string conversion. I reimplemented that part manually, because it seems overkill to depend on Asio just for a simple string conversion. (But at some point, we might want to properly reimplement `plNetAddress` based on `asio::ip::address`.)

### DIFF
--- a/Sources/Plasma/NucleusLib/pnNetCommon/CMakeLists.txt
+++ b/Sources/Plasma/NucleusLib/pnNetCommon/CMakeLists.txt
@@ -37,8 +37,6 @@ target_link_libraries(
     PRIVATE
         pnMessage
         pnSceneObject
-        $<$<PLATFORM_ID:SunOS>:socket>
-        $<$<PLATFORM_ID:Windows>:ws2_32>
 )
 
 source_group("Header Files" FILES ${pnNetCommon_HEADERS})
--- a/Sources/Plasma/NucleusLib/pnNetCommon/plNetAddress.cpp
+++ b/Sources/Plasma/NucleusLib/pnNetCommon/plNetAddress.cpp
@@ -41,20 +41,27 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *==LICENSE==*/
 
 #include <array>
+#include <string_theory/format>
 #include <string_theory/string_stream>
 
 #include "plNetAddress.h"
-#include "pnNetCommon.h"
+
+ST::string plNetAddress::GetIPv4AddressAsString(uint32_t ip4addr)
+{
+    static_assert(sizeof(ip4addr) == 4*sizeof(uint8_t));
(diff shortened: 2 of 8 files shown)


Hoikas (commented):
Have you considered writing tests for the two functions you re-implemented by hand in `plNetAddress`? The code looks correct and unlikely to fail, but it's always nice to have low-hanging tests.

Otherwise, this seems fine to me. It's kind of a shame to lose cross platform network code, but, aside from the client (which uses asio), no one needs it. We can always find it again in the git history.

dgelessus:
> Have you considered writing tests for the two functions you re-implemented by hand in `plNetAddress`?

I considered it, but was tired/lazy 🤪 Yeah, I should add some tests while I'm here.

> It's kind of a shame to lose cross platform network code, but, aside from the client (which uses asio), no one needs it.

It is cross-platform (mostly), but so is Asio, so it seemed like needless duplication to me... Or is there a possible future use for "Asio-less" DNS resolution?

My other problem with this code is the implicit blocking DNS query when passing a string that's not an IP address. Currently, all of our DNS queries are non-blocking (via Asio), so I'd rather not have a blocking DNS API that's easy to use by accident. (Concretely: I was debugging a UI freeze on the login dialog when connecting to an invalid server, and I wanted to eliminate this blocking code as a possible cause.)

Oh, and it doesn't do any proper error handling if the DNS query fails. I could fix that, but it didn't seem worth the effort for unused code.

Hoikas:
> (Concretely: I was debugging a UI freeze on the login dialog when connecting to an invalid server, and I wanted to eliminate this blocking code as a possible cause.)

I don't know if you've found the cause yet, but I seem to remember that this is related to a blocking lookup somewhere in NetClientComm. IIRC an async API is called, but there's a busy loop waiting on the result.

dgelessus:
Yeah, I think I determined later that it's caused (at least partially) by one of those "wait loops". But the blocking DNS resolution is what I looked into first 😄

dpogue (approved):
This seems fine to me.

I initially had concerns about the endianness of the ip4addr when stringifying, but I see a comment further down in the file confirming that plNetAddress is always in network order (BE) format. :+1:

Hoikas:
@dgelessus Would you prefer to add tests to this PR now defer that to a new PR?

dgelessus:
I was meaning to add the tests into this PR (so I can verify that they pass with both the previous and new implementations). But if I don't get around to it Soon™, feel free to merge if there are no objections - I can always add the tests later.

PR #1776, opened 2025-09-05
Helpers for Darwin Autorelease stuff

dpogue:
In older version of Cocoa with manual memory management, you'd create `NSAutoreleasePool*` objects in code and drain the pool at the end of the function/loop/whatever. You'd call `[obj autorelease]` to associate an object with a pool and it would automatically be destroyed when the pool was cleared.

In mac OS X 10.7, Apple introduced Automatic Reference Counting which forbids calling `[obj autorelease]` in Objective-C and also forbids creating `NSAutoreleasePool` objects manually. Instead, syntax sugar was introduced to the Clang compiler to automatically add autorelease pools using `@autoreleasepool { }` blocks. Internally, Clang turns these into calls to `objc_autoreleasePoolPush()` and `objc_autoreleasePoolPop()`.

That left CoreFoundation objects (which are C API but also bridged with Objective-C) out of participating in autorelease pools, so in OS X 10.9 they introduced `CFAutorelease()` for CF objects.

To this day, you have the choice of compiling with Automatic Reference Counting or without Automatic Reference Counting for Objective-C objects, so we need to handle both cases.

This PR introduces:
* `bridge_cast<T>()` to cast to Objective-C types (without changing the reference count if compiled with ARC)
* `hsAutorelease()` to call either `CFAutorelease()` or `[obj autorelease]` (or nothing, in ARC mode) depending on the object type, with a polyfill for `CFAutorelease()` for older OS X versions
* `hsAutoreleasePool` as a RAII type that will push an autorelease pool when it is created and drain the pool when it leaves scope
* `hsAutoreleasingScope` as a quick way to create an anonymous `hsAutoreleasePool` for the duration of a block

It's somewhat weird to me that `CFAutorelease` was introduced, but there is no public documented API for creating autorelease pools in C. Supposedly there are `_CFAutoreleasePoolPush()` and `_CFAutoreleasePoolPop()` private functions that just call the Clang helpers directly, but they're also only available from OS X 10.7 onwards, so I opted to just use `objc_msgSend` to create an `NSAutoreleasePool` if we're on an older version or not using a compiler with the syntax sugar.


dpogue:
One thing that I found confusing or non-obvious, is that this code crashes (but it also crashes if I use `@autoreleasepool` and `[s autorelease]` so it's not a bug in Plasma code)

```objc
{
    hsAutoreleasingScope;
    NSString* s = [NSString stringWithFormat:@"Hello %s", "world"];
    hsAutorelease(s);
}
```

I guess the `NSString*` returned via the Create Rule is already automatically added to the autorelease pool? It doesn't crash if I can `[s retain]` on it before calling `hsAutorelease(s)`.

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -43,12 +43,84 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#endif
+
+
+inline void hsAutorelease(CFTypeRef ptr) {
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1090
+    CFAutorelease(ptr);
+#else
+    SEL autorelease = sel_registerName("autorelease");
+    IMP imp = class_getMethodImplementation(object_getClass(bridge_cast<id>(const_cast<void*>(ptr))), autorelease);
+    ((CFTypeRef (*)(CFTypeRef, SEL))imp)(ptr, autorelease);
```
dpogue: This is jumping through several hoops to void the case where ARC tries to be overly helpful and causes double-free issues: https://www.mikeash.com/pyblog/friday-qa-2014-05-09-when-an-autorelease-isnt.html

colincornaby:
> It's somewhat weird to me that CFAutorelease was introduced, but there is no public documented API for creating autorelease pools in C. Supposedly there are _CFAutoreleasePoolPush() and _CFAutoreleasePoolPop() private functions that just call the Clang helpers directly, but they're also only available from OS X 10.7 onwards, so I opted to just use objc_msgSend to create an NSAutoreleasePool if we're on an older version or not using a compiler with the syntax sugar.

Autorelease pools were not supposed to be part of CoreFoundation. CoreFoundation shipped on Mac OS 8 under Carbon - before Cocoa - so it's not 1:1. Autorelease pools were a uniquely Cocoa concept.

Casting to an NSObject was usually the best way this was handled. But pure CoreFoundation code should probably avoid Autorelease pools.

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
```diff
@@ -89,15 +89,15 @@ bool pfApplePasswordStore::SetPassword(const ST::string& username, const ST::str
     CFStringRef serviceName = CFStringCreateWithSTString(service);
     CFDataRef   passwordData = CFDataCreate(nullptr, (const UInt8*)password.c_str(), password.size());
 
-    CFAutorelease(accountName);
-    CFAutorelease(serviceName);
-    CFAutorelease(passwordData);
+    hsAutorelease(accountName);
+    hsAutorelease(serviceName);
+    hsAutorelease(passwordData);
```
colincornaby: My guess is that Autorelease was just used here because it's convenient. We might need to change some of the ordering but we could use 'CFRelease' here to be more portable?

(Same with the other uses.)

Sources/Tests/CoreTests/test_hsDarwin_NS.mm
```diff
@@ -47,55 +47,45 @@
 
 TEST(hsDarwin_Foundation, converts_to_ST_string)
 {
-    NSAutoreleasePool* pool = [[NSAutoreleasePool alloc] init];
+    hsAutoreleasingScope;
```
colincornaby, line 50: One option with this scope type is we could just turn off ARC in places where the code needs to be shared. I'm not sure what your plan is with the client in since that targets a newer SDK.

dpogue: Not sure about the client, but I was thinking this would replace the hacking around pools I did in hsMessageBox

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
dpogue: I don't remember if these were originally `CFRelease` and got changed to `CFAutorelease` when you refactored the security APIs (which don't compile on Leopard btw). There's no explicit autorelease pool here so I don't know what benefit `CFAutorelease` is giving us.

I'm fine with doing the releasing manually here since all the object lifecycles should be contained to within this function.

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -43,12 +43,84 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 #define _hsDarwin_inc_
 
+#include "HeadSpin.h"
 #include <string_theory/string>
 #include <string_theory/format>
 
 #ifdef HS_BUILD_FOR_APPLE
 #include <CoreFoundation/CoreFoundation.h>
 
+template<typename T, typename U> inline T bridge_cast(U* obj)
```
Hoikas:
```suggestion
template<typename T, typename U>
inline T bridge_cast(U* obj)
```

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -43,12 +43,84 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+#ifdef HAS_AUTORELEASEPOOL_ARC
+        objc_autoreleasePoolPop(fPool);
+#else
+        objc_msgSend(bridge_cast<id>(fPool), sel_registerName("drain"));
+#endif
+    }
+
+private:
+    void* const fPool;
```
Hoikas: Move this above the `public:` block - you're using `fPool` above.

dpogue:
> The rule is if the function starts with init/new/copy (working from memory here) it's retained. If it doesn't, it's autoreleased.

This is a bit outside the scope of this PR, but trying to wrap my head about this: Does [our `[NSString stringWithSTString:]` class-level helper](https://github.com/H-uru/Plasma/blob/37f564a8c86cbf4a82bb8654d8e5fcf5b24c49b2/Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString%2BStringTheory.mm#L48-L51) in plClient get this rule applied automatically by the compiler, or do we need to do something to mark it as returning an autoreleased string vs a retained string? Should that be wrapped in `hsAutorelease` so it's marked for autoreleasing even in non-ARC?

colincornaby:
I'm going to do a deeper review on this tonight, leaving a few comments in the meantime...

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
colincornaby: I'd like to trim it down because I want to see what's really leaning on autorelease. The abstraction could be avoided if nothing needs it. I can do a more full pass on that later.

Generally - compiling something without ARC and putting back in the backwards compatibility memory management would be preferred to creating a bunch of new abstraction IMO. I use ARC in the client because we had some weird memory management going on and it was easier just to turn on ARC. But we don't need to maintain ARC support in every file.

colincornaby: (Along with that - I should warn there is always the possibility we'll need to adopt Swift for parts of the client in the future. So if we're thinking about backwards compatibility we may also want to decide what is core to Plasma and what could diverge for older macOS support.)

dpogue:
> No - ARC will automatically apply the autorelease. ARC reads the function names and automatically applies the correct ruleset. When calling ARC from ARC it doesn't really matter, the rule set application is only done so non-ARC code can call ARC code and get the matching behavior.

What about non-ARC because it sounds like `NSString` `stringWithFormat:` and `NSString` `stringWithSTString:` will have different behaviour with one returning an autoreleased value and one returning a retained value.

With `hsAutorelease` on an Obj-C (`id`) type, in ARC it's a no-op, and in non-ARC it calls `[obj autorelease]`. That _seems_ (to my naive understanding anyways) like what we want for `NSString` `stringWithSTString:`

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
dpogue: It sounds like there's general agreement to move the pfPasswordStore stuff to not using autorelease at all, which makes sense to me. I'm happy to make that change if you don't see any concerns with how the code is currently structured

colincornaby:
> What about non-ARC because it sounds like NSString stringWithFormat: and NSString stringWithSTString: will have different behaviour with one returning an autoreleased value and one returning a retained value.

Both should return an autoreleased reference in since neither function starts with the magic init/create/copy words.

For completeness I looked it up here:
https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html#//apple_ref/doc/uid/20000994-SW1

>You own any object you create
>You create an object using a method whose name begins with “alloc”, “new”, “copy”, or “mutableCopy” (for example, [alloc]).

dpogue:
> Both should return an autoreleased reference in since neither function starts with the magic init/create/copy words.

Okay, so in ARC this already worked because we return a `CFBridgingRelease()` to transfer ownership to ARC (when ARC is enabled). I've now added `hsAutorelease()` to the non-ARC case so that it should pick up the right behaviour.

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
colincornaby: Let's go ahead and make that change. I'd prefer to not define custom scaffolding where possible - and it doesn't hurt us in any way on present macOS versions.

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -43,12 +43,85 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+#endif
+
+
+inline CFTypeRef hsAutorelease(CFTypeRef ptr) {
+#if MAC_OS_X_VERSION_MIN_REQUIRED >= 1090
+    return CFAutorelease(ptr);
+#else
+    SEL autorelease = sel_registerName("autorelease");
+    IMP imp = class_getMethodImplementation(object_getClass(bridge_cast<id>(const_cast<void*>(ptr))), autorelease);
+    return ((CFTypeRef (*)(CFTypeRef, SEL))imp)(ptr, autorelease);
```
dgelessus: It's generally not a good idea to call method implementation pointers directly, because that may not work correctly for classes that do "dynamic things" with method calls. Is there a particular reason not to use `objc_msgSend` here (cast to the correct signature, as discussed below)?

I saw your other comment about ARC messing with the autoreleased value, but if I'm reading the blog post correctly, the solution was to declare the return type of the `autorelease` call as `CFTypeRef` instead of `id`, so that ARC doesn't touch it. (The blog post also calls the method implementation directly, but never explains why. I would assume that `objc_msgSend` would work just as well.)

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -43,12 +43,85 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+private:
+    void* const fPool;
+
+public:
+    hsAutoreleasePool()
+#ifdef HAS_AUTORELEASEPOOL_ARC
+        : fPool(objc_autoreleasePoolPush())
+#else
+        : fPool(objc_msgSend(
+                    bridge_cast<id>(objc_msgSend(
```
dgelessus: On modern platforms (x86_64 or ARM), you cannot "just call" `objc_msgSend` - it's declared as `void objc_msgSend(void)`, so I would expect this to be a compile error. The correct procedure is to cast `objc_msgSend` to a function pointer with the signature of the method you want to call (in this case `id (*)(id, SEL)`) and then call this correctly typed pointer.

On older platforms (32-bit x86 and PowerPC, and probably also 64-bit PowerPC?), it was declared as `id objc_msgSend(id, SEL, ...)`, so calls like this worked. The problem is that not all calling conventions use the same argument passing rules for varargs as for fixed arguments (aarch64 in particular doesn't, IIRC), so this declaration doesn't work in the general case.

If we really want to use `objc_msgSend` rather than actual Objective-C code, it would be good to write a helper macro/template for this casting logic.

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -89,17 +162,13 @@ inline NSString* NSStringCreateWithSTString(const ST::string& str)
 #if __has_feature(objc_arc)
     return (NSString*)CFBridgingRelease(CFStringCreateWithSTString(str));
 #else
-    return (NSString*)CFStringCreateWithSTString(str);
+    return (NSString*)hsAutorelease(CFStringCreateWithSTString(str));
```
dgelessus: This is incorrect. Our function `NSStringCreateWithSTString` has "Create" in the name, so according to [the Create Rule](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029), it must return a retained/owned reference, not an autoreleased one. (The function is also declared as `__attribute__((ns_returns_retained))` accordingly.)

dgelessus (commented):
I agree with @colincornaby that *if* we can completely avoid autoreleases in the non-Objective-C code, we should do that rather than introducing a custom compatibility/abstraction layer like this. (In that case, most of my comments below would become obsolete.)

But assuming that we do want this abstraction: the autoreleasepool code could be simplified a lot by implementing it in Objective-C. This would require making the functions not `inline` and moving the implementation out of the header, so that the header remains plain C++. (That may reduce performance, but if you're creating autoreleasepools in a tight loop, you're doing something wrong.)

Sources/Plasma/CoreLib/hsDarwin.h
dpogue: I thought this was gonna be safe because the "functions" I'm calling don't take any varargs, but given we've opted to move away from autoreleasing in pure C++ code, I'm liking your idea of just having this bit implemented in ObjC

Sources/Plasma/CoreLib/hsDarwin.h
dpogue: I believe it's because you're not allowed to send that message in ARC mode. Per https://developer.apple.com/library/archive/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226:

> You cannot explicitly invoke `dealloc`, or implement or invoke `retain`, `release`, `retainCount`, or `autorelease`.
>
> The prohibition extends to using `@selector(retain)`, `@selector(release)`, and so on.

Per https://clang.llvm.org/docs/AutomaticReferenceCounting.html#memory-management-methods:

> A program is ill-formed if it contains a method definition, message send, or `@selector` expression for any of the following selectors:
> * `autorelease`
> * `release`
> * `retain`
> * `retainCount`

Sources/Plasma/CoreLib/hsDarwin.h
dpogue: Okay... that makes sense... so the `hsAutorelease` should go in the NSString `stringWithSTString:` category extension so that it follows expected ObjC semantics of static constructors returning autoreleased values?

dgelessus: I'm not sure it matters, because plClient is currently built with ARC enabled, right? But to make that code work correctly *without* ARC, the autorelease would have to go into the `+[NSString stringWithSTString:]` implementation, yes.

Sources/Plasma/CoreLib/hsDarwin.h
dgelessus: In this particular case, I assume the `objc_msgSend` declaration with varargs would work (passing no arguments is probably quite safe). The problem is that when building on modern architectures, Apple's `<objc/message.h>` doesn't give you that declaration and always requires a cast.

(There is a macro that you can define to get the old varargs declaration on architectures where it's not the default, but on arm64, even that is no longer allowed.)

Sources/Plasma/CoreLib/hsDarwin.h
dgelessus: Have you tested it? My guess would be that only `@selector(autorelease)` is a compile error in ARC mode, and the `objc_msgSend` would work fine if you get the selector another way (as we do here).

*Officially*, I think neither way of indirectly calling `autorelease` is allowed in ARC mode. We're just circumventing the compile-time checks here on purpose.

We could also simply disallow this branch in ARC mode and say that if you're building for a macOS target older than 10.9, all the code has to be built without ARC.

dpogue: I did a test with `objc_msgSend`, and it seems to work in ARC mode (as long as it's typed as returning `CFTypeRef` and not `id`), so I've moved this over to using that

Sources/Plasma/CoreLib/hsDarwin.h
dpogue: I've changed this over so that `hsAutoreleasePool` is only defined in ObjC mode (and thus only usable from ObjC/ObjC++ code) and just calls `NSAutoreleasePool` directly.

Will test that on Leopard tonight and confirm that it's happy (or, as happy as an OS from 2007 being forced to confront modern technology can be)

Sources/Plasma/Apps/plClient/Mac-Cocoa/NSString+StringTheory.mm
```diff
@@ -47,7 +47,7 @@ @implementation NSString (StringTheory)
 
 + (id)stringWithSTString:(const ST::string&)string
 {
-    return NSStringCreateWithSTString(string);
+    return hsAutorelease(NSStringCreateWithSTString(string));
```
colincornaby: I'm leaning towards this file just turning off ARC instead of adding hsAutorelease.

Sources/Plasma/FeatureLib/pfPasswordStore/pfPasswordStore_Apple.cpp
colincornaby, line 1: This changes look good.

dpogue:
I've removed the `hsAutorelease` stuff here so this focuses just on implementing the autoreleasepool helper (which is needed to fix hsMessageBox on OS X 10.5)

Sources/Plasma/CoreLib/hsDarwin.h
```diff
@@ -95,18 +106,47 @@ inline NSString* NSStringCreateWithSTString(const ST::string& str)
...
+    {
+#if __has_feature(objc_arc) || (MAC_OS_X_VERSION_MIN_REQUIRED >= 1070 && defined(__clang__))
+        objc_autoreleasePoolPop(fPool);
+#else
+        [bridge_cast<NSAutoreleasePool*>(fPool) drain];
+#endif
+    }
+};
+
+#define hsAutoreleasingScope hsAutoreleasePool hsUniqueIdentifier(_AutoreleasePool_);
```
dgelessus: To ensure that users of this macro have to put a semicolon after it:

```suggestion
#define hsAutoreleasingScope hsAutoreleasePool hsUniqueIdentifier(_AutoreleasePool_)
```

Sources/Tests/FeatureTests/pfPasswordStoreTest/test_pfPasswordStore.cpp
dgelessus, line 1: The platform password store tests should call `SetServerDisplayName` so that the credentials get saved to a sensible place. IIRC, at least on Windows, the code that saves the username in the registry behaves a little oddly if the server display name is an empty string.

dpogue: Thanks for the warning!

Sources/Tests/FeatureTests/pfPasswordStoreTest/CMakeLists.txt
dgelessus, line 1: Late comment: I just noticed that this doesn't use our custom `plasma_test` function, unlike all the other tests' CMakeLists. Was there a particular reason for that, or just an oversight?

(Not a big issue - I only noticed because `test_pfPasswordStore` isn't in the same IDE folder as all the other tests.)

dpogue: Just an oversight. I resurrected these tests from a branch I had back in like 2014 that I could never get completely passing in CI.

I'll fix that up.

PR #1777, opened 2025-09-08
Resolve Max Textures During Export (Part Deux)

Hoikas:
Closes #1623 by either using the safer, new in Max 2022 `BMMGetFullFilename()` or by falling back to copying the `BitmapInfo` object in older versions of Max. I also centralized all of the conversions to `plPlasmaMAXLayer::GetBitmapFileName()`.


dpogue (approved):
LGTM

PR #1779, opened 2025-09-09
Journal Page Remembering (Part Deux)

Hoikas:
Superceeds and closes #1679 by addressing concerns I had and fixing some disclosed drawbacks with the previous implementation.

Journals should now remember their last position, for the most part. I've retained the 1024 character limit for the chronicle to ensure nothing bad happens. This PR does not use JSON at all and also hashes the journal content to ensure that we don't assume that page numbers from old journal contents correspond to the new contents. Further, with this PR, the auto turn mechanic can tell the difference between closing the journal on the cover (the cover reopens) or closing the journal on the first page (the first page reopens).

Amusingly, I didn't realize it was possible to turn back to the cover in a journal until I was looking at this. For posterity, you can turn back to the cover by clicking on where the left turn button should be on the first page.

As a bonus, you get some code cleanups in xJournalBookGUIPopup.

### DIFF
--- a/Scripts/Python/plasma/PlasmaConstants.py
+++ b/Scripts/Python/plasma/PlasmaConstants.py
@@ -103,6 +103,7 @@ class PtBookEventTypes:
     kNotifyPreviousPage = 4
     kNotifyCheckUnchecked = 5
     kNotifyClose = 6
+    kNotifyOpen = 7
 
 class PtBrainModes:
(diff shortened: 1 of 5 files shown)


TheScarFr:
Thank you very much Hazado and Adam, I was waiting for this feature since long time ago🎉

PR #1780, opened 2025-09-09
Windows/ARM support

colincornaby:
This PR has been tested on Visual Studio on top of ARM Windows running in VMWare on a MacBook Pro. Requires Python 3.12 - so this PR is being opened post that merge. _This change only targets ARM64._

Changes were required:
- We have to take ownership of the PhysX port in since we have to patch it for ARM Windows. I've added the patch files necessary.
- I added the correct stack frame offsets for ARM64.
- I fixed the CPUID tests which were not behaving correctly on ARM.

Note that I did not add NEON acceleration - Plasma will currently fall back to software for matrix operations on ARM.

Also the CairoSVG wheel is not hosted already for Python 3.12 - and I had a lot of trouble getting it to compile locally. For now - I've disabled the CairoSVG dependency on ARM. I think later versions of Python have that wheel precompiled so I think that will be fixed in the future.

### DIFF
--- a/Scripts/Ports/physx/fix-compiler-flag.patch
+++ b/Scripts/Ports/physx/fix-compiler-flag.patch
@@ -1,0 +1,75 @@
+diff --git a/physx/compiler/public/CMakeLists.txt b/physx/compiler/public/CMakeLists.txt
+index 77776ca7..bd7b496d 100644
+--- a/physx/compiler/public/CMakeLists.txt
++++ b/physx/compiler/public/CMakeLists.txt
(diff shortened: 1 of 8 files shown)


dpogue:
Build failures look to be caused by https://gitlab.xiph.org/ being down (temporarily, we hope), rather than any of the changes here, but I'd prefer to see green checkmarks before merging this.

It might be worth, as a test once CI is passing, to add a Windows Arm runner to the GitHub workflows and confirm that it builds properly there (and then likely revert that so we don't keep expanding the CI matrix in all directions)

PR #1782, opened 2025-09-11
Use prefixes and suffixes for manifest names to indicate platform.

Hoikas:
This formalizes a system of prefixing and suffixing manifest names to indicate the platform. Given that the current examples of production manifests are:

`External` -> external client, windows, 32-bit
`macExternal` -> external client, macOS cider wrapper, 32-bit Intel

The baseline rule is that no prefix means Windows and no suffix means 32-bit intel. Therefore, all other operating systems and architectures must be suffixed. I included PowerPC suffixes for @dpogue whom I know is tinkering with PowerPC macs. Here is what I came up with:

Suffixes:
- ` `: (no suffix) x86
- `AMD64`: AMD64/x86_64
- `ARM`: 32-bit arm (eg NVIDIA Tegra)
- `ARM64`: 64-bit arm (eg Apple Silicon, Snapdragon X)
- `PPC`: 32-bit PowerPC (eg iMac G4)
- `PPC64`: 64-bit PowerPC (eg PowerMac G5)

Prefixes:
- ` `: (no prefix) Windows
- `macos`: macOS (**not** `mac` because this conflicts with the old cider wrapper)
- `linux`: Linux (presumably glibc based)

So, some sample manifests:
- `External64`: x64 Windows external client image manifest
- `linuxThinInternal`: x86 Linux internal client manifest
- `macosInternalPatcherARM64`: Apple Silicon macOS internal release patcher manifest

Currently, we do not build any universal/fat binaries for macOS, therefore nothing has been added for any kind of universal mac binary. I'm somewhat concerned about the UX if/when an architecture gets dropped and a user on that architecture downloads a new fat binary without that architecture. Aside from that, the best way to handle universal binaries, IMO, would be to add a `Universal` or `Fat` suffix (remember, no suffix is already established as 32-bit intel) and detect if the current binary is fat. If so, use the fat binary. If not, use the arch-specific one. This would match the Windows behavior where anyone running a 32-bit client on 64-bit Windows will continue to receive 32-bit clients, even though they can run a 64-bit client. However, as stated previously, we currently do not build universal binaries, so this is currently only a tangential point.

~~This proposal matches what I'm already doing in UruManifest - it assumes that amd64 windows binaries should go into a manifest suffixed with `64`.~~ The proposed mac changes to UruManifest will need adjustments because it wrongly reassigns the meaning of `macExternal` to "64-bit intel". There's also some other oddness in there about a mac bundle manifest that doesn't make any sense. All client files should go into the client manifest.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
@@ -41,6 +41,8 @@ Mead, WA   99021
 *==LICENSE==*/
 
 #include "plManifests.h"
+
+#include "hsConfig.h"
 #include "plFileSystem.h"
 
 #include <string_theory/string>
@@ -62,7 +64,9 @@ Mead, WA   99021
 #   define EXECUTABLE_SUFFIX ""
 #endif
 
-#if (defined(__GNUC__) && defined(__x86_64__)) || (defined(_MSC_VER) && defined(_M_X64))
+#if !defined(USE_PLATFORM_MANIFESTS)
+#   define MANIFEST_SUFFIX ""
+#elif (defined(__GNUC__) && defined(__x86_64__)) || (defined(_MSC_VER) && defined(_M_X64))
 #   define MANIFEST_SUFFIX "AMD64"
 #elif (defined(__GNUC__) && defined(__i386__)) || (defined(_MSC_VER) && defined(_M_IX86))
 #   define MANIFEST_SUFFIX ""
@@ -78,7 +82,9 @@ Mead, WA   99021
 #   error "Unknown architecture in plManifest"
 #endif
 
-#if defined(HS_BUILD_FOR_WIN32)
+#if !defined(USE_PLATFORM_MANIFESTS)
+#   define MANIFEST_PREFIX ""
+#elif defined(HS_BUILD_FOR_WIN32)
 #   define MANIFEST_PREFIX ""
 #elif defined(HS_BUILD_FOR_MACOS)
     // "mac" was used by the old Cider wrappers. To prevent silently overwriting these clients,
--- a/Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
@@ -48,8 +48,10 @@ Mead, WA   99021
 // Helper that returns the appropriate string per build
 #ifdef PLASMA_EXTERNAL_RELEASE
 #   define MANIFEST(in, ex) ex
+#   define MANIFEST_RELEASE_TYPE "External"
 #else
 #   define MANIFEST(in, ex) in
+#   define MANIFEST_RELEASE_TYPE "Internal"
 #endif // PLASMA_EXTERNAL_RELEASE
 
 #if defined(HS_BUILD_FOR_APPLE)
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
```diff
@@ -60,41 +62,71 @@ Mead, WA   99021
...
+#elif (defined(__GNUC__) && defined(__ppc__))
+#   define MANIFEST_SUFFIX "PPC"
+#else
+#   error "Unknown architecture in plManifest"
+#endif
+
+#if defined(HS_BUILD_FOR_MACOS)
+#   define MANIFEST_PREFIX "mac"
+#elif defined(HS_BUILD_FOR_UNIX)
+#   define MANIFEST_PREFIX "nix"
```
dpogue: Strictly speaking, this is probably not going to work for the whole set of Linux/Solaris/BSD(s) that fall under "Unix-like", but I also imagine that beyond Linux nobody is going to be providing pre-built clients, so maybe this isn't something we should care about here.

I would maybe suggest using "lin" as a prefix for Linux builds, but we'd need to pipe that through as a define I guess...

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
```diff
@@ -60,41 +62,71 @@ Mead, WA   99021
...
+#elif (defined(__GNUC__) && defined(__i386__)) || (defined(_MSC_VER) && defined(_M_IX86))
+#   define MANIFEST_SUFFIX ""
+#elif (defined(__GNUC__) && defined(__arm__) || (defined(_MSC_VER) && defined(_M_ARM))
+#   define MANIFEST_SUFFIX "ARM"
+#elif (defined(__GNUC__) && defined(__aarch64__) || (defined(_MSC_VER) && defined(_M_ARM64))
+#   define MANIFEST_SUFFIX "ARM64"
+#elif (defined(__GNUC__) && defined(__powerpc64__))
+#   define MANIFEST_SUFFIX "PPC64"
+#elif (defined(__GNUC__) && defined(__ppc__))
+#   define MANIFEST_SUFFIX "PPC"
```
dpogue, line 80: There's no harm in having these, but again I'm not sure anyone is realistically going to be distributing prebuilt PPC clients (and even I haven't been adventurous enough to poke at PPC64 😛)

I wonder what the best way to handle a universal (i.e., x86_64 and arm64 in the same file) macOS build would be? Separate manifests pointing to the same files might be least hassle and avoid needing to teach the client too much?

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
```diff
@@ -60,41 +62,71 @@ Mead, WA   99021
...
+    return ST_LITERAL(
+        MANIFEST("plClient", "UruExplorer")
+        EXECUTABLE_SUFFIX
+    );
 }
 
 plFileName plManifest::PatcherExecutable()
 {
+    // On macOS, due to the chaos of .app bundles, the client and
+    // patcher are the same executable at this time.
```
dpogue, line 110: I'd be interested in hearing if there's a compelling reason to not unify them on other OSes...

dpogue (commented):
Overall this looks good.

My one minor complaint (which isn't new) is that because this is hardcoded at build-time, there's no way to do something like ask plFilePatcher to download the files for another manifest to inspect them

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
```diff
@@ -60,41 +62,71 @@ Mead, WA   99021
 #   define EXECUTABLE_SUFFIX ""
 #endif
 
+#if (defined(__GNUC__) && defined(__x86_64__)) || (defined(_MSC_VER) && defined(_M_X64))
+#   define MANIFEST_SUFFIX "64"
+#elif (defined(__GNUC__) && defined(__i386__)) || (defined(_MSC_VER) && defined(_M_IX86))
+#   define MANIFEST_SUFFIX ""
+#elif (defined(__GNUC__) && defined(__arm__) || (defined(_MSC_VER) && defined(_M_ARM))
```
dpogue:
```suggestion
#elif (defined(__GNUC__) && defined(__arm__)) || (defined(_MSC_VER) && defined(_M_ARM))
```

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
```diff
@@ -60,41 +62,71 @@ Mead, WA   99021
...
+#   define MANIFEST_SUFFIX "ARM"
+#elif (defined(__GNUC__) && defined(__aarch64__) || (defined(_MSC_VER) && defined(_M_ARM64))
```
dpogue:
```suggestion
#elif (defined(__GNUC__) && defined(__aarch64__)) || (defined(_MSC_VER) && defined(_M_ARM64))
```

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
Hoikas: I've been evolving toward pretty strongly preferring that the patcher and client remain separate, especially on Windows. The patcher's job is (IMO) to ensure that the client is in a state where it can launch correctly. Sometimes, that means that the patcher has to install new redistributables or do other rather clever tasks.

Case in point, when OU upgraded the compiler from VS .NET to VS 2010, the patcher did not know how to install redistributables. So, I added this functionality to the patcher, which was recompiled with VS 2003, but the client was compiled with VS 2010. The MOULa server shipped down the VS .NET compiled patcher for quite some time so that players could quietly have the VC++2010 redistributable installed before the VS 2010 client launched. If the client and patcher are unified, these kinds of tricks become much more difficult. Although our patcher now has redist installing built in, I am wary of closing the door for these kinds of tricks.

Hoikas: Also, I still have the hope of plClient.exe not spawning any OS gui at all and all logging in related tasks occuring in the StartUp Age. Having a separate patcher makes that delineation of "basically no native GUI stuff" and "oh, look, dialogs" pretty clear.

dpogue: Ahh, that does sound like a pretty compelling reason to keep it that way (at least on Windows)

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
Hoikas: As I mentioned in the PR, I think the universal manifest should be suffixed "Universal," and if the client detects that it's a fat binary, it can request the fat manifest. A bit of a special case, but so is all of macOS, IMO. Another hot take: just because a manifest is listed here doesn't mean it *has* to exist. If we only ever distribute a universal client, that's totally fine. It's just not what the build system spits out natively, so we need to handle the actual build output.

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
Hoikas: I was wondering about that. I know that gaming on FreeBSD is a thing, but 🤷‍♂️

Sources/Plasma/FeatureLib/pfPatcher/plManifests.cpp
colincornaby: FWIW - I'm open to this discussion on macOS as well, but was driven by a few things:

- It's an unusual arrangement on macOS.
- I don't think we have a similar runtime problem - but there is the potential the client patches itself to a client with higher OS requirements, and then can no longer launch itself.

dpogue: macOS doesn't really have installable newer runtimes. You get what Apple shipped and you will never get a newer version without upgrading your OS.

dgelessus:
> Do you remember when the PhysX 2.6.4 update happened on MOULa?

Perhaps I'm misremembering - it might not have been the switch to PhysX 2.6.4, but rather when the PhysX_Setup.exe was changed from the old Ageia installer to the newer Nvidia "legacy" installer. But it definitely had something to do with the PhysX redistributable.

I have some old files from late May 2020 where I tried to retrofit the new PhysX files into the Cider wrapper (it actually worked IIRC, but I never uploaded the result anywhere). I wrote "OpenUru 159" in the file names, and Chogon's release notes for that build (from 2020-05-27) list "Physics update" as one of the changes, so that checks out.

In short: it's all at least 5 years ago.

Hoikas:
I've changed the macOS prefix to `macos` to avoid conflicting with the Cider wrapper. I've also added a configure option `USE_PLATFORM_MANIFESTS` that will turn these names off and just use the old `External`, `ThinExternal`, etc. manifest names. The new platform specific names default to on.

dgelessus (approved):
This all looks good to me now. (I haven't tested the code myself though.)

colincornaby:
> I've changed the macOS prefix to `macos` to avoid conflicting with the Cider wrapper. I've also added a configure option `USE_PLATFORM_MANIFESTS` that will turn these names off and just use the old `External`, `ThinExternal`, etc. manifest names. The new platform specific names default to on.

Does this let the macOS client connect and use the Window manifests? Not interested in that as a user feature - just would like to understand it for testing purposes.

Hoikas:
> Does this let the macOS client connect and use the Window manifests?

No, rather, this makes the macOS client request a different manifest that should never reference any Windows specific files.

PR #1783, opened 2025-09-11
Add missing double click handler for macOS.

Hoikas:
These are needed to select a player in the StartUp Age by double clicking on its entry in the list.


dpogue (approved):
This looks good to me, but might be worth getting @colincornaby to look in case there's some weird non-obvious behaviour around click counts

Technically you could do `event.clickCount` instead of `[event clickCount]` (since it's a property) but functionally it's identical

dpogue:
> For my information - where in Uru are double click events used?

I think possibly only on the player select screen, which was added in https://github.com/H-uru/Plasma/pull/1654

Hoikas:
Thanks for the suggestions - I know basically nothing about ObjC 😅

dpogue (approved):

Hoikas:
Pushed the suggested change.

PR #1784, opened 2025-09-14
Reduce copy-paste in console command implementation code

dgelessus:
A few declarations and macros were copy-pasted into each console command implementation file. This PR pulls them out into a proper shared header.

Also finally adds a named type for the `PrintFunc`s that get passed all around the console code.

### DIFF
--- a/Sources/MaxPlugin/MaxMain/GlobalUtility.cpp
+++ b/Sources/MaxPlugin/MaxMain/GlobalUtility.cpp
@@ -94,14 +94,6 @@ public:
 static PlasmaMaxClassDesc PlasmaMaxCD;
 ClassDesc* GetGUPDesc() { return &PlasmaMaxCD; }
 
-//////////////////////////////////////////
-
-// This function is from the console.  This dummy version is here so that plNetLinkingMgr will build.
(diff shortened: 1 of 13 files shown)


Sources/MaxPlugin/MaxMain/GlobalUtility.cpp
```diff
@@ -94,14 +94,6 @@ class PlasmaMaxClassDesc : public plMaxClassDesc<ClassDesc>
 static PlasmaMaxClassDesc PlasmaMaxCD;
 ClassDesc* GetGUPDesc() { return &PlasmaMaxCD; }
 
-//////////////////////////////////////////
-
-// This function is from the console.  This dummy version is here so that plNetLinkingMgr will build.
-plKey FindSceneObjectByName(const ST::string& name, const ST::string& ageName, char* statusStr, bool subString)
```
dgelessus, line 100: As usual, I have no way to test the Max plugin. I'm pretty sure this is safe to remove, but perhaps someone should check that the Max plugin still loads and doesn't complain about a missing function or something.

Hoikas: I don't think this should be a problem.

PR #1786, opened 2025-09-15
Fix spawn point argument of `Net.LinkWithOriginalBook`

dgelessus:
In short, the existing code was broken. The usual call `Net.LinkWithOriginalBook AGENAME Default:LinkInPointDefault:;` only worked sort of by accident and left some junk in the vault (mostly harmless junk, thankfully). See the commit message for details.

This PR makes the spawn point argument work properly, with checks against obviously incorrect inputs. It also makes the argument optional, so simply `Net.LinkWithOriginalBook AGENAME` will now link to the default spawn point.


Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommandsNet.cpp
```diff
@@ -361,12 +402,26 @@ PF_CONSOLE_CMD( Net,        // groupName
...
+        return;
+    }
+
     plAgeLinkStruct link;
     link.GetAgeInfo()->SetAgeFilename(params[0]);
-    link.SpawnPoint() = plSpawnPointInfo(params[1], params[1]);
+
+    if (numParams >= 2) {
+        ST::string error;
+        link.SetSpawnPoint(TryParseSpawnPointInfo(params[1], error));
```
Hoikas: I don't really love the error string out parameter check. It just seems like odd semantics, especially when most libraries I'm familiar with return a status and use an out parameter for the result. That's basically the opposite of what we have here, unfortunately. I really wish we could use [`std::expected`](https://en.cppreference.com/w/cpp/utility/expected.html) for this, but that's a C++23 feature that requires bumping our minimum Visual Studio requirement to VS 2022.

If you're feeling ambitious, you might consider using an std::expected polyfill to make these semantics clearer. It looks like Sy Brand with MS has a [polyfill](https://github.com/TartanLlama/expected) and martinmoene has [one](https://github.com/martinmoene/expected-lite) as well. Both of these are available as vcpkg ports and both appear to generate CMake package config files, so bringing in either of them should be trivial.

Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp
```diff
@@ -2256,6 +2256,7 @@ void cyMisc::RebuildCameraStack(const ST::string& name, const ST::string& ageNam
             plVirtualCam1::Instance()->PushThirdPerson();
             ST::string errmsg = ST::format("Sceneobject {} not found", name);
             PyErr_SetString(PyExc_NameError, errmsg.c_str());
+            PYTHON_RETURN_ERROR;
```
Hoikas, line 2259: Well, that's fun.

Deledrius:
I really appreciate setting a sane default on this to save all of the redundant (error-prone) typing.  I've never felt I understood most of this well enough to make the changes myself. :)

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommandsNet.cpp
Hoikas: I should probably clarify that I don't mean "return the error string" because I'm not a fan of that either, as we've previously discussed. I'd rather see an exception or something like std::expected.

I wonder if it would be worth having an hsExpected for this pattern?

dgelessus: Yeah, I don't exactly like it either. I just didn't want to add another custom exception just for this helper function. `std::expected` or equivalent really would be nice for these cases (especially as it's coming up again and again).

As a quick solution for this case, maybe I'll just return a `std::variant<plSpawnPointInfo, ST::string>` - not ideal, but it works here, because the success and error types are distinct. Then you or I can make a separate PR for adding something like `hsExpected` or an existing third-party implementation.

Hoikas: I think that sounds like a fair compromise for now. `std::expected` is basically a fancy `std::variant` anyway.

Hoikas (approved):

PR #1788, opened 2025-09-16
Implement `hsExpected`.

Hoikas:
This implements `hsExpected`, a minimal polyfill for `std::expected` and converts `TryParseSpawnPointInfo` from #1786 to it instead of `std::variant`. Alternatively, we could pull in a more full-featured 3rd party polyfill, but the code is simple enough for a minimal polyfill that we should probably just vendor it until we decide to bump to C++23.


dgelessus:
In terms of complexity, this seems much nicer than the third-party solutions you linked before, which are both multiple thousand lines long. I'm sure they have more features, but to me it felt overkill to pull all of that in when we only need the basics.

If the intention is to follow the `std::expected` API closely, perhaps the method names should be in `snake_case` like in the original? It doesn't match the Plasma naming style, but it would simplify switching to the real `std::expected` (or a similar third-party implementation) in the future.

(No idea about the implicit conversion errors with GCC and clang, sorry.)

Deledrius:
I think in that case being different is okay, and won't pose too much extra work if/when the official implementation becomes available.

zrax:
> I think in that case being different is okay, and won't pose too much extra work if/when the official implementation becomes available.

More importantly, it makes it easier to identify areas which might need a closer look when porting to a different implementation.  For that reason alone, I'm in favor of keeping the "incompatible" naming.

Sources/Plasma/CoreLib/hsExpected.h
```diff
@@ -0,0 +1,292 @@
...
+    const E&& Value() const&&
+    {
+        return std::move(fValue);
+    }
+};
+
+template<typename T, typename E>
+class hsExpected
+{
+    static_assert(!(std::is_void_v<T> || std::is_void_v<E>), "hsExpected cannot use void");
```
zrax, line 132: It seems like `hsExpected<void, Err>` would be a useful use case, but I suppose that would add a lot of unnecessary complexity here...

dgelessus: There's always [`std::monostate`](https://cppreference.com/w/cpp/utility/variant/monostate.html) as a workaround.

PR #1791, opened 2025-09-17
Skip clang-tidy for targets with no real source files

dgelessus:
For header-only targets like `pfFeatureInc`, the generated tidy target called clang-tidy with no source files, which clang-tidy considers an error (checked with clang-tidy 21.1.0). So create each tidy target only if there are any source files.


cmake/Sanitizers.cmake
```diff
@@ -92,16 +92,21 @@ function(plasma_sanitize_target TARGET)
...
-                COMMAND_EXPAND_LISTS
-                COMMAND ${CLANG_TIDY_EXE} --quiet -p "${CMAKE_BINARY_DIR}" "-export-fixes=${_TARGET_FIXES_FILE}" "${_TIDY_SOURCES}"
-                WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-                BYPRODUCTS "${_TARGET_FIXES_FILE}"
-            )
-            add_dependencies(tidy tidy_${TARGET})
+            # clang-tidy gives an error if no source files are passed,
+            # so create the target only if there are any source files.
+            # This is relevant for targets that only contain header files, such as pfFeatureInc.
+            if(NOT "${_TIDY_SOURCES}" STREQUAL "")
```
Hoikas: Could this just be

```suggestion
            if(_TIDY_SOURCES)
```

dgelessus: Yeah, I guess I'm being paranoid about CMake's very liberal definition of "false". We shouldn't have any source file names that could cause problems here.

Hoikas: Hope no one ever makes a source file named `Plasma.cpp-NOTFOUND`

Deledrius: That _should_ be a safe hope.

Hoikas (approved):

PR #1792, opened 2025-09-17
Factoring PLSLoginController out of the window controller

colincornaby:
This separates the business logic of logging in from the UI for logging in. This is part of the iOS changes (in since that platform uses SwiftUI and not AppKit - so it's helpful not to have the business logic in the UI.)

This business logic would be something that would be helpful to have in C++ and shared between the platforms - but this PR merely separates the existing Cocoa business logic out.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginController.h
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginController.h
@@ -40,22 +40,27 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 
-#import <Cocoa/Cocoa.h>
+#import <Foundation/Foundation.h>
 
-#include "plNetClient/plNetClientMgr.h"
-
-#import "PLSLoginController.h"
+#import "pnNetBase/pnNbError.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class PLSLoginWindowController;
+@interface PLSLoginParameters : NSObject
+@property NSString* username;
+@property NSString* password;
+@property BOOL rememberPassword;
 
-@protocol PLSLoginWindowControllerDelegate <NSObject>
-- (void)loginWindowControllerDidLogin:(PLSLoginWindowController*)sender;
+- (void)load;
+- (void)save;
+- (void)makeCurrent;
+
 @end
 
-@interface PLSLoginWindowController : NSWindowController
-@property(weak) id<PLSLoginWindowControllerDelegate> delegate;
+@interface PLSLoginController : NSObject
+
++ (void)attemptLogin:(void (^)(ENetError))completion;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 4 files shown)


Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginController.h
```diff
@@ -0,0 +1,66 @@
...
+
+- (void)load;
+- (void)save;
+- (void)makeCurrent;
+
+@end
+
+@interface PLSLoginController : NSObject
+
++ (void)attemptLogin:(void (^)(ENetError))completion;
```
colincornaby, line 62: This function would be more difficult to replace with a pure C++ one in since Swift does not support C++ lambdas (yet). But PLSLoginController could pretty easily be reduced to a thin wrapper if the underlying code was refactored into a shared C++ class.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginController.mm
```diff
@@ -0,0 +1,183 @@
...
+#define FAKE_PASS_STRING @"********"
+
+static NSOperationQueue* _loginQueue = nil;
+
+@implementation PLSLoginController
+
++ (void)initialize
+{
+    _loginQueue = [NSOperationQueue new];
+    _loginQueue.maxConcurrentOperationCount = 1;
```
colincornaby, line 65: I stared at this for a bit - an operation queue is overkill but _is proper_ in this situation. NetComm is a singleton, so we need a singleton operation queue to safely wrap access. The only part that's really overkill is I think it would be difficult to have several requests going at once that would require an operation queue to control access to the singleton. But it could be helpful when debugging SwiftUI.

Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSLoginController.mm
```diff
@@ -0,0 +1,183 @@
...
+    NSString* serverName = [NSString stringWithSTString:GetServerDisplayName()];
+    NSMutableDictionary* settingsDictionary =
+        [[[NSUserDefaults standardUserDefaults] dictionaryForKey:serverName] mutableCopy];
+    if (!settingsDictionary)
+        settingsDictionary = [NSMutableDictionary dictionary];
+    [settingsDictionary setObject:self.username forKey:@"LastAccountName"];
+    [settingsDictionary setObject:[NSNumber numberWithBool:self.rememberPassword]
+                           forKey:@"RememberPassword"];
+    [[NSUserDefaults standardUserDefaults] setObject:settingsDictionary forKey:serverName];
+    [[NSUserDefaults standardUserDefaults] synchronize];
```
colincornaby, line 118: Oh right - deciding when to save the password talks to user defaults which is correct, but also platform specific. That specifically could be tossed back over to the UI side if there was deeper refactoring to share code here. I don't know if it's worth baking user defaults into Plasma itself.

colincornaby:
@dgelessus - I can't add you as a reviewer but this seems up your alley.

colincornaby:
@dpogue - You might be the most relevant reviewer here. Sorry to put you on the spot with some Obj-C.

PR #1796, opened 2025-09-18
Use `std::move` around `plAgeLinkStruct` and friends

dgelessus:
Plus a couple of minor cleanups. None of this should be particularly exciting.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/pyVaultAgeLinkNode.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/pyVaultAgeLinkNode.cpp
@@ -166,18 +166,17 @@ bool pyVaultAgeLinkNode::HasSpawnPoint( const ST::string & spawnPtName ) const
 
 PyObject * pyVaultAgeLinkNode::GetSpawnPoints() const
 {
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfPython/pyVaultAgeLinkNode.cpp
```diff
@@ -174,8 +174,8 @@ PyObject * pyVaultAgeLinkNode::GetSpawnPoints() const
     plSpawnPointVec points;
     VaultAgeLinkNode access(fNode);
     access.GetSpawnPoints(&points);
-    for (unsigned i = 0; i < points.size(); ++i) {
-        PyObject* elementObj = pySpawnPointInfo::New(points[i]);
```
Hoikas, line 178: I know the point of this PR is to use `std::move()`, but it would be more efficient to leave this an index based loop, create the list with the final size, and insert the elements into place.

dgelessus: Done. Also did the same optimization for 2 other list-returning methods where it was also possible.

PR #1797, opened 2025-09-20
Revive `ptVaultPlayerNode`, sort of

dgelessus:
`ptVaultPlayerNode` was implemented, but didn't get added to the `Plasma` module. I guess no one noticed, because nothing ever called `ptVaultNode.upcastToPlayerNode`? Oops.

(This was the only class with this issue. I checked the other Python bindings and they all get added correctly.)

Many `ptVaultPlayerNode` methods were "global getters" that already existed almost identically in `ptVault`, so I removed the duplicates here.

I wasn't able to fully test these changes - I couldn't find any API for getting a player node `ptVaultNode` from Python... But I did confirm that the `ptVaultPlayerNode` class appears at runtime now.


Scripts/Python/plasma/Plasma.py
```diff
@@ -5383,6 +5383,49 @@ def playerSetOnline(self, state):
...
+        ...
+
+    def isDisabled(self):
+        """Is the avatar disabled?"""
+        ...
+
+    def isExplorer(self):
+        """Returns true for 'explorer', false for 'visitor'."""
+        ...
+
```
Hoikas: I would go ahead and remove this one too. We instituted a change about a decade ago that removed the concept of visitor accounts.

dgelessus: Hmm, this is a thin wrapper around the C++ `VaultPlayerNode`, which still has `Int32_2` mapped to `Explorer`, so I would have kept it... Similarly, `ptVaultAgeInfoNode` has `getCzar`/`getCzarID` even though the "czar" concept doesn't exist anymore.

I'm not hugely opposed to removing it though - not like it will have any practical use in future scripts. (I would keep it on the C++ side though, just so we're aware of the former meaning if we ever want to reuse that field on player nodes.)

dgelessus: ... I noticed now that the czar methods are stubbed out, so perhaps that's not a meaningful precedent, lol. I've removed `isExplorer`/`setExplorer` now.

PR #1800, opened 2025-09-24
Various compiler warning fixes

dpogue:
Fixes various warnings (mostly in CoreLib) from Xcode and Visual Studio around uninitialized local variables, unused values, improper use of `getcwd`, and some (but not all) signed/unsigned and int sizing warnings in hsStream.

This also enables [CMP0156](https://cmake.org/cmake/help/latest/policy/CMP0156.html) which is supposed to deduplicate linked libraries to avoid warnings from Xcode.

I've tried to keep each commit reasonably self-contained, so we can drop anything that we don't have agreement on.

### DIFF
--- a/Sources/Plasma/CoreLib/hsStream.cpp
+++ b/Sources/Plasma/CoreLib/hsStream.cpp
@@ -201,9 +201,8 @@ bool hsStream::IsTokenSeparator(char c)
 
 bool hsStream::GetToken(char *s, uint32_t maxLen, const char beginComment, const char endComment)
 {
-    char c;
-    char endCom;
-        endCom = endComment;
+    char c = 0;
+    char endCom = endComment;
 
     while( true )
     {
@@ -235,9 +234,8 @@ bool hsStream::GetToken(char *s, uint32_t maxLen, const char beginComment, const
 
 bool hsStream::ReadLn(char *s, uint32_t maxLen, const char beginComment, const char endComment)
 {
-    char c;
-    char endCom;
-        endCom = endComment;
+    char c = 0;
+    char endCom = endComment;
 
     while( true )
     {
@@ -270,7 +268,7 @@ bool hsStream::ReadLn(char *s, uint32_t maxLen, const char beginComment, const c
 bool hsStream::ReadLn(ST::string& s, const char beginComment, const char endComment)
 {
     ST::string_stream ss;
-    char c;
+    char c = 0;
     char endCom = endComment;
 
     while (true) {
--- a/Sources/Plasma/CoreLib/hsSystemInfo.cpp
+++ b/Sources/Plasma/CoreLib/hsSystemInfo.cpp
@@ -122,8 +122,8 @@ static inline ST::string IGetAppleCPUVendor()
     ST::string result;
 
 #ifdef HAVE_SYSCTL
-    size_t bufsize = 100;
-    char buffer[bufsize];
+    char buffer[100];
+    size_t bufsize = sizeof(buffer);
 
     if (sysctlbyname("machdep.cpu.brand_string", &buffer, &bufsize, nullptr, 0) == 0) {
         result = buffer;
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -471,7 +471,7 @@ plFileName plFileSystem::GetUserDataPath()
 #if HS_BUILD_FOR_WIN32
         wchar_t path[MAX_PATH];
         if (!SHGetSpecialFolderPathW(nullptr, path, CSIDL_LOCAL_APPDATA, TRUE))
-            return "";
+            return {};
 
         _userData = plFileName::Join(ST::string::from_wchar(path), plProduct::LongName());
 #elif HS_BUILD_FOR_APPLE
@@ -481,6 +481,8 @@ plFileName plFileSystem::GetUserDataPath()
             sysdir_search_path_enumeration_state state;
             state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_APPLICATION_SUPPORT, SYSDIR_DOMAIN_MASK_USER);
             state = sysdir_get_next_search_path_enumeration(state, path);
+            if (state == 0)
+                return {};
         }
         else
 #endif
@@ -490,13 +492,19 @@ plFileName plFileSystem::GetUserDataPath()
             NSSearchPathEnumerationState state;
             state = NSStartSearchPathEnumeration(NSApplicationSupportDirectory, NSUserDomainMask);
             state = NSGetNextSearchPathEnumeration(state, path);
+            if (state == 0)
+                return {};
 
             IGNORE_WARNINGS_END
         }
 
         if (path[0] == '~') {
+            const char* homedir = getenv("HOME");
+            if (!homedir)
+                return {};
+
             char home[PATH_MAX] {};
-            strlcat(home, getenv("HOME"), sizeof(home));
+            strlcat(home, homedir, sizeof(home));
             strlcat(home, &path[1], sizeof(home));
 
             _userData = plFileName::Join(home, plProduct::LongName());
@@ -510,7 +518,7 @@ plFileName plFileSystem::GetUserDataPath()
         } else {
             homedir = getenv("HOME");
             if (!homedir)
-                return "";
+                return {};
 
             _userData = plFileName::Join(homedir, ".config", plProduct::LongName());
         }
--- a/Sources/Plasma/CoreLib/plFileSystem.cpp
+++ b/Sources/Plasma/CoreLib/plFileSystem.cpp
@@ -262,10 +262,10 @@ plFileName plFileSystem::GetCWD()
         cwd = ST::string::from_wchar(cwd_sm);
     }
 #else
-    char *cwd_a = getcwd(nullptr, 0);
-    hsAssert(cwd_a, "Failure to get working directory (unsupported libc?)");
+    char cwd_a[PATH_MAX];
+    const char* res = getcwd(cwd_a, sizeof(cwd_a));
+    hsAssert(res, "Failure to get working directory (unsupported libc?)");
     cwd = cwd_a;
-    free(cwd_a);
 #endif
 
     return cwd;
--- a/Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp
@@ -2237,47 +2237,42 @@ PyObject* cyMisc::RebuildCameraStack(const ST::string& name, const ST::string& a
     if (name.compare("empty") == 0)
         PYTHON_RETURN_NONE;
 
-    if ( !name.empty() )
-    {
-        key=plKeyFinder::Instance().StupidSearch("", "", plSceneObject::Index(), name, false);
-    }
-    if (key == nullptr)
-    {
+    if (!name.empty())
+        key = plKeyFinder::Instance().StupidSearch("", "", plSceneObject::Index(), name, false);
+
+    if (key == nullptr) {
         // try and use this new hack method to find it
-        if (!plVirtualCam1::Instance()->RestoreFromName(name))
-        {
+        if (!plVirtualCam1::Instance()->RestoreFromName(name)) {
             // give up and force built in 3rd person
             plVirtualCam1::Instance()->PushThirdPerson();
             ST::string errmsg = ST::format("Sceneobject {} not found", name);
             PyErr_SetString(PyExc_NameError, errmsg.c_str());
             PYTHON_RETURN_ERROR;
         }
-    }
-    else
-    {
+
+        PYTHON_RETURN_NONE;
+    } else {
         // now we have the scene object, look for it's camera modifier
         const plCameraModifier1* pMod = nullptr;
         plSceneObject* pObj = plSceneObject::ConvertNoRef(key->ObjectIsLoaded());
-        if (pObj)
-        {
-            for (size_t i = 1; i < pObj->GetNumModifiers(); i++)
-            {
+        if (pObj) {
+            for (size_t i = 1; i < pObj->GetNumModifiers(); i++) {
                 pMod = plCameraModifier1::ConvertNoRef(pObj->GetModifier(i));
                 if (pMod)
                     break;
             }
-            if (pMod)
-            {   
+
+            if (pMod) {
                 plVirtualCam1::Instance()->RebuildStack(pMod->GetKey());
                 PYTHON_RETURN_NONE;
             }
         }
+
         plVirtualCam1::Instance()->PushThirdPerson();
         ST::string errmsg = ST::format("Sceneobject {} has no camera modifier", name);
         PyErr_SetString(PyExc_NameError, errmsg.c_str());
         PYTHON_RETURN_ERROR;
     }
-    
 }
 
 void cyMisc::PyClearCameraStack()
--- a/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
+++ b/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
@@ -45,6 +45,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsStream.h"
 
 #include <cstring>
+#include <string_theory/codecs>
 
 struct _InitOpenSSL
 {
@@ -217,23 +218,11 @@ void plMD5Checksum::Finish()
     fContext = nullptr;
 }
 
-const char* plMD5Checksum::GetAsHexString() const
+ST::string plMD5Checksum::GetAsHexString() const
 {
-    const int   kHexStringSize = (2 * MD5_DIGEST_LENGTH) + 1;
-    static char tempString[kHexStringSize];
-
-    int     i;
-    char    *ptr;
-
-
     hsAssert(fValid, "Trying to get string version of invalid checksum");
 
-    for (i = 0, ptr = tempString; i < sizeof(fChecksum); i++, ptr += 2)
-        sprintf(ptr, "%02x", fChecksum[i]);
-
-    *ptr = 0;
-
-    return tempString;
+    return ST::hex_encode(fChecksum, sizeof(fChecksum));
 }
 
 void plMD5Checksum::SetFromHexString(const char* string)
@@ -370,22 +359,11 @@ void plSHAChecksum::Finish()
     fValid = true;
 }
 
-const char* plSHAChecksum::GetAsHexString() const
+ST::string plSHAChecksum::GetAsHexString() const
 {
-    const int kHexStringSize = (2 * SHA_DIGEST_LENGTH) + 1;
-    static char tempString[kHexStringSize];
-
-    int i;
-    char* ptr;
-
     hsAssert(fValid, "Trying to get string version of invalid checksum");
 
-    for (i = 0, ptr = tempString; i < sizeof(fChecksum); i++, ptr += 2)
-        sprintf(ptr, "%02x", fChecksum[i]);
-
-    *ptr = 0;
-
-    return tempString;
+    return ST::hex_encode(fChecksum, sizeof(fChecksum));
 }
 
 void plSHAChecksum::SetFromHexString(const char* string)
@@ -512,22 +490,11 @@ void plSHA1Checksum::Finish()
     fContext = nullptr;
 }
 
-const char* plSHA1Checksum::GetAsHexString() const
+ST::string plSHA1Checksum::GetAsHexString() const
 {
-    const int kHexStringSize = (2 * SHA_DIGEST_LENGTH) + 1;
-    static char tempString[kHexStringSize];
-
-    int i;
-    char* ptr;
-
     hsAssert(fValid, "Trying to get string version of invalid checksum");
 
-    for (i = 0, ptr = tempString; i < sizeof(fChecksum); i++, ptr += 2)
-        sprintf(ptr, "%02x", fChecksum[i]);
-
-    *ptr = 0;
-
-    return tempString;
+    return ST::hex_encode(fChecksum, sizeof(fChecksum));
 }
 
 void plSHA1Checksum::SetFromHexString(const char* string)
--- a/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
+++ b/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
@@ -96,9 +96,7 @@ class plMD5Checksum
         // Backdoor for cached checksums (ie, if you loaded it off disk)
         void SetValue(uint8_t* checksum);
 
-        // Note: GetAsHexString() returns a pointer to a static string;
-        // do not rely on the contents of this string between calls!
-        const char* GetAsHexString() const;
+        ST::string GetAsHexString() const;
         void SetFromHexString(const char* string);
 
         bool operator==(const plMD5Checksum& rhs) const;
@@ -142,9 +140,7 @@ class plSHAChecksum
         // Backdoor for cached checksums (ie, if you loaded it off disk)
         void SetValue(uint8_t* checksum);
 
-        // Note: GetAsHexString() returns a pointer to a static string;
-        // do not rely on the contents of this string between calls!
-        const char* GetAsHexString() const;
+        ST::string GetAsHexString() const;
         void SetFromHexString(const char* string);
 
         bool operator==(const plSHAChecksum& rhs) const;
@@ -182,9 +178,7 @@ class plSHA1Checksum
         // Backdoor for cached checksums (ie, if you loaded it off disk)
         void SetValue(uint8_t* checksum);
 
-        // Note: GetAsHexString() returns a pointer to a static string;
-        // do not rely on the contents of this string between calls!
-        const char* GetAsHexString() const;
+        ST::string GetAsHexString() const;
         void SetFromHexString(const char* string);
 
         bool operator==(const plSHA1Checksum& rhs) const;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,10 +18,17 @@ if(POLICY CMP0127)
     cmake_policy(SET CMP0127 OLD)
 endif()
 
-
 # When cross-compiling, we want CMake to find the host system's Python interpreter
 if(POLICY CMP0190)
     cmake_policy(SET CMP0190 NEW)
+endif()
+
+# Deduplicate libraries on linker commands
+if(POLICY CMP0156)
+    cmake_policy(SET CMP0156 NEW)
+    if(POLICY CMP0179)
+        cmake_policy(SET CMP0179 NEW)
+    endif()
 endif()
 
 include(CheckLanguage)
--- a/Sources/Plasma/CoreLib/hsBounds.cpp
+++ b/Sources/Plasma/CoreLib/hsBounds.cpp
@@ -2203,7 +2203,7 @@ bool hsBounds3Ext::ISectBoxBS(const hsBounds3Ext &other, const hsVector3 &myVel,
     hsAssert(fBounds3Flags & kCenterValid, "Sphere set but not center (BoxBS(vel))");
 
     hsVector3 minAxis;
-    float minDepth;
+    float minDepth = -1.f;
     bool haveAxis = false;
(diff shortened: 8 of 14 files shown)


Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -489,14 +489,18 @@ plFileName plFileSystem::GetUserDataPath()
...
-            state = NSGetNextSearchPathEnumeration(state, path);
+            NSGetNextSearchPathEnumeration(state, path);
 
             IGNORE_WARNINGS_END
         }
 
         if (path[0] == '~') {
+            const char* homedir = getenv("HOME");
+            if (!homedir)
+                return "";
```
Hoikas:
```suggestion
                return {};
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -229,7 +229,7 @@ const char* plMD5Checksum::GetAsHexString() const
     hsAssert(fValid, "Trying to get string version of invalid checksum");
 
     for (i = 0, ptr = tempString; i < sizeof(fChecksum); i++, ptr += 2)
-        sprintf(ptr, "%02x", fChecksum[i]);
+        snprintf(ptr, 3, "%02x", fChecksum[i]);
```
Hoikas: Would it be nicer to remove the loop and convert this to `ST::format()`?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -381,7 +381,7 @@ const char* plSHAChecksum::GetAsHexString() const
...
```
Hoikas: Would it be nicer to remove the loop and convert this to `ST::format()`?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -523,7 +523,7 @@ const char* plSHA1Checksum::GetAsHexString() const
...
```
Hoikas: Would it be nicer to remove the loop and convert this to `ST::format()`?

Sources/Plasma/CoreLib/hsSystemInfo.cpp
```diff
@@ -123,7 +123,7 @@ static inline ST::string IGetAppleCPUVendor()
 
 #ifdef HAVE_SYSCTL
     size_t bufsize = 100;
-    char buffer[bufsize];
+    char buffer[100];
```
Hoikas, line 125:
```suggestion
    char buffer[100];
    size_t bufsize = sizeof(buffer);
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
dpogue: Turns out string_theory has a built-in hexdigest method to turn a buffer of bytes into a hex string, so I've updated it to use that.

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -262,10 +262,10 @@ plFileName plFileSystem::GetCWD()
         cwd = ST::string::from_wchar(cwd_sm);
     }
 #else
-    char *cwd_a = getcwd(nullptr, 0);
+    char cwd_a[1024];
```
dgelessus:
```suggestion
    char cwd_a[PATH_MAX];
```

The super-ideal POSIX solution would be to call `pathconf` to dynamically get `PATH_MAX` for the path `"."`, but we already use the `PATH_MAX` macro constant in a few other places, so for now this is fine I think. (Plus, [according to gnulib](https://www.gnu.org/software/gnulib/manual/html_node/getcwd.html), many `getcwd` implementations don't work properly for paths longer than `PATH_MAX` anyway.)

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -262,10 +262,10 @@ plFileName plFileSystem::GetCWD()
...
+    getcwd(cwd_a, sizeof(cwd_a));
     hsAssert(cwd_a, "Failure to get working directory (unsupported libc?)");
```
dgelessus: The assert needs to be updated so it checks the return value, not the buffer (which is now a local array and thus never `nullptr`).

```suggestion
    const char* res = getcwd(cwd_a, sizeof(cwd_a));
    hsAssert(res, "Failure to get working directory (unsupported libc?)");
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -480,7 +480,7 @@ plFileName plFileSystem::GetUserDataPath()
         if (__builtin_available(macOS 10.12, *)) {
             sysdir_search_path_enumeration_state state;
             state = sysdir_start_search_path_enumeration(SYSDIR_DIRECTORY_APPLICATION_SUPPORT, SYSDIR_DOMAIN_MASK_USER);
-            state = sysdir_get_next_search_path_enumeration(state, path);
+            sysdir_get_next_search_path_enumeration(state, path);
```
dgelessus: What is the issue here? That the value of the second assignment to `state` isn't actually used?

According to the documentation, `sysdir_get_next_search_path_enumeration` returns 0 when it's done iterating. That *hopefully* never happens here on the first iteration, but to be safe, I think this would be the correct fix:

```suggestion
            state = sysdir_get_next_search_path_enumeration(state, path);
            if (state == 0) {
                return {};
            }
```

Sources/Plasma/CoreLib/plFileSystem.cpp
```diff
@@ -489,14 +489,18 @@ plFileName plFileSystem::GetUserDataPath()
 
             NSSearchPathEnumerationState state;
             state = NSStartSearchPathEnumeration(NSApplicationSupportDirectory, NSUserDomainMask);
-            state = NSGetNextSearchPathEnumeration(state, path);
+            NSGetNextSearchPathEnumeration(state, path);
```
dgelessus: As above:

```suggestion
            state = NSGetNextSearchPathEnumeration(state, path);
            if (state == 0) {
                return {};
            }
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
```diff
@@ -96,9 +96,7 @@ class plMD5Checksum
         // Backdoor for cached checksums (ie, if you loaded it off disk)
         void SetValue(uint8_t* checksum);
 
-        // Note: GetAsHexString() returns a pointer to a static string;
-        // do not rely on the contents of this string between calls!
```
dgelessus, line 100: good riddance!

Sources/Plasma/FeatureLib/pfPython/cyMisc.cpp
```diff
@@ -2283,7 +2283,8 @@ PyObject* cyMisc::RebuildCameraStack(const ST::string& name, const ST::string& a
         PyErr_SetString(PyExc_NameError, errmsg.c_str());
         PYTHON_RETURN_ERROR;
     }
-    
+
+    PYTHON_RETURN_NONE;
```
dgelessus: oops, that was me, thanks for catching that!

I think this would fit better into the `if (key == nullptr)` branch specifically, to avoid any accidental "successful" fallthrough.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -86,7 +86,7 @@ uint32_t hsStream::GetSizeLeft()
 
 //////////////////////////////////////////////////////////////////////////////////
 
-uint32_t hsStream::WriteSafeString(const ST::string &string)
+size_t hsStream::WriteSafeString(const ST::string &string)
```
dgelessus: GitHub doesn't let me comment on it, but there are `uint32_t i` loop variables in `WriteSafeString` and `WriteSafeWString`, which should also be changed to `size_t`.

I'm pretty sure that `WriteSafeWString` is also missing a `return i;` somewhere - currently it always returns 0.

Perhaps we should just make these methods return `void`? As far as I can tell, the return value is just the size of the string in the respective encoding, which you can also get from the string itself if you really need it. I couldn't find any code that uses the return value currently.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -440,7 +438,7 @@ bool hsUNIXStream::Open(const plFileName &name, const char *mode)
     return (fRef) ? true : false;
 }
 
-uint32_t hsUNIXStream::Read(uint32_t bytes,  void* buffer)
+size_t hsUNIXStream::Read(size_t bytes,  void* buffer)
```
dgelessus:
```suggestion
size_t hsUNIXStream::Read(size_t bytes, void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -558,20 +556,20 @@ bool  plReadOnlySubStream::AtEnd()
     return false;
 }
 
-uint32_t  plReadOnlySubStream::Read(uint32_t byteCount, void* buffer)
+size_t  plReadOnlySubStream::Read(size_t byteCount, void* buffer)
```
dgelessus:
```suggestion
size_t plReadOnlySubStream::Read(size_t byteCount, void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -558,20 +556,20 @@ bool  plReadOnlySubStream::AtEnd()
...
     }
 
-    uint32_t read = fBase->Read( byteCount, buffer );
+    size_t read = fBase->Read( byteCount, buffer );
     IFixPosition();
     return read;
 }
 
-uint32_t  plReadOnlySubStream::Write(uint32_t byteCount, const void* buffer)
+size_t  plReadOnlySubStream::Write(size_t byteCount, const void* buffer)
```
dgelessus:
```suggestion
size_t plReadOnlySubStream::Write(size_t byteCount, const void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -612,7 +610,7 @@ bool hsRAMStream::AtEnd()
     return (fPosition >= fVector.size());
 }
 
-uint32_t hsRAMStream::Read(uint32_t byteCount, void * buffer)
+size_t hsRAMStream::Read(size_t byteCount, void * buffer)
```
dgelessus:
```suggestion
size_t hsRAMStream::Read(size_t byteCount, void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -700,13 +698,13 @@ bool hsNullStream::AtEnd()
     return true;
 }
 
-uint32_t hsNullStream::Read(uint32_t byteCount, void * buffer)
+size_t hsNullStream::Read(size_t byteCount, void * buffer)
```
dgelessus:
```suggestion
size_t hsNullStream::Read(size_t byteCount, void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -859,13 +857,13 @@ hsQueueStream::~hsQueueStream()
     delete [] fQueue;
 }
 
-uint32_t hsQueueStream::Read(uint32_t byteCount, void * buffer)
+size_t hsQueueStream::Read(size_t byteCount, void * buffer)
```
dgelessus:
```suggestion
size_t hsQueueStream::Read(size_t byteCount, void* buffer)
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -859,13 +857,13 @@ hsQueueStream::~hsQueueStream()
...
 {
     hsAssert(fWriteCursor < fSize,"hsQueueStream: WriteCursor out of range.");
     hsAssert(fReadCursor < fSize,"hsQueueStream: ReadCursor out of range.");
 
-    int32_t limit, length, total;
-    
+    size_t limit, length, total;
```
dgelessus: Can we move these declarations down to their first assignments?

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -885,12 +883,12 @@ uint32_t hsQueueStream::Read(uint32_t byteCount, void * buffer)
...
-uint32_t hsQueueStream::Write(uint32_t byteCount, const void* buffer)
+size_t hsQueueStream::Write(size_t byteCount, const void* buffer)
 {
     hsAssert(fWriteCursor < fSize,"hsQueueStream: WriteCursor out of range.");
     hsAssert(fReadCursor < fSize,"hsQueueStream: ReadCursor out of range.");
 
-    int32_t length;
+    size_t length;
 
     length = std::min(fSize-fWriteCursor, byteCount);
```
dgelessus: As above:

```suggestion
    size_t length = std::min(fSize-fWriteCursor, byteCount);
```

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -916,16 +914,16 @@ uint32_t hsQueueStream::Write(uint32_t byteCount, const void* buffer)
 
 void hsQueueStream::Skip(uint32_t deltaByteCount)
 {
-    int32_t limit, length;
+    size_t limit, length;
```
dgelessus: As above, should be moved to the first assignments.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -916,16 +914,16 @@ uint32_t hsQueueStream::Write(uint32_t byteCount, const void* buffer)
...
     
     limit = fWriteCursor >= fReadCursor ? fWriteCursor : fSize;
-    length = std::min(limit-fReadCursor, deltaByteCount);
+    length = std::min(limit-fReadCursor, static_cast<size_t>(deltaByteCount));
```
dgelessus: I wonder why this cast is necessary? `uint32_t` fits into `size_t` on any system we care about.

Sources/Plasma/CoreLib/hsStream.cpp
```diff
@@ -1157,15 +1155,15 @@ void hsBufferedStream::Skip(uint32_t delta)
...
 
         // We've got data in the buffer, see if we can just skip in that
         if (fBufferLen > 0)
         {
-            int32_t newBufferPos = int32_t(fPosition % kBufferSize) + int32_t(delta);
+            size_t newBufferPos = (fPosition % kBufferSize) + delta;
 
             // If we skipped outside of our buffer, invalidate it
-            if (newBufferPos < 0 || uint32_t(newBufferPos) >= fBufferLen)
+            if (newBufferPos >= fBufferLen)
```
dgelessus: This isn't an appropriate use of `size_t`, because it's a position/offset within a file, not a memory buffer size.

Because we're using standard C `fseek` here, technically the correct type would be `long` (ugh!). But I think *for now* we should stick with `uint32_t` for the offsets, because that's what all the other stream code does. Fixing that limitation is a larger issue.

If the previous code here is giving you warnings, try changing `int32_t newBufferPos` to `uint32_t`. I don't know why that variable is signed - all the other values here are unsigned.

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -63,8 +63,8 @@ enum {
     virtual     ~hsStream() { }
 
     virtual bool      AtEnd() = 0;
-    virtual uint32_t  Read(uint32_t byteCount, void * buffer) = 0;
-    virtual uint32_t  Write(uint32_t byteCount, const void* buffer) = 0;
+    virtual size_t    Read(size_t byteCount, void * buffer) = 0;
```
dgelessus:
```suggestion
    virtual size_t    Read(size_t byteCount, void* buffer) = 0;
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -222,8 +222,8 @@ class hsRAMStream : public hsStream
     hsRAMStream(uint32_t chunkSize) { fVector.reserve(chunkSize); }
 
     bool      AtEnd() override;
-    uint32_t  Read(uint32_t byteCount, void * buffer) override;
-    uint32_t  Write(uint32_t byteCount, const void* buffer) override;
+    size_t    Read(size_t byteCount, void * buffer) override;
```
dgelessus:
```suggestion
    size_t    Read(size_t byteCount, void* buffer) override;
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -246,8 +246,8 @@ class hsRAMStream : public hsStream
 class hsNullStream : public hsStream {
 public:
     bool AtEnd() override;
-    uint32_t  Read(uint32_t byteCount, void * buffer) override;  // throws exception
-    uint32_t  Write(uint32_t byteCount, const void* buffer) override;
+    size_t    Read(size_t byteCount, void * buffer) override;  // throws exception
```
dgelessus:
```suggestion
    size_t    Read(size_t byteCount, void* buffer) override;  // throws exception
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -271,8 +271,8 @@ class hsReadOnlyStream : public hsStream {
     {}
 
     bool      AtEnd() override;
-    uint32_t  Read(uint32_t byteCount, void * buffer) override;
-    uint32_t  Write(uint32_t byteCount, const void* buffer) override;    // throws exception
+    size_t    Read(size_t byteCount, void * buffer) override;
```
dgelessus:
```suggestion
    size_t    Read(size_t byteCount, void* buffer) override;
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -301,8 +301,8 @@ class hsWriteOnlyStream : public hsStream {
     hsWriteOnlyStream& operator=(hsWriteOnlyStream&& other) = delete;
 
     bool      AtEnd() override;
-    uint32_t  Read(uint32_t byteCount, void * buffer) override;  // throws exception
-    uint32_t  Write(uint32_t byteCount, const void* buffer) override;
+    size_t    Read(size_t byteCount, void * buffer) override;  // throws exception
```
dgelessus:
```suggestion
    size_t    Read(size_t byteCount, void* buffer) override;  // throws exception
```

Sources/Plasma/CoreLib/hsStream.h
```diff
@@ -315,32 +315,32 @@ class hsWriteOnlyStream : public hsStream {
...
     hsQueueStream(const hsQueueStream& other) = delete;
     hsQueueStream(hsQueueStream&& other) = delete;
     ~hsQueueStream();
 
     const hsQueueStream& operator=(const hsQueueStream& other) = delete;
     hsQueueStream& operator=(hsQueueStream&& other) = delete;
 
-    uint32_t  Read(uint32_t byteCount, void * buffer) override;
-    uint32_t  Write(uint32_t byteCount, const void* buffer) override;
+    size_t    Read(size_t byteCount, void * buffer) override;
```
dgelessus:
```suggestion
    size_t    Read(size_t byteCount, void* buffer) override;
```

Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -269,26 +269,26 @@ uint32_t plEncryptedStream::GetEOF()
...
-    uint32_t startAmt = (startChunkPos != 0) ? std::min(kEncryptChunkSize - startChunkPos, bytes) : 0;
+    size_t startAmt = (startChunkPos != 0) ? std::min(kEncryptChunkSize - startChunkPos, bytes) : 0;
 
-    uint32_t totalNumRead = IRead(bytes, buffer);
+    size_t totalNumRead = IRead(bytes, buffer);
 
-    uint32_t numMidChunks = (totalNumRead - startAmt) / kEncryptChunkSize;
-    uint32_t endAmt = (totalNumRead - startAmt) % kEncryptChunkSize;
+    size_t numMidChunks = (totalNumRead - startAmt) / kEncryptChunkSize;
+    size_t endAmt = (totalNumRead - startAmt) % kEncryptChunkSize;
```
dgelessus: As above, anything that is an offset/position within a stream (rather than a buffer size) should stay as `uint32_t` for now.

(`plEncryptedStream` will always be limited to 32-bit sizes anyway, because of the file format.)

Sources/Plasma/PubUtilLib/plFile/plSecureStream.cpp
```diff
@@ -409,26 +409,26 @@ uint32_t plSecureStream::GetEOF()
...
-    uint32_t startAmt = (startChunkPos != 0) ? std::min(kEncryptChunkSize - startChunkPos, bytes) : 0;
+    size_t startAmt = (startChunkPos != 0) ? std::min(kEncryptChunkSize - startChunkPos, bytes) : 0;
 
-    uint32_t totalNumRead = IRead(bytes, buffer);
+    size_t totalNumRead = IRead(bytes, buffer);
 
-    uint32_t numMidChunks = (totalNumRead - startAmt) / kEncryptChunkSize;
-    uint32_t endAmt = (totalNumRead - startAmt) % kEncryptChunkSize;
+    size_t numMidChunks = (totalNumRead - startAmt) / kEncryptChunkSize;
+    size_t endAmt = (totalNumRead - startAmt) % kEncryptChunkSize;
```
dgelessus: Same story as for `plEncryptedStream` - offsets should stay `uint32_t` for now.

Sources/Tools/plPythonPack/main.cpp
```diff
@@ -101,18 +101,18 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
     ST::printf(out, "== Packing {}, ", fileName);
 
     pyStream.FastFwd();
-    uint32_t pyFileSize = pyStream.GetPosition();
+    size_t pyFileSize = pyStream.GetPosition();
```
dgelessus:
```suggestion
    uint32_t pyFileSize = pyStream.GetPosition();
```

Sources/Tools/plPythonPack/main.cpp
```diff
@@ -101,18 +101,18 @@ void WritePythonFile(const plFileName &fileName, const plFileName &path, hsStrea
...
     pyStream.Rewind();
 
     glueStream.FastFwd();
-    uint32_t glueFileSize = glueStream.GetPosition();
+    size_t glueFileSize = glueStream.GetPosition();
```
dgelessus:
```suggestion
    uint32_t glueFileSize = glueStream.GetPosition();
```

Sources/Plasma/NucleusLib/pnSceneObject/plSceneObject.cpp
```diff
@@ -133,7 +133,7 @@ void plSceneObject::Read(hsStream* stream, hsResMgr* mgr)
     fModifiers.resize(nOldMods + nNewMods);    // reserve space for new modifiers+existing modifiers
     for (size_t i = nOldMods; i < nOldMods + nNewMods; i++)
     {
-        refMsg = new plObjRefMsg(GetKey(), plRefMsg::kOnCreate, i, plObjRefMsg::kModifier);
+        refMsg = new plObjRefMsg(GetKey(), plRefMsg::kOnCreate, static_cast<int8_t>(i), plObjRefMsg::kModifier);
```
dgelessus: Feels like it would be better to allow larger indices in `plObjRefMsg` itself?

Sources/Plasma/PubUtilLib/plScene/plSceneNode.cpp
```diff
@@ -94,7 +94,7 @@ void plSceneNode::Read(hsStream* s, hsResMgr* mgr)
     hsKeyedObject::Read(s, mgr);
 
     uint32_t n;
-    int i;
+    uint32_t i;
```
dgelessus: Could be made a local variable in each `for` loop (and also `n` could be merged with its first assignment).

dgelessus (changes requested):
This generally looks good.

I'm a little cautious about `size_t` in the context of file IO, because when done wrong, it can introduce problems with large files that go unnoticed in 64-bit builds and only affect 32-bit builds.

This isn't a realistic problem for us right now, because so much of the `hsStream` code hardcodes file sizes/offsets as `uint32_t` (and fixing that isn't the topic of this PR). But to avoid future issues, we should use `size_t` in IO code only for sizes/offsets of *memory* buffers (as for `Read`/`Write` for example). Sizes/offsets within files/streams should *not* be `size_t`. For now, unfortunately, the appropriate type is `uint32_t`, but at some point we should add an equivalent of POSIX `off_t` for file sizes/offsets. I've commented on the specific cases where I think `size_t` *isn't* appropriate.

Sources/Plasma/NucleusLib/pnSceneObject/plSceneObject.cpp
dpogue: We can't change this without changing the stream layout for the message. There really shouldn't be any RefMsgs serialized anywhere, but it's not impossible, so we need to be careful about that.

Hoikas: I think that we have, in a few places, widened the internal storage and added something like `hsAssert(fValue <=std::numeric_limits<oldType_t>::max(), "overflow");` to the serialization routine [(example)](https://github.com/H-uru/Plasma/blob/84ad757fba7c57d61131de0678abc8cc23249695/Sources/Plasma/PubUtilLib/plScene/plOccluder.cpp#L308). It might not be a bad idea to do that.

`plAttachMsg` is a `plRefMsg` that can go over the wire, btw. It's pretty trivial to send those from Python. Thankfully, `plAttachMsg` is not a subclass of `plObjRefMsg`.

dgelessus:
Looking at this again, would it make sense to split the integer size and signedness commits into a separate PR? All the other commits look like straightforward, independent fixes - they don't need to be blocked on figuring out the details of the integer types.

In my last review, there's only one remaining comment that's *not* about integer types, which is that the result array for `getcwd` should use `PATH_MAX` as the length, instead of hardcoding `1024`.

dpogue:
> would it make sense to split the integer size and signedness commits into a separate PR?

yeah, I've pulled those out onto an `int-warnings` branch in my fork for now. To be revisited at some point.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -217,23 +218,11 @@ void plMD5Checksum::Finish()
...
-
     hsAssert(fValid, "Trying to get string version of invalid checksum");
 
-    for (i = 0, ptr = tempString; i < sizeof(fChecksum); i++, ptr += 2)
-        sprintf(ptr, "%02x", fChecksum[i]);
-
-    *ptr = 0;
-
-    return tempString;
+    return ST::hex_encode(fChecksum, sizeof(fChecksum)).c_str();
```
dgelessus:
```suggestion
    return ST::hex_encode(fChecksum, sizeof(fChecksum));
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -370,22 +359,11 @@ void plSHAChecksum::Finish()
...
```
dgelessus:
```suggestion
    return ST::hex_encode(fChecksum, sizeof(fChecksum));
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -512,22 +490,11 @@ void plSHA1Checksum::Finish()
...
```
dgelessus:
```suggestion
    return ST::hex_encode(fChecksum, sizeof(fChecksum));
```

dgelessus (changes requested):
I noticed a few unnecessary `.c_str()` calls that I missed before. Otherwise, everything looks good now. Sorry for the late response!

dpogue:
Good catch, updated and rebased

dgelessus (approved):

PR #1801, opened 2025-09-27
Don't enable Metal pipeline if we don't have a Metal compiler

dpogue:
It is possible (in certain contrived situations) to have the Metal framework headers but not a working Metal compiler (and vice versa, since the Metal compiler is actually available for Windows as well).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -322,7 +322,7 @@ jobs:
             -DPLASMA_BUILD_TESTS=ON \
             -DPLASMA_BUILD_TOOLS=ON \
             -DPLASMA_EXTERNAL_RELEASE=${{ matrix.cfg.external }} \
-            -DQt5_ROOT=$(brew --prefix qt@${{ matrix.platform.qt-version }}) \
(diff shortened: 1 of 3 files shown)


dpogue:
My original use case here was different (cross-compiling with clang on Linux but targeting Darwin), but I think there's a possibility of this happening with Xcode 26 as well, since the Metal Toolchain is now a separate add-on that needs to be explicitly installed through Xcode (or using `xcodebuild -downloadComponent MetalToolchain`).

It is pre-installed on the macos-26 GitHub runners, but not be for people building on their own machines. In that case, I don't know that just silently disabling the Metal pipeline is the right approach.

colincornaby:
> My original use case here was different (cross-compiling with clang on Linux but targeting Darwin), but I think there's a possibility of this happening with Xcode 26 as well, since the Metal Toolchain is now a separate add-on that needs to be explicitly installed through Xcode (or using `xcodebuild -downloadComponent MetalToolchain`).

Noted. I'm not sure what we should do about this yet - I think the current behavior is fine for now. In the future this should be documented - or we could force it back on in the presence of a macOS SDK higher than a certain version.

colincornaby (approved):
My CMake is somewhat unsteady so I'm approving based on a reading with a working understanding of CMake and what the improvement is doing.

PR #1805, opened 2025-11-09
Fixing Metal validation error around lighting

colincornaby:
The Metal validator has started complaining that we’re passing a buffer that’s smaller than the argument size. This validation is tripped even if I change the argument in the shader to a pointer.

The idea was that I could pass a partial structure of lights to save on bandwidth costs - but Metal no longer seems to like that. I’ll have to pass the full array of lights whenever they are updated.

This came up after the merge of the Metal per pixel lighting code. I'm not sure why the validator has decided to complain now - but I've done an Xcode upgrade since then.


dpogue (commented):
Are the unused portions of fLights zeroed out or otherwise known to be ignored by the shader, or are we potentially passing uninitialized/stale data to the shader here?

PR #1806, opened 2025-11-13
Fixing plTextFont on Mac

colincornaby:
Fixing a myriad of issues that affected font rendering (and in some cases all rendering):
- The Mac client, on windowed resolution change, was setting the window height using a method that includied the title bar. Changed to a method that doesn’t include the title bar. This caused a compressed output image in windowed mode that visually distorted fonts.
- Deleting dead didChangeOutputSize callback. Was used for live resizing of the client window.
- Text shader now draws in center of pixels to prevent blurriness.
- plTextFont now queries the DPI on Mac. Note: This doesn’t respond to DPI changes due to the window moving right now. But Plasma has a lot of known issues with multi display systems.

This should clear up font rendering issues on macOS.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -208,7 +208,7 @@ void plClient::IResizeNativeDisplayDevice(int width, int height, bool windowed)
 {
     // Client load is threaded so we might not receive this on the main thread
     // We _need_ to do this on the main thread becuase it has AppKit calls.
-    if(NSThread.currentThread != NSThread.mainThread) {
+    if (NSThread.currentThread != NSThread.mainThread) {
         // Even though I used NSThread above this is cleaner to send to
         // the main thread in libDispatch
         dispatch_sync(dispatch_get_main_queue(), ^{


Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
```diff
@@ -149,7 +149,10 @@ uint16_t  *plTextFont::IInitFontTexture()
     CFRelease(fontURL);
     CFRelease(fileSystemPath);
     
-    FT_UInt freeTypeResolution = 192;
+    // Windows uses a base DPI of 96, while Mac uses a DPI of 72.
+    // To make sure everything aligns match Windows.
+    // FIXME: We don't update when the DPI changes at runtime
```
dpogue, line 154: Presumably this would involve something (the pipeline?) registering to handle the plDisplayScaleChangedMsg message and passing that to plTextFont?

dpogue (approved):

Sources/Plasma/PubUtilLib/plPipeline/plTextFont.cpp
colincornaby: Yeah. I'm not sure if the best move would be to destroy the font and recreate it each time the DPI changes or cache so if the window moves back and forth there are multiple fonts already stored.

Plasma really doesn't do well with the window being moved between multiple displays right now. The Windows version only queries the main display for properties like renderers and resolutions. And I don't think the other platforms handle that at all.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -206,28 +206,41 @@ @interface AppDelegate : NSWindowController <NSApplicationDelegate,
...
-        // Toggle full screen if we need to
-        if (((appDelegate.window.styleMask & NSWindowStyleMaskFullScreen) > 0) == windowed) {
-            [appDelegate.window toggleFullScreen:nil];
-        }
-        auto* msg = new plDisplayScaleChangedMsg(appDelegate.window.backingScaleFactor);
-        msg->Send();
-    });
+    // Client load is threaded so we might not receive this on the main thread
+    // We _need_ to do this on the main thread becuase it has AppKit calls.
+    if(NSThread.currentThread != NSThread.mainThread) {
```
Hoikas:
```suggestion
    if (NSThread.currentThread != NSThread.mainThread) {
```

PR #1808, opened 2025-11-16
Metal bump mapping

colincornaby:
This PR extends the per pixel lighting pull request that was merged earlier with bump mapping for Metal. This handles the production bumpmaps - it does not handle BUMPENVMAPLUMINANCE which was never used.

Plasma actually uses normal mapping assets, which will be relevant. There is a good tutorial on normal mapping here that describes important concepts like tangent space:
https://learnopengl.com/Advanced-Lighting/Normal-Mapping

Normal maps in Plasma do not work like normal maps.  There appears to have been some back and forth with the artists about this. The key differences are:

- Normal maps are part of lighting calculations - and they reduce contributions of lights that aren't facing the material described by the map. D3D Plasma does not work this way. It renders the material normally with all lighting contributions. It then loads the normal map and then does a second pass doing a lighting highlight on the material with the map. This means normal maps in Plasma can only add light, they can't subtract it.
- Plasma zeros the blue component of the normal map/N component of the tangent space vector. The reasoning is described here: https://github.com/H-uru/Plasma/blob/de71f00f5e90c1de83e5f59a37449acadf8fcad8/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp#L6688-L6693
Because the normal map is applied as a glow - the blue channel of the map would cause the top of anything "flat" to glow. The N component is zeroed so a glow is only applied in the T and B directions so it's more subtle.

Plasma encodes the bump map as four layers.

- Two layers contain uv coordinates that represent the T and B vectors.
- One layer contain an operation the loads the normal into the computation for the N vector
- One layer contains the normal map.

The T/B/N layers can probably be encoded in any order - but the normal map layer likely has to always be last.

I'm attaching the D3D normal mapping state captured from the PIX debugger: [State12178.htm](https://github.com/user-attachments/files/23566523/State12178.htm). This document describes the configuration of texturing stages that are set for normal mapping. D3D loads passthrough textures so the UV coordinates from the T and B layers - and the normal value - can be loaded as colors into the texture pipeline. Stages 0-3 load the T/B/N. Stage 3 dots it against the normal map. And Stage 4 multiplies it by the diffuse color from vertex lighting. UV matrices are loaded to perform the transform of the T/B/N vectors based on light direction. This computation is described here:
https://github.com/H-uru/Plasma/blob/de71f00f5e90c1de83e5f59a37449acadf8fcad8/Sources/Plasma/FeatureLib/pfDXPipeline/plDXPipeline.cpp#L6537-L6597

The Metal version has a few key differences:
- The D3D version can only handle one light direction - so all light directions are averaged together by the D3D pipeline. The Metal version supports multiple lights so it can calculate the contribution of each light individually.
- D3D uses vertex lighting but calculates the bump per pixel. Metal calculates both the light and the bump per pixel. Bump mapped materials will go through the per pixel lighting path.
- The Metal version skips the entire step of constructing transformation matrices for the light direction in the pipeline. The shader calculating the contribution of each light individually means we don't need to pre-do this work.

These enhancements to the Metal version should be visually subtle and not distract from the original artists intent. They make it easier to implement bump mapping and open the door to future improvements.

The Metal version also includes a bool constant in the shader: `bumpMapIsAdditive`. This controls the emulation of Plasma's bump mapping behavior. When `bumpMapIsAdditive` is true the normal map is treated as a second glow pass - emulating the D3D pipeline. When `bumpMapIsAdditive` is set to false - the normal map texture is factored into the normal lighting process instead of being applied as a glow. This includes using the normal map to reduce the contributions of lights that are not facing the normal map direction. This mode performs accurate normal mapping but is visually distinct from what shipped. `bumpMapIsAdditive` is set to true - but the full normal mapping path has been left in the code for possible future use.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
@@ -1720,7 +1720,7 @@ void plMetalPipeline::IBindLights()
 
     if (fLightsDirty) {
         if (fLightingPerPixel) {
-            fDevice.CurrentRenderCommandEncoder()->setFragmentBytes(fLights->data(), sizeof(plMetalShaderActiveLight) * fLights->size(), ShaderActiveLights);
+            fDevice.CurrentRenderCommandEncoder()->setFragmentBytes(fLights->data(), bindSize, ShaderActiveLights);
             fDevice.CurrentRenderCommandEncoder()->setFragmentBytes(&lightSize, sizeof(uint), ShaderActiveLightCount);
         } else {
             fDevice.CurrentRenderCommandEncoder()->setVertexBytes(fLights->data(), bindSize, ShaderActiveLights);
@@ -2506,6 +2506,7 @@ void plMetalPipeline::ISetEnablePerPixelLighting(const bool enable)
 {
     if (fLightingPerPixel != enable) {
         fLightingPerPixel = enable;
+        fLightsDirty = true;
 
         // These states need to be reset for a change in lighting technique
         fState.fBoundMaterialProperties.reset();
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/ShaderTypes.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/ShaderTypes.h
@@ -78,9 +78,10 @@ enum plMetalShaderArgument
     /// Texture is a legacy argument for the simpler plate shader
     FragmentShaderArgumentTexture = 1,
     /// Fragment uniforms
-    FragmentShaderArgumentShadowCastUniforms = 4,
+    FragmentShaderArgumentShadowCastUniforms = 5,
+    // FIXME: Plate shader is using a hardcoded argument in slot 6
     /// Legacy argument buffer
-    FragmentShaderArgumentUniforms = 5,
+    FragmentShaderArgumentUniforms = 7,
     /// Layer index of alpha for shadow fragment shader
     FragmentShaderArgumentShadowCastAlphaSrc = 8,
     /// Material properties for vertex lighting


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -277,6 +280,12 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
         (&out.texCoord1)[layer] = uniforms.sampleLocation(layer, &in.texCoord1, cameraSpaceNormal, vCamPosition);
     }
 
+    
+    if(bumpMap) {
+        out.T = normalize( uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo.dTangentUIndex], 0.f)). xyz;
+        out.B = normalize( uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo.dTangentVIndex], 0.f)). xyz;
```
colincornaby: `texCoord1` is a pointer to the beginning of the array. I should have called it `texCoord0` - I have regrets. Maybe something for a future PR.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -198,6 +206,13 @@ void plMetalMaterialShaderRef::EncodeArguments(MTL::RenderCommandEncoder* encode
             return postEncodeTransform(layer, index);
         }
     );
+    
+    if (fBumps[pass].has_value())
+    {
+        plLayerInterface *bumpLayer = fMaterial->GetLayer(fMaterial->GetNumLayers()-1);
+        auto texture = (plMetalTextureRef*)bumpLayer->GetTexture()->GetDeviceRef();
+        encoder->setFragmentTexture(texture->fTexture, 7);
```
colincornaby, line 213: I might need to look into this. Metal supports 8 texture passes - so I'm shoving the normal map into the last slot for convenience. But there are comments that imply multiple texture maps are possible - so I might need something more nuanced. Also hardcoded number bad.

dpogue: DirectX renders the object multiple times to achieve multiple passes, and always renders the bump mapping as its own pass. That probably only works because it's just adding light.

colincornaby: Yeah, I was thinking about that. If bump maps could be unlimited I would have to break each out into its own pass.

There is a comment somewhere else that hints multiple bump maps could be used for water. I haven't seen that used anywhere, but I guess I should could against that.

I'd probably keep support for single pass normal maps in the shader in since we've discussed having that available - but I'd have to separate out the glow maps a bit more to get them into proper additional stages.

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -277,6 +280,12 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
...
+    }
```
Hoikas:
```suggestion
    if (bumpMap) {
        out.T = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo.dTangentUIndex], 0.f)). xyz;
        out.B = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo.dTangentVIndex], 0.f)). xyz;
    }
```

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -442,9 +451,43 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
...
+     of the normal map file.
+     - If bumpMapIsAdditive is false, full normal mapping is performed on a per
+     pixel basis. This is not how the engine shipped and will produce different
+     results than the original age artist may have expected. But it will be
+     proper normal mapping.
+     */
+    constexpr bool bumpMapIsAdditive = true;
+    constexpr bool performBaseLighting = bumpMapIsAdditive || !bumpMap;
+    
+    if(performBaseLighting) {
```
Hoikas:
```suggestion
    if (performBaseLighting) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -442,9 +451,43 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
...
+        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lights, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
+    }
+    
+    if(bumpMap) {
```
Hoikas:
```suggestion
    if (bumpMap) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -442,9 +451,43 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
...
+        float3 sampleCoord = in.texCoord1;
+        half3 bumpNormal = bumpTexture.sample(fragmentShaderArgs.samplers, sampleCoord.xy).rgb;
+        
+        bumpNormal -= 0.5f;
+        bumpNormal *= 2.f;
+        
+        float3x3 TBN = float3x3(in.T, in.B, in.normal);
+        bumpNormal = half3(normalize(TBN * float3(bumpNormal)));
+        
+        if(performBaseLighting) {
```
Hoikas:
```suggestion
        if (performBaseLighting) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -163,6 +163,14 @@ void plMetalMaterialShaderRef::FastEncodeArguments(MTL::RenderCommandEncoder* en
     }
 
     encoder->setFragmentBuffer(fPassArgumentBuffers[pass], 0, FragmentShaderArgumentUniforms);
+    
+    if (fBumps[pass].has_value())
+    {
+        plLayerInterface *bumpLayer = fMaterial->GetLayer(fMaterial->GetNumLayers()-1);
```
Hoikas:
```suggestion
        plLayerInterface* bumpLayer = fMaterial->GetLayer(fMaterial->GetNumLayers()-1);
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -198,6 +206,13 @@ void plMetalMaterialShaderRef::EncodeArguments(MTL::RenderCommandEncoder* encode
             return postEncodeTransform(layer, index);
         }
     );
+    
+    if (fBumps[pass].has_value())
+    {
```
Hoikas:
```suggestion
    if (fBumps[pass].has_value()) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -198,6 +206,13 @@ void plMetalMaterialShaderRef::EncodeArguments(MTL::RenderCommandEncoder* encode
...
+        plLayerInterface *bumpLayer = fMaterial->GetLayer(fMaterial->GetNumLayers()-1);
```
Hoikas:
```suggestion
        plLayerInterface* bumpLayer = fMaterial->GetLayer(fMaterial->GetNumLayers()-1);
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -293,6 +307,18 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
 
         fPassIndices.push_back(currLayer);
         fPassLengths.push_back(j - currLayer);
+        
+        auto bumpMap = IEatBumpmapLayers(j);
+        fBumps.push_back(bumpMap);
+        
+        if (bumpMap.has_value())
+        {
```
Hoikas:
```suggestion
        if (bumpMap.has_value()) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
     }
 }
 
+std::optional<plMetalBumpMapping> plMetalMaterialShaderRef::IEatBumpmapLayers( uint32_t& layerIdx )
```
Hoikas:
```suggestion
std::optional<plMetalBumpMapping> plMetalMaterialShaderRef::IEatBumpmapLayers(uint32_t& layerIdx)
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+{
+    
+    std::optional<plMetalBumpMapping> bumpMapping;
+    
+    // If there aren't enough layers left to support a bump map, return
+    if (!(layerIdx + 3 < fMaterial->GetNumLayers()))
+    {
```
Hoikas:
```suggestion
    if (!(layerIdx + 3 < fMaterial->GetNumLayers())) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        return bumpMapping;
+    }
+    
+    // Does the next layer imply a bump map?
+    if (!(fMaterial->GetLayer(layerIdx)->GetMiscFlags() & hsGMatState::kMiscBumpChans))
+    {
```
Hoikas:
```suggestion
    if (!(fMaterial->GetLayer(layerIdx)->GetMiscFlags() & hsGMatState::kMiscBumpChans)) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        return bumpMapping;
+    }
+
+    bumpMapping = plMetalBumpMapping();
+    // We have a bump map and it should occupy the next four layers
+    for (size_t bumpLayer = 0; bumpLayer < 4; bumpLayer++)
+    {
```
Hoikas:
```suggestion
    for (size_t bumpLayer = 0; bumpLayer < 4; bumpLayer++) {
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        plLayerInterface* layer = fMaterial->GetLayer(layerIdx + bumpLayer);
+        uint32_t miscFlags = layer->GetMiscFlags();
+        switch( miscFlags & hsGMatState::kMiscBumpChans )
```
Hoikas:
```suggestion
        switch (miscFlags & hsGMatState::kMiscBumpChans)
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+            bumpMapping.value().dTangentUIndex = layer->GetUVWSrc();
+        case hsGMatState::kMiscBumpDv:
+            bumpMapping.value().dTangentVIndex = layer->GetUVWSrc();
+        case hsGMatState::kMiscBumpDw:
+        default:
+            break;
+        }
+    }
+    
+    layerIdx+=4;
```
Hoikas:
```suggestion
    layerIdx += 4;
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
     }
 }
 
+std::optional<plMetalBumpMapping> plMetalMaterialShaderRef::IEatBumpmapLayers( uint32_t& layerIdx )
+{
+    
+    std::optional<plMetalBumpMapping> bumpMapping;
+
```
Hoikas:
```suggestion
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
 
+std::optional<plMetalBumpMapping> plMetalMaterialShaderRef::IEatBumpmapLayers( uint32_t& layerIdx )
+{
+    
+    std::optional<plMetalBumpMapping> bumpMapping;
+    
+    // If there aren't enough layers left to support a bump map, return
+    if (!(layerIdx + 3 < fMaterial->GetNumLayers()))
+    {
+        return bumpMapping;
```
Hoikas:
```suggestion
        return std::nullopt;
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +327,45 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+    }
+    
+    // Does the next layer imply a bump map?
+    if (!(fMaterial->GetLayer(layerIdx)->GetMiscFlags() & hsGMatState::kMiscBumpChans))
+    {
+        return bumpMapping;
+    }
+
+    bumpMapping = plMetalBumpMapping();
```
Hoikas:
```suggestion
    plMetalBumpMapping bumpMapping;
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.h
```diff
@@ -118,10 +118,13 @@ class plMetalMaterialShaderRef : public plMetalDeviceRef
                              const std::function<plLayerInterface* (plLayerInterface*, uint32_t)>& postEncodeTransform);
     bool     ICanEatLayer(plLayerInterface* lay);
     uint32_t ILayersAtOnce(uint32_t which);
+    
+    std::optional<plMetalBumpMapping> IEatBumpmapLayers( uint32_t& layer );
```
Hoikas:
```suggestion
    std::optional<plMetalBumpMapping> IEatBumpmapLayers(uint32_t& layer);
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1659,7 +1659,11 @@ bool plMetalPipeline::IHandleMaterialPass(hsGMaterial* material, uint32_t pass,
                                   postEncodeTransform);
         }
         
-        fragmentShaderDescription.fUsePerPixelLighting = PLASMA_FORCE_PER_PIXEL_LIGHTING;
+        if(!fragmentShaderDescription.fUsePerPixelLighting)
```
```suggestion
        if (!fragmentShaderDescription.fUsePerPixelLighting)
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.h
```diff
@@ -148,12 +148,13 @@ struct plMetalFragmentShaderDescription
...
     uint8_t  fNumLayers;
     bool     fUsePerPixelLighting;
+    bool     fHasBumpMap;
 
     size_t hash;
 
     bool operator==(const plMetalFragmentShaderDescription& p) const
     {
-        bool match = fNumLayers == p.fNumLayers && memcmp(fPassTypes, p.fPassTypes, sizeof(fPassTypes)) == 0 && memcmp(fBlendModes, p.fBlendModes, sizeof(fBlendModes)) == 0 && memcmp(fMiscFlags, p.fMiscFlags, sizeof(fMiscFlags)) == 0 && fUsePerPixelLighting == p.fUsePerPixelLighting;
+        bool match = fNumLayers == p.fNumLayers && memcmp(fPassTypes, p.fPassTypes, sizeof(fPassTypes)) == 0 && memcmp(fBlendModes, p.fBlendModes, sizeof(fBlendModes)) == 0 && memcmp(fMiscFlags, p.fMiscFlags, sizeof(fMiscFlags)) == 0 && fUsePerPixelLighting == p.fUsePerPixelLighting && fHasBumpMap == p.fHasBumpMap;
```
Hoikas, line 157: Can we.... not?

Hoikas (commented):
What's going on with the code style here?

colincornaby:
I need to review several of the proposed changes for logic.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.h
colincornaby: I'm not sure if there is a better way to implement equality on this class cluster. Any suggestions?

Hoikas: Please break this up into multiple lines.

dpogue: Are there any properties that aren't being compared here? Can we just compare the entirety of `this` against `p`?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +325,39 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+    // We have a bump map and it should occupy the next four layers
+    for (size_t bumpLayer = 0; bumpLayer < 4; bumpLayer++) {
+        plLayerInterface* layer = fMaterial->GetLayer(layerIdx + bumpLayer);
+        uint32_t miscFlags = layer->GetMiscFlags();
+        switch (miscFlags & hsGMatState::kMiscBumpChans)
+        {
+        case hsGMatState::kMiscBumpDu:
+            bumpMapping.value().dTangentUIndex = layer->GetUVWSrc();
```
Hoikas:
```suggestion
            bumpMapping.dTangentUIndex = layer->GetUVWSrc();
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +325,39 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        case hsGMatState::kMiscBumpDv:
+            bumpMapping.value().dTangentVIndex = layer->GetUVWSrc();
```
Hoikas:
```suggestion
            bumpMapping.dTangentVIndex = layer->GetUVWSrc();
```

Hoikas (commented):
I should have included these suggestions with https://github.com/H-uru/Plasma/pull/1808/files#r2532515865. Sorry about the oversight.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -309,10 +308,9 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
         fPassLengths.push_back(j - currLayer);
         
         auto bumpMap = IEatBumpmapLayers(j);
-        fBumps.push_back(bumpMap);
         
-        if (bumpMap.has_value())
-        {
+        if (bumpMap.has_value()) {
+            fBumps.push_back(bumpMap);
```
Hoikas: This is making a copy that we could probably debounce.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -310,11 +310,13 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
         auto bumpMap = IEatBumpmapLayers(j);
         
         if (bumpMap.has_value()) {
-            fBumps.push_back(bumpMap);
             passDescription.fUsePerPixelLighting = true;
             passDescription.fHasBumpMap = true;
         }
         
+        // This needs to be added regardless so each pass has a value
+        fBumps.push_back(std::move(bumpMap));
```
Hoikas, line 318: This should be `emplace_back` to properly move.  But, since `plMetalBumpMapping` is a trivial type, this move is still basically a copy. I would just leave it as `.push_back(bumpMap)` until a construct-in-place strategy can be devised.

colincornaby:
Ok, I've been researching the Windows source and 3DS Plugin to see how viable multi bump is in practice. I think it is possible so I'll need to account for it. I may establish an upper bound. The way Windows does it as a second pass is expensive so I'd like to avoid that.

colincornaby:
I have another pull request I'd like people to look at that this is starting to stack with, I'm going to demote this back to a draft so I can get the other one through.

PR #1811, opened 2025-11-24
Add xForcedViewCamera Python script

### LINKED ISSUE #1804: Telescope Camera Issues
A bug I came across in Pod12 while trying to create closeup cameras for bookshelf items using the cleftRS telescope script as a trigger mechanism (likely present in the generic telescope script as well).

It turns out if you have a clickable that triggers a telescope camera change, and that clickable is *still* within the view of the telescope's camera, allowing you to click on the same clickable that triggered the telescope camera change WHILE you are in the telescope camera change, it can result in the third person camera *randomly* jumping back to that camera after exiting the camera, in various places in the age. 

So to replicate:

Create a Standard Telescope Modifier Setup > Have the telescope camera point at the telescope's triggering clickable > click on the clickable > enter telescope > click on the clickable again > 3rd person cameras will now be sticky upon exit and randomly jump back to this camera when entering other camera regions. 

@Hoikas requested I make an issue, so here it is. :)

dpogue:
Script to force a camera viewpoint from a clickable, with an optional vignette.

This is like a hybrid of the `xTelescope` and `clftRS` scripts, but with the logic simplified and some shortcomings addressed.

* No telescope multistage behaviour mod
* Clickable activator is locally disabled while the camera is active, to prevent breaking the camera stack
* Optional GUI vignette dialog
* Exit with escape or movement keys

### DIFF
--- a/Scripts/Python/xForcedViewCamera.py
+++ b/Scripts/Python/xForcedViewCamera.py
@@ -1,0 +1,174 @@
+# /*==LICENSE==*
+#
+# CyanWorlds.com Engine - MMOG client, server and tools
+# Copyright (C) 2011  Cyan Worlds, Inc.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Additional permissions under GNU GPL version 3 section 7
+#
+# If you modify this Program, or any covered work, by linking or
+# combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+# NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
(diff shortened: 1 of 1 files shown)


Scripts/Python/xForcedViewCamera.py
```diff
@@ -0,0 +1,172 @@
...
+    multiple players can activate the viewpoint camera concurrently).
+    """
+
+    def __init__(self):
+        ptModifier.__init__(self)
+        self.id = 231
+        version = 1
+        minorVersion = 0
+        self.version = version
+        PtDebugPrint("__init__xForcedViewCamera v%d.%d" % (version, minorVersion))
```
Hoikas:
```suggestion
        PtDebugPrint("__init__xForcedViewCamera v{version}.{minorVersion}")
```

Scripts/Python/xForcedViewCamera.py
```diff
@@ -0,0 +1,172 @@
...
+        self.inputInterface = ptInputInterface()
```
Hoikas: This will run when 3ds Max inits. I'd prefer to see this happen later, eg in `OnInit()`

dpogue:
FYI @Traveler263 if you want to give this a test on Destiny

Traveler263:
> FYI @Traveler263 if you want to give this a test on Destiny

Ill try to poke at this either tomorrow or Tuesday.

TheScarFr:
Why not adding an optional two-states responder and an optional SDLQuitCameraView, like in the script from [link removed]?
It would be usefull.

Hoikas (approved):

dpogue:
@Traveler263 Out of curiosity, did this ever get tested in game and confirmed to work the way you expected it to?

Traveler263:
> @Traveler263 Out of curiosity, did this ever get tested in game and confirmed to work the way you expected it to?

It worked on HURU, yes. I don't think the code was coded for the OU client so I never tested it there. But yes, it worked as expected on HURU.

PR #1812, opened 2025-11-24
Refactored Metal Light Encoding

colincornaby:
This change refactors the Metal lighting system to improve CPU usage by around 7% for each frame. A lot of previous changes - including some improvements during per pixel lighting and refactoring how shadows interacted with the light system - made this possible. This should incrementally improve performance - especially on older Macs.

Before, the Metal renderer used to send the full attributes for all lights that affect a material during each material pass. This was expensive both in CPU and bandwidth to the GPU. I tried to mitigate this by sending partial buffers but this performance optimization was lost in #1805. I also had to track any changes to the lights using memcmps. We couldn't cache the lights in a buffer because buffer loads are not in sequence with render operations in Metal - meaning the buffer pretty much has to stay consistent for the entire render. D3D would do things like change light state during the render for passes like shadows.

Metal now loads all currently registered lights into a buffer at the beginning of a render. This buffer must be triple buffered in since our frames are triple buffered. Each material pass is then encoded along with a list of light indices, along with scales for each light. Scale is still sent for each material pass because the scale can change per material.

Because lights are also considered immutable for each render the light state tracking has been greatly simplified. We now only need to track if the active light list for a material pass has changed, or if the scale has changed.

The Metal renderer also now listens for the same register/unregister events that the D3D renderer does.

The buffers sent to the shader now follow array semantics instead of struct semantics. This undoes the damage caused by #1805. The light count is now sent as a standalone uniform instead of an entry in a light structure. The Metal renderer now technically supports unlimited lights - but this has been hard capped at 256 for buffer allocation reasons.

This change assumes - except for scale - that no light attribute will change mid frame render. I've looked through the engine and I'm fairly certain this is the case. Any mid frame changes to lights in the D3D render were part of the renderers shadow system. Metal has decoupled the lighting structures from shadow rendering.

### Future Directions

Now that the light list is a GPU resource this opens up a few possibilities like deferred shading or forward plus shading. With forward plus specifically - light culling/selection would move onto a GPU compute shader and the CPU wouldn't even need to perform `ISelectLights` anymore. But the current forward lighting system could be left in place.

The shader hash table and texture binding are now the larger contributors in the Metal render loop. I've also been thinking about these parts of the renderer - but these are harder to optimize. The Metal renderer has basically abandoned UBOs because of features like piggybacks - so we're rebinding all the textures individually each material pass. These bindings should be saved in buffers on the GPU. The shader hash table is expensive because of the number of states it's storing - although I've considered different cache layouts that are more static and not as reliant on virtual functions.

Metal still has to manage a light stack for saving and restoring light state when projections need to be rendered. I'd eventually like to eliminate this. Previously this was a proper push/pop which meant copying data. I've reduced this to a save/restore which gets rid of the memcpy so it should be a little lighter now. We're also saving/restoring without regard to if we actually need to render projections. I tried to fix that in this commit but there's still something causing light state to get messed up if I don't save the light state before calling `ISelectLights` to find the projection lights.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
@@ -2485,7 +2485,7 @@ void plMetalPipeline::ILoadLight(plLightInfo* light)
     auto metalLight = static_cast<plMetalLightRef*>(light->GetDeviceRef());
     // Track the index of the light for this pass so we can find it again
     metalLight->fPassIndex = fLights->size();
-    fLights->push_back({metalLight->fBufferIndex, 1.f});
+    fLights->emplace_back(metalLight->fBufferIndex, 1.f);
     fLightsDirty = true;
 }
 
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/ShaderTypes.h
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/ShaderTypes.h
@@ -167,6 +167,12 @@ struct plMetalShaderActiveLight
 {
     uint index;
     __fp16 scale;
+    
+    plMetalShaderActiveLight(uint indexIn, float scaleIn)
+    {
+        index = indexIn;
+        scale = scaleIn;
+    }
 };
 #ifndef __METAL_VERSION__
 static_assert(std::is_trivial_v<plMetalShaderLightSource>, "plMetalShaderLightSource must be a trivial type!");


Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalLightRef.h
```diff
@@ -0,0 +1,77 @@
...
+
+    void             Link(plMetalLightRef** back) { plMetalDeviceRef::Link((plMetalDeviceRef**)back); }
+    plMetalLightRef* GetNext() { return (plMetalLightRef*)fNext; }
+
+    plMetalLightRef()
+        : fOwner()
+    {
+    }
+
+    virtual ~plMetalLightRef();
```
Hoikas:
```suggestion
    ~plMetalLightRef();
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -123,6 +124,13 @@ plProfile_CreateCounter("NumSkin", "PipeC", NumSkin);
 #define PLASMA_FORCE_PER_PIXEL_LIGHTING 0
 #endif
 
+// Hypothetically we have no light limit, but Metal will be upset
+// if we encode too much data directly into the command buffer.
+// Command buffer encode limit is 4096 bytes, lets undercut that a bit
+// and max at 256 lights.
+
+#define kMetalMaxLightCount 256
```
Hoikas: `constexpr`?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -1676,15 +1704,23 @@ bool plMetalPipeline::IHandleMaterialPass(hsGMaterial* material, uint32_t pass,
...
-
-    if (!(fState.fBoundLights.has_value() && memcmp(&fState.fBoundLights, &fLights, lightSize) == 0)) {
-        memcpy(&fState.fBoundLights, &fLights, lightSize);
+    // We do need at least some data, Metal gets mad on zero length
+    // data. We reserve at least one light's worth of data elsewhere so
+    // bind at least one light. It's ok if it's garbage, ShaderLightCount
+    // will still be 0.
+    // lightSize needs to be a uint to load onto into the shader
+    uint lightSize = uint(fLights->size());
+    int  bindSize = sizeof(plMetalShaderActiveLight) * fmax(fmin(lightSize, kMetalMaxLightCount), 1);
```
Hoikas: Should probably use `std::clamp`

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -2548,20 +2592,23 @@ void plMetalPipeline::IDrawPlate(plPlate* plate)
...
 // In Metal, when a part of the pipeline wants to own lights
 // we'll just let them push/pop the current state.
-void plMetalPipeline::PushCurrentLightSources()
+void plMetalPipeline::SaveCurrentLightSources()
 {
-    plMetalLights* lightSources = new plMetalLights();
-    memcpy(lightSources, &fLights, sizeof(plMetalLights));
-    fLightSourceStack.emplace_back(lightSources);
+    fLightSourceStack.emplace_back();
+    fLights = &fLightSourceStack.back();
```
```suggestion
    fLights = &fLightSourceStack.emplace_back();
```

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -171,14 +171,20 @@ typedef struct
     half4 fogColor;
 } ColorInOut;
 
+struct Lighting
+{
+    constant plMetalShaderLightSource * lights      [[ buffer(ShaderLights) ]];
+    constant plMetalShaderActiveLight * activeLights      [[ buffer(ShaderActiveLights) ]];
+    constant uint & lightCount [[ buffer(ShaderActiveLightCount)  ]];
```
Hoikas:
```suggestion
    constant plMetalShaderLightSource* lights      [[ buffer(ShaderLights) ]];
    constant plMetalShaderActiveLight* activeLights      [[ buffer(ShaderActiveLights) ]];
    constant uint& lightCount [[ buffer(ShaderActiveLightCount)  ]];
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDeviceRefs.cpp
```diff
@@ -157,3 +158,70 @@ void plMetalRenderTargetRef::Release()
...
+    plDirectionalLightInfo* dirLight = nullptr;
+    plOmniLightInfo*        omniLight = nullptr;
+    plSpotLightInfo*        spotLight = nullptr;
+
+    constexpr float kMaxRange = 32767.f;
+    dev->range = kMaxRange;
+
+    if ((dirLight = plDirectionalLightInfo::ConvertNoRef(fOwner)) != nullptr) {
+        hsVector3 lightDir = dirLight->GetWorldDirection();
+        dev->position = {lightDir.fX, lightDir.fY, lightDir.fZ, 0.0};
```
Hoikas:
```suggestion
        dev->position = {lightDir.fX, lightDir.fY, lightDir.fZ, 0.0f};
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -2378,87 +2478,31 @@ void plMetalPipeline::ISelectLights(const plSpan* span, plMetalMaterialShaderRef
...
-        } else {
-            fLights.lampSources[i].spotProps = {0.0f, 0.0f, 0.0f};
-        }
-    } else {
-        IDisableLight(i);
-    }
+    auto metalLight = static_cast<plMetalLightRef*>(light->GetDeviceRef());
+    // Track the index of the light for this pass so we can find it again
+    metalLight->fPassIndex = fLights->size();
+    fLights->push_back({metalLight->fBufferIndex,
```
```suggestion
    fLights->emplace_back({metalLight->fBufferIndex,
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
```diff
@@ -2378,87 +2478,31 @@ void plMetalPipeline::ISelectLights(const plSpan* span, plMetalMaterialShaderRef
...
-    fLights.lampSources[i].quadAtten = {0.0f};
-    fLights.lampSources[i].scale = {0.0f};
-}
+    auto metalLight = static_cast<plMetalLightRef*>(light->GetDeviceRef());
 
-void plMetalPipeline::IScaleLight(size_t i, float scale)
-{
     scale = int(scale * 1.e1f) * 1.e-1f;
-    fLights.lampSources[i].scale = scale;
+    fLights->at(metalLight->fPassIndex).scale = scale;
```
```suggestion
    fLights[metalLight->fPassIndex].scale = scale;
```

`at()` uses bounds checking, which is a tiny amount slower. I'm not sure if `at()` is intentional over `[]`, but I thought I'd leave the suggestion.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.h
```diff
@@ -263,16 +268,21 @@ class plMetalPipeline : public pl3DPipeline<plMetalDevice>
...
+
+    std::vector<plMetalShaderActiveLight>*             fLights;
+    bool                                               fLightingPerPixel;
+    std::vector<std::vector<plMetalShaderActiveLight>> fLightSourceStack;
 
     static plMetalEnumerate enumerator;
 
     plTextFont* fTextFontRefList;
+    plMetalLightRef* fLightRefList;
+    bool fLightsDirty = true;
```
Hoikas: This should be done in the constructor.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.cpp
colincornaby: emplace_back doesn't support aggregate types until C++20. I can use emplace_back, but I'd need to add an initializer. I can make that change - I would just need to make sure it doesn't cause issues in since the header that defines the type is shared with the shader. (Metal Shading Language is a derivative of C++ so it hopefully shouldn't hurt anything....)

Hoikas: Hopefully that won't cause a problem. IIRC, according to the standard, this `push_back` call should construct the object, then make a copy of it in the vector, which is not ideal. OTOH `emplace_back` will construct the object in place in the vector. On MSVC, `push_back` just calls `emplace_back`, so you get the most performant path by default, but that's not helpful here. I don't know what other STLs do.

Hoikas:
This LGTM, but I'm not sure what that's worth since Mac/Metal aren't exactly areas of expertise for me.

colincornaby:
I've been testing this and so far it seems fine. I do have a bigger, better version of Metal bump mapping nearly ready that would require this. So I'm eager to get it in.

PR #1814, opened 2025-11-26
xConsoleCamera

TheScarFr:
Add a fairly simple script to handle looking at a console-like object.
Like for the "Kadish Tolesa's Vault Console" for example, but without any puzzle element in the script.

Detailed setup instructions:

```
1. activator
The clickable which triggers looking at the console.

2. camera
The camera we switch to when "activator" is clicked.

3. respUsingConsole (optional)
A two-state responder triggered when starting or stopping to look at the console.
The first state is called when the avatar stops looking at the console.
The second state is called when the avatar starts looking at the console.

4. sdlQuitConsole (optional)
The name of an SDL variable which, when changed, forces any player to stop interacting with the console.
```

This script, from @Jrius, works on both POTS and Destiny.
It may need some ajustments to be accepted on H-uru, but like I'm not a code-dev, I'm counting on code-devs to tell me what to do, or to do the ajustments.


Hoikas:
This is basically a duplicate of #1811 with the added ability to fire an arbitrary responder from the script. If the ability to fire a responder is needed from the script, I would suggest adding that functionality to the already accepted code.

TheScarFr:
Hoikas wrote:
> This is basically a duplicate of #1811 with the added ability to fire an arbitrary responder from the script. If the ability to fire a responder is needed from the script, I would suggest adding that functionality to the already accepted code.

TheScarFr wrote:
> Why not adding an optional two-states responder and an optional SDLQuitCameraView

dpogue wrote:
> This is intended to be a simple script for forcing a camera perspective, not for complex Age-specific situations that have a bunch of other state associated with them. More complicated scenarios probably have other Python logic they will need, and proper consideration regarding multiplayer interactions. The xTelescope script for example restricts a single player to using the telescope at a time, but this script allows multiple players to trigger to viewpoint camera.

Hum, come to an agreement, guys...
I need the xConsoleCamera.py script in my Ages, but it can be used in any Age.
It currently works in CC and on Destiny.

Hoikas:
You can add the functionality to the code that already exists without exceeding the scope that @dpogue mentioned. Python is an object oriented language, and it is trivial to extend code that is already present without modifying that code. Alternatively, the responder can be triggered outside of the script via the other logic objects. You surely have some way to trigger the telescope, after all.

PR #1815, opened 2025-11-26
Allow longer duration KI light and randomization

Hazado:
Allows longer duration KI Light and randomizing the time for puzzles

Python File Mod #5 is an Integer that sets the duration of the light when unrepaired, default of 5 seconds
Python File Mod #6 is an Integer that sets the duration of the light when repaired, default of 60 seconds

byteKILightFunc now has a third option for random light time between the short and long time

### DIFF
--- a/Scripts/Python/dsntKILightMachine.py
+++ b/Scripts/Python/dsntKILightMachine.py
@@ -76,8 +76,6 @@ lightOn = 0
 #----------
 # constants
 #----------
-kLightTimeShort = intKILightTimeShort.value
-kLightTimeLong = intKILightTimeLong.value
 listLightResps = ["respKILightOff","respKILightOn"]
 kLightStopID = 1
 kKILightShortSFXRespName = "respSFX-KILight-Short"
@@ -166,11 +164,11 @@ class dsntKILightMachine(ptModifier):
         if not byteKILightFunc:
             return
         elif byteKILightFunc == 1:
-            lightStop = (lightStart + kLightTimeShort)
+            lightStop = (lightStart + intKILightTimeShort.value)
         elif byteKILightFunc == 2:
-            lightStop = (lightStart + kLightTimeLong)
+            lightStop = (lightStart + intKILightTimeLong.value)
         elif byteKILightFunc == 3:
-            lightStop = (lightStart + random.randint(kLightTimeShort,kLightTimeLong))
+            lightStop = (lightStart + random.randint(intKILightTimeShort.value,intKILightTimeLong.value))
 
         timeRemaining = (lightStop - lightStart)
         PtDebugPrint("timer set, light will shut off in ",timeRemaining," seconds; lightStop = ",lightStop)
@@ -241,7 +239,7 @@ class dsntKILightMachine(ptModifier):
                         self.SetKILightChron(remaining)
                         lightOn = 0
                         PtSetLightAnimStart(avatarKey, KILightObjectName, False)
-                elif resp.getName() == kKILightShortSFXRespName and remaining == kLightTimeShort:
+                elif resp.getName() == kKILightShortSFXRespName and remaining == intKILightTimeShort.value:
                     PtDebugPrint("dsntKILightMachine.DoKILight():\tRunning short KI Light SFX")
                     sndResp = ptAttribResponder(43)
                     sndResp.__setvalue__(resp)


Scripts/Python/dsntKILightMachine.py
```diff
@@ -57,6 +58,8 @@
 respDispensor = ptAttribResponder(2,"resp: KI dispensor")
 respGotKI = ptAttribResponder(3,"resp: got KI")
 sdlKILightFunc = ptAttribString(4,"sdl: KI light func")
+intKILightTime = ptAttribInt(5,"int: Length that Ki light lasts in seconds",0)
+boolKIRandom = ptAttribBoolean(6,"bool: Randomize time light last, longest time is intKILightTime, Shortest is 5s",False)
```
Hoikas: I wonder if it would make more sense/be more flexible to change the `kLightTimeShort` and `kLightTimeLong` constants to attributes with the defaults that the constants currently have. IMO, the random in range operation should be a new KI light func mode (`byteKILightFunc`).

Scripts/Python/dsntKILightMachine.py
```diff
@@ -157,6 +163,11 @@ def SetKILightTime(self,on):
         global lightOn
         PtDebugPrint("dsntKILightMachine.SetKILightTime(): byteKILightFunc = ",byteKILightFunc)
         lightStart = PtGetDniTime()
+        
+        if intKILightTime.value:
+            kLightTimeLong = intKILightTime.value
+        if boolKIRandom.value:
+            kLightTimeLong = xRandom.random.randint(kLightTimeShort,kLightTimeLong)
```
Hoikas: You're using the Python standard module `random` here instead of `xRandom`'s special `randint()` method. We could therefore just use `random` instead of Cyan's special `xRandom` module.

Scripts/Python/dsntKILightMachine.py
```diff
@@ -73,8 +76,8 @@
 #----------
 # constants
 #----------
-kLightTimeShort = 5
-kLightTimeLong = 60
+kLightTimeShort = intKILightTimeShort.value
+kLightTimeLong = intKILightTimeLong.value
```
Hoikas: This won't work, unfortunately. The attributes' values are at the default when this code runs. You'll want to remove these constants and change the code to use the attribute values everywhere the constants were referenced.

Hazado: Should that be set in the ``__init__`` or the ``OnServerInitComplete``?

Hoikas: The attribute values are available after `OnInit` gets called (different than `__init__`). They are also available in `OnServerInitComplete`

Hazado: Right, I understand what your saying now.

EhrenCG:
> I'm having trouble following your comment, @EhrenCG, so - correct me if I'm wrong - we also need to adjust a PythonFileMod in Descent so that the new state 3 matches up visually with a functioning KI machine?

Sorry, the reason I didn't reply to this, is because I was being a weenie...  I felt insulted at the time that you didn't understand something that I took way too long to type up as clear as possible.  I figured saying nothing was better than just going off about how angry I was.  I've had unfortunate outbursts, but I do put actual effort into trying not to do that.

It was dumb of me to be like that, but I think this is most relevant to mention now:

Yes, some PythonFileMods needed to be adjusted so that the main (and currently only) KI light machine would enable the proper objects at the new var of 3, though I guess it could be argued that since the Cyan one should never be set past 2, it was irrelevant to make that particular change to understand 3 at all, but my mindset was just that since the base code changed I wanted it to be fully functional and testable for the TBS KI light machine.

But the action needed was on my end, on the special page that has the visual effects, particles being the most important.  I already adjusted for this change and I also told Cal what to change for the clickable object on his end so the job has already been done.  (For example the clickable object had to be told to not be disabled for 3, which the show hide code normally would do since it needs to be told the values in which it should enable the clickable.)

So, yes, this has my explicit approval as it's already been adjusted for.  So, even if I'm still not making sense to you, at least know that part.

Hoikas:
Thanks for the clarification. I think that we should adjust the PFM that's already in the game to cope with this change. That's something I can do, I just wasn't sure what exactly was going on. It takes me probably a solid two weeks right now to look at all of our open tasks, so it might take me a bit to get to that.

FWIW, I haven't been as engaged with everything that's going on over the last few months, so I'm really just getting bits and pieces of what's going on. Also, as time goes on, I find my ability to catch on to what people are talking about is not as great as it used to be, so please don't take my not understanding as a failure on your part. I'm just not a quick study anymore 😉

PR #1817, opened 2025-11-29
Fix `OpenEncryptedFile`/`OpenSecureFile` not reporting failure

dgelessus:
Even though these methods can report failure, they practically never did, because they silently ignored any error returns from `plFileSystemStream::Open`. This fixes that. Also adds error checks to a couple of callers, but most of them already had error handling.

Main practical benefit of this change: Linking to a nonexistant age now gives an assertion error with a clear message instead of staying at the loading screen forever.

### DIFF
--- a/Sources/Plasma/Apps/plClient/plClient.cpp
+++ b/Sources/Plasma/Apps/plClient/plClient.cpp
@@ -2102,6 +2102,10 @@ void plClient::IDetectAudioVideoSettings()
 void plClient::IWriteDefaultAudioSettings(const plFileName& destFile)
 {
     std::unique_ptr<hsStream> stream = plEncryptedStream::OpenEncryptedFileWrite(destFile);
+    if (!stream) {
+        return;
+    }
+
     WriteBool(stream.get(), "Audio.Initialize",  true);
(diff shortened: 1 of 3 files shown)


Sources/Plasma/FeatureLib/pfPython/pyStream.cpp
```diff
@@ -87,15 +87,19 @@ bool pyStream::Open(const plFileName& fileName, const ST::string& flags)
                 if (encryptflag)
                 {
                     auto stream = std::make_unique<plEncryptedStream>();
-                    stream->Open(fileName, "wb");
-                    fStream = std::move(stream);
+                    if (stream->Open(fileName, "wb")) {
+                        fStream = std::move(stream);
+                    } else {
+                        fStream = nullptr;
+                    }
```
Hoikas: `fStream` should already be null, I think.

dgelessus: I think you're right - the start of the method calls `Close`, which clears out `fStream`. I was extra cautious here, because this code path previously always overwrote `fStream`.

PR #1818, opened 2025-11-29
Remove `hsThread::StartDetached`, which is generally unsafe

dgelessus:
Follow-up to #1809.

After `StartDetached` is called, the `hsThread` can be destroyed while the underlying thread is still running. This is unsafe in the general case, because `hsThread`'s thread proc captures the `this` pointer and calls virtual methods (such as `OnQuit`) on the potentially-destroyed `hsThread`!

There is exactly one case where this method is safe, namely if the `hsThread` is destroyed at the very end of its own `OnQuit` method. This was done by `pfPatcherWorker`, the only `hsThread` implementation for which `StartDetached` was used. `pfPatcherWorker` is now reimplemented using a detached `std::thread` that owns all the data it needs, so the unsafe `hsThread::StartDetached` method is no longer needed.

This also refactors the `pfPatcher` internals to use `std::unique_ptr`s instead of bare pointers to `pfPatcher` and `pfPatcherWorker`, to make it clearer who currently owns the patcher state.

Because my local test shard doesn't have a file server, I can only do limited testing of this change myself. @Hoikas Could you please put this PR on TrollLand when you have the time? Then I can test that patching still works everywhere in the launcher and the game itself.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSPatcher.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSPatcher.mm
@@ -100,19 +100,19 @@ public:
 {
     [[NSRunLoop mainRunLoop] addTimer:self.networkPumpTimer forMode:NSDefaultRunLoopMode];
 
-    auto patcher = std::make_unique<pfPatcher>();
-    patcher->OnFileDownloadBegin(
+    pfPatcher patcher;
+    patcher.OnFileDownloadBegin(
         std::bind(&Patcher::IOnDownloadBegin, _cppPatcher, std::placeholders::_1));
-    patcher->OnProgressTick(std::bind(&Patcher::IOnProgressTick, _cppPatcher,
-                                      std::placeholders::_1, std::placeholders::_2,
-                                      std::placeholders::_3));
-    patcher->OnCompletion(std::bind(&Patcher::IOnPatchComplete, _cppPatcher, std::placeholders::_1,
-                                    std::placeholders::_2));
-    patcher->OnFileDownloadDesired(IApproveDownload);
-    patcher->OnSelfPatch(std::bind(&Patcher::ISelfPatch, _cppPatcher, std::placeholders::_1));
-    patcher->OnFindBundleExe(std::bind(&Patcher::IFindBundleExe, _cppPatcher, std::placeholders::_1));
-    patcher->RequestManifest(plManifest::ClientManifest());
-    patcher->Start();
+    patcher.OnProgressTick(std::bind(&Patcher::IOnProgressTick, _cppPatcher,
+                                     std::placeholders::_1, std::placeholders::_2,
+                                     std::placeholders::_3));
+    patcher.OnCompletion(std::bind(&Patcher::IOnPatchComplete, _cppPatcher, std::placeholders::_1,
+                                   std::placeholders::_2));
+    patcher.OnFileDownloadDesired(IApproveDownload);
+    patcher.OnSelfPatch(std::bind(&Patcher::ISelfPatch, _cppPatcher, std::placeholders::_1));
+    patcher.OnFindBundleExe(std::bind(&Patcher::IFindBundleExe, _cppPatcher, std::placeholders::_1));
+    patcher.RequestManifest(plManifest::ClientManifest());
+    patcher.Start();
 }
 
 - (NSURL *)completeSelfPatch:(NSError **)error;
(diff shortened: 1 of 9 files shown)


Sources/Plasma/Apps/plClient/Mac-Cocoa/PLSPatcher.mm
```diff
@@ -71,7 +71,6 @@
 
 @interface PLSPatcher ()
 @property BOOL selfPatched;
-@property pfPatcher* patcher;
```
dgelessus, line 74: I hope it's safe to convert this to a local variable like I did. I couldn't find any place that really needs to use the `pfPatcher` outside of the `start` method, but I'm not super familiar with the Mac patcher code.

dgelessus:
Did some quick tests on the new TrollLand and the patcher seems to still work fine. I've tested the launcher and in-game updating with both internal and external builds of the Windows client.

Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.h
```diff
@@ -146,10 +154,10 @@ class pfPatcher
     void RequestManifest(const ST::string& mfs);
     void RequestManifest(const std::vector<ST::string>& mfs);
 
-    /** Start patching the requested manifests. Please note that after calling this, you should
-     *  discard your pointer to this object as it will memory-manage itself.
+    /** Start patching the requested manifests. This consumes a unique_ptr,
+     * because pfPatcher will memory-manage itself from this point on.
      */
-    bool Start();
+    static void Start(std::unique_ptr<pfPatcher> patcher);
```
Hoikas: I wonder if the argument here should be an r-value reference since it doesn't really make sense to take an l-value and/or make a copy.

Sources/Plasma/PubUtilLib/plAgeLoader/plResPatcher.cpp
```diff
@@ -176,18 +176,18 @@ plResPatcher::~plResPatcher()
 void plResPatcher::Update(const std::vector<ST::string>& manifests)
 {
     InitProgress();
-    pfPatcher* patcher = CreatePatcher();
+    auto patcher = CreatePatcher();
     if (!gDataServerLocal)
         patcher->RequestManifest(manifests);
-    patcher->Start(); // whoosh... off it goes
+    pfPatcher::Start(std::move(patcher)); // whoosh... off it goes
```
Hoikas: I understand why you've changed this - it makes the ownership of the `pfPatcher` object much clearer and prevents folks from touching it after its ownership has been stolen. I feel like `pfPatcher::Start(std::move(patcher));` is a bit of a cumbersome incantation to have to do any time we want to start a patch operation. It looks like this pattern is repeated a few times in the PR. The nice thing about the old `->Start()` method was that it was easy to use.

I wonder if it would be better to, from `pfPatcher::Create()`, return some kind of wrapper object that dereferenced to the patcher and had a safe start method. So something like:

```cpp
class pfPatcherFoobar
{
    friend class pfPatcher;
    std::unique_ptr<pfPatcher> fPatcher;

protected:
    pfPatcherFoobar(std::unique_ptr<pfPatcher>&& patcher)
        : fPatcher(patcher)
    { }

public:
    pfPatcher* ->operator() const { return fPatcher.get(); }
    void Start() { pfPatcher::Start(std::move(fPatcher)); }
};
```

That way, we can basically retain the easy `patcher->Start()` invocation and gain the ownership clarity.

dgelessus: Yeah, the `Start` calls became a bit unwieldy from using only standard C++ idioms to express the ownership transfer... A custom wrapper like you suggest would probably be good, as long as it remains clear that once the patcher has been started, it manages itself and you're not allowed to touch it anymore.

Looking at the code again now... is there even a reason why `pfPatcher` needs to stay alive while the patcher is running? It looks like all `pfPatcher` methods store everything directly into the underlying `pfPatcherWorker`, so it should be okay to destroy the `pfPatcher` after the worker is started and manages itself. I think `pfPatcher` doesn't even need to be heap-allocated - it could just become a convenience wrapper around `std::unique_ptr<pfPatcherWorker>`.

Hoikas: The whole `pfPatcher`/`pfPatcherWorker` thing was my first experiment with PIMPL, with the complication that the object needs to memory manage itself.  I think `pfPatcher` becoming a stack allocated convenience wrapper around `std::unqiue_ptr<pfPatcherWorker>` sounds good.

PR #1819, opened 2025-12-07
plEncryptedStream no longer asserts on EOF

colincornaby:
Returning EOF to a log event. Not sure if EOF shouldbe a log event either, but in the original engine it was. Reaching EOF before a buffer fills should be expected behavior.

Putting this up mostly as a way to kickstart the conversation. I'm not sure if I restored the old logging behavior correctly or if I'm attached to the right log.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plFile/CMakeLists.txt
+++ b/Sources/Plasma/PubUtilLib/plFile/CMakeLists.txt
@@ -23,7 +23,6 @@ target_link_libraries(
     PUBLIC
         CoreLib
     PRIVATE
-        plNetClient
         $<$<PLATFORM_ID:Windows>:shell32>
 )
 
--- a/Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
+++ b/Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
@@ -40,8 +40,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 #include "plEncryptedStream.h"
-#include "pnNetCommon/plNetApp.h"
-#include "pnNetCommon/pnNetCommon.h"
 
 #include <ctime>
 #include <string_theory/format>
@@ -180,13 +178,11 @@ uint32_t plEncryptedStream::IRead(uint32_t bytes, void* buffer)
     size_t numItems = fread(buffer, 1 /*size*/, bytes /*count*/, fRef);
     fPosition += numItems;
     if (numItems < bytes) {
-        if (feof(fRef)) {
-            hsLogEntry(plNetApp::StaticDebugMsg(ST::format("Hit EOF on UNIX Read, only read {} out of requested {} bytes", numItems, bytes).c_str()));
-        } else {
+        if (!feof(fRef)) {
             hsAssert(false, ST::format("Error on UNIX Read (ferror = {})", ferror(fRef)).c_str());
         }
     }
-    return uint32_t(numItems);
+    return static_cast<uint32_t>(numItems);
 }
 
 void plEncryptedStream::IBufferFile()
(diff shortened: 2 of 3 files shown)


Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -179,12 +181,12 @@ uint32_t plEncryptedStream::IRead(uint32_t bytes, void* buffer)
     fPosition += numItems;
     if (numItems < bytes) {
         if (feof(fRef)) {
-            hsAssert(false, ST::format("Hit EOF on UNIX Read, only read {} out of requested {} bytes", numItems, bytes).c_str());
+            hsLogEntry(plNetApp::StaticDebugMsg(ST::format("Hit EOF on UNIX Read, only read {} out of requested {} bytes", numItems, bytes).c_str()));
```
dgelessus: `hsStatusMessage` may be better so we don't have to depend on `plNetClient` just for this. network.log isn't really the right place for this message anyway, IMO.

But if this is clearly not an error condition, I think we can take out the log message entirely and leave just the assert on `ferror` below.

Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -179,12 +181,12 @@ uint32_t plEncryptedStream::IRead(uint32_t bytes, void* buffer)
...
         } else {
             hsAssert(false, ST::format("Error on UNIX Read (ferror = {})", ferror(fRef)).c_str());
         }
     }
-    return numItems;
+    return uint32_t(numItems);
```
dgelessus:
```suggestion
    return static_cast<uint32_t>(numItems);
```

Sources/Plasma/PubUtilLib/plFile/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ target_link_libraries(
     PUBLIC
         CoreLib
     PRIVATE
+        plNetClient
```
Hoikas: `plNetClient` isn't being used anywhere.

Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -40,6 +40,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 *==LICENSE==*/
 #include "plEncryptedStream.h"
+#include "pnNetCommon/plNetApp.h"
+#include "pnNetCommon/pnNetCommon.h"
```
Hoikas: Dependency is missing from `target_link_libraries`.

Sources/Plasma/PubUtilLib/plFile/CMakeLists.txt
colincornaby: Yeah. Something is weird here - plNetClient worked as the dependency locally but it's not on Actions. But it's not the right dependency - and if we removing the logging entirely this won't be necessary.

dgelessus: I think the linker errors are because the use of `plNetApp` is indirectly pulling in all the infrastructure around `plCreatable`, and some executables (`plFileEncrypt` and `plUruLauncher`) that use `plFile` don't have the magic "AllCreatables" includes that are needed to make creatables work.

(I'm a little surprised though. Did the launcher really not use *any* creatables, keyed objects, or messages before?)

Hoikas: > (I'm a little surprised though. Did the launcher really not use any creatables, keyed objects, or messages before?)

Nope, it doesn't need to 😉

Sources/Plasma/PubUtilLib/plFile/plEncryptedStream.cpp
```diff
@@ -178,13 +178,11 @@ uint32_t plEncryptedStream::IRead(uint32_t bytes, void* buffer)
     size_t numItems = fread(buffer, 1 /*size*/, bytes /*count*/, fRef);
     fPosition += numItems;
     if (numItems < bytes) {
-        if (feof(fRef)) {
-            hsAssert(false, ST::format("Hit EOF on UNIX Read, only read {} out of requested {} bytes", numItems, bytes).c_str());
-        } else {
+        if (!feof(fRef)) {
```
colincornaby, line 181: After I committed I stared at this for a second - not sure if it's worth compressing the two ifs into one now....

PR #1820, opened 2025-12-08
Fixing Metal rendering when the vertex color contains an alpha

colincornaby:
This is a regression from the per pixel lighting changes. Re-multiplying lightingContributionColor is not supposed to include the alpha.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -473,7 +473,7 @@ fragment half4 pipelineFragmentShader(ColorInOut in [[stage_in]],
             }
         }
         
-        currentColor = lightingContributionColor * currentColor;
+        currentColor = half4(lightingContributionColor.rgb, 1.h) * currentColor;
     }
     
     currentColor.rgb = mix(in.fogColor.rgb, currentColor.rgb, (float)clamp(in.fogColor.a, 0.0h, 1.0h));


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -473,7 +473,7 @@ fragment half4 pipelineFragmentShader(ColorInOut in [[stage_in]],
             }
         }
         
-        currentColor = lightingContributionColor * currentColor;
+        currentColor = half4(lightingContributionColor.rgb, 1.h) * currentColor;
```
colincornaby, line 476: @dpogue - I got this from the OpenGL code... I think I asked before why we remultiply the vertex color. But I can't remember why. Could you explain this again? I'll put a comment into the code so I don't forget again.

dpogue: I think it's because if you have a texture, the texture has now overridden the vertex colour so it needs to be multiplied back in

colincornaby: That might be it. I'm still thinking it through. This is _correct_, at least where I've checked, it's matching the D3D output... But I'm not quite sure why it would be done this way.

In vertex shading the color is going to come through pre-lit, so I think the problem is we then need to light the texture too. A blend wouldn't do that correctly. But I'm not sure why I wouldn't want to multiply the lighting before the blend.

The vertex color is also baked in with the lighting which makes it a little more weird, but maybe that's part of the answer here.

colincornaby: Ok, if I'm reading `blendFirst` correctly it looks like we should always overwrite the initial value which means that the initial set of currentColor is ignored (unless we don't enter the if at all because there are no texture layers, which is the correct behavior.)

The only possible weird edge cases is if layer 0 has `kBlendNoTexColor` or `kBlendNoTexAlpha` set in which case the vertex color would be blended into later layers. I don't know if this would cause a double blend once we hit this line.

PR #1821, opened 2025-12-13
Add xLinkEventTrigger.py.

Hoikas:
This new general purpose global script enables triggering either responders or an arbitrary logic modifier when a player spawns or links, depending on rules set on the script attributes. This is useful because, presently, the only way to achieve anything remotely like this effect is to auto-start an animation or sound on age load. Age load and link in don't exactly correspond, so that can result in things starting before the player is even in the Age to observe the action. There is no general mechanism for firing on link out or spawn out.

The downside to this is that it makes it trivial for someone to wire up things that have gotchas. For example, having a Python file mod that triggers when the local player spawns out is not actually useful for regular Ages. The Python file mod would likely be unloaded before such a thing ever fired.

### DIFF
--- a/Scripts/Python/xLinkEventTrigger.py
+++ b/Scripts/Python/xLinkEventTrigger.py
@@ -68,7 +68,6 @@ from Plasma import *
 from PlasmaTypes import *
 
 import enum
-import itertools
 
 # Fire this "responder" on link in. In 3ds Max or Korman logic nodes, this can be a responder
 # with some messages that you fire on link in. You need to be careful about doing that, though,


Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,240 @@
...
+        # Now actually handle business. We may trigger now or when link effects play. There are
+        # pros and cons to each. The spawn happens right before the link effects play, but that
+        # may cause some effect you want to happen for the player to be partially cut off due to it
+        # starting before they're actually in the Age. Further, if this is firing for a remote
+        # player linking in, the remote player (for the most part) decides when they link in, which
+        # means the delay is dependent on their loading speed. Relying on link trigger, however, is
+        # probably not the *best* solution though because internal clients (CSR) can go into
+        # "stealth mode," which might mean they never play the link in effects at all. You wouldn't
+        # want to do gameplay logic in a link effects trigger. To get the best of both worlds, it
+        # would be nice if we could urigger (state==1.0) when the interesting event starts and
```
dpogue:
```suggestion
        # would be nice if we could trigger (state==1.0) when the interesting event starts and
```

Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,240 @@
...
+import itertools
+
+# This is anticipated to be a kMultiTrigger plLogicModifier with a plActivatorActivatorConditionalObject
+# set as its only condition. plAACO seems to not be exposed to Max, but it's very useful. It tells
+# its LM to trigger when it receives a state==1.0 notification (if not already triggered) and to
+# untrigger when it receives a state==0.0 (if already triggered). Since this is a kMultiTrigger,
+# any time we send a notify (state==1.0) to the LM, the plAACO will trigger the LM, sending off
+# whatever notifies it has stored up. This can be used to activate other PFMs to do things like
+# set SDL variables (xAgeSDLBoolSet) on link in...
+actActivator = ptAttribActivator(1, "Activator: An ActivatorActivator to send notifications to")
```
Hoikas: Since the expected use case here is an ActivatorActivator, it would be possible to simplify this script by simply deleting this attribute and tossing the ActivatorActivator's LogicMod key into attribute 2. The ActivatorActivator fires when it receives a notify message, which is what `ptAttribResponder.run()` does. It just seems a little odd to toss a LogicMod key into a Responder attribute, but the only thing that cares about the object type in an attribute is the Korman and 3ds Max UI.

With the simplification, we only have one output attribute, can remove the dependency on `itertools`, and don't have to manually send notify messages. I think for some other Korman high level logic nodes, I will need to do either this or generate some thunk responders. Anyway, feedback would be appreciated.

dpogue: Condensing it down to a single attribute seems okay to me, provided that doesn't cause problems for either korman or PlasmaMax

Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,285 @@
...
+from Plasma import *
+from PlasmaTypes import *
+
+import enum
+import itertools
+
+# Fire this "responder" on link in. In 3ds Max or Korman logic nodes, this can be a responder
+# with some messages that you fire on link in. You need to be careful about doing that, though,
+# because it's easy to craft something that won't synchronize well in multiplayer. In reality, my
+# thinking is that, although this is labelled a responder, it should really hold a key to a kMultiTrigger
```
Deledrius:
```suggestion
# thinking is that, although this is labeled a responder, it should really hold a key to a kMultiTrigger
```

I assume we prefer American English spelling rather than British English. ;)

Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,285 @@
...
+        # probably not the *best* solution though because internal clients (CSR) can go into
+        # "stealth mode," which might mean they never play the link in effects at all. You wouldn't
+        # want to do gameplay logic in a link effects trigger. To get the best of both worlds, it
+        # would be nice if we could trigger (state==1.0) when the interesting event starts and
+        # untrigger (state==0.0) when it ends. So, for example, if we wanted to trigger on link in,
+        # trigger (state==1.0) when the link effect starts and untrigger (state==0.0) when the
+        # link in effect stops. Unfortunately, that's not going to work.
+        #
+        # Why won't it work, you ask? Well, gather 'round and let Hoikas explain. Because players
+        # might be linking/spawning in/out simultaneously, if we're using an the actActivator
```
Deledrius:
```suggestion
        # might be linking/spawning in/out simultaneously, if we're using an actActivator
```

Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,285 @@
...
+        playerID = PtGetClientIDFromAvatarKey(avObj.getKey())
+        inout = "in" if pageIn else "out"
+        if self._triggerOn == TriggerOn.NOTHING:
+            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn {inout} (we trigger for nothing)", level=kDebugDumpLevel)
+            return
+        if self._triggerFor == TriggerFor.NOBODY:
+            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn {inout} (we trigger for nobody)", level=kDebugDumpLevel)
+            return
+        if self._triggerFor == TriggerFor.ME and not self.isLocalAvatar(avObj):
+            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn in (we trigger for the local player only)", level=kDebugDumpLevel)
```
Deledrius:
```suggestion
            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn {inout} (we trigger for the local player only)", level=kDebugDumpLevel)
```

Presumably it's possible for a non-local avatar to trigger this in both cases?

Scripts/Python/xLinkEventTrigger.py
```diff
@@ -0,0 +1,285 @@
...
+            return
+        if self._triggerFor == TriggerFor.HUMANS and not avObj.isHuman():
+            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn {inout} (we trigger for humans only", level=kDebugDumpLevel)
```
Deledrius:
```suggestion
            PtDebugPrint(f"xLinkEventTrigger.AvatarPage(): {self.sceneObjectName=} {playerID=} ignoring spawn {inout} (we trigger for humans only)", level=kDebugDumpLevel)
```

PR #1823, opened 2025-12-14
Adding Metal 3 shader compilation and check

colincornaby:
This commit adds Metal 3 support to Plasma while maintaining support for Metal 2. For context - I'm working on some features that would reduce CPU consumption on macOS. These features would work best in Metal 3. I'll still be bringing those features to Metal 2 - but Metal 2 Macs will have to use less optimal paths.

I'm adding a `METAL_3_SDK` preprocessor flag so the renderer can include or drop those code paths based on the presence of Metal 3 in the SDK. This should allow Plasma to compiling in versions of Xcode older than 13 - which was the first Xcode to ship with the Metal 3 SDK. Metal 3.0 SDK presence is assumed when Xcode is not available.

For clarity - Apple lists the requirements for Metal 3 as:

> macOS Ventura 13 or later
> MacBook Pro introduced in 2017 or later*
> MacBook Air introduced in 2018 or later
> MacBook introduced in 2017 or later
> Mac mini introduced in 2018 or later
> iMac introduced in 2020 or later, plus iMac (21.5-inch, 2017) and iMac Pro
> Mac Studio introduced in 2022 or later
> Mac Pro introduced in 2019 or later*

Any older Mac will continue using Metal 2 paths in the renderer.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/CMakeLists.txt
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/CMakeLists.txt
@@ -115,7 +115,7 @@ set_target_properties(pfMetalPipelineShadersMSL23 PROPERTIES
     FOLDER FeatureLib
 )
 
-if (NOT DEFINED CMAKE_XCODE_VERSION_MAJOR OR
+if(NOT DEFINED CMAKE_XCODE_VERSION_MAJOR OR
     CMAKE_XCODE_VERSION_MAJOR VERSION_GREATER_EQUAL 13)
 
     add_metal_shader_library(pfMetalPipelineShadersMSL30


colincornaby:
This PR is open to anyone but looking for feedback from @dpogue in since this touches Xcode-less Metal builds.

Sources/Plasma/FeatureLib/pfMetalPipeline/CMakeLists.txt
```diff
@@ -114,3 +114,23 @@ target_include_directories(pfMetalPipelineShadersMSL23
 set_target_properties(pfMetalPipelineShadersMSL23 PROPERTIES
     FOLDER FeatureLib
 )
+
+if (NOT DEFINED CMAKE_XCODE_VERSION_MAJOR OR
```
Hoikas:
```suggestion
if(NOT DEFINED CMAKE_XCODE_VERSION_MAJOR OR
```

dpogue (approved):
I still have hesitation around the Xcode version based check, but if that turns out to cause problems we can sort that out later

colincornaby:
I think the Xcode version check will be fine in since the Xcode version and the current SDK have been linked for a while now. And it's a pretty niche case - I can't imagine someone is going to try to load a version of Xcode and a specific SDK in that very specific combination. Worst case - it will just compile without Metal 3.

PR #1825, opened 2025-12-19
Argument buffers and Metal bump mapping support

colincornaby:
This PR implements argument buffers in the Metal renderer - and implements them first through bump mapping. Eventually I would like argument buffers to replace the entire layer system in the Metal shaders. I'm actively working on that - will be a future PR. If anyone is interested in rendering - please take your time on this one. Any improvements/feedback here will be very helpful for future work.

This came about because @dpogue asked about multiple bump maps in the original PR. There aren't an unlimited number of blind points in the shaders and the bind points have to be predefined - so supporting multiple bump maps would have been complicated.

Argument buffers replace bind points in shaders. I bind resources to the argument buffers instead, and then pass the buffers to the shader. The advantage is that now the bind points for textures, samplers, and other resources don't need to be predefined in the shader. And because they don't need to be predefined I can pass an array of argument buffers of any length and the shader can iterate over them. That means I can effectively bind an unlimited number of resources within shaders. This solves the problem of how we deliver an indeterminate amount of bump maps to the shader.

There is also a performance pickup. I can leave the argument buffers on the GPU, and instead of having to bind textures and samplers individually, I can just rebind the buffer I have cached. I have to be careful doing this around features like push over layers and piggyback layers. I'm doing some state tracking and comparisons right now to guard against changes and update the argument buffers when necessary. _This will get messier when layers themselves move over._

While this change lets us effectively support an unlimited number of bump maps - the structure passed between the vertex and fragment stages is still compile time defined. This has constrained the number of bump maps that can be supported by the shader to 8. (Which seems totally reasonable to me.)

Argument buffers work differently in Metal 3 and Metal 2. This code supports both approaches.

- Tier 2 argument buffers (Metal 3) give direct access to the underlying buffer. I can directly store texture/sampler/etc addresses into those buffers.
- Tier 1 argument buffers (Metal 2) have to support hardware with an inconsistent memory layout. An encoder object must be created that manages those differences - which adds some overhead.

Not all contributor Macs will support Metal 3, Apple lists the Metal 3 requirements here:
https://support.apple.com/en-us/102894

Tier 1 and Tier 2 buffers are also a little different in the shader. Tier 1 buffers still need ids for each binding so the encoder knows what is being bound where. Tier 2 argument buffers give me direct access - so there is no need to binding ids.

All previous conversations about the quality of the bump map still apply here. The code is present for full quality bump maps. But we're emulating the D3D Plasma behavior of ignoring the "height" component of the normal map.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -212,7 +212,7 @@ struct plTier1Bumpmap
     char2 dTangentIndex         [[ id(dTangentIndexID) ]];
     texture2d<half> bumpTexture [[ id(textureID) ]];
     sampler bumpTextureSampler  [[ id(samplerID) ]];
-    float3 scale                [[ id(dScaleID) ]];
+    float scale                 [[ id(dScaleID) ]];
 };
     
 #if __METAL_VERSION__ >= 300
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -331,8 +331,10 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
 
     
     for (int bumpMap=0; bumpMap<numBumpMaps; bumpMap++) {
-        (&out.T)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[0]], 0.f)). xyz;
-        (&out.B)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[1]], 0.f)). xyz;
+        const float4 T = float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[0]], 0.f);
+        const float4 B = float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[1]], 0.f);
+        (&out.T)[bumpMap] = normalize(uniforms.localToWorldMatrix * T). xyz;
+        (&out.B)[bumpMap] = normalize(uniforms.localToWorldMatrix * B).xyz;
     }
     
     out.position = vCamPosition * uniforms.projectionMatrix;
@@ -519,12 +521,18 @@ fragment half4 pipelineFragmentShader(ColorInOut in [[stage_in]],
     constexpr bool performBaseLighting = bumpMapIsAdditive || !bumpMap;
     
     if (performBaseLighting) {
-        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
+        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(
+            lighting,
+            in.vtxColor,
+            materialLighting,
+            in.worldPos,
+            in.normal
+       ) : in.vtxColor;
     }
     
     if (bumpMap) {
         float3 normal = 0.f;
-        for (size_t bumpIndex=0; bumpIndex<numBumpMaps; bumpIndex++) {
+        for (int bumpIndex=0; bumpIndex<numBumpMaps; bumpIndex++) {
             device ShaderBumpMapType& bump = bumpInfo[bumpIndex];
             float3 sampleCoord = in.texCoord1;
             half3 bumpNormal = bump.bumpTexture.sample(bump.bumpTextureSampler, sampleCoord.xy).rgb;
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -69,8 +69,10 @@ enum plUVWSrcModifiers: uint32_t
 
 using namespace metal;
 
-constant const bool perPixelLighting [[ function_constant(FunctionConstantPerPixelLighting)    ]];
-constant const bool perVertexLighting = !perPixelLighting;
+constant const bool     perPixelLighting [[ function_constant(FunctionConstantPerPixelLighting)    ]];
+constant const bool     perVertexLighting = !perPixelLighting;
+constant const uint8_t  numBumpMaps [[ function_constant(FunctionConstantNumBumpMaps)    ]];
+constant const bool     bumpMap = numBumpMaps > 0;
 
 constant const uint8_t sourceType1 [[ function_constant(FunctionConstantSources + 0)    ]];
 constant const uint8_t sourceType2 [[ function_constant(FunctionConstantSources + 1)    ]];
@@ -122,6 +124,15 @@ constant const bool hasCubicTexture5 = (sourceType5 == PassTypeCubicTexture && h
 constant const bool hasCubicTexture6 = (sourceType6 == PassTypeCubicTexture && hasLayer6);
 constant const bool hasCubicTexture7 = (sourceType7 == PassTypeCubicTexture && hasLayer7);
 constant const bool hasCubicTexture8 = (sourceType8 == PassTypeCubicTexture && hasLayer8);
+    
+constant const bool hasBumpMap1 = (numBumpMaps > 0);
+constant const bool hasBumpMap2 = (numBumpMaps > 1);
+constant const bool hasBumpMap3 = (numBumpMaps > 2);
+constant const bool hasBumpMap4 = (numBumpMaps > 3);
+constant const bool hasBumpMap5 = (numBumpMaps > 4);
+constant const bool hasBumpMap6 = (numBumpMaps > 5);
+constant const bool hasBumpMap7 = (numBumpMaps > 6);
+constant const bool hasBumpMap8 = (numBumpMaps > 7);
 
 struct FragmentShaderArguments
 {
@@ -167,6 +178,24 @@ typedef struct
     float3 texCoord6 [[ function_constant(hasLayer6) ]];
     float3 texCoord7 [[ function_constant(hasLayer7) ]];
     float3 texCoord8 [[ function_constant(hasLayer8) ]];
+    
+    float3 T    [[ function_constant(hasBumpMap1) ]];
+    float3 T1   [[ function_constant(hasBumpMap2) ]];
+    float3 T2   [[ function_constant(hasBumpMap3) ]];
+    float3 T3   [[ function_constant(hasBumpMap4) ]];
+    float3 T4   [[ function_constant(hasBumpMap5) ]];
+    float3 T5   [[ function_constant(hasBumpMap6) ]];
+    float3 T6   [[ function_constant(hasBumpMap7) ]];
+    float3 T7   [[ function_constant(hasBumpMap8) ]];
+    float3 B    [[ function_constant(hasBumpMap1) ]];
+    float3 B1   [[ function_constant(hasBumpMap2) ]];
+    float3 B2   [[ function_constant(hasBumpMap3) ]];
+    float3 B3   [[ function_constant(hasBumpMap4) ]];
+    float3 B4   [[ function_constant(hasBumpMap5) ]];
+    float3 B5   [[ function_constant(hasBumpMap6) ]];
+    float3 B6   [[ function_constant(hasBumpMap7) ]];
+    float3 B7   [[ function_constant(hasBumpMap8) ]];
+    
     half4 vtxColor   [[ centroid_perspective ]];
     half4 fogColor;
 } ColorInOut;
@@ -177,6 +206,19 @@ struct Lighting
     constant plMetalShaderActiveLight* activeLights      [[ buffer(ShaderActiveLights) ]];
     constant uint& lightCount [[ buffer(ShaderActiveLightCount)  ]];
 };
+    
+struct plTier1Bumpmap
+{
+    char2 dTangentIndex [[ id(dTangentIndexID) ]];
+    texture2d<half> bumpTexture [[ id(textureID) ]];
+    float3 scale [[ id(dScaleID) ]];
+};
+    
+#if __METAL_VERSION__ >= 300
+typedef plMetalBumpmap ShaderBumpMapType;
+#else
+typedef plTier1Bumpmap ShaderBumpMapType;
+#endif
 
 typedef struct
 {
@@ -250,7 +292,8 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
                                        constant VertexUniforms & uniforms   [[ buffer(VertexShaderArgumentFixedFunctionUniforms) ]],
                                        Lighting lighting [[ function_constant(perVertexLighting) ]],
                                        constant plMaterialLightingDescriptor & materialLighting   [[ buffer(VertexShaderArgumentMaterialLighting), function_constant(perVertexLighting) ]],
-                                       constant float4x4 & blendMatrix1     [[ buffer(VertexShaderArgumentBlendMatrix1), function_constant(temp_hasOnlyWeight1) ]])
+                                       constant float4x4 & blendMatrix1     [[ buffer(VertexShaderArgumentBlendMatrix1), function_constant(temp_hasOnlyWeight1) ]],
+                                       device ShaderBumpMapType *bumpInfo [[ buffer(BumpState), function_constant(bumpMap) ]])
 {
     ColorInOut out;
     // FVF ARGB color is in little endian order, so we need to swizzle it into RGBA
@@ -281,10 +324,16 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
 
     const float4 cameraSpaceNormal = normalize(((float4(in.normal, 0.f) * uniforms.localToWorldMatrix) * uniforms.worldToCameraMatrix));
 
-    for (size_t layer=0; layer<num_layers; layer++) {
+    for (int layer=0; layer<num_layers; layer++) {
         (&out.texCoord1)[layer] = uniforms.sampleLocation(layer, &in.texCoord1, cameraSpaceNormal, vCamPosition);
     }
 
+    
+    for (int bumpMap=0; bumpMap<numBumpMaps; bumpMap++) {
+        (&out.T)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[0]], 0.f)). xyz;
+        (&out.B)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[1]], 0.f)). xyz;
+    }
+    
     out.position = vCamPosition * uniforms.projectionMatrix;
 
     return out;
@@ -450,9 +499,47 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
 fragment half4 pipelineFragmentShader(ColorInOut in [[stage_in]],
                                       const FragmentShaderArguments fragmentShaderArgs,
                                       Lighting lighting [[ function_constant(perPixelLighting) ]],
-                                      constant plMaterialLightingDescriptor & materialLighting   [[ buffer(FragmentShaderArgumentMaterialLighting), function_constant(perPixelLighting) ]])
+                                      constant plMaterialLightingDescriptor & materialLighting   [[ buffer(FragmentShaderArgumentMaterialLighting), function_constant(perPixelLighting) ]],
+                                      device ShaderBumpMapType *bumpInfo [[ buffer(BumpState), function_constant(bumpMap) ]])
 {
-    const half4 lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
+    half4 lightingContributionColor = half4(0.h);
+    
+    /*
+     This controls the bumpmap style:
+     - Additive bump maps do a secondary hilight pass on the material. This is
+     how Cyan shipped bump maps in Uru. These bump maps ignore the blue channel
+     of the normal map file.
+     - If bumpMapIsAdditive is false, full normal mapping is performed on a per
+     pixel basis. This is not how the engine shipped and will produce different
+     results than the original age artist may have expected. But it will be
+     proper normal mapping.
+     */
+    constexpr bool bumpMapIsAdditive = true;
+    constexpr bool performBaseLighting = bumpMapIsAdditive || !bumpMap;
+    
+    if (performBaseLighting) {
+        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
+    }
+    
+    if (bumpMap) {
+        float3 normal = 0.f;
+        for (size_t bump=0; bump<numBumpMaps; bump++) {
+            float3 sampleCoord = in.texCoord1;
+            half3 bumpNormal = bumpInfo[bump].bumpTexture.sample(fragmentShaderArgs.samplers, sampleCoord.xy).rgb;
+            
+            bumpNormal -= 0.5f;
+            bumpNormal *= 2.f;
+            
+            float3x3 TBN = float3x3(in.T, in.B, in.normal);
+            normal += TBN * float3(bumpNormal) * bumpInfo[bump].scale;
+        }
+        
+        normal = normalize(normal);
+        if (performBaseLighting) {
+            normal.z = 0.f;
+        }
+        lightingContributionColor += perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, normal) : in.vtxColor;
+    }
     half4 currentColor = lightingContributionColor;
 
     /*
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
@@ -209,9 +209,10 @@ struct Lighting
     
 struct plTier1Bumpmap
 {
-    char2 dTangentIndex [[ id(dTangentIndexID) ]];
+    char2 dTangentIndex         [[ id(dTangentIndexID) ]];
     texture2d<half> bumpTexture [[ id(textureID) ]];
-    float3 scale [[ id(dScaleID) ]];
+    sampler bumpTextureSampler  [[ id(samplerID) ]];
+    float3 scale                [[ id(dScaleID) ]];
 };
     
 #if __METAL_VERSION__ >= 300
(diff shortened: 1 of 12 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -449,9 +499,48 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
...
+    constexpr bool bumpMapIsAdditive = true;
+    constexpr bool performBaseLighting = bumpMapIsAdditive || !bumpMap;
+    
+    if (performBaseLighting) {
+        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
+    }
+    
+    if (bumpMap) {
+        float3 normal = 0.f;
+        for (size_t bumpIndex=0; bumpIndex<numBumpMaps; bumpIndex++) {
```
colincornaby: Nit picking on myself - I don't think I'm supposed to use size_t in a Metal shader for iterating in a loop, but I need to look that one up.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +322,64 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        fBumps.emplace_back(std::nullopt);
+    }
+    auto& buffer = fBumps[pass];
+    auto  bumpCount = bumps.size();
+    if (bumpCount == 0) {
+        buffer.reset();
+        return;
+    }
+    if (!buffer.has_value() || buffer.value()->GetNumElements() != bumpCount) {
+        buffer = std::make_shared<plMetalBumpArgumentBuffer>(fDevice, bumps.size());
```
colincornaby: Nitpicking myself: Could this use std::unique_ptr instead?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.cpp
```diff
@@ -0,0 +1,140 @@
...
+    _bumps.resize(numElements);
+}
+
+void plMetalBumpArgumentBuffer::Set(const std::vector<plMetalBumpMapping>& bumps)
+{
+    if (CheckBuffer(bumps)) {
+        return;
+    }
+    ConfigureBuffer();
+    int i = 0;
```
Hoikas: If you need an index, we should probably not be using a range based for. Instead, do an old fashioned `for (int = 0; i < bumps.size(); ++i)` and get the vector element you need.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.cpp
```diff
@@ -0,0 +1,140 @@
...
+    encoder->setVertexBuffer(GetBuffer(), 0, BumpState);
+    encoder->setFragmentBuffer(GetBuffer(), 0, BumpState);
+}
+
+bool plMetalBumpArgumentBuffer::CheckBuffer(const std::vector<plMetalBumpMapping>& bumps)
+{
+    if (bumps.size() != fNumElements || GetBuffer() == nullptr) {
+        return false;
+    }
+    int i = 0;
```
Hoikas: Remove range based for

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,177 @@
...
+    // hardware.
+    Tier2 = MTL::ArgumentBuffersTier2,
+#endif
+};
+
+template <class T>
+class plMetalArgumentBuffer
+{
+public:
+    plMetalArgumentBuffer(MTL::Device* device, size_t numElements) : fDevice(device), fNumElements(numElements), fEncoder(nullptr)
```
Hoikas:
```suggestion
    plMetalArgumentBuffer(MTL::Device* device, size_t numElements)
        : fDevice(device), fNumElements(numElements), fEncoder()
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,177 @@
...
+    }
+    T*     ValueAt(size_t i) { return fValue[i]; }
+    size_t GetNumElements() { return fNumElements; }
+    ~plMetalArgumentBuffer()
+    {
+        fBuffer[0]->release();
+        fBuffer[1]->release();
+        fBuffer[2]->release();
+        fEncoder->release();
+    }
```
Hoikas: Please add blank lines

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,177 @@
...
+    };
+    constexpr MTL::Buffer* GetBuffer()
+    {
+        if (fCurrentBufferIndex == -1) {
+            return nullptr;
+        }
+        return fBuffer[fCurrentBufferIndex];
+    }
+    T*     ValueAt(size_t i) { return fValue[i]; }
+    size_t GetNumElements() { return fNumElements; }
```
Hoikas:
```suggestion
    size_t GetNumElements() const { return fNumElements; }
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
Hoikas: Continue this line of investigation. I strongly perfer to not see `std::shared_ptr`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +322,64 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        plLayerInterface* layer = fMaterial->GetLayer(layerIdx + bumpLayer);
+        uint32_t          miscFlags = layer->GetMiscFlags();
+        switch (miscFlags & hsGMatState::kMiscBumpChans) {
+            case hsGMatState::kMiscBumpDu:
+                bumpMapping.dTangentUIndex = layer->GetUVWSrc();
+            case hsGMatState::kMiscBumpDv:
+                bumpMapping.dTangentVIndex = layer->GetUVWSrc();
+            case hsGMatState::kMiscBumpDw:
+            default:
+                break;
```
Hoikas: Should this be a `DEFAULT_FATAL`?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.h
```diff
@@ -119,10 +122,14 @@ class plMetalMaterialShaderRef : public plMetalDeviceRef
...
-    void                                                 EncodeTransform(const plLayerInterface* layer, UVOutDescriptor *transform);
-    std::vector<std::vector<plLayerInterface*>>          fPasses;
-    std::vector<struct plMetalFragmentShaderDescription> fFragmentShaderDescriptions;
+    void IEatBumpmapLayers(uint32_t& layer, std::vector<plMetalBumpMapping>& bumpsOut);
+    void IEncodeBumpmapLayers(const std::vector<plMetalBumpMapping>& bumps, uint32_t pass);
+    void IBuildLayerTexture(MTL::RenderCommandEncoder* encoder, const uint32_t offsetFromRootLayer, plLayerInterface* layer);
+    void EncodeTransform(const plLayerInterface* layer, UVOutDescriptor* transform);
+
+    std::vector<std::vector<plLayerInterface*>>                            fPasses;
+    std::vector<std::optional<std::shared_ptr<plMetalBumpArgumentBuffer>>> fBumps;
```
Hoikas: If it's a pointer, we don't need to be std::optional, IMO.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipelineState.h
```diff
@@ -148,12 +148,18 @@ struct plMetalFragmentShaderDescription
...
 
     bool operator==(const plMetalFragmentShaderDescription& p) const
     {
-        bool match = fNumLayers == p.fNumLayers && memcmp(fPassTypes, p.fPassTypes, sizeof(fPassTypes)) == 0 && memcmp(fBlendModes, p.fBlendModes, sizeof(fBlendModes)) == 0 && memcmp(fMiscFlags, p.fMiscFlags, sizeof(fMiscFlags)) == 0 && fUsePerPixelLighting == p.fUsePerPixelLighting;
+        bool match = fNumLayers == p.fNumLayers &&
+                                    memcmp(fPassTypes, p.fPassTypes, sizeof(fPassTypes)) == 0 &&
+                                    memcmp(fBlendModes, p.fBlendModes, sizeof(fBlendModes)) == 0 &&
+                                    memcmp(fMiscFlags, p.fMiscFlags, sizeof(fMiscFlags)) == 0 &&
+                                    fUsePerPixelLighting == p.fUsePerPixelLighting &&
+                                    fNumBumpMaps == p.fNumBumpMaps;
```
Hoikas: This is so long that it wraps in split diff mode.

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -449,9 +499,48 @@ half4 FragmentShaderArguments::sampleLayer(const size_t index, const half4 verte
...
+     - If bumpMapIsAdditive is false, full normal mapping is performed on a per
+     pixel basis. This is not how the engine shipped and will produce different
+     results than the original age artist may have expected. But it will be
+     proper normal mapping.
+     */
+    constexpr bool bumpMapIsAdditive = true;
+    constexpr bool performBaseLighting = bumpMapIsAdditive || !bumpMap;
+    
+    if (performBaseLighting) {
+        lightingContributionColor = perPixelLighting ? calcLitMaterialColor(lighting, in.vtxColor, materialLighting, in.worldPos, in.normal) : in.vtxColor;
```
Hoikas: This line is so long it wraps in split diff mode.

Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/FixedPipelineShaders.metal
```diff
@@ -280,10 +324,16 @@ vertex ColorInOut pipelineVertexShader(Vertex in [[stage_in]],
...
 
-    for (size_t layer=0; layer<num_layers; layer++) {
+    for (int layer=0; layer<num_layers; layer++) {
         (&out.texCoord1)[layer] = uniforms.sampleLocation(layer, &in.texCoord1, cameraSpaceNormal, vCamPosition);
     }
 
+    
+    for (int bumpMap=0; bumpMap<numBumpMaps; bumpMap++) {
+        (&out.T)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[0]], 0.f)). xyz;
+        (&out.B)[bumpMap] = normalize(uniforms.localToWorldMatrix * float4((&in.texCoord1)[bumpInfo[bumpMap].dTangentIndex[1]], 0.f)). xyz;
```
Hoikas: Wraps in split diff

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
colincornaby: The switch is complete so I removed the default case.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
```diff
@@ -301,6 +322,63 @@ void plMetalMaterialShaderRef::ILoopOverLayers()
...
+        }
+    }
+    plLayerInterface* bumpTextureLayer = fMaterial->GetLayer(layerIdx + 3);
+    bumpMapping.scale = bumpTextureLayer->GetRuntimeColor().r;
+    fPipeline->CheckTextureRef(bumpTextureLayer);
+    bumpMapping.texture = static_cast<plMetalTextureRef*>(bumpTextureLayer->GetTexture()->GetDeviceRef())->fTexture;
+    MTL::SamplerState* samplerState = fPipeline->fDevice.SampleStateForClampFlags(hsGMatState::hsGMatClampFlags(bumpTextureLayer->GetClampFlags()));
+    bumpMapping.sampler = samplerState;
+    bumpsOut.emplace_back(bumpMapping);
```
Hoikas: It would be nice if we were doing `plMetalBumpMapping& bumpMapping = bumpsOut.emplace_back()` so that the bump mapping object is only constructed once.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    descriptors[3] = MTL::ArgumentDescriptor::argumentDescriptor();
+    descriptors[3]->setIndex(dScaleID);
+    descriptors[3]->setDataType(MTL::DataTypeFloat);
+
+    NS::Array* array = NS::Array::array((const NS::Object* const*)descriptors, 4);
+    return array;
+}
+
+plMetalBumpArgumentBuffer::plMetalBumpArgumentBuffer(MTL::Device* device, size_t numElements) : plMetalArgumentBuffer<plMetalBumpmap>(device, numElements)
```
Hoikas:
```suggestion
plMetalBumpArgumentBuffer::plMetalBumpArgumentBuffer(MTL::Device* device, size_t numElements)
    : plMetalArgumentBuffer<plMetalBumpmap>(device, numElements)
```

I'm not sure how I missed this before

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalMaterialShaderRef.cpp
Hoikas: Looks like this is still outstanding

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    Tier2 = MTL::ArgumentBuffersTier2,
+#endif
+};
+
+template <class T>
+class plMetalArgumentBuffer
+{
+public:
+    plMetalArgumentBuffer(plMetalDevice* device, size_t numElements)
+        : fDevice(device), fNumElements(numElements), fEncoder(nullptr)
```
Hoikas:
```suggestion
        : fDevice(device), fNumElements(numElements), fEncoder()
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    {
+        fCurrentBufferIndex = -1;
+        auto tier = std::min(device->ArgumentBuffersTier(), MTL::ArgumentBuffersTier2);
+        fTier = plMetalArgumentBufferTier(tier);
+        fBuffer[0] = nullptr;
+        fBuffer[1] = nullptr;
+        fBuffer[2] = nullptr;
```
Hoikas: Shouldn't this be done in the initializer?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    };
+    
+    constexpr MTL::Buffer* GetBuffer()
```
Hoikas:
```suggestion
    MTL::Buffer* GetBuffer() const
```

I don't know that this function being `constexpr`  will actually have any impact. `fCurrentBufferIndex` is determined at runtime, right?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    
+    T*     ValueAt(size_t i) { return fValue[i]; }
+    
+    size_t GetNumElements() const { return fNumElements; }
+    
+    ~plMetalArgumentBuffer()
+    {
+        fBuffer[0]->release();
+        fBuffer[1]->release();
+        fBuffer[2]->release();
```
Hoikas:
```suggestion
        for (auto& buffer : fBuffer) {
            if (buffer)
                buffer->release();
        }
```

C style arrays can use range based for loops. Further, each element of `fBuffer` is initialized to `nullptr`, so we need to be cautious about dereferencing them.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+        fEncoder->release();
```
Hoikas:
```suggestion
        if (fEncoder)
            fEncoder->release();
```

This is initialized to `nullptr`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    // Host Plasma device for buffers
+    plMetalDevice*            fDevice;
+    // Encoder for tier 1 buffers, don't use in tier 2
+    MTL::ArgumentEncoder*     fEncoder;
+    // Number of array elements supported per buffer
+    size_t                    fNumElements;
+    // Tier of argument buffers targeted by this instance
+    plMetalArgumentBufferTier fTier;
+    // Current buffer in the triple buffer rotation
+    size_t                    fCurrentBufferIndex;
```
Hoikas: These should be declared before they are used.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+        fCurrentBufferIndex = (fCurrentBufferIndex + 1) % 3;
+        switch (fTier) {
+            case plMetalArgumentBufferTier::Tier1:
+                ConfigureTier1();
+                break;
+#ifdef METAL_3_SDK
+            case plMetalArgumentBufferTier::Tier2:
+                ConfigureTier2();
+                break;
+#endif
```
Hoikas: It feels like `Tier2` should be defined to a dummy/impossible value if the Metal 3 SDK is not available. Then we can avoid these excessive ifdefs.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    void Set(const std::vector<plMetalBumpMapping>& bumps);
+    plMetalBumpArgumentBuffer(plMetalDevice* device, size_t numElements);
+    void Bind(MTL::RenderCommandEncoder* encoder);
+    bool CheckBuffer(const std::vector<plMetalBumpMapping>& bumps);
+
+protected:
+    virtual NS::Array* GetArgumentDescriptors() override;
+
+private:
+    std::vector<plMetalBumpMapping> _bumps;
```
Hoikas: Fields should begin with `f`, not `_`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+    // Number of array elements supported per buffer
+    size_t                    fNumElements;
+    // Tier of argument buffers targeted by this instance
+    plMetalArgumentBufferTier fTier;
+    // Current buffer in the triple buffer rotation
+    size_t                    fCurrentBufferIndex;
+    // Subclasses must override this to create an encoder for
+    // tier 1 buffers. Encoder creation needs to know the
+    // buffer layout.
+    virtual NS::Array*        GetArgumentDescriptors() = 0;
```
Hoikas: This looks like it should probably be a `const` method.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+
+    void ConfigureBuffer()
+    {
+        fCurrentBufferIndex = (fCurrentBufferIndex + 1) % 3;
```
Hoikas: Instead of hardcoding the 3, should we do something like


```suggestion
        fCurrentBufferIndex = (fCurrentBufferIndex + 1) % std::size(fBuffer);
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,185 @@
...
+#endif
+};
+
+struct plMetalBumpMapping
+{
+    uint8_t             dTangentUIndex;
+    uint8_t             dTangentVIndex;
+    float               scale;
+    MTL::Texture*       texture;
+    MTL::SamplerState*  sampler;
```
Hoikas: Fields should begin with `f`

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
colincornaby: I'm being careful here because if the Metal 3 SDK is not available these symbols _may_ not be available to the compiler. Technically the Metal header will dynamically look up symbols blah blah blah... but Tier 2 symbols won't technically be available to the compiler.

I could require everyone build against the Metal 3 SDK but that will move the minimum required build toolchain to Xcode 14.

Hoikas: If you declare dummy values, what the SDK provides is irrelevant.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
colincornaby: Calling functions on null pointer Obj-C types is a language feature. Metal-cpp inherits this behavior because it's just a thing wrapper over Obj-C. The if check is unnecessary.

(Annoying that Metal-cpp causes Obj-Cisms to become a thing in C++. But at least anywhere you see a retain/release it can be assumed that it's ok to call that on a nullptr.)

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
colincornaby: Same feedback here. I can implement the loop, but in since this is Obj-C, it's ok to send messages/call functions on nullptr.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
Hoikas: That sounds like we'd be depending on an implementation detail, and I can't say I'm a fan of that.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
Hoikas: My purpose for suggesting the loop is to avoid hardcoding the existence of specifically three buffers.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
colincornaby: This is an overtly stated "feature" of the library:

> Similar to Objective-C, it is legal to call any method, including `retain()` and `release()`, on `nullptr` "objects". While calling methods on `nullptr` still does incur in function call overhead, the effective result is equivalent of a NOP.

Unfortunately can't directly link to the readme in since it's inside a zip file. :(

colincornaby:
I'm working through clearing my pull requests. This one is important because it finalizes bump mapping in the Metal renderer. It's also holding up another PR with performance improvements - getting concerned I'll lose my mental context on the changes I made in that PR. Hoping we can get this one through soon.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
colincornaby: Hey @Hoikas - I've been thinking about this conversation and I had a realization. Please check out the last commit. I think it should resolve the issues you had here.

We're sort of deadlocked on the retain/release thing. But I remembered that Metal-cpp has a NS::SharedPtr type. It adds C++ SharedPtr semantics to Obj-C types with no extra overhead. That _should_ resolve the issue. I don't have to manually release the objects any more - so there is not more discussion on potentially sending a release to a nullptr. It also simplifies the code.

NS::SharedPtr wasn't originally part of Metal-cpp, so it wasn't originally adopted by this renderer. I think we should adopt it now.

There is no NS::UniquePtr - so I'll have to ask for a bit of understanding there. But all Obj-C pointer types already have SharedPtr mechanics built in - so there wouldn't be a point in NS::UniquePtr.

Hoikas: Yes, this looks a lot better. I see a few very, very, very minor things left, otherwise this will be a 👍 from me. I'd like for @dpogue to look behind me, though, if possible. My mac and Metal knowledge is still quite poor.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+void plMetalBumpArgumentBuffer::Set(const std::vector<plMetalBumpMapping>& bumps)
+{
+    if (CheckBuffer(bumps)) {
+        return;
+    }
+    ConfigureBuffer();
+    if (fTier == plMetalArgumentBufferTier::Tier1) {
+        // Tier 1, because Tier 1 doesn't guarantee memory layout, go through the encoder
+        for (int i = 0; i < bumps.size(); ++i) {
```
Hoikas: These two for loops - should the `i`ndex be typed `size_t` instead of `int`?

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalArgumentBuffer.h
```diff
@@ -0,0 +1,174 @@
...
+    // Host Plasma device for buffers
+    plMetalDevice*              fDevice;
+    // Encoder for tier 1 buffers, don't use in tier 2
+    NS::SharedPtr<MTL::ArgumentEncoder>   fEncoder;
+    // Number of array elements supported per buffer
+    size_t                      fNumElements;
+    // Tier of argument buffers targeted by this instance
+    plMetalArgumentBufferTier fTier;
+    // Current buffer in the triple buffer rotation
+    size_t                      fCurrentBufferIndex;
```
Hoikas: It would be nice to move these members up to the top.

dpogue (approved):
LGTM

PR #1830, opened 2026-01-01
Support for VP9 decoding with Apple's VideoToolbox

colincornaby:
This PR adds VideoToolbox support alongside the exist libwebm support. It can optionally replace the libvpx decoder. This decoder is only available on Apple devices that feature hardware decoding support - Apple does not include a software decoder and this code will fallback to libvpx in that case. Apple has not provided hardware decoding drivers for all hardware that could support it - this functionality seems mostly limited to Apple Silicon devices.

For other platforms like iOS and visionOS where hardware decode support is guaranteed - this turns libvpx into an optional dependency. This is especially helpful for visionOS where there are no working builds of libvpx.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/CMakeLists.txt
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/CMakeLists.txt
@@ -32,19 +32,5 @@ target_link_libraries(
         $<$<BOOL:${USE_VIDEOTOOLBOX}>:VideoToolbox::VideoToolbox>
 )
 
-if(USE_VPX)
-    target_compile_definitions(pfMoviePlayer
-        PRIVATE
-        USE_VPX
-    )
-endif(USE_VPX)
-
-if(USE_VIDEOTOOLBOX)
-    target_compile_definitions(pfMoviePlayer
-        PRIVATE
-        USE_VIDEOTOOLBOX
-    )
-endif(USE_VIDEOTOOLBOX)
-
 source_group("Source Files" FILES ${pfMoviePlayer_SOURCES})
 source_group("Header Files" FILES ${pfMoviePlayer_HEADERS})
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
@@ -54,6 +54,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #   define WEBM_CODECID_VP9 "V_VP9"
 #   define WEBM_CODECID_OPUS "A_OPUS"
+#   ifdef USE_VIDEOTOOLBOX
+#       include "plVTDecoder.h"
+#   endif
 #endif
 
 #include "hsConfig.h"
@@ -70,7 +73,6 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plPlanarImage.h"
 #include "plResMgr/plLocalization.h"
 #include "plStatusLog/plStatusLog.h"
-#include "plVTDecoder.h"
 
 #define SAFE_OP(x, err) \
 { \
@@ -83,6 +85,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 }
 
 // =====================================================
+
+#ifdef USE_VPX
+class plVPXMovieFrame : public plMovieFrame
+{
+public:
+    plVPXMovieFrame(vpx_image_t* image) : fImage(image)
+    {
+    }
+
+    uint32_t GetWidth() override
+    {
+        return fImage->d_w;
+    }
+
+    uint32_t GetHeight() override
+    {
+        return fImage->d_h;
+    }
+
+    int32_t* GetStrides() override
+    {
+        return fImage->stride;
+    }
+
+    unsigned char** GetPlanes() override
+    {
+        return fImage->planes;
+    }
+
+    Format GetFormat() override
+    {
+        return plMovieFrame::Format::I420;
+    }
+
+    ~plVPXMovieFrame()
+    {
+        vpx_img_free(fImage);
+    }
+
+private:
+    vpx_image_t* fImage;
+};
+#endif
 
 class VPX
 {
@@ -110,12 +155,12 @@ public:
         return instance;
     }
 
-    plMovieFrameRef Decode(uint8_t* buf, uint32_t size)
+    std::unique_ptr<plMovieFrame> Decode(uint8_t* buf, uint32_t size)
     {
         if (vpx_codec_decode(&codec, buf, size, nullptr, 0) != VPX_CODEC_OK) {
             const char* detail = vpx_codec_error_detail(&codec);
             hsAssert(false, detail ? detail : "unspecified decode error");
-            return std::nullopt;
+            return nullptr;
         }
 
         vpx_codec_iter_t  iter = nullptr;
@@ -124,31 +169,13 @@ public:
         auto vpxFrame = vpx_codec_get_frame(&codec, &iter);
         
         if (vpxFrame->fmt != VPX_IMG_FMT_I420) {
-            // We only handle YUX 420 right now...
+            // We only handle YUV 420 right now...
+            hsAssert(0, "VPX decoded a frame in an unexpected format");
             vpx_img_free(vpxFrame);
-            return std::nullopt;
+            return nullptr;
         }
 
-        plMovieFrame* frame = new plMovieFrame();
-        frame->fFormat = plMovieFrame::Format::I420;
-        frame->fWidth = vpxFrame->d_w;
-        frame->fHeight = vpxFrame->d_h;
-
-        frame->fPlanes[0] = vpxFrame->planes[0];
-        frame->fPlanes[1] = vpxFrame->planes[1];
-        frame->fPlanes[2] = vpxFrame->planes[2];
-        frame->fPlanes[3] = vpxFrame->planes[3];
-
-        frame->fStride[0] = vpxFrame->stride[0];
-        frame->fStride[1] = vpxFrame->stride[1];
-        frame->fStride[2] = vpxFrame->stride[2];
-        frame->fStride[3] = vpxFrame->stride[3];
-
-        return std::unique_ptr<plMovieFrame, void (*)(plMovieFrame*)>(frame, [](plMovieFrame* ptr) {
-            auto imageBuffer = static_cast<vpx_image*>(ptr->fContext);
-            vpx_img_free(imageBuffer);
-            delete ptr;
-        });
+        return std::make_unique<plVPXMovieFrame>(vpxFrame);
     }
 #else
     VPX() { }
@@ -174,7 +201,7 @@ public:
 #ifdef USE_WEBM
         // If we have no block yet, grab the first one
         if (!fCurrentBlock)
-            fStatus = fTrack->GetFirst(fCurrentBlock);
+            fStatus = int32_t(fTrack->GetFirst(fCurrentBlock));
 
         // Continue through the blocks until our current movie time
         while (fCurrentBlock && fStatus == 0) {
@@ -189,7 +216,7 @@ public:
                     data.Read(reader, buf);
                     frames.push_back(std::make_tuple(std::unique_ptr<uint8_t>(buf), static_cast<int32_t>(data.len)));
                 }
-                fStatus = fTrack->GetNext(fCurrentBlock, fCurrentBlock);
+                fStatus = int32_t(fTrack->GetNext(fCurrentBlock, fCurrentBlock));
             } else {
                 // We've got all frames that have to play... come back for more later!
                 return true;
@@ -345,35 +372,38 @@ bool plMoviePlayer::ICheckLanguage(const mkvparser::Track* track)
 void plMoviePlayer::IProcessVideoFrame(const std::vector<blkbuf_t>& frames)
 {
 #ifdef USE_WEBM
-    plMovieFrameRef img;
+    std::unique_ptr<plMovieFrame> img;
 
     // We have to decode all the frames, but we only want to display the most recent one to the user.
     for (const auto& frame : frames) {
         const std::unique_ptr<uint8_t>& buf = std::get<0>(frame);
         uint32_t size = static_cast<uint32_t>(std::get<1>(frame));
+        // Assume only one decoder is active here
 #ifdef USE_VIDEOTOOLBOX
-        if (!img && fVt)
+        if (fVt)
             img = fVt->DecodeNextFrame(buf.get(), size);
 #endif
 #ifdef USE_VPX
-        if (!img && fVpx)
+        if (fVpx)
             img = fVpx->Decode(buf.get(), size);
 #endif
     }
 
+    bool decoded = false;
     if (img) {
         fLastImg = std::move(img);
+        decoded = true;
     }
     
     // We need to refill the buffer if there is a new image or a new buffer
-    if (img || (!fTexture->GetDeviceRef() && fLastImg)) {
+    if (decoded || (!fTexture->GetDeviceRef() && fLastImg)) {
         // According to VideoLAN[1], I420 is the most common image format in videos. I am inclined to believe this as our
         // attemps to convert the common Uru videos use I420 image data. So, as a shortcut, we will only implement that format.
         // If for some reason we need other formats, please, be my guest!
         // [1] = http://wiki.videolan.org/YUV#YUV_4:2:0_.28I420.2FJ420.2FYV12.29
-        switch (fLastImg->get()->fFormat) {
+        switch (fLastImg->GetFormat()) {
         case plMovieFrame::Format::I420:
-            plPlanarImage::Yuv420ToRgba(fLastImg->get()->fWidth, fLastImg->get()->fHeight, fLastImg->get()->fStride, fLastImg->get()->fPlanes, reinterpret_cast<uint8_t*>(fTexture->GetImage()));
+            plPlanarImage::Yuv420ToRgba(fLastImg->GetWidth(), fLastImg->GetHeight(), fLastImg->GetStrides(), fLastImg->GetPlanes(), reinterpret_cast<uint8_t*>(fTexture->GetImage()));
             break;
 
         DEFAULT_FATAL(fLastImg->fmt);
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
@@ -66,22 +66,22 @@ namespace mkvparser
 typedef std::tuple<std::unique_ptr<uint8_t>, int32_t> blkbuf_t;
 typedef struct vpx_image vpx_image_t;
 
-struct plMovieFrame
+class plMovieFrame
 {
+public:
     enum class Format
     {
         I420
     };
 
-    uint32_t       fWidth;
-    uint32_t       fHeight;
-    int32_t        fStride[4];
-    unsigned char* fPlanes[4];
-    Format         fFormat;
-    void*          fContext;
-};
+    virtual uint32_t        GetWidth() = 0;
+    virtual uint32_t        GetHeight() = 0;
+    virtual int32_t*        GetStrides() = 0;
+    virtual unsigned char** GetPlanes() = 0;
+    virtual Format          GetFormat() = 0;
 
-typedef std::optional<std::unique_ptr<plMovieFrame, void (*)(plMovieFrame*)>> plMovieFrameRef;
+    virtual ~plMovieFrame() {}
+};
 
 class plMoviePlayer
 {
@@ -92,12 +92,16 @@ protected:
 #ifdef USE_WEBM
     mkvparser::MkvReader* fReader;
     std::unique_ptr<mkvparser::Segment> fSegment;
-    plMovieFrameRef fLastImg;
+    std::unique_ptr<plMovieFrame>       fLastImg;
 #endif
     std::unique_ptr<class TrackMgr> fAudioTrack, fVideoTrack; // TODO: vector of tracks?
     std::unique_ptr<class plWin32VideoSound> fAudioSound;
+#ifdef USE_VPX
     std::unique_ptr<class VPX> fVpx;
+#endif
+#ifdef USE_VIDEOTOOLBOX
     std::unique_ptr<class plVTDecoder> fVt;
+#endif
 
     int64_t fMovieTime, fLastFrameTime; // in ms
     hsPoint2 fPosition, fScale;
--- a/Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
+++ b/Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
@@ -49,7 +49,62 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plMoviePlayer.h"
 #include "plPlanarImage.h"
 
-OSStatus CreateFormatDescriptionFromVP9Track(
+class plVTMovieFrame : public plMovieFrame
+{
+public:
+    plVTMovieFrame(CVPixelBufferRef pixelBuffer) : fPixelBuffer(pixelBuffer)
+    {
+        CVPixelBufferLockBaseAddress(fPixelBuffer, 0);
+        CVPixelBufferRetain(fPixelBuffer);
+        fPlanes[0] = static_cast<unsigned char*>(CVPixelBufferGetBaseAddressOfPlane(pixelBuffer, 0));
+        fPlanes[1] = static_cast<unsigned char*>(CVPixelBufferGetBaseAddressOfPlane(pixelBuffer, 1));
+        fPlanes[2] = static_cast<unsigned char*>(CVPixelBufferGetBaseAddressOfPlane(pixelBuffer, 2));
+        fPlanes[3] = nullptr;
+
+        fStrides[0] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 0));
+        fStrides[1] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 1));
+        fStrides[2] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 2));
+        fStrides[3] = 0;
+    }
+
+    uint32_t GetWidth() override
+    {
+        return uint32_t(CVPixelBufferGetWidth(fPixelBuffer));
+    }
+
+    uint32_t GetHeight() override
+    {
+        return uint32_t(CVPixelBufferGetHeight(fPixelBuffer));
+    }
+
+    int32_t* GetStrides() override
+    {
+        return fStrides;
+    }
+
+    unsigned char** GetPlanes() override
+    {
+        return fPlanes;
+    }
+
+    Format GetFormat() override
+    {
+        return plMovieFrame::Format::I420;
+    }
+
+    ~plVTMovieFrame()
+    {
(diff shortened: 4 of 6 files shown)


Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -66,6 +67,23 @@ namespace mkvparser
...
+        I420
+    };
+
+    uint32_t       fWidth;
+    uint32_t       fHeight;
+    int32_t        fStride[4];
+    unsigned char* fPlanes[4];
+    Format         fFormat;
+    void*          fContext;
+};
```
colincornaby, line 84: Tried to be sensitive here in since I know this code is already earmarked for a refactor. Basic abstraction to keep plMoviePlayer away for decoder specifics.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -66,6 +67,23 @@ namespace mkvparser
...
+
+typedef std::optional<std::unique_ptr<plMovieFrame, void (*)(plMovieFrame*)>> plMovieFrameRef;
```
colincornaby: The optional might be overkill in since `std::unique_ptr` can be null.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+OSStatus CreateFormatDescriptionFromVP9Track(
+    const mkvparser::VideoTrack* video_track,
+    CMFormatDescriptionRef*      format_desc_out)
+{
+    // We need to reconstruct the ISO/IEC 14496 atom
+    // that would prepend the track in an MP4 container
+
+    // The full box container as specified by
+    // ISO/IEC 14496-12
```
colincornaby, line 115: This ISO doc is paywall so I can't link to it. However WebKit implements the atom that needs to be passed to the decoder in the same way with the same structure. The WebM documentation is public and assumes you know what a "full box" is.

Sources/Plasma/FeatureLib/pfMoviePlayer/CMakeLists.txt
```diff
@@ -26,7 +28,23 @@ target_link_libraries(
...
+        "$<$<BOOL:${USE_VIDEOTOOLBOX}>:-framework CoreMedia>"
+        $<$<BOOL:${USE_VIDEOTOOLBOX}>:VideoToolbox::VideoToolbox>
 )
 
+if(USE_VPX)
+    target_compile_definitions(pfMoviePlayer
+        PUBLIC
+        USE_VPX
+    )
+endif(USE_VPX)
```
Hoikas: This is already defined in `hsConfig.h`. Please use `hsConfig.h`.

Sources/Plasma/FeatureLib/pfMoviePlayer/CMakeLists.txt
```diff
@@ -26,7 +28,23 @@ target_link_libraries(
...
+
+if(USE_VIDEOTOOLBOX)
+    target_compile_definitions(pfMoviePlayer
+        PUBLIC
+        USE_VIDEOTOOLBOX
+    )
+endif(USE_VIDEOTOOLBOX)
```
Hoikas: Please add a cmakedefine to `hsConfig.h.cmake` for this.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -110,18 +112,45 @@ class VPX
...
+
+        frame->fStride[0] = vpxFrame->stride[0];
+        frame->fStride[1] = vpxFrame->stride[1];
+        frame->fStride[2] = vpxFrame->stride[2];
+        frame->fStride[3] = vpxFrame->stride[3];
+
+        return std::unique_ptr<plMovieFrame, void (*)(plMovieFrame*)>(frame, [](plMovieFrame* ptr) {
+            auto imageBuffer = static_cast<vpx_image*>(ptr->fContext);
+            vpx_img_free(imageBuffer);
+            delete ptr;
```
Hoikas: Woah. This isn't idiomatic at all. We should be doing this in the destructor of `plMovieFrame`.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
Hoikas: Agree. Remove `std::optional`. Prefer to remove this typedef entirely.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include "plVTDecoder.h"
+
+#include <mkvparser/mkvparser.h>
+#include <mkvparser/mkvreader.h>
+
+#include "plGImage/plMipmap.h"
+#include "plMoviePlayer.h"
+#include "plPlanarImage.h"
+
+OSStatus CreateFormatDescriptionFromVP9Track(
```
Hoikas:
```suggestion
static OSStatus CreateFormatDescriptionFromVP9Track(
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+    CFMutableDictionaryRef atoms = CFDictionaryCreateMutable(kCFAllocatorDefault,
+                                                             1,
+                                                             &kCFTypeDictionaryKeyCallBacks,
+                                                             &kCFTypeDictionaryValueCallBacks);
+
+    CFDictionarySetValue(atoms, CFSTR("vpcC"), boxData);
+
+    CFDictionarySetValue(extensions,
+                         kCMFormatDescriptionExtension_SampleDescriptionExtensionAtoms,
+                         atoms);
```
Hoikas: These lines are so long that they wrap when using split diff. Please use a single handing indent.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return new plVTDecoder(track);
+}
+
+plVTDecoder::plVTDecoder(const mkvparser::VideoTrack* track)
+{
+    OSStatus               err;
+    CFMutableDictionaryRef decoderOptions = CFDictionaryCreateMutable(kCFAllocatorDefault,
+                                                                      1,
+                                                                      &kCFTypeDictionaryKeyCallBacks,
+                                                                      &kCFTypeDictionaryValueCallBacks);
```
Hoikas: Wrapping

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+    CFNumberRef PixelFormatTypeNumber = CFNumberCreate(kCFAllocatorDefault, kCFNumberSInt32Type, &PixelFormatTypeValue);
+
+    const void* outputKeys[] = {kCVPixelBufferPixelFormatTypeKey};
+    const void* outputValues[] = {PixelFormatTypeNumber};
+    auto        outputDescription = CFDictionaryCreate(
+        kCFAllocatorDefault, outputKeys, outputValues, 1,
+        &kCFTypeDictionaryKeyCallBacks, &kCFTypeDictionaryValueCallBacks);
+
+    err = VTDecompressionSessionCreate(kCFAllocatorDefault, fFormat, decoderOptions, outputDescription, nullptr, &fDecompressionSession);
+    hsAssert(err = noErr, "Decoder creation failed");
```
Hoikas: Did you intend to assign to `err`? The assignment will have no effect in release.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+}
+
+plMovieFrameRef plVTDecoder::DecodeNextFrame(uint8_t* frameData, const size_t size)
+{
+    CMSampleBufferRef sampleBuffer;
+    CMBlockBufferRef  buffer;
+
+    OSStatus err;
+    err = CMBlockBufferCreateWithMemoryBlock(nullptr, frameData, size, nullptr, nullptr, 0, size, 0, &buffer);
+    hsAssert(err = noErr, "Could not create block buffer for frame data");
```
Hoikas: Did you intend to assign to `err`? The assignment will have no effect in release.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+    err = CMSampleBufferCreate(nullptr, buffer, true, nullptr, nullptr, fFormat, 1, 0, nullptr, 1, &size, &sampleBuffer);
+    hsAssert(err = noErr, "Could not create sample buffer");
```
Hoikas: Did you intend to assign to `err`? The assignment will have no effect in release.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,220 @@
...
+        frame->fStride[1] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(imageBuffer, 1));
+        frame->fStride[2] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(imageBuffer, 2));
+        frame->fStride[3] = 0;
+    });
+    hsAssert(err = noErr, "Decoding failed");
+
+    return std::unique_ptr<plMovieFrame, void (*)(plMovieFrame*)>(frame, [](plMovieFrame* ptr) {
+        CVImageBufferRef imageBuffer = static_cast<CVImageBufferRef>(ptr->fContext);
+        CVPixelBufferUnlockBaseAddress(imageBuffer, 0);
+        delete ptr;
```
Hoikas: Should be done in the destructor. I recommend subclassing `plMovieFrame` for each provider.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.h
```diff
@@ -0,0 +1,71 @@
...
+*==LICENSE==*/
+
+#include <VideoToolbox/VideoToolbox.h>
+
+#include "plGImage/plMipmap.h"
+#include "plMoviePlayer.h"
+
+namespace mkvparser
+{
+class VideoTrack;
```
Hoikas: Indent

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
colincornaby: We can't. plMovieFrame could contain a VideoToolbox buffer or a VPX buffer - so either plMovieFrame needs to be a virtual or it needs a custom destructor.

Sources/Plasma/FeatureLib/pfMoviePlayer/CMakeLists.txt
colincornaby: VPX would be an optional feature so it can't be hardcoded in hsConfig.h. Is there an idiomatic way for the CMake to inject this into hsConfig.h?

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
colincornaby: This seems to be the answer for above - move plMovieFrame to a virtual pattern.

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.h
dgelessus: For forward declarations of a single class in a namespace, our usual convention seems to be to put it all on one line:

```cpp
namespace mkvparser { class VideoTrack; }
```

colincornaby:
Two things:
The `VTDecompressionSessionDecodeFrameWithOutputHandler` is a monster but I'm not sure how to break it up because it includes a function block. It might be Stockholm syndrome in since Apple developers tend to be used to the long lines and verbose API.

I am extremely rarely hitting this during the intro movie:

> ASSERTION FAILED:
> File: /Users/colincornaby/Documents/Plasma/PlasmaSrc/Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp   Line: 163
> Message: Failed to decode tile data

This is odd because this is the layer between libwebm and libvpx - which I didn't touch. The Mac Pro I'm testing this on doesn't activate the VideoToolbox path either (hardware isn't compatible.) I'm checking into it. Address Sanitizer is not finding any memory corruption issues.

dgelessus:
A while ago I encountered a different crash when spamming Esc to skip the intro (fixed in #1718). That was in a completely different part of the code than your assertion failure, so I assume the two aren't related, but perhaps `plMoviePlayer` has a similar logic issue if it receives multiple "stop requests" at (almost) the same time?

colincornaby:
It's not clear to me. I can see the decoder provides a compressed frame that is under 100 bytes which seems... small. Although maybe the first frame fits. `codec` has a `VPX_CODEC_CORRUPT_FRAME` error state.

Nothing in the path seems input triggered to me right now. It's the normal decode loop. Trying to see how input could influence things...

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -54,6 +54,9 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #   define WEBM_CODECID_VP9 "V_VP9"
 #   define WEBM_CODECID_OPUS "A_OPUS"
+#   ifdef USE_VIDEOTOOLBOX
+#   include "plVTDecoder.h"
+#   endif
```
Hoikas, line 59:
```suggestion
#   ifdef USE_VIDEOTOOLBOX
#       include "plVTDecoder.h"
#   endif
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
+#ifdef USE_VPX
+class plVPXMovieFrame : public plMovieFrame
+{
+public:
+    plVPXMovieFrame(vpx_image_t* image) : fImage(image)
+    {
+    }
+
+    virtual uint32_t GetWidth() override
```
Hoikas:
```suggestion
    uint32_t GetWidth() override
```

`virtual` is redundant

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+        return fImage->d_w;
+    }
+
+    virtual uint32_t GetHeight() override
```
Hoikas:
```suggestion
    uint32_t GetHeight() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+        return fImage->d_h;
+    }
+
+    virtual int32_t* GetStrides() override
```
Hoikas:
```suggestion
    int32_t* GetStrides() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+        return fImage->stride;
+    }
+
+    virtual unsigned char** GetPlanes() override
```
Hoikas:
```suggestion
    unsigned char** GetPlanes() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+        return fImage->planes;
+    }
+
+    virtual Format GetFormat() override
```
Hoikas:
```suggestion
    Format GetFormat() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -84,6 +86,49 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+    {
+        return plMovieFrame::Format::I420;
```
Hoikas, line 119: I realize that, in practice, I420 is basically the only format that is seen in VP9 movies. I do hope we're still checking this assumption somewhere?

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -50,6 +50,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "hsRefCnt.h"
 
 #include <memory>
+#include <optional>
```
Hoikas:
```suggestion
```

I don't think we're using `std::optional` any more, correct?

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+        fPlanes[2] = static_cast<unsigned char*>(CVPixelBufferGetBaseAddressOfPlane(pixelBuffer, 2));
+        fPlanes[3] = nullptr;
+
+        fStrides[0] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 0));
+        fStrides[1] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 1));
+        fStrides[2] = static_cast<int32_t>(CVPixelBufferGetBytesPerRowOfPlane(pixelBuffer, 2));
+        fStrides[3] = 0;
+    }
+
+    virtual uint32_t GetWidth() override
```
Hoikas:
```suggestion
    uint32_t GetWidth() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        return uint32_t(CVPixelBufferGetWidth(fPixelBuffer));
+    }
+
+    virtual uint32_t GetHeight() override
```
Hoikas:
```suggestion
    uint32_t GetHeight() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        return uint32_t(CVPixelBufferGetHeight(fPixelBuffer));
+    }
+
+    virtual int32_t* GetStrides() override
```
Hoikas:
```suggestion
    int32_t* GetStrides() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        return fStrides;
+    }
+
+    virtual unsigned char** GetPlanes() override
```
Hoikas:
```suggestion
    unsigned char** GetPlanes() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        return fPlanes;
+    }
+
+    virtual Format GetFormat() override
```
Hoikas:
```suggestion
    Format GetFormat() override
```

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    CFDictionarySetValue(
+        decoderOptions,
+        kVTVideoDecoderSpecification_EnableHardwareAcceleratedVideoDecoder,
+        kCFBooleanTrue);
+
+    CreateFormatDescriptionFromVP9Track(track, &fFormat);
+
+    SInt32      PixelFormatTypeValue = kCVPixelFormatType_420YpCbCr8PlanarFullRange;
+    CFNumberRef PixelFormatTypeNumber = CFNumberCreate(kCFAllocatorDefault, kCFNumberSInt32Type, &PixelFormatTypeValue);
```
Hoikas: This is another one that wraps in split diff mode

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    hsAssert(err == noErr, "Decoder creation failed");
+}
+
+std::unique_ptr<plMovieFrame> plVTDecoder::DecodeNextFrame(uint8_t* frameData, const size_t size)
+{
+    CMSampleBufferRef sampleBuffer;
+    CMBlockBufferRef  buffer;
+
+    OSStatus err;
+    err = CMBlockBufferCreateWithMemoryBlock(nullptr, frameData, size, nullptr, nullptr, 0, size, 0, &buffer);
```
Hoikas: wrapping

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    hsAssert(err == noErr, "Could not create block buffer for frame data");
+    err = CMSampleBufferCreate(nullptr, buffer, true, nullptr, nullptr, fFormat, 1, 0, nullptr, 1, &size, &sampleBuffer);
```
Hoikas: wrapping

Sources/Plasma/FeatureLib/pfMoviePlayer/plVTDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    hsAssert(err == noErr, "Could not create sample buffer");
+    __block std::unique_ptr<plMovieFrame> frame;
+    err = VTDecompressionSessionDecodeFrameWithOutputHandler(fDecompressionSession, sampleBuffer, 0, nullptr, ^(OSStatus status, VTDecodeInfoFlags infoFlags, CVImageBufferRef _Nullable imageBuffer, CMTime presentationTimeStamp, CMTime presentationDuration) {
```
Hoikas: wrapping

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -66,6 +67,21 @@ namespace mkvparser
...
+    {
+        I420
+    };
+
+    virtual uint32_t        GetWidth() = 0;
+    virtual uint32_t        GetHeight() = 0;
+    virtual int32_t*        GetStrides() = 0;
+    virtual unsigned char** GetPlanes() = 0;
+    virtual Format          GetFormat() = 0;
+};
```
Hoikas, line 84: Since this now has subclasses, we need a virtual destructor in this class to ensure the subclass destructors are executed.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
colincornaby: Here: https://github.com/H-uru/Plasma/pull/1830/changes#diff-77cf34ec018800cf6984dbebfc701726dfc4ac655a91dd7c5a9b83968bb2b41fR171

It doesn't assert but it does bail from the frame decode.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.cpp
```diff
@@ -121,7 +166,15 @@ class VPX
         vpx_codec_iter_t  iter = nullptr;
         // ASSUMPTION: only one image per frame
         // if this proves false, move decoder function into IProcessVideoFrame
-        return vpx_codec_get_frame(&codec, &iter);
+        auto vpxFrame = vpx_codec_get_frame(&codec, &iter);
+        
+        if (vpxFrame->fmt != VPX_IMG_FMT_I420) {
+            // We only handle YUX 420 right now...
```
Hoikas:
```suggestion
            // We only handle YUV 420 right now...
```

Should probably also think about an assertion or debug message.

Sources/Plasma/FeatureLib/pfMoviePlayer/plMoviePlayer.h
```diff
@@ -66,6 +66,23 @@ namespace mkvparser
...
+        I420
+    };
+
+    virtual uint32_t        GetWidth() = 0;
+    virtual uint32_t        GetHeight() = 0;
+    virtual int32_t*        GetStrides() = 0;
+    virtual unsigned char** GetPlanes() = 0;
+    virtual Format          GetFormat() = 0;
+
+    virtual ~plMovieFrame() {} ;
```
Hoikas:
```suggestion
    virtual ~plMovieFrame() {}
```

cmake/FindVideoToolbox.cmake
```diff
@@ -0,0 +1,17 @@
+include(FindPackageHandleStandardArgs)
+
+find_library(
+    VideoToolbox_LIBRARY
+    NAMES VideoToolbox
+)
+
+find_package_handle_standard_args(VideoToolbox REQUIRED_VARS VideoToolbox_LIBRARY)
+
+if(VideoToolbox_FOUND AND NOT TARGET VideoToolbox::VideoToolbox)
```
dpogue, line 10: Is VideoToolbox a framework? Do we need to create a module for it and not just treat it like any other framework (e.g., Metal or Security or CoreFoundation)?

colincornaby: It depends. VideoToolbox exists as of 10.8 - so for the existing Mac build we could assume it exists. For any older Mac builds you're working on you would not be able to assume its existence and we'd need to treat it like an optional component.

colincornaby: @dpogue - Following up here, I think the merge is holding waiting on this to be resolved. I'm fine resolving this later though! In since we're just potentially overbuilt here I don't think it's a huge concern.

dpogue: I don't think there's a downside to doing it this way, it's just different from how other frameworks on macOS have been handled (but more consistent with how all the other library dependencies are handled)

colincornaby: Yeah. I think as I revisit things - this is the way optional frameworks on macOS should be handled. VideoToolbox is technically optional in since there are efforts to back port the client to older versions of macOS.

PR #1832, opened 2026-01-01
Launcher language

Filtik:
Launcher language change automaticly if language changing


dgelessus:
(repeating what I told Filtik via chat, so others can also read it)

Ideally, the login dialog would use the same XML translation mechanism as the rest of the game, so that the translations can be shared across OSes. This isn't possible at the moment, because `pfLocalizationMgr` is only initialized after login (I vaguely remember that it's because parsing all the XML files causes a short but noticeable delay, but I may be wrong...). If we can initialize `pfLocalizationMgr` earlier, we can convert the existing hardcoded translations to XML files, and then these new translations should also be done in XML.

colincornaby:
Ah - this is localizing the launcher itself. I wonder if our internal mechanism could be plugged into Apple's. But I don't know if we'd need to have platform parity on this.

Filtik:
> Yes, it would be nice to select the system language by default (instead of just defaulting to English) if the player hasn't made a language choice yet. That can be implemented with or without this PR though - the language selector already exists, just the login dialog is always in English.

That might be something you mean.

`
LANGID systemLang = GetUserDefaultUILanguage();
WORD primaryLang = PRIMARYLANGID(systemLang);

plLocalization::Language startLang;

if (primaryLang == LANG_GERMAN) {
	startLang = plLocalization::kGerman;
}
else {
	startLang = plLocalization::kEnglish;
}

plLocalization::SetLanguage(startLang);
SetWindowsUILanguage(startLang);

HWND hCombo = GetDlgItem(hwndDlg, IDC_LANGUAGE);
SendMessage(hCombo, CB_SETCURSEL, (WPARAM)startLang, 0);

UpdateDialogTexts(hwndDlg);
`

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -579,6 +579,13 @@ INT_PTR CALLBACK AuthFailedDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, L
     return FALSE;
 }
 
+static const char* LoadLocalizedString(UINT nID)
+{
+    static char buffer[256];
+    LoadStringA(gHInst, nID, buffer, sizeof(buffer));
+    return buffer;
+}
```
Hoikas: In order to best support all languages, we should probably use unicode here. Also, per the documentation, we can ask for a read only pointer to the string and avoid the copy.


```suggestion
static const wchar_t* GetLocalizedString(UINT nID)
{
    wchar_t* buffer = nullptr;
    LoadStringW(gHInst, nID, reinterpret_cast<LPWSTR>(&buffer), 0);
    return buffer;
}
```

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -598,16 +605,19 @@ INT_PTR CALLBACK UruTOSDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARA
 
                 SetDlgItemTextW(hwndDlg, IDC_URULOGIN_EULATEXT,
                                 ST::string(eula, ST::substitute_invalid).to_wchar().data());
+
+                SetDlgItemText(hwndDlg, IDC_BUTTON_ACCEPT, LoadLocalizedString(IDC_TEXT_ACCEPT));
+                SetDlgItemText(hwndDlg, IDC_BUTTON_DECLINE, LoadLocalizedString(IDC_TEXT_DECLINE));
```
Hoikas:
```suggestion
                SetDlgItemTextW(hwndDlg, IDC_BUTTON_ACCEPT, GetLocalizedString(IDC_TEXT_ACCEPT));
                SetDlgItemTextW(hwndDlg, IDC_BUTTON_DECLINE, GetLocalizedString(IDC_TEXT_DECLINE));
```

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -722,6 +732,23 @@ static size_t CurlCallback(void *buffer, size_t size, size_t nmemb, void *param)
     return size * nmemb;
 }
 
+void SetWindowsUILanguage(plLocalization::Language lang)
```
Hoikas:
```suggestion
static void SetWindowsUILanguage(plLocalization::Language lang)
```

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -896,6 +923,16 @@ INT_PTR CALLBACK UruLoginDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPA
 
                 return TRUE;
             }
+            else if (HIWORD(wParam) == CBN_SELCHANGE && LOWORD(wParam) == IDC_LANGUAGE) 
+            {
+                HWND hCombo = (HWND)lParam;
+                int  currentIndex = (int)SendMessage(hCombo, CB_GETCURSEL, 0, 0);
```
Hoikas:
```suggestion
                int currentIndex = ComboBox_GetCurSel(hCombo);
```

Sources/Plasma/Apps/plClient/win32/winmain.cpp
```diff
@@ -896,6 +923,16 @@ INT_PTR CALLBACK UruLoginDialogProc( HWND hwndDlg, UINT uMsg, WPARAM wParam, LPA
...
+
+                if (currentIndex != CB_ERR) {
+                    plLocalization::Language new_language = (plLocalization::Language)SendMessage(GetDlgItem(hwndDlg, IDC_LANGUAGE), CB_GETCURSEL, 0, 0L);
```
Hoikas: I think we could just cast `currentIndex` to `plLocalization::Lanuage` instead of asking for the current selection again.

Sources/Plasma/Apps/plClient/win32/res/resource_de.rc
```diff
@@ -0,0 +1,22 @@
+#include "resource.h"
+#include <windows.h>
```
Hoikas:
```suggestion
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
```

Sources/Plasma/Apps/plClient/win32/res/resource_en.rc
```diff
@@ -0,0 +1,22 @@
+#include "resource.h"
+#include <windows.h>
```
Hoikas:
```suggestion
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
```

PR #1850, opened 2026-03-07
Fix endianness and alignment assumptions in `pnNpCommon` IO code

dgelessus:
My goal was just to remove the `StrCopy` calls, but while I was here, I also fixed the endianness and alignment issues in this code. Perhaps it will help @dpogue on PowerPC :)

This adds some helper functions for converting little-endian UTF-16 directly from/to `ST::string`, which could also be useful for other code that reads/writes UTF-16 strings manually, like file server manifests and the GameMgr.

### DIFF
--- a/Sources/Plasma/CoreLib/hsEndian.cpp
+++ b/Sources/Plasma/CoreLib/hsEndian.cpp
@@ -58,21 +58,21 @@ ST::string hsSTStringFromUTF16LE(const void* buffer, size_t char16Count)
 ST::string hsSTStringFromTerminatedUTF16LE(const void* buffer, size_t bufferSize, size_t& consumedSize)
 {
     auto byteBuffer = static_cast<const uint8_t*>(buffer);
-    // consumedChar16Count may be equal to or one higher than utf16Buffer.size(),
-    // depending on whether a terminator was found before the end of the buffer.
+    // Count how many char16_ts there are in the string.
+    // char16Count only counts the actual string contents,
+    // consumedChar16Count also counts the terminator, if present
+    // (there is no terminator if the end of the buffer was reached unexpectedly).
     size_t consumedChar16Count = 0;
-    // Can't pre-allocate anything, because we don't know the string length yet...
-    std::vector<char16_t> utf16Buffer;
+    size_t char16Count = 0;
     for (size_t i = 0; i < bufferSize / sizeof(char16_t); i++) {
-        char16_t c = byteBuffer[2*i] | byteBuffer[2*i + 1] << 8;
         consumedChar16Count++;
-        if (c == 0) {
+        if (byteBuffer[2*i] == 0 && byteBuffer[2*i + 1] == 0) {
             break;
         }
-        utf16Buffer.emplace_back(c);
+        char16Count++;
     }
     consumedSize = consumedChar16Count * sizeof(char16_t);
-    return ST::string::from_utf16(utf16Buffer.data(), utf16Buffer.size());
+    return hsSTStringFromUTF16LE(buffer, char16Count);
 }
 
 std::vector<uint8_t> hsSTStringToUTF16LE(const ST::string& string)
--- a/Sources/Plasma/CoreLib/hsEndian.cpp
+++ b/Sources/Plasma/CoreLib/hsEndian.cpp
@@ -67,7 +67,8 @@ ST::string hsSTStringFromTerminatedUTF16LE(const void* buffer, size_t bufferSize
         }
         utf16Buffer.emplace_back(c);
     }
-    consumedSize = utf16Buffer.size() * sizeof(char16_t);
+    // consumedSize includes the terminator, which is not stored in utf16Buffer
+    consumedSize = (utf16Buffer.size() + 1) * sizeof(char16_t);
     return ST::string::from_utf16(utf16Buffer.data(), utf16Buffer.size());
 }
 
--- a/Sources/Plasma/CoreLib/hsEndian.cpp
+++ b/Sources/Plasma/CoreLib/hsEndian.cpp
@@ -1,0 +1,85 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
+PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+licensors of this Program grant you additional
+permission to convey the resulting work. Corresponding Source for a
+non-source form of such a combination shall include the source code for
+the parts of OpenSSL and IJG JPEG Library used as well as that of the covered
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "hsEndian.h"
+
+#include <string_theory/string>
+
+ST::string hsSTStringFromUTF16LE(const void* buffer, size_t char16Count)
+{
+    auto byteBuffer = static_cast<const uint8_t*>(buffer);
+    ST::utf16_buffer utf16Buffer;
+    utf16Buffer.allocate(char16Count);
+    for (size_t i = 0; i < char16Count; i++) {
+        utf16Buffer[i] = byteBuffer[2*i] | byteBuffer[2*i + 1] << 8;
+    }
+    return ST::string::from_utf16(utf16Buffer);
+}
+
+ST::string hsSTStringFromTerminatedUTF16LE(const void* buffer, size_t bufferSize, size_t& consumedSize)
+{
+    auto byteBuffer = static_cast<const uint8_t*>(buffer);
+    // Can't pre-allocate anything, because we don't know the string length yet...
+    std::vector<char16_t> utf16Buffer;
+    for (size_t i = 0; i < bufferSize / sizeof(char16_t); i++) {
+        char16_t c = byteBuffer[2*i] | byteBuffer[2*i + 1] << 8;
+        if (c == 0) {
+            break;
+        }
+        utf16Buffer.emplace_back(c);
+    }
+    consumedSize = utf16Buffer.size() * sizeof(char16_t);
+    return ST::string::from_utf16(utf16Buffer.data(), utf16Buffer.size());
+}
+
+std::vector<uint8_t> hsSTStringToUTF16LE(const ST::string& string)
+{
+    ST::utf16_buffer utf16Buffer = string.to_utf16();
+    std::vector<uint8_t> buffer;
+    buffer.resize(utf16Buffer.size() * 2);
+    for (size_t i = 0; i < utf16Buffer.size(); i++) {
+        char16_t c = utf16Buffer[i];
+        buffer[2*i] = c & 0xff;
+        buffer[2*i + 1] = c >> 8 & 0xff;
+    }
+    return buffer;
+}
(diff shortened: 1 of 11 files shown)


dpogue (approved):
This is definitely nicer than the hacky string copy stuff I had to do in plNglFile

Sources/Plasma/CoreLib/hsEndian.h
```diff
@@ -45,6 +45,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include "HeadSpin.h"
 
+#include <string_theory/string>
```
Hoikas: Could we use forward declarations instead?

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
```diff
@@ -69,62 +69,56 @@ const unsigned kNumBlobFields   = 4;
 template <typename T>
 inline void IReadValue (T * value, uint8_t ** buffer, unsigned * bufsz) {
     ASSERT(*bufsz >= sizeof(T));
-    *value = *(T *)*buffer;
+    T leValue;
+    memcpy(&leValue, *buffer, sizeof(T));
+    *value = hsToLE(leValue);
```
Hoikas, line 74: I don't think `hsToLE` modifies in place. It would be nice to avoid the `memcpy` and perform one copy instead of two on LE systems.

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
```diff
@@ -69,62 +69,56 @@ const unsigned kNumBlobFields   = 4;
...
 //============================================================================
 template <typename T>
 inline void IWriteValue (const T & value, std::vector<uint8_t> * buffer) {
     static_assert(std::is_trivially_copyable_v<T>,
                   "IWriteValue can only be used on trivially copyable types");
     const size_t endPos = buffer->size();
     buffer->resize(endPos + sizeof(T));
-    memcpy(buffer->data() + endPos, &value, sizeof(T));
+    T leValue = hsToLE(value);
+    memcpy(buffer->data() + endPos, &leValue, sizeof(T));
```
Hoikas, line 108: Same RE extra copy.

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
```diff
@@ -369,8 +353,9 @@ inline bool IRead(const uint8_t*& buf, size_t& bufsz, T& dest)
     if (bufsz < sizeof(T)) {
         return false;
     }
-    const T* ptr = reinterpret_cast<const T*>(buf);
-    dest = *ptr;
+    T leValue;
+    memcpy(&leValue, buf, sizeof(T));
+    dest = hsToLE(leValue);
```
Hoikas, line 360: Another potentially spurious copy

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
```diff
@@ -468,13 +462,14 @@ inline void IWrite(std::vector<uint8_t>* buffer, const T& value)
                   "IWrite can only be used on trivially copyable types");
     const size_t oldSize = buffer->size();
     buffer->resize(oldSize + sizeof(T));
-    memcpy(buffer->data() + oldSize, &value, sizeof(T));
+    T leValue = hsToLE(value);
+    memcpy(buffer->data() + oldSize, &leValue, sizeof(T));
```
Hoikas, line 470: Another potentially spurious copy

Sources/Plasma/CoreLib/hsEndian.h
dgelessus: Probably. I was too lazy to type out the template declaration needed for `ST::char16_buffer` :D

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
dgelessus: The `memcpy` is technically required to avoid UB, because the source pointer might not be properly aligned for a direct cast to `T*`. The expectation is that the compiler will optimize the `memcpy` to something more efficient - and indeed, both MSVC and GCC at `-O2` on x86_64 turn the entire thing into a simple `mov`, just like the previous pointer-casting code (see [Godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjEIACsAQAOqAqETgwe3r7%2BpClpjgIhYZEsMXGJtpj2hQxCBEzEBFk%2BfgF2mA4Z9Y0ExRHRsQntDU0tOSO9/aXlwwCUtqhexMjsHOYAzKHI3lgA1CYbbsgKBPiCh9gmGgCCm9u7mAdHJ2dOl9d3ZlsMO177h2Op2IoWAH1un3uVCwVD2AH0ALJCNxwgBq2AASp89jiDt8YaEnggFEIAO5MJLYBj4QxcABsEAAbmI5vCogBPAiYBTkpJwrwKBAkAhMlnY3GbAlhPbEskUqk0hgbMyi2isuEcrk8in82gCYCqubinGSzA0aWy3kKvCGOmSQ1sznc3n80IEO2GyHfap4WFSzDoCBwuG0WiMljB1kmADsbhjbj2EH9geDAHFwnIUXCo2Y6eY6Ymg3D05ng3sAFR7LgaDQ5izw4sZlEIgCS4QA8hjI88rhsACJ7SQaAAcRtuuLxG39MpJVupNoY9IdwaiXjojgYGu1SSXzLVxsn08t8vnhmVy41a9oG63vPPe7HNwnpvNRNnJ8VHofDdX69Ct4pL8xQhb1aCUA9NliYgSDxMwUgaIhzDML0NlcX1IRArZoTNPYAAkhDhCwW1TOFsHCPsWxucIIPxM1CRnAAVVAgmwJcGFZCcJ2PSlT0XBl2Joqc6ItBQmJY892M4rj3x4xUJMfZ9aNfRjmOwD1JKknFuOtW17QEzDqnA8cJSU%2BjiTE1j%2BI4zSIH0p8TKE5TzNU%2BTNJxWyFIco9RNU9TrKkjyULQqgMJuLkWCSAwuWeNwCHZJJGFYJ4GMuPZQloeiGJUliICyvcvEwKN%2Bz2LB6C5Q4rFucLIqYaLAVS9L6OQBBGmy7AIGa1r8sKg5o3rYhMAIZYGD2bqKt6vtPmqqKnnqjZsDShgMulNJgDCdA9k64g2ogVb1s2lrtu6qM%2Br2AahuIEaxo2esY0mqrMAimaYoapb6K8Bg9oDA7WucnKPq%2BjattGsQCpO/rBuGkHvEwca7qmx6arqo5XuWp43XpOECB2jG6Sx6Gwd6iGLpGiBcaxuY/ssqA10ETGCDmY64eje6wsR565oWxrpVpgh6Z23n%2BeOomzshy62t3UHCuZ1nptq2aUfmxa0cWghlXxqmycEdXseFmNiahrW1bMCmqfPCBeZ1xmpaNG6JoRp75ZepXuaeS2TexzX3fxvXTvOqGzZVJm7fhh7HeRuMXbe6U3TtDWfJy2PJB9m2Rf98WjbjhmqY9C2k4p4PbpZh2kYVyOuejt3889hP2t5rOCZ6/XRZJtqgJh22i9ZkuZsBOKEuYNg9hS%2BbPld0bUDwDaWwxTAmHQVEpcTLLK260g9l54d8fLStVyoKhYnBg8svoReYYqg82BYZAknZCB81PqX1/LPeD7IPY0gAL0wVAqFyuZO4HnLN1Z4A5NaPw7hfYyOIAD0MC9ieCBhgdgB44HAKXocAc5ZcoVjmC/Lw%2B9YhQLuMXCEtw3R7BYEwUIRs9iNGAMgdeW0d50OIMARkR9oEbzdFvAgu8CFv1AaLN0sQkjnThKIU4gJN5Y3LJcCA9DGQmHiBYLgyi%2ByAK4d7XWUtiEThnnPBeUt765jXniOkr9D56NxOnK6uibqQhZhwBYtBODxF4H4DgWhSCoE4HGSw1gP5LBWLNb4PBSAEE0M4hYABrBIGh9CcEkB4qJPjOC8AUCABJkSvHONIHAWASA0ARToLEcglBilJFKXERkN8%2BSMi4AATjhGYZpqIhCYy4FwOEqg7R8HXLETJEAoipKiKERo7JODhLGcwYg7J2xRG0J0HJ4TilsEEO2JakzcmkCwKuYAbgxBgSmbwLAVCjDiB2fgAaXRGTclSZgVQnQvBchOeQQQ1RUkZSiMQCZHgsCpIICCFgby7nECiKkTAfZHqGGAMtUAuSFhUAMMABQqI8CYFJO2Aebz%2BCCBEGIdgUgZCCEUCodQOzdBcH0LClA1hrD6DwFETJkAFioCSLUTJHAAC07YzC8FQGCkEWAWUQAWB0LozgICuHGH4alwRCQzCGNS/I6QBCyr0Kq2o0xBhxGpRK2oPQxieFaHoA13RRh9EVbqs1lqNX6stTqsoyrxXBNWBIFxbiUk7N8RwPYvTJCUIULU0aTSAB0rS9jtPpGGrpiZcCEBgpsLgcxeA5K0AA0gcT4gJNcRwZJpBPHeN9RkrJESomZrzfywtqSS3lsRQsMFaRnCSCAA%3D%3D)).

It does lead to less optimal code in unoptimized builds, but I think that's acceptable, because those are not used by players, and this IO code isn't used in a tight loop during normal gameplay or anything.

Sources/Plasma/CoreLib/hsEndian.cpp
```diff
@@ -0,0 +1,82 @@
...
+        utf16Buffer[i] = byteBuffer[2*i] | byteBuffer[2*i + 1] << 8;
+    }
+    return ST::string::from_utf16(utf16Buffer);
+}
+
+ST::string hsSTStringFromTerminatedUTF16LE(const void* buffer, size_t bufferSize, size_t& consumedSize)
+{
+    auto byteBuffer = static_cast<const uint8_t*>(buffer);
+    // Can't pre-allocate anything, because we don't know the string length yet...
+    std::vector<char16_t> utf16Buffer;
```
zrax: You can, however, _reserve_ `bufferSize/sizeof(char16_t)` to flatten the number of actual memory allocations to 1.

Sources/Plasma/CoreLib/hsEndian.cpp
```diff
@@ -0,0 +1,82 @@
...
+        if (c == 0) {
+            break;
+        }
+        utf16Buffer.emplace_back(c);
+    }
+    consumedSize = utf16Buffer.size() * sizeof(char16_t);
+    return ST::string::from_utf16(utf16Buffer.data(), utf16Buffer.size());
+}
+
+ST::utf16_buffer hsSTStringToUTF16LE(const ST::string& string)
```
zrax: Swapping the endianness of the buffer implies that this is no longer a legal `utf16_buffer`...  It may make more sense logically to return as a `vector<uint16_t>`

Sources/Plasma/CoreLib/hsEndian.h
zrax: Changing the return type of `hsSTStringToUTF16LE` as suggested above would also resolve this ;)

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
```diff
@@ -70,62 +69,56 @@ const unsigned kNumBlobFields   = 4;
...
+{
     uint32_t bytes;
     IReadValue(&bytes, buffer, bufsz);
-    ASSERT(bytes % sizeof(T) == 0);
-    *elems = bytes / sizeof(T);
-    T * src = (T *)*buffer;
-    free(*buf);
-    *buf = (T *)malloc(bytes);
-    memcpy(*buf, src, bytes);
+    ASSERT(bytes % sizeof(char16_t) == 0);
```
zrax, line 84: Might want to also assert that `string` is not null, since it's used unconditionally below...

Sources/Plasma/CoreLib/hsEndian.cpp
dgelessus: hmm, I don't think that's a good idea in the general case. For something like the file server manifests, where we have potentially hundreds of zero-terminated strings concatenated, we don't want to reserve the size of the entire manifest for each string.

zrax: Ah, good point, I wasn't considering the case where the buffer is significantly larger than the string to extract from it...

Sources/Plasma/CoreLib/hsEndian.cpp
dgelessus: I made it a `std::vector<uint8_t>` now, so we don't have any incorrect endianness values at all. This actually also works better for the calling code.

Sources/Plasma/NucleusLib/pnNetProtocol/pnNpCommon.cpp
dgelessus: None of the other out pointers are checked either - I say just let it crash...

Sources/Plasma/CoreLib/hsEndian.cpp
dpogue: Would something like 128 work as an initial size that's big enough to handle most of the strings we'd expect over the network without incurring reallocations but also not incurring a big chunk of memory?

dgelessus: I rewrote the code so it counts the string length beforehand. This allows reusing `hsSTStringFromUTF16LE`, which allocates a buffer with the right size.

zrax (approved):

PR #1857, opened 2026-03-27
Small improvements to CMake folders

dgelessus:
Fixes the folder for the Scripts target to avoid this suboptimal display in CLion:

CLion's Run Configurations menu, showing a folder with an empty name containing a target named Scripts

and moves the many, many targets related to clang-tidy into a folder named `tidy` so that they don't fill up the `CMakePredefinedTargets` default folder.

With my build configuration, the folder `CMakePredefinedTargets` now still contains the targets `check`, `gmock`, `gmock_main`, `gtest`, `gtest_main`, and `tools`. Should I move any of those into different folders too (`Tests` and `Tools` presumably)? None of these targets come from CMake, so it's a bit surprising that they're in `CMakePredefinedTargets`.

(For context: the CMake folder structure was originally added in #1612.)

### DIFF
--- a/Sources/Tests/CMakeLists.txt
+++ b/Sources/Tests/CMakeLists.txt
@@ -16,9 +16,17 @@ set(gtest_build_tests OFF CACHE INTERNAL "Override gtest default")
 set(gtest_disable_pthreads OFF CACHE INTERNAL "Override gtest default")
 set(gtest_hide_internal_symbols ON CACHE INTERNAL "Override gtest default")
 
+# Put gtest's targets in the Tests folder,
+# not our default "CMakePredefinedTargets" folder.
+set(_PREV_CMAKE_FOLDER ${CMAKE_FOLDER})
+set(CMAKE_FOLDER "Tests")
+
 set(GOOGLETEST_VERSION 1.12.1)
 set(INSTALL_GTEST OFF CACHE INTERNAL "Override gtest default" FORCE)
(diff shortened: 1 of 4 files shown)


Hoikas:
CI failure should be fixed by #1859.

> With my build configuration, the folder CMakePredefinedTargets now still contains the targets check, gmock, gmock_main, gtest, gtest_main, and tools. Should I move any of those into different folders too (Tests and Tools presumably)? None of these targets come from CMake, so it's a bit surprising that they're in CMakePredefinedTargets.

That seems like a good idea.

Sources/Tests/CMakeLists.txt
```diff
@@ -16,10 +16,18 @@ set(gtest_build_tests OFF CACHE INTERNAL "Override gtest default")
 set(gtest_disable_pthreads OFF CACHE INTERNAL "Override gtest default")
 set(gtest_hide_internal_symbols ON CACHE INTERNAL "Override gtest default")
 
+# Put gtest's targets in the Tests folder,
+# not our default "CMakePredefinedTargets" folder.
+set(_PREV_CMAKE_FOLDER ${CMAKE_FOLDER})
```
Hoikas, line 21: If we ever bump to CMake 3.25+, this would be a great place for `block(SCOPE_FOR VARIABLES)`

Hoikas (approved):

PR #1859, opened 2026-04-06
Update pillow to fix macOS CI.

Hoikas:
Speculative fix. Verified that resource.dat is still generated correctly.


Hoikas:
Looks like this will fix the problem.

dpogue (approved):

PR #1860, opened 2026-04-12
Metal render destinations

colincornaby:
When the Metal renderer was first written - details of the window server were allowed to leak into the renderer. This held up ok at first - Metal is only available on Apple platforms. But as Apple has added more variety to their operating systems it's no longer safe to allow window server details into the renderer. For example - stereoscopic rendering on Vision Pro requires integration with the RealityKit based window server. This is a very different window server than we use on macOS.

Another example of where this has fallen apart is CAMetalDisplayLink - which is the modern version of vsync for Metal. It monitors our frame rate and adjusts our frame rate in response to load. For these reasons CAMetalDisplayLink sits between us and the window server - and we're not allowed to talk directly to the window server.

This PR starts to pull that relationship apart by adding a new `plMetalRenderDestination` type that sits in between the client and the renderer - and allows the renderer to start to abstract these relationships.

The `plDirectMetalRenderDestination` provided with this PR is a very simple implementation that directly pulls framebuffers from Metal on a traditional window server - which is how Plasma has pulled framebuffers from the window server so far. This is part of a series of pull requests - and I do have a CAMetalDisplayLink frame provider in my back pocket. But I'm splitting up this work because CAMetalDisplayLink controls both vsync timing and the frame buffer destination, so it will require a few separate abstractions.

Another twist is that visionOS only provides a Swift API to RealityKit and the RealityKit window server. There are certain things - detailed in this PR - that I'm doing to prepare for that. For example - Swift can't subclass C++ types, but we can template Swift on C++ types. I'm not doing any Swift/C++ integration in this PR but assume I also have that back pocket and I'm getting ready for that in this PR.

### DIFF
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -221,6 +221,9 @@ static uint32_t ParseRendererArgument(const ST::string& requested)
 @property NSWindow* gameWindow;
 @property plMetalRenderDestinationType*    renderDestination;
 
+@property plMetalPipeline* metalPipeline;
+@property plGLPipeline*    glPipeline;
+
 @end
 
 void plClient::IResizeNativeDisplayDevice(int width, int height, bool windowed)
@@ -262,7 +265,17 @@ void plClient::IResizeNativeDisplayDevice(int width, int height, bool windowed)
 
 void plClient::IChangeResolution(int width, int height) {}
 void plClient::IUpdateProgressIndicator(plOperationProgress* progress) {}
-void plClient::ShowClientWindow() {}
+void plClient::ShowClientWindow()
+{
+    AppDelegate* appDelegate = (AppDelegate*)[NSApp delegate];
+    plPipeline* pipeline = appDelegate->gClient->GetPipeline();
+    appDelegate.metalPipeline = dynamic_cast<plMetalPipeline*>(pipeline);
+    appDelegate.glPipeline = dynamic_cast<plGLPipeline*>(pipeline);
+
+    if (appDelegate.metalPipeline) {
+        appDelegate.metalPipeline->SetRenderDestination(appDelegate.renderDestination);
+    }
+}
 void plClient::FlashWindow()
 {
     dispatch_async(dispatch_get_main_queue(), ^{
@@ -329,15 +342,14 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
     _displayHelper = new plMacDisplayHelper();
     plDisplayHelper::SetInstance(_displayHelper);
 
-    gClient.SetClientWindow((__bridge void*)view.layer);
-    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
-
     // Start the game with a direct render destination
     // Components like the intro movie may want to draw
     // before we're ready to start a vsync managed render
     // destination.
     self.renderDestination = new plMetalRenderDestination<plDirectMetalRenderDestination*>((CAMetalLayer*)self.plsView.layer);
-    self.renderDestination->MakeCurrentRenderDestination();
+
+    gClient.SetClientWindow(&_renderDestination);
+    gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
 
     self = [super initWithWindow:window];
     self.window.acceptsMouseMovedEvents = YES;
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -271,10 +271,6 @@ void plClient::ShowClientWindow()
     plPipeline* pipeline = appDelegate->gClient->GetPipeline();
     appDelegate.metalPipeline = plMetalPipeline::ConvertNoRef(pipeline);
     appDelegate.glPipeline = plGLPipeline::ConvertNoRef(pipeline);
-
-    if (appDelegate.metalPipeline) {
-        appDelegate.metalPipeline->SetRenderDestination(appDelegate.renderDestination);
-    }
 }
 
 void plClient::FlashWindow()
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -276,6 +276,7 @@ void plClient::ShowClientWindow()
         appDelegate.metalPipeline->SetRenderDestination(appDelegate.renderDestination);
     }
 }
+
 void plClient::FlashWindow()
 {
     dispatch_async(dispatch_get_main_queue(), ^{
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -59,6 +59,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #import "PLSKeyboardEventMonitor.h"
 #import "PLSLoginWindowController.h"
 #import "PLSPatcherWindowController.h"
+#import "plDirectMetalRenderDestination.h"
 #import "PLSServerStatus.h"
 #import "PLSView.h"
 
@@ -218,6 +219,7 @@ static uint32_t ParseRendererArgument(const ST::string& requested)
 @property PLSPatcher* patcher;
 @property PLSLoginWindowController* loginWindow;
 @property NSWindow* gameWindow;
+@property plMetalRenderDestinationType*    renderDestination;
 
 @end
 
@@ -329,6 +331,13 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
 
     gClient.SetClientWindow((__bridge void*)view.layer);
     gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
+
+    // Start the game with a direct render destination
+    // Components like the intro movie may want to draw
+    // before we're ready to start a vsync managed render
+    // destination.
+    self.renderDestination = new plMetalRenderDestination<plDirectMetalRenderDestination*>((CAMetalLayer*)self.plsView.layer);
+    self.renderDestination->MakeCurrentRenderDestination();
 
     self = [super initWithWindow:window];
     self.window.acceptsMouseMovedEvents = YES;
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -196,6 +196,24 @@ static uint32_t ParseRendererArgument(const ST::string& requested)
     return hsG3DDeviceSelector::kDevTypeUnknown;
 }
 
+@class AppDelegate;
+
+// Stub placeholder for actual plClientWindow implementation
+class plClientWindow {
+
+};
+
+class plMacClientWindow: public plClientWindow, public plMetalWindow
+{
+public:
+    plMetalRenderDestinationType& metalRenderDestination() override {
+        return *fMetalRenderDestination;
+    }
+
+    std::unique_ptr<plMetalRenderDestinationType> fMetalRenderDestination;
+    ~plMacClientWindow() = default;
+};
+
 @interface AppDelegate : NSWindowController <NSApplicationDelegate,
                                              NSWindowDelegate,
                                              PLSViewDelegate,
@@ -207,6 +225,7 @@ static uint32_t ParseRendererArgument(const ST::string& requested)
     dispatch_source_t   _displaySource;
     plMacDisplayHelper* _displayHelper;
     PLSWakeLockHolder   _wakeLockHolder;
+    plMacClientWindow   _plasmaWindow;
 }
 
 @property(retain) PLSKeyboardEventMonitor* eventMonitor;
@@ -219,7 +238,6 @@ static uint32_t ParseRendererArgument(const ST::string& requested)
 @property PLSPatcher* patcher;
 @property PLSLoginWindowController* loginWindow;
 @property NSWindow* gameWindow;
-@property plMetalRenderDestinationType*    renderDestination;
 
 @property plMetalPipeline* metalPipeline;
 @property plGLPipeline*    glPipeline;
@@ -343,9 +361,9 @@ static void* const DeviceDidChangeContext = (void*)&DeviceDidChangeContext;
     // Components like the intro movie may want to draw
     // before we're ready to start a vsync managed render
     // destination.
-    self.renderDestination = new plMetalRenderDestination<plDirectMetalRenderDestination*>((CAMetalLayer*)self.plsView.layer);
+    _plasmaWindow.fMetalRenderDestination = std::make_unique<plMetalRenderDestination<plDirectMetalRenderDestination*>>((CAMetalLayer*)self.plsView.layer);
 
-    gClient.SetClientWindow(&_renderDestination);
+    gClient.SetClientWindow(&_plasmaWindow);
     gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
 
     self = [super initWithWindow:window];
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -239,9 +239,6 @@ public:
 @property PLSLoginWindowController* loginWindow;
 @property NSWindow* gameWindow;
 
-@property plMetalPipeline* metalPipeline;
-@property plGLPipeline*    glPipeline;
-
 @end
 
 void plClient::IResizeNativeDisplayDevice(int width, int height, bool windowed)
@@ -283,13 +280,7 @@ void plClient::IResizeNativeDisplayDevice(int width, int height, bool windowed)
 
 void plClient::IChangeResolution(int width, int height) {}
 void plClient::IUpdateProgressIndicator(plOperationProgress* progress) {}
-void plClient::ShowClientWindow()
-{
-    AppDelegate* appDelegate = (AppDelegate*)[NSApp delegate];
-    plPipeline* pipeline = appDelegate->gClient->GetPipeline();
-    appDelegate.metalPipeline = plMetalPipeline::ConvertNoRef(pipeline);
-    appDelegate.glPipeline = plGLPipeline::ConvertNoRef(pipeline);
-}
+void plClient::ShowClientWindow() {}
 
 void plClient::FlashWindow()
 {
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
@@ -269,8 +269,8 @@ void plClient::ShowClientWindow()
 {
     AppDelegate* appDelegate = (AppDelegate*)[NSApp delegate];
     plPipeline* pipeline = appDelegate->gClient->GetPipeline();
-    appDelegate.metalPipeline = dynamic_cast<plMetalPipeline*>(pipeline);
-    appDelegate.glPipeline = dynamic_cast<plGLPipeline*>(pipeline);
+    appDelegate.metalPipeline = plMetalPipeline::ConvertNoRef(pipeline);
+    appDelegate.glPipeline = plGLPipeline::ConvertNoRef(pipeline);
 
     if (appDelegate.metalPipeline) {
         appDelegate.metalPipeline->SetRenderDestination(appDelegate.renderDestination);
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.h
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.h
@@ -50,10 +50,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 class plDirectMetalRenderDestination
 {
 public:
-    plDirectMetalRenderDestination(CAMetalLayer *layer);
+    plDirectMetalRenderDestination(CAMetalLayer* layer);
     std::optional<plMetalRenderSurface> GetNextRenderSurface(MTL::CommandBuffer* buffer);
     void SetOutputSize(CGSize size);
 
 private:
-    CAMetalLayer *fLayer;
+    CAMetalLayer* fLayer;
 };
--- a/Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.h
+++ b/Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.h
@@ -1,0 +1,59 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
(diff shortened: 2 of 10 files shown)


colincornaby:
As an example where I'm aiming - here is the visionOS type that is used as a Metal stereoscopic render destination. It's not connected to the layer system at all:
https://developer.apple.com/documentation/realitykit/lowleveltexture

We also have to call this replace function to actually do the frame buffer swap:
https://developer.apple.com/documentation/realitykit/lowleveltexture/replace(using:)

colincornaby:
Build is failing somewhere in the Cairo dependency step - not sure why.

Hoikas:
Fixed by #1859, please rebase.

colincornaby:
Ah, I had rebased on my fork of master instead of upstream. :( Sorry about that.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -330,6 +332,13 @@ - (id)init
     gClient.SetClientWindow((__bridge void*)view.layer);
     gClient.SetClientDisplay([window.screen.deviceDescription[@"NSScreenNumber"] unsignedIntValue]);
 
+    // Start the game with a direct render destination
+    // Components like the intro movie may want to draw
+    // before we're ready to start a vsync managed render
+    // destination.
+    self.renderDestination = new plMetalRenderDestination<plDirectMetalRenderDestination*>((CAMetalLayer*)self.plsView.layer);
+    self.renderDestination->MakeCurrentRenderDestination();
```
dpogue, line 340: We can sort this out later, but just noting that this is baking an assumption on the Metal renderer into the macOS client which won't hold true with OpenGL

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+class plMetalRenderDestinationType
+{
+public:
+    virtual plOptionalMetalRenderSurface GetNextRenderSurface(MTL::CommandBuffer* buffer) = 0;
+    virtual void SetOutputSize(CGSize size) = 0;
+
+    void MakeCurrentRenderDestination() { fCurrentRenderDestination = this; }
+    static plMetalRenderDestinationType* CurrentRenderDestination() { return fCurrentRenderDestination; }
```
dpogue, line 152: Is there a way that this could be owned by the pipeline, rather than as a self-owned singleton?

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
colincornaby: Yeah, I've put quite a bit of thought into that. `CAMetalDisplayLink` is odd in that it's _both_ a vsync timer and the framebuffer provider in since it needs to do both for it's functionality. That would be bad for GL - we don't want to couple the idea of the render loop and framebuffer provider together. So I've split it so that the framebuffer provider and render loop are two separate abstract interfaces. I think the next PR will deal with the vsync timer side in a way that's compatible with GL.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
colincornaby: The intro movie is the problem. The client loader creates the pipeline and immediately starts rendering before the client is even aware the pipeline exists. I also don't like this hack, but there would need to be a way to get the client loader to create a pipeline with arguments.

The client may also need to change the render destination on the fly (hint for the future) but that doesn't require a singleton either.

colincornaby:
I've made changes to eliminate the singleton. However - the Metal render destination still needs to be changeable on the fly. The render destination depends on if vsync is on or off. We make one of those transitions at the beginning because the intro movie is not vsynced. So the render destination is stored as a pointer to a pointer.

Thinking through alternatives - I don't think a shared_ptr will work here in since you can't change the reference for all instances of that shared_ptr. But I'd be open to any more modern C++ facing alternatives.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -260,7 +265,17 @@ @interface AppDelegate : NSWindowController <NSApplicationDelegate,
 
 void plClient::IChangeResolution(int width, int height) {}
 void plClient::IUpdateProgressIndicator(plOperationProgress* progress) {}
-void plClient::ShowClientWindow() {}
+void plClient::ShowClientWindow()
+{
+    AppDelegate* appDelegate = (AppDelegate*)[NSApp delegate];
+    plPipeline* pipeline = appDelegate->gClient->GetPipeline();
+    appDelegate.metalPipeline = dynamic_cast<plMetalPipeline*>(pipeline);
+    appDelegate.glPipeline = dynamic_cast<plGLPipeline*>(pipeline);
```
dpogue: Are you wanting one of these to be null based on what the actual type of the pipeline is? The way we'd typically do that in Plasma is with `plGLPipeline::ConvertNoRef(pipeline)` rather than `dynamic_cast<plGLPipeline*>(pipeline)`

dpogue (commented):
I still don't think I have a good sense of how this pattern would be applied to OpenGL, but we don't strictly need to solve that here.

colincornaby:
> I still don't think I have a good sense of how this pattern would be applied to OpenGL, but we don't strictly need to solve that here.

There's a second set of changes that will be applicable to GL. I think this set is mostly focused around keeping the pipeline from directly knowing about the layer - which is a required change for future work.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
```diff
@@ -0,0 +1,196 @@
...
+private:
+    T _provider;
+};
+
+template <typename T>
+class plMetalRenderDestination<T*> : public plMetalRenderDestinationType
+{
+public:
+    template <typename... Args>
+    plMetalRenderDestination(Args&&... args) : _provider(new T(std::forward<Args>(args)...))
```
Hoikas: Should probably be using `std::make_unique` here

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
```diff
@@ -0,0 +1,196 @@
...
+        other.colorTexture = nullptr;
+        other.drawable = nullptr;
+    }
+
+    // Copy constructor - retain to create independent copy
+    plMetalRenderSurface(const plMetalRenderSurface& other)
+    {
+        this->colorTexture = other.colorTexture->retain();
+        this->drawable = other.drawable->retain();
+    }
```
Hoikas: What's the deal with the uses of `this` and not using initializers?

Hoikas (commented):
This PR does not seem to utilize the [preferred code style](https://github.com/H-uru/Plasma/wiki/Preferred-Code-Style).

colincornaby:
> This PR does not seem to utilize the [preferred code style](https://github.com/H-uru/Plasma/wiki/Preferred-Code-Style).

Thanks - I'm still juggling multiple C++ code styles in my head. It looks like they got conflated again.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
colincornaby: I honestly don't remember and I've been wracking my brain. This code is about six months old - but it started as a Swift class. And then it migrated to C++ when it looked like having the base code written in C++ would be easier than writing this component in Swift. But I don't remember what I did when I was bringing this code over from Swift - this looks closer to a direct translation.

I should have caught this sort of refinement before opening the PR. That's on me.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
colincornaby: Changed.

colincornaby:
I've gone over the code and I think I've brought it in line with the coding standards. There's a chance I missed something.

I have not run a full clang-format pass in since we've agreed that may not be helpful.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -260,7 +265,17 @@ @interface AppDelegate : NSWindowController <NSApplicationDelegate,
...
+    AppDelegate* appDelegate = (AppDelegate*)[NSApp delegate];
+    plPipeline* pipeline = appDelegate->gClient->GetPipeline();
+    appDelegate.metalPipeline = plMetalPipeline::ConvertNoRef(pipeline);
+    appDelegate.glPipeline = plGLPipeline::ConvertNoRef(pipeline);
+
+    if (appDelegate.metalPipeline) {
+        appDelegate.metalPipeline->SetRenderDestination(appDelegate.renderDestination);
+    }
+}
 void plClient::FlashWindow()
```
Hoikas, line 285: Minor nitpick: blank lines between functions.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.h
```diff
@@ -169,13 +169,18 @@ class plMetalPipeline : public pl3DPipeline<plMetalDevice>
     MTL::PixelFormat GetFramebufferFormat() { return fDevice.GetFramebufferFormat(); };
     void             SetFramebufferFormat(MTL::PixelFormat format) { fDevice.SetFramebufferFormat(format); };
 
+    void                          SetRenderDestination(plMetalRenderDestinationType* renderDestination) { fRenderDestination = renderDestination; }
+    plMetalRenderDestinationType* GetRenderDestination() { return fRenderDestination; }
```
Hoikas: This feels like a good candidate to be `const`.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
```diff
@@ -0,0 +1,192 @@
...
+ but they behave like references - similar to std::shared_ptr.
+ That means we need to have a version of plMetalRenderDestination
+ for value types, and another for pointer types.
+
+ The pointer types version is used when we have a real C++
+ type that we need to specialize on. The value type is used
+ when this template specializes on a Swift type.
+ */
+
+template <typename T>
```
Hoikas:
```suggestion
template<typename T>
```

Hoikas (commented):
I still see some code style issues, specifically around pointers needing to bind to the type and incorrect templates. I left a few specific notes, but those are not all inclusive.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
colincornaby: I'll make the change.  But I think the code standards have a space here!

colincornaby:
I can also squash this down when it's ready to go.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
Hoikas: Interesting - so the wiki page does! It looks like new code uses a mixture of both `template<` and `template <`. Sigh.

Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.h
```diff
@@ -0,0 +1,59 @@
...
+#import <Metal/Metal.hpp>
+
+#import <optional>
+
+#import "pfMetalPipeline/plMetalRenderDestination.h"
+
+class plDirectMetalRenderDestination
+{
+public:
+    plDirectMetalRenderDestination(CAMetalLayer *layer);
```
Hoikas:
```suggestion
    plDirectMetalRenderDestination(CAMetalLayer* layer);
```

Sources/Plasma/Apps/plClient/Mac-Cocoa/plDirectMetalRenderDestination.mm
```diff
@@ -0,0 +1,68 @@
...
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#import <Metal/Metal.h>
+
+#import "plDirectMetalRenderDestination.h"
+
+plDirectMetalRenderDestination::plDirectMetalRenderDestination(CAMetalLayer *layer) : fLayer(layer)
```
Hoikas, line 47:
```suggestion
plDirectMetalRenderDestination::plDirectMetalRenderDestination(CAMetalLayer* layer) : fLayer(layer)
```

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalRenderDestination.h
```diff
@@ -0,0 +1,192 @@
...
+#import <Metal/Metal.hpp>
+#import <QuartzCore/QuartzCore.hpp>
+#import <optional>
+
+struct plMetalRenderSurface
+{
+    // Required - the color texture to draw to.
+    MTL::Texture*      colorTexture;
+    // Optional - only provided for a CA based windowing system
+    CA::MetalDrawable* drawable;
```
Hoikas: These still need to use the field prefix.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalPipeline.h
```diff
@@ -169,13 +169,18 @@ class plMetalPipeline : public pl3DPipeline<plMetalDevice>
     MTL::PixelFormat GetFramebufferFormat() { return fDevice.GetFramebufferFormat(); };
     void             SetFramebufferFormat(MTL::PixelFormat format) { fDevice.SetFramebufferFormat(format); };
 
+    void SetRenderDestination(plMetalRenderDestinationType* renderDestination) { fRenderDestination = renderDestination; }
+    const plMetalRenderDestinationType* GetRenderDestination() { return fRenderDestination; }
```
Hoikas: I was thinking more along the lines of


```suggestion
    plMetalRenderDestinationType* GetRenderDestination() const { return fRenderDestination; }
```

to be honest.

Sources/Plasma/FeatureLib/pfMetalPipeline/plMetalDevice.h
```diff
@@ -102,6 +104,8 @@ class plMetalDevice
     MTL::CommandQueue*         fCommandQueue;
     MTL::Buffer*               fCurrentIndexBuffer;
 
+    plMetalRenderDestinationType** fRenderDestination;
```
Hoikas: I'm not really sure I can put my finger on why this needs to be a pointer to a pointer. Can you clarify?

colincornaby: This comes in as part of the next branch. The client needs to be able to change the render destination between frames. Thought maybe an easy way to make that indirect was to make a pointer to a pointer.

Modern Metal vsync has a callback you receive a frame one. Asking for a frame will cause an exception. But that's not compatible with the intro movie playback phase of the client startup. The intro movie startup loop implicitly asks for a frame buffer to render into for each frame of the movie.

So in the final shape of this code, the Mac client:

- Starts with a render destination that is not vsynced.
- Completes the intro movie and allows that movie draw loop to ask for frames
- After the intro movie is done - setupRunLoop is called and the client swaps the render destination for a vsync'd render frame.
- The run loop of the game is now triggered by vsync.

Hoikas: Is the "next branch" #1877? I'm still struggling with conceptualizing why we need to use a pointer to a pointer - that feels like a very C level thing that is out of place in modern C++, which we are trying to encourage new code to use. It feels to me like, since C++ has classes, a pointer to the render destination object is sufficient, and its insides can be plumbed as appropriate. Mind you, I don't have the full picture here, and maybe I'm missing something or am on an island.

colincornaby: I'll back up and explain the problem. Maybe instead of focusing on the code that is here it would be helpful to go back to here's the problem/how would we fix it? This is what is coming in #1877.

In modern Metal - Apple changed how things work depending on if vsync is on or off.
- When vsync is off, getting a frame buffer is on the pull model. You can ask for and receive a framebuffer to render to.
- When vsync is on, getting a frame buffer is on the push mode. You must wait for a framebuffer to be delivered to you through a callback.

This means there are going to be two different render destinations. One managed by vsync and the other one unmanaged.

The intro movie _requires_ the pull model. It runs on its own loop and demands frame buffers to render into. So we cannot use the vsync render destination with the intro movie. The vsync render destination works on the push model.

The Plasma startup sequence turns into the following:

- Plasma starts up with "pull" render destination
- The intro movie starts and beings pulling frames
- Intro movie finishes
- The client destroys the "pull" render destination
- The render loop setup begins
- The clients creates a vsync/"push" render destination
- Each push of a frame buffer triggers the run loop

The pointer to a pointer is covering this transition. Instead of passing a pointer to the "pull" frame buffer at startup - I'm passing a pointer to the pointer to whatever the active render destination is.

This isn't the only place this could happen. The macOS client always has vsync on right now. But if I implemented vsync on/off I would need to do the same transition between the two kinds of frame buffer sources.

Hoikas: That is what I ascertained from your previous comment. I would prefer to focus on code because I do not think a pointer to pointer is the right choice for new code in the engine. Pointers to pointers are very much a C thing and can be difficult to reason about. This is a C++ project that we're trying to modernize. IMO, a more idiomatic C++ approach is a wrapper object holding the OS render destination resource that the client owns and the device holds a pointer to. When vsync is on, the client can swap the OS render destination resource when the OS pushes a new one, without exposing `plMetalDevice`'s internal state to arbitrary modification.

colincornaby: Thanks. Let me think about what the clearest terminology is for that API and I'll implement it. I'm passing the render destination through the display argument to the pipeline. It might be a good time to bridge how a display relates to a render surface through this wrapper type.

colincornaby:
Ok. I've been thinking about the problem @Hoikas pointed out all week. What he's pointing at is a shared problem between the OpenGL and Metal renderers. I could write a quick Metal wrapper and pass it in through the window parameter. I hate the idea of expanding this PR - but the solutions @dpogue and I are coming up with here are divergent between Metal and OpenGL. So it might be time to talk it out.

@dpogue - I think you can probably see the issue here. What he's pointing to is the API for how the layer should talk to the renderer. I'm thinking through if a wrapper type should be more generic. A wrapper type could allow the renderer to configure either OpenGL or Metal in the macOS renderer. But it would be weird to do that just for macOS and make macOS some special case. I think?

I don't know if we need to step back and decide:
- How multiplatform renderers ask for the window server integration bits.
- If those window server bits should be owned by the client or the renderer.

dpogue:
On the OpenGL side, for non-macOS there's not really a concept of layers:

* EGL creates a rendering surface from a window and a display device (the exact values/type of those vary depending on the OS and display server)
* GLX creates a rendering surface from an X11 window
* WGL creates a rendering surface from a Win32 window and display device context
* CGL **can** render directly to a window but it's considered better practice to go through a layer like `CAOpenGLLayer` or `NSOpenGLView`

Direct3D (as far as I can tell) just renders directly to the window.

I [previously experimented](https://github.com/H-uru/Plasma/pull/1554) with making the pipeline provide the macOS layer instance, but you flagged concerns about that due to non-macOS platforms having different preconceptions about how rendering should be done.

---

The other aspect here seems to be around VSync and keeping rendering tied to the refresh rate. I haven't implemented that at all in OpenGL yet, but [a quick search](https://stackoverflow.com/a/38914829) seems to imply that it's the sort of thing that can be set up with OpenGL extensions that don't need anything from the window display server. On macOS there's even `kCGLCPSwapInterval` that supposedly handles it.

So it's possible that this particular type of complexity is in fact just Apple-specific, but it would be good to make sure Metal and OpenGL are approaching it the same way.

colincornaby:
> I [previously experimented](https://github.com/H-uru/Plasma/pull/1554) with making the pipeline provide the macOS layer instance, but you flagged concerns about that due to non-macOS platforms having different preconceptions about how rendering should be done.

I took a look at that code. I thought the approach of having subclassed devices was pretty elegant. But I struggled a bit with how to get the layer back up to the client. I wasn't sure if that was solved in your branch. It's also pushing Obj-C down into the renderer (or Swift) which has been a pain point.

For your vsync question - where this started was with the issue @Hoikas found where the main draw loop was too aggressive and was preventing other code on the main thread from being surfaced. [CAMetalDisplayLink](https://developer.apple.com/documentation/quartzcore/cametaldisplaylink?language=objc) seems to be the way we're meant to solve this issue. But there are two issues:

- Not compatible with GL at all - so I can't only implement a CAMetalDisplayLink. I'm hoping to get OpenGL back on the Mac soon to help support development, even if it's just the stub.
- When a CAMetalDisplayLink is active, calling [nextDrawable](https://developer.apple.com/documentation/quartzcore/cametallayer/nextdrawable()?language=objc) on the CAMetalLayer to get a frame will cause an exception. I need to wait for a [delegate callback](https://developer.apple.com/documentation/quartzcore/cametaldisplaylinkdelegate/metaldisplaylink(_:needsupdate:)?language=objc) from the link to get a framebuffer.

The intro movie pulls frame buffers from the layer so I can't have the CAMetalDisplayLink active during the intro movie.

> On macOS there's even kCGLCPSwapInterval that supposedly handles it.

Yeah, I think we're trying to solve the same problem. In Metal - I need to reach back into either the timer or the layer to enable/disable vsync. In OpenGL you have to reach back into macOS's CGL.

One reason I'm considering OpenGL right now is because it might be helpful to align on whatever that wrapper is out of the engine and into the window server.

> flagged concerns about that due to non-macOS platforms having different preconceptions about how rendering should be done.

For transparency - I'm specifically thinking about RealityKit's [LowLevelTexture](https://developer.apple.com/documentation/realitykit/lowleveltexture) which is required for stereoscopic rendering on visionOS. It is Swift only and has a [very different model](https://developer.apple.com/documentation/realitykit/lowleveltexture/replace(using:)) for retrieving a texture to render to.

dpogue:
> When a CAMetalDisplayLink is active, calling [nextDrawable](https://developer.apple.com/documentation/quartzcore/cametallayer/nextdrawable()?language=objc) on the CAMetalLayer to get a frame will cause an exception. I need to wait for a [delegate callback](https://developer.apple.com/documentation/quartzcore/cametaldisplaylinkdelegate/metaldisplaylink(_:needsupdate:)?language=objc) from the link to get a framebuffer.

yeah, this delegate callback thing was the part where my plans kinda fell apart...

Presumably CADisplayLink is the non-Metal version of CAMetalDisplayLink, though that's only available on macOS 14+. Prior to that, it seems like there was a C-based CVDisplayLink API.

WebKit has some code that acted as a wrapper over both CADisplayLink and CVDisplayLink in case that's useful for inspiration: https://github.com/WebKit/WebKit/commit/c206b6b473adbc878d70f41e79ec4b2dbe5dc157 (they also have some sort of implementation for Linux, but it seems to be timer based)

colincornaby:
The follow on to this PR - #1877 implements wrappers over all three. Including CVDisplayLink to support development for older macOS versions.

One thing I like about your design is it lets the pipeline control which renderer integration is setup. In the direction this is going the client needs to do a bunch of if'ing to decide which layer goes with which pipeline.

I don't know if that's the right approach. It is designed around different clients having different requirements - which then leans on the client to make those decisions.

I wish the intro movie run on a render loop instead of driving its own loop. That would let me set vsync from the beginning and not transition between the two renderers. I think every other movie in the engine runs on the normal render loop.

colincornaby:
Ok, I have a wrapper in. @dpogue - your review would be helpful here. This is based on the changes in #1325. I'm not branched from those changes - but I did add a `plClientWindow` stub. I also added a `plMetalWindow` and then a `plMacClientWindow` that inherits from both. My thinking is at some point there would be a `plGLWindow` that `plMacClientWindow` would also inherit from.

What I don't like - and why this took me so long - is the possibility of a diamond inheritance problem. Ideally - `plMetalWindow` would be a subclass of `plClientWindow`. But then we could end up in a situation where:

```text
plClientWindow
      |
      +-----> plMetalWindow -----+
      |                          |
      +-----> plGLWindow --------+-----> plMacClientWindow
```

I tried clever ways to avoid this sort of multiple inheritance thing. So far `plMetalWindow` and `plClientWindow` are pure virtuals - so maybe thats a little less dangerous. I think C++20 concepts could be helpful here. But we don't support C++20 yet.

Because the concept of a `plClientWindow` is not available in the pipeline, `plMetalWindow` doesn't inherit from it yet. But if we intend to pass client windows through the window pointer we'll need to come up with ideas for that.

PR #1864, opened 2026-05-01
Allow certain safe console commands through net message screener again

### LINKED ISSUE #1831: Python PtConsoleNet it's not working properly
Hi

This used to work correctly. However, it has since been discovered that this is no longer the case.
The change is applied to the user who executes it.
But it is not applied to the other user.

Example:
`PtConsoleNet('Graphics.Renderer.Fog.SetDefColor 1 1 1', 1)`

dgelessus:
This un-breaks `PtConsoleNet` for visual effects in multiplayer, as used by TOC-MOUL and bots. Fixes #1831.

@Filtik (and perhaps @Mirphak): Could you test this to confirm that it fixes what you need?

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetClient/plNetClientMsgScreener.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetClient/plNetClientMsgScreener.cpp
@@ -41,6 +41,10 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 *==LICENSE==*/
 
 #include "plNetClientMsgScreener.h"
+
+#include <string_theory/string>
+#include <unordered_set>
+
 #include "plCreatableIndex.h"
 #include "plNetLinkingMgr.h"
 
@@ -57,11 +61,14 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include "plAvatar/plCoopCoordinator.h"
 #include "plMessage/plAvCoopMsg.h"
 #include "plMessage/plAvatarMsg.h"
+#include "plMessage/plInputEventMsg.h"
 #include "plMessage/plLoadAvatarMsg.h"
 #include "plMessage/plLoadCloneMsg.h"
 #include "plStatusLog/plStatusLog.h"
 
 #include "pfMessage/pfKIMsg.h"
+
+using namespace ST::literals;
 
 ///////////////////////////////////////////////////////////////
 // CLIENT Version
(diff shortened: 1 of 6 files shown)


Sources/Tests/PubUtilTests/plNetClientTest/test_plNetClientMsgScreener.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "pfAudio/pfAudioCreatable.h"
+#include "pfCamera/pfCameraCreatable.h"
+#include "pfCharacter/pfCharacterCreatable.h"
+#include "pfConditional/plConditionalObjectCreatable.h"
+#include "pfGameGUIMgr/pfGameGUIMgrCreatable.h"
+#include "pfGameMgr/pfGameMgrCreatable.h"
+#include "pfJournalBook/pfJournalBookCreatable.h"
+#include "pfMessage/pfMessageCreatable.h"
+#include "pfPython/pfPythonCreatable.h"
+#include "pfSurface/pfSurfaceCreatable.h"
```
dgelessus: I'm not very happy with this. Linking in all these libraries makes the test .exe about 50 MB large (and the .ilk and .pdb even larger) and increases link and startup times. But without these libraries, the test fails to link at all. I guess something here depends on pfPython and that pulls in everything else.

Filtik:
Works perfectly

Sources/Tests/PubUtilTests/plNetClientTest/test_plNetClientMsgScreener.cpp
dgelessus: With #1866, I could drop `pfCharacter`, `pfGameGUIMgr`, `pfGameMgr`, `pfJournalBook`, `pfPython`, and `pfSurface`. That improves things a bit, but not much.

Hoikas: Have you tried turning off unity and seeing if that impacts the binary size? IME, MSVC has a harder time dropping unused code from the final executable when unity is enabled. This is irrelevant for client builds because the client needs basically all of the code, but for smaller tools it makes a huge difference.

Sources/Plasma/PubUtilLib/plNetClient/plNetClientMsgScreener.cpp
```diff
@@ -136,6 +140,74 @@ bool plNetClientMsgScreener::AllowOutgoingMessage(const plMessage* msg) const
...
+ * because some of them write files to disk (such as Graphics.Renderer.GrabCubeMap).
+ */
+static const std::unordered_set<ST::string> kSafeConsoleCommandNames = {
+    "Graphics.Renderer.Fog.SetDefColor",
+    "Graphics.Renderer.Fog.SetDefExp",
+    "Graphics.Renderer.Fog.SetDefExp2",
+    "Graphics.Renderer.Fog.SetDefLinear",
+    "Graphics.Renderer.SetClearColor",
+    "Graphics.Renderer.SetYon",
+    "Graphics.Renderer.Setyon", // non-standard capitalization used at least by Mir-o-bot as of 2026-04
```
Hoikas: Could we instead use a case insensitive set?

Sources/Tests/PubUtilTests/plNetClientTest/test_plNetClientMsgScreener.cpp
dgelessus: I already have unity builds disabled :( I also have LTO disabled - does that increase binary size much?

I would guess that most of the size comes from real dependencies though - there are lots of small cross-library dependencies in PubUtilLib...

Sources/Plasma/PubUtilLib/plNetClient/plNetClientMsgScreener.cpp
dgelessus: Probably yes - or just a `for` + `compare_i`, because the list is so small. I just wanted to avoid unneeded flexibility here (all commands except this one seem to be capitalized correctly).

Mirphak: I see I wrote Setyon instead of SetYon because it's writen Setyon in the .fni files. I can correct my code.

I took a look at the PtConsoleNet I use in Mir-o-Bot scripts :
Graphics.Renderer.SetClearColor
Graphics.Renderer.Fog.SetDefColor
Graphics.Renderer.Fog.SetDefLinear
Graphics.Renderer.Setyon
Graphics.Renderer.SetYon
Graphics.SetDebugFlag noFog
Avatar.Spawn.DontPanic
Nav.PageInNode
Nav.PageOutNode

and Mystitech use also (I don't think all need to be net propagated) :
Avatar.ShowLocations
Avatar.Spawn.DontPanic
Avatar.Spawn.Go
Avatar.Warp.WarpPlayerHere
Avatar.Warp.WarpToPlayer
Avatar.Warp.WarpToXYZ
Camera.SetFOV
Camera.SetGlobalVelocity
Camera.UseSpeedOverrides
Net.LinkPlayerHere
Net.LinkPlayerToPrevAge
Net.LinkToPlayersAge

dgelessus: Thanks, I'll add the other commands for Mir-o-Bot too. By the way, some of the commands are disabled in external builds (`Avatar.Spawn.DontPanic`, `Graphics.SetDebugFlag`, `Nav.PageOutNode`), so they won't work in normal gameplay even if I add them to the safe commands list.

Yes, the .fni files are inconsistent about `SetYon`/`Setyon`. If you could change Mir-o-Bot to use `SetYon`, that would simplify things 🙂

Sources/Tests/PubUtilTests/plNetClientTest/test_plNetClientMsgScreener.cpp
Hoikas: I find that LTO is very successful in eliminating unused code from cross-library dependencies. It does increase the link time pretty significantly, though. When I was more actively doing engine work, I would turn off LTO for certain configurations (like debug) so I could iterate quickly but have smaller release-like builds.

Sources/Plasma/PubUtilLib/plNetClient/plNetClientMsgScreener.cpp
Hoikas: To my mind, it would be surprising that a command that's accepted by the console itself would be filtered out due to different casing. I would probably just change this to an `std::unordered_set<ST::string. ST::hash_i, ST::equal_i>`. We probably already have (or will eventually have) other instances of that container in the code.

Hoikas (approved):

dgelessus:
Rebased after #1866 to reduce the test dependencies, [as discussed above](https://github.com/H-uru/Plasma/pull/1864#pullrequestreview-4213750933). No other changes.

Hoikas (approved):

PR #1867, opened 2026-05-02
Assorted improvements/cleanups in `plAgeDescription`

dgelessus:
The most significant change: rewrote the .age file parser using manual string splitting instead of `plInitFileReader`. The format is so simple that the parser is easier to understand this way IMO. An .ini parser seemed odd here anyway - an .age file isn't supposed to have sections.

This was actually the last use of `plInitFileReader`, so I removed it entirely. I don't think we'll need it again, because the engine no longer uses any real .ini files. But if anyone wants to keep it, let me know.

Other small improvements:

* Added some syntax checking to the parser (using `hsAssert`, so works in debug builds only).
* Allowed iterating over pages using a normal range-`for` loop instead of that `SeekFirstPage`/`GetNextPage` nonsense.
* Added some useful getter/setter variants.
* Fixed moveability.


Sources/Plasma/PubUtilLib/plAgeDescription/plAgeDescription.cpp
```diff
@@ -292,84 +290,59 @@ void plAgeDescription::Write(hsStream* stream) const
...
+        hsAssert(parts.size() == 2, "Syntax error in .age file: missing equals sign");
+        ST::string name = std::move(parts[0]);
+        ST::string value = std::move(parts[1]);
+
+        ST::conversion_result result;
+        if (name.compare_i("StartDateTime") == 0) {
+            // Cyan's original code copied the StartDateTime into a 10-char buffer,
+            // silently truncating any digits after the first 10...
+            // Not sure if anything relies on this by accident, so just in case:
+            hsAssert(value.size() <= 10, "Syntax warning in .age file: StartDateTime is longer than 10 characters - it will be truncated for compatibility!");
```
Hoikas, line 298: I seem to remember that there's something about either the cavern day length or cavern epoch that isn't or wasn't quite right. Probably due to something like this. I don't remember anything more specific than that, though.

Sources/Plasma/PubUtilLib/plAgeDescription/plAgeDescription.cpp
```diff
@@ -117,19 +117,13 @@ char        plAgeDescription::kAgeDescPath[] = { "dat" PATH_SEPARATOR_STR };
 const char* plAgeDescription::fCommonPages[] = { "Textures", "BuiltIn" };
 
 plAgeDescription::plAgeDescription()
+    : fDayLength(24.0f),
+      fMaxCapacity(-1),
+      fLingerTime(180), // seconds
+      fSeqPrefix(0),
+      fReleaseVersion(0)
```
Hoikas:
```suggestion
      fSeqPrefix(),
      fReleaseVersion()
```

Sources/Plasma/PubUtilLib/plAgeDescription/plAgeDescription.cpp
dgelessus: I vaguely remember *some* discussion about digits getting truncated, and I *think* it was something similar to what you say... which is why I decided to put this hack in until someone remembers :D

All the .age files in this repo and in MOULa-current-content have a `StartDateTime` of exactly 10 digits, so I'm not sure where such a problem might have happened...

Hoikas: I think the discussion might have centered around us trying to add more digits and finding they had no effect. But, yeah, I think leaving in the current limitation and documenting it more clearly is good for now.

Hoikas (approved):

PR #1869, opened 2026-05-06
Cleaning up Metal Matrix Multiplies

colincornaby:
Some of the HLSL translated code is manually performing matrix muls with dot products. The Metal code also carried that in by declaring parts of the buffers as individual float4 rows instead of matrices. Moving code over to normal matrix muls for clarity.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveSet7.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveSet7.metal
@@ -65,9 +65,7 @@ typedef struct
     float4 WindRot;
     float4 EnvAdjust;
     float4 EnvTint;
-    float4 LocalToWorldRow1;
-    float4 LocalToWorldRow2;
-    float4 LocalToWorldRow3;
+    float3x4 LocalToWorld;
     float4 Lengths;
     float4 WaterLevel;
     float4 DepthFalloff;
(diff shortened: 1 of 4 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveSet7.metal
```diff
@@ -101,20 +99,7 @@ vertex vs_WaveFixedFin7InOut vs_WaveFixedFin7(Vertex in                     [[st
...
-    matrix_float4x3 localToWorld;
-    localToWorld[0] = column1;
-    localToWorld[1] = column2;
-    localToWorld[2] = column3;
-    localToWorld[3] = column4;
-
-    float4 worldPosition = float4(localToWorld * float4(in.position, 1.0), 1.0);
-
-    //
+    float4 worldPosition = float4(float4(in.position, 1.f) * uniforms.LocalToWorld, 1.f);
```
dpogue, line 102: Does the order of the multiplication here matter? It looks like that changed

colincornaby: It does. D3D is row major, Metal is column major. The old code manually flipped the matrix. But that's kind of silly. By flipping the multiply order the shader can use row major matrices.

So yes - but it's intentional.

dpogue (approved):

PR #1870, opened 2026-05-06
Remove last uses of plContainer

dgelessus:
plContainer mainly consists of `plConfigInfo`, which is another .ini file reader/writer that I don't think we'll need again, because we don't use any real .ini files anymore. (Just like `plInitFileReader`, which I removed in #1867.)

(Actually, it's a bit more generic than *just* .ini parsing: there's also support for reading settings from other sources, like environment variables. None of this is ever used though, and because it's so .ini-centric, I don't think it's a useful abstraction for us.)

The only thing that uses `plConfigInfo` is the netcode for sending CCR petitions, which serializes the petition into .ini format. This is easy to reimplement using manual string building. The code was already broken though, because it calls an unimplemented NetCliAuth function. So I replaced the entire petition sending code path with an unconditional assert. (I'm not removing any other code related to CCR petitions, so we can theoretically fix them in the future by implementing new netcode here.)

Everything else deleted in this PR is support code that was only used in plContainer and nowhere else.


Hoikas (approved):
I was initially thinking that this was all good except [1e39c50](https://github.com/H-uru/Plasma/pull/1870/commits/1e39c5079dbc60f2180a44b8f0711481d29372b9) should be stripped out. But, honestly, all of that CCR petition stuff is from the original Uru Live, and the ini format for support requests is weird. If anyone ever needs it back, that commit is trivial to revert.

PR #1872, opened 2026-05-08
Improve `hsRef` and simplify ref management in a few places

dgelessus:
Replaces `hsRefCnt_Safe[Un]Ref` calls by direct calls to `Ref`/`UnRef` if it's clear from context that the pointer cannot be null - because it was already checked, or because it comes directly from `new`, or because the code path already assumes that the pointer is not null.

This fixes some false positive warnings in CLion about missing null checks - CLion takes the null check inside the "safe" macros as an indication that the pointer can indeed be null, and then warns about any other uses of the same pointer variable without an explicit null check.

### DIFF
--- a/Sources/MaxPlugin/MaxComponent/plAnimEventComponent.cpp
+++ b/Sources/MaxPlugin/MaxComponent/plAnimEventComponent.cpp
@@ -167,7 +167,7 @@ plEventCallbackMsg* CreateCallbackMsg(plAnimCmdMsg* animMsg, plKey modKey)
     eventMsg->fRepeats = -1;
 
     animMsg->AddCallback(eventMsg);
-    hsRefCnt_SafeUnRef(eventMsg);   // AddCallback adds it's own ref, so remove ours (the default of 1)
+    eventMsg->UnRef(); // AddCallback adds it's own ref, so remove ours (the default of 1)
 
     return eventMsg;
 }
--- a/Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
+++ b/Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
@@ -688,7 +688,7 @@ public:
         kRefFontShadowed
     };
 
-    static void     ConvertScheme( IParamBlock2 *pb, pfGUIColorScheme *destScheme, plErrorMsg *pErrMsg );
+    static void ConvertScheme(IParamBlock2* pb, hsWeakRef<pfGUIColorScheme> destScheme, plErrorMsg* pErrMsg);
 };
 
 //Max desc stuff necessary below.
@@ -977,9 +977,9 @@ bool plGUIColorSchemeComp::Convert(plMaxNode *node, plErrorMsg *pErrMsg)
     pfGUIControlMod *ctrl = plGUIControlBase::GrabControlFromObject( node );
     if (ctrl != nullptr)
     {
-        pfGUIColorScheme *cs = new pfGUIColorScheme;
+        hsRef cs(new pfGUIColorScheme(), hsStealRef);
         ConvertScheme( fCompPB, cs, pErrMsg );
-        ctrl->SetColorScheme( cs );
+        ctrl->SetColorScheme(std::move(cs));
     }
     else
     {
@@ -996,7 +996,7 @@ void    SMaxRGBAToPlasmaRGBA( Color maxRGB, hsColorRGBA &plasmaRGBA )
     plasmaRGBA.Set( maxRGB.r, maxRGB.g, maxRGB.b, 1.f );
 }
 
-void    plGUIColorSchemeComp::ConvertScheme( IParamBlock2 *pb, pfGUIColorScheme *destScheme, plErrorMsg *pErrMsg )
+void plGUIColorSchemeComp::ConvertScheme(IParamBlock2* pb, hsWeakRef<pfGUIColorScheme> destScheme, plErrorMsg* pErrMsg)
 {
     SMaxRGBAToPlasmaRGBA( pb->GetColor( kRefForeColor ), destScheme->fForeColor );
     SMaxRGBAToPlasmaRGBA( pb->GetColor( kRefBackColor ), destScheme->fBackColor );
--- a/Sources/Plasma/CoreLib/hsRefCnt.h
+++ b/Sources/Plasma/CoreLib/hsRefCnt.h
@@ -90,19 +90,42 @@ class hsRef
 public:
     hsRef() : fObj() { }
     hsRef(std::nullptr_t) : fObj() { }
-    hsRef(_Ref *obj, hsStealRef_Type) : fObj(obj) { }
-    hsRef(const hsRef<_Ref> &copy) : fObj(copy.fObj) { if (fObj) fObj->Ref(); }
-    hsRef(hsRef<_Ref> &&move) : fObj(move.fObj) { move.fObj = nullptr; }
+
+    hsRef(_Ref* obj, hsStealRef_Type) : fObj(obj) {}
+
+    // Needed to allow passing an S* into an hsRef<T> parameter (where S is a subclass of T).
+    hsRef(_Ref* copy) : fObj(copy) { if (fObj) fObj->Ref(); }
+    // This is redundant with the next overload, but allows template argument deduction for hsRef(otherRef).
+    hsRef(const hsRef<_Ref>& copy) : fObj(copy.Get()) { if (fObj) fObj->Ref(); }
+    // Needed to allow passing an hsRef<S> into an hsRef<T> parameter (where S is a subclass of T).
+    template<class _OtherRef>
+    hsRef(const hsRef<_OtherRef>& copy) : fObj(copy.Get()) { if (fObj) fObj->Ref(); }
+    // Most of the other constructor and operator= overloads are duplicated for the same reasons.
+
+    hsRef(hsRef<_Ref>&& move) : fObj(move.Release()) {}
+    template<class _OtherRef>
+    hsRef(hsRef<_OtherRef>&& move) : fObj(move.Release()) {}
 
     ~hsRef() { if (fObj) fObj->UnRef(); }
 
     hsRef<_Ref> &operator=(const hsRef<_Ref> &copy)
     {
-        if (copy.fObj)
-            copy.fObj->Ref();
+        if (copy)
+            copy.Get()->Ref();
         if (fObj)
             fObj->UnRef();
-        fObj = copy.fObj;
+        fObj = copy.Get();
+        return *this;
+    }
+
+    template<class _OtherRef>
+    hsRef<_Ref>& operator=(const hsRef<_OtherRef>& copy)
+    {
+        if (copy)
+            copy.Get()->Ref();
+        if (fObj)
+            fObj->UnRef();
+        fObj = copy.Get();
         return *this;
     }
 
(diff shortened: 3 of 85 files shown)


Sources/MaxPlugin/MaxComponent/plAnimEventComponent.cpp
```diff
@@ -167,7 +167,7 @@ plEventCallbackMsg* CreateCallbackMsg(plAnimCmdMsg* animMsg, plKey modKey)
     eventMsg->fRepeats = -1;
 
     animMsg->AddCallback(eventMsg);
-    hsRefCnt_SafeUnRef(eventMsg);   // AddCallback adds it's own ref, so remove ours (the default of 1)
```
Hoikas, line 170: There's quite a bit of this anti-pattern repeated. I wonder if it would make more sense to change `AddCallback` to accept an `hsRef` and change all of these call sites to create an hsRef of whatever message is being sent as the callback and `std::move` it as the callback. That's potentially a larger change though, and I don't think that should be a requirement.

Hoikas (commented):
I think this is fine, but I wonder if we should take this as an opportunity to use `hsRef` instead. There are several places where we have vectors of `hsRefCnt` and `hsRefCnt` members that we're manually performing the ref count management on, and I think changing those to `hsRef` would be a good code quality improvement.

dgelessus:
Yes, `hsRef` would definitely be the nicer solution in many cases. I've changed some of the raw pointers to `hsRef`/`hsWeakRef`, in a few places where it doesn't require too many changes. I'm not going to `hsRef`-ify everything touched by this PR, otherwise I'd be busy for a while with the rendering pipelines :D

Sources/MaxPlugin/MaxComponent/plAnimEventComponent.cpp
dgelessus: I'm wary of using `hsRef` with messages in particular, because then calling `msg->Send()` as usual will cause a double unref - you have to remember to use `msg->SendAndKeep()` instead. Which is unfortunate, because `hsRef` would definitely be useful for non-straightforward message fiddling code.

I wonder if we should add a dedicated `plMessageRef` class with its own `Send` method that automatically nulls out the ref after sending. Sadly, that still doesn't prevent you from calling `msg->Send()` (wrong) instead of `msg.Send()` (correct).

Sources/Plasma/PubUtilLib/plModifier/plAxisAnimModifier.cpp
```diff
@@ -88,7 +88,7 @@ class plAxisInputInterface : public plInputInterface
                 {
                     // Remove ourselves from the stack
                     plInputIfaceMgrMsg *msg = new plInputIfaceMgrMsg( plInputIfaceMgrMsg::kRemoveInterface );
-                    msg->SetIFace( this );
+                    msg->SetIFace(hsRef<plInputInterface>(this));
```
dgelessus: The implicit conversion from a raw pointer to `hsRef` doesn't seem to work if the raw pointer points to a *subclass* of the expected class. Adding an explicit constructor call like this fixes it for some reason (I'm calling exactly the type expected by the method).

Sources/Plasma/PubUtilLib/plMessage/plInputIfaceMgrMsg.h
```diff
@@ -112,8 +105,8 @@ class plInputIfaceMgrMsg : public plMessage
         const plUUID&     GetAgeInstanceGuid() const { return ageInstanceGuid; }
         uint8_t           GetCommand() const { return fCommand; }
         uint32_t          GetPageID() const { return fPageID; }
-        void              SetIFace(plInputInterface *iface);
-        plInputInterface* GetIFace() const { return fInterface; }
+        void              SetIFace(hsRef<plInputInterface> iface);
```
Hoikas, line 108: This will only work if you're passing `hsRef` of `plInputInterface` - instead, you'll have to do something like
```cpp
template<typename _IFaceT>
void SetIFace(hsRef<_IFaceT> iface)
{
    static_assert(std::is_base_of_v<plInputInterface, _IFaceT>);
    // ... code ...
}
```
or (this is probably not preferred, but for completeness)
```cpp
template<typename _IFaceT>
std::enable_if_t<std::is_base_of_v<plInputInterface, _IFaceT>, void>
SetIFace(hsRef<_IFaceT> iface)
{
    // ... code ...
}
```
to allow the derived input interface type through.

dgelessus: hmm, is there a way we can put this into the `hsRef` move constructor? Otherwise we'd have to repeat this in many places if we want to use `hsRef` more extensively...

Hoikas: Hmmm... It might be worth a shot to try writing a conversion move constructor, but I'm not sure the compiler would be able to follow the conversion rules. Maybe @zrax would have ideas?

zrax: It looks like you should be able to add a new copy constructor (and optionally copy operator) to hsRef to handle this:
```c++
    template <class _CRefT, typename = std::enable_if_t<std::is_base_of_v<_Ref, _CRefT>, void>>
    hsRef(const hsRef<_CRefT> &copy) : fObj(copy.Get()) { if (fObj) fObj->Ref(); }

    template <class _CRefT, typename = std::enable_if_t<std::is_base_of_v<_Ref, _CRefT>, void>>
    hsRef<_Ref> &operator=(const hsRef<_CRefT> &copy)
    {
        if (copy.Get())
            copy.Get()->Ref();
        if (fObj)
            fObj->UnRef();
        fObj = copy.Get();
        return *this;
    }
```

I tried replacing the non-template copy constructor with this, but it didn't seem to like that (probably due to the move constructor/operator causing implicit deletion of the default copy constructor).

(NOTE: The use of `Get()` is intentional in the above, since `hsRef<Derived>` is not a friend of `hsRef<Base>` in the current implementation)

Another option would be to provide an explicit conversion method (along the lines of rust's `Into` trait), but I think the above is probably clean/safe enough, especially with the `is_base_of` constraint.

dgelessus: Do we need the manual `is_base_of` check at all? If something passes an incompatible ref, that should already cause a type error at the `fObj = copy.Get()` part, right?

I now added overloads with an unconstrainted `template<class _OtherRef>` for all the constructors/assignment operators, and that seems to work (after replacing some `fObj` accesses, as you describe). I then noticed that `hsWeakRef` also does exactly this already!

If I *replace* the existing constructors with templated versions, I get a "template argument deduction failed" in the vault code (for `device = hsRef(node)` in a lambda capture list). But if I keep the same-type constructors and *add* templated overloads, everything seems to work.

Sources/Plasma/PubUtilLib/plModifier/plAxisAnimModifier.cpp
dgelessus: I fixed this by adding an explicit `hsRef(_Ref* copy)` constructor overload. That works even if the pointer type doesn't match exactly (unlike the indirect conversion via `hsWeakRef` that was previously used here).

Sources/Plasma/PubUtilLib/plMessage/plInputIfaceMgrMsg.h
dgelessus: If we do decide to add an explicit check: [cppreference.com notes that](https://cppreference.com/cpp/types/is_base_of#Notes) "In many situations, `std::is_convertible<B*, A*>` is the more appropriate test" than `std::is_base_of<A, B>`.

Sources/MaxPlugin/MaxComponent/plAnimEventComponent.cpp
Hoikas: That's an interesting idea. It's a big enough change that I think we should keep that in mind for future work, though. We could potentially prevent `msg->Send()` instead of `plMessageRef::Send()` by not exposing `operator->` and instead only offering `Message* Get() const;` access to the message or by having `operator->` return another special wrapper that inherits from the message type and deletes the `Send()` method. That last idea is getting pretty evil, though.

Hoikas (commented):
I think this looks good. I'd appreciate a double check from @zrax, though.

dgelessus:
Friendly bump: are there any remaining issues here that I should address?

This PR touches many different files, so it would be good to not keep it open for a long time, to avoid merge conflicts.

Sources/Plasma/CoreLib/hsRefCnt.h
```diff
@@ -90,28 +90,59 @@ class hsRef
...
     hsRef() : fObj() { }
     hsRef(std::nullptr_t) : fObj() { }
-    hsRef(_Ref *obj, hsStealRef_Type) : fObj(obj) { }
-    hsRef(const hsRef<_Ref> &copy) : fObj(copy.fObj) { if (fObj) fObj->Ref(); }
-    hsRef(hsRef<_Ref> &&move) : fObj(move.fObj) { move.fObj = nullptr; }
+
+    hsRef(_Ref* obj, hsStealRef_Type) : fObj(obj) {}
+
+    // Needed to allow passing an S* into an hsRef<T> parameter (where S is a subclass of T).
+    hsRef(_Ref* copy) : fObj(copy) { if (fObj) fObj->Ref(); }
```
Hoikas, line 97: I think we (well, @zrax) omitted this constructor for a specific reason, but I'm not sure offhand. I think we didn't want to allow implicit conversions. But, since we have separate steal and new reference constructors, I'm ok with this.

Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
```diff
@@ -977,9 +977,9 @@ bool plGUIColorSchemeComp::Convert(plMaxNode *node, plErrorMsg *pErrMsg)
     pfGUIControlMod *ctrl = plGUIControlBase::GrabControlFromObject( node );
     if (ctrl != nullptr)
     {
-        pfGUIColorScheme *cs = new pfGUIColorScheme;
+        hsRef cs(new pfGUIColorScheme(), hsStealRef);
```
Hoikas, line 980: I wonder if the uses of `hsRef` and `hsWeakRef` in this commit (d025ea6db952fe35477c65bf4e713f9156887184) should be changed to include the template parameter - eg `hsRef<pfGUIColorScheme>`. This one is pretty obvious what the underlying type is, but in other places it's a bit more implicit. The rest of this PR seems to prefer listing the template parameter explicitly.

Hoikas (commented):
Yeah, this has been hanging out for awhile... I was hoping that someone other than me would look at this one. Just one observation and one question, then I'd be inclined to accept and merge.

Sources/MaxPlugin/MaxComponent/plGUIComponents.cpp
dgelessus: Yeah that's valid. I've now made the types explicit everywhere, except for cases where the type is already spelled out on the right-hand side.

Hoikas (approved):
I'm going to leave this one open until tomorrow (North America) to see if anyone else wants to chime in. I'll press the shiny green button tomorrow

PR #1877, opened 2026-05-31
Mac render loops

colincornaby:
PR requires #1860 

This PR introduces render loops into the Mac client. The Mac client had an existing render loop before - but it was Mac only and based on a legacy framework. This code modularizes the render loops so we can support more platforms and more APIs.

This PR enables Swift in the Mac code base. Some of the visionOS APIs around render loops are Swift only - so this PR starts to build towards that. It establishes an interop pattern between C++ and Swift using templates that seems to work well. The Swift types in this PR are bridging to C++ as first class C++ types.

There are three kinds of render loops added in this PR:

- plLegacyRenderLoop: Based on CVDisplayLink just like our existing render loop. Supports OpenGL and Metal.
- RenderLoop: Replacement for plLegacyRenderLoop when running on macOS 14 or higher. Based on CADisplayLink. Supports OpenGL and Metal. CVDisplayLink is only available on macOS - this render loop can be used on iOS or visionOS.
- MetalRenderLoop: Metal specific render loop based on CAMetalDisplayLink. CAMetalDisplayLink is a specialized display link provided by Apple just for Metal. It includes deeper run loop management and better timing. It needs to manage the frame buffers so this render loop is both a loop and a render destination. When this render loop is active asking the layer for a drawable will result in an exception from the system frameworks.

RenderLoop/MetalRenderLoop are Swift types. They don't have a three letter prefix like our Obj-C types because Swift has namespaces. They don't have a pl prefix like our C++ types because I'm not sure how to manage the Swift namespace. Prefixes are uncommon with Swift classes but not unheard of. Especially if we're bridging C++ types.


colincornaby:
Hi @dpogue - I think you were curious where things were going with the Mac render destinations PR and how it relates to OpenGL. This PR is the follow on PR - I'm leaving it as draft until the first PR is merged. You should be able to look at the last commit.

Let me know if you have any questions. This is the first PR that involves Swift code - and the changes make use of the C++/Swift interop. I'll need to make sure Actions can successfully build this - and I'm guessing there will be more conversation about Swift. Some of the visionOS API is Swift only - so this is the first step that direction. I'm only writing Swift in places you would not need to back port to older versions of macOS.

colincornaby:
I'm still looking at the build issues here.

I mentioned it in IRC - but there is a new wrinkle in the builds. Our Intel build requires macOS 10.14.0. Swift's modules didn't come preinstalled until 10.14.4. So we need to bundle the dylibs. Xcode does that for us automatically. But builds outside of the Xcode generator won't handle the bundling of those libraries. Apple provides a `swift-stdlib-tool` (invokable with `xcrun swift-stdlib-tool`) that lets us bundle the libraries with a third party toolchain.

This problem doesn't affect our Apple Silicon builds because macOS didn't support Apple Silicon until macOS 11. (Also related to why only the Intel builds are failing in this branch.)

We could also drop support for macOS 10.14.0 and require 10.14.4. That's a patch release that worked on all hardware that supported 10.14.0. I'd be surprised if anyone on the 10.14 release wasn't running the latest patch release.

Hoikas:
I would say that we need to do whatever we must to avoid bringing in dylibs.

colincornaby:
> I would say that we need to do whatever we must to avoid bringing in dylibs.

I think we should move to a minimum of macOS 10.14.4 then.

We can't statically link the Swift stdlib and friends on macOS. In 10.14.4 and higher they put Swift stdlib into the system dynamic library cache. So they intentionally want to relink us against what the system ships with - and so the toolchain doesn't support static linking for macOS.

There are similar issues with the C++ stdlib - which also ships with macOS and is dynamically linked. Hasn't been a problem because our C++ standard has not moved in a while. But if we need a newer C++ stdlib we might also need to move our minimum requirement forward.

Issue #1874, opened 2026-05-08
3dsMax Plugin crashing Max 2022


Filtik:
Hi,
Unfortunately, 3dsMax 2022 crashes immediately after startup with the latest plugin.
I was able to trace the crash back to this commit:
https://github.com/H-uru/Plasma/commit/20d265cfb355f153ec7bfb18861a6560bf0e0b62

Before that, it starts without any issues.

dgelessus:
Filtik investigated a bit more and told me via chat what the problem seems to be. To summarize, apparently:

* The Max plugin loads all Python files *before* it initializes the ResMgr.
* Some ages call `PtPageInNode` inside `__init__`, which makes that call happen during the Max plugin loading process.
* Before my PR #1845, `PtPageInNode`/`PtPageOutNode` silently did nothing when called while there is no ResMgr. In my PR, I added some code in those functions that unconditionally uses the ResMgr. This seems to be what's causing the crash.

(Some of Filtik's investigation was with the use of an AI coding tool, so I'm a bit skeptical about the analysis - but looking at the Max plugin startup code, the explanation seems to be generally correct.)

The question is what to do about this. The easy fix is to make `PtPageInNode`/`PtPageOutNode` safely do nothing if there's no ResMgr, like they did before. (I'll submit a quick PR for this once Filtik has confirmed that my fix works.) This feels more like a workaround than a proper solution though.

I'm wondering if there's a good reason why the Max plugin loads the Python files *before* initializing the ResMgr. That's backwards compared to what the client does, and it feels like that could cause issues elsewhere too. If the plugin would initialize the ResMgr first, then `PtPageInNode` would work normally. On the other hand, I assume we don't actually *want* random age scripts to page in their PRPs during Max startup. So even if we change the initialization order, it's not a good fix on its own.

It seems to me like `__init__` is a bad place to call `PtPageInNode`. The Max plugin *has* to instantiate every Python module's "main class" at startup to get the `id` and `version` (which sadly are instance attributes and not class or module attributes), so it seems like the `__init__` should do as little as possible. That is indeed how most of the scripts are written, but a few (such as `city`, `Cleft`, `grsnPageMaster`) do complex setup inside `__init__` - and sometimes that involves `PtPageInNode` calls, which are apparently the problem here. Perhaps the proper fix would be to move all non-trivial code from `__init__` to a later callback method? `OnInit` might be a good candidate, because that's called very soon after `__init__`.

dpogue:
[...]

PR #1881, opened 2026-06-12
Python: Move work from __init__ to OnInit

dpogue:
Refs #1874.

The `__init__` method is called by the PlasmaMax plugin because it needs the script ID and version values to be set. If we try to do other work there (such as loading pages) it might cause issues because the full client and all of its machinery are not necessarily available. To be safe, we move that work to `OnInit` which is called when the PythonFileMod is initialized by the client.

Some of this might be unnecessary, and I'm not sure if it's better to do all the variable assignments with default values in `__init__` or if it's fine to wait until `OnInit`, but this is a starting point at least.

### DIFF
--- a/Scripts/Python/city.py
+++ b/Scripts/Python/city.py
@@ -68,6 +68,7 @@ class city(ptResponder):
         self.id = 5026
         self.version = 1
 
+    def OnInit(self):
         global IsPublic
         global IsKadishGallery
 
@@ -102,8 +103,8 @@ class city(ptResponder):
             spTitle = "title unknown"
             spName = "spawn point unknown"
 
-        PtDebugPrint("city.__init__(): spTitle = ",spTitle)
-        PtDebugPrint("city.__init__(): spName = ",spName)
+        PtDebugPrint("city.OnInit(): spTitle = ",spTitle)
+        PtDebugPrint("city.OnInit(): spName = ",spName)
 
         ## NOT USING THIS FOR NOW - MAY NEED TO LOAD IN SPECIFIC PAGE(S) FOR FUTURE CITY AREAS...
 #        if spTitle == "KadishGallery":
--- a/Scripts/Python/Cleft.py
+++ b/Scripts/Python/Cleft.py
@@ -72,6 +72,7 @@ class Cleft(ptResponder):
         self.id = 5209
         self.version = 22
 
+    def OnInit(self):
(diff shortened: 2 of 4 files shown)


Hoikas:
I think this is overly aggressive. We really just need to ensure that engine features are not being called until no sooner than `OnInit()`. I would prefer to see things like default values for variables in `__init__()`, because Python language servers have special handling for `__init__()`, as opposed to `OnInit()`, which makes the development experience much nicer.

dpogue:
Is `random.seed()` safe to call from `__init__` or is that going to do some global state stuff (which arguably probably doesn't matter if we're not actually running the scripts)?

Hoikas:
`random.seed()` is safe. It does modify global state, but it's a built-in Python module that doesn't depend on Plasma.

Scripts/Python/grsnWallEventHandler.py
```diff
@@ -74,6 +74,8 @@ def __init__(self):
         self.BlockersHit = 0
         self.startTime = 0
         self.BlockerSfxBlock = True
+
+    def OnInit(self):
         InitEventHandler(self)
```
Hoikas, line 77: This is also a variable setter

dgelessus:
... I actually already did this on a local branch, but then got busy/distracted and never finished and pushed it :D Sorry!

I'll have to compare this with my branch, but it looks correct at first glance.

Scripts/Python/city.py
```diff
@@ -102,8 +103,8 @@ def __init__(self):
             spTitle = "title unknown"
             spName = "spawn point unknown"
 
-        PtDebugPrint("city.__init__(): spTitle = ",spTitle)
-        PtDebugPrint("city.__init__(): spName = ",spName)
+        PtDebugPrint("city.OnInit(): spTitle = ",spTitle)
+        PtDebugPrint("city.OnInit(): spName = ",spName)
```
dgelessus, line 107: There are a couple more prints above these that should also be adjusted.

(Also, most of this is actually dead code, because the "are we linking to Kadish Gallery" check is commented out and `IsKadishGallery` is hardcoded to 0. We should clean that up at some point.)

dgelessus (commented):
I checked now and this matches my unpushed local change. (I also moved the `random.seed()` calls, but I'm fine with leaving those where they are - they should indeed not cause any problems for the Max plugin.)

Hoikas (approved):

PR #1882, opened 2026-06-12
Fix `PtPageInNode` crashing when called during Max plugin startup

### LINKED ISSUE #1874: 3dsMax Plugin crashing Max 2022

dgelessus:
Fixes #1874. I can't test this myself, but @Filtik told me that this fixes the crash.

As discussed in #1874, the proper solution is to not call `PtPageInNode` inside PythonFileMod `__init__` methods at all. #1881 by @dpogue changes the Python scripts accordingly. It's probably still good to have this workaround on the C++ side though - it's not very complex, and it prevents the plugin crashing for no clear reason if one of the old faulty scripts is present. (Which may well happen by accident, because Cyan's scripts have done the wrong thing for a long time with no negative effects, and some fan ages, such as TOC's City of Dimensions, have copied it not knowing any better.)

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
+++ b/Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
@@ -46,6 +46,8 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 #include <string_view>
 #include <vector>
 
+#include "hsResMgr.h"
+
 #include "plMessage/plConfirmationMsg.h"
 #include "plMessage/plLOSRequestMsg.h"
 #include "plNetClientComm/plNetClientComm.h"
@@ -323,6 +325,17 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtPageInNode, args, "Params: nodeName, netForce=
         PYTHON_RETURN_ERROR;
     }
 
+    // In the past, some age scripts called this function in their __init__,
+    // which gets called during Max plugin startup where there is no ResMgr yet.
+    // This "worked" in the past, but isn't supported anymore.
+    // We still guard against this case so that scripts that haven't been updated
+    // will only cause Python errors and not hard-crash the Max plugin on startup.
+    // See: https://github.com/H-uru/Plasma/issues/1874
+    if (hsgResMgr::ResMgr() == nullptr) {
+        PyErr_SetString(PyExc_RuntimeError, "No ResMgr present. PtPageInNode needs a ResMgr to find pages. If you are calling PtPageInNode from your Python modifier __init__, move the call into an OnInit callback to fix this error.");
+        PYTHON_RETURN_ERROR;
+    }
+
     std::vector<plLocation> nodeLocs;
     for (const auto& nodeName : nodeNames) {
         plLocation nodeLoc = plKeyFinder::Instance().FindLocation(ageName, nodeName);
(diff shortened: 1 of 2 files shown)


Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
```diff
@@ -323,6 +325,14 @@ PYTHON_GLOBAL_METHOD_DEFINITION(PtPageInNode, args, "Params: nodeName, netForce=
         PYTHON_RETURN_ERROR;
     }
 
+    // Some age scripts call this function in their __init__,
+    // which gets called during Max plugin startup where there is no ResMgr yet.
+    // Guard against that to prevent crashes.
+    if (hsgResMgr::ResMgr() == nullptr) {
+        // No ResMgr - assume we were called during Max plugin startup and silently do nothing.
+        PYTHON_RETURN_NONE;
+    }
```
Hoikas, line 337: This isn't a pattern that we generally see in the Python glue code. I wonder if it would make more sense to protect `plKeyFinder::FindLocation()` instead, which is I presume where the crash is actually happening. This would make the code more resilient in general.

dgelessus: Yes, the check is unusual, but it was also unusual before I moved it here. Blame Cyan :P My guess is that Cyan ran into a crash similar to ours and worked around it by adding the check to just these two functions.

We could change `plKeyFinder` to return a "nothing found" result if there is no ResMgr, but that would make the `PtPageInNode` glue code throw an exception, which would also cause problems for the Max plugin (though not as bad as a crash). To get the desired "if in Max, silently do nothing" behavior, we need *some* kind of special case here, I think.

I also think it wouldn't be a good design choice to make `plKeyFinder` silently "work" without a ResMgr. Searching for keys/locations fundamentally requires a ResMgr, so if there is none, that should be reported and not silently ignored. I suppose we could add an explicit assert, for better feedback than a hard crash?

Hoikas: Hmm... I'm wondering if the "if in Max, silently do nothing" is actually an undesirable behavior. It's invalid for most Plasma APIs except for `PtDebugPrint` to be called from an `__init__()` method. The previous behavior of these functions made it non-obvious that invalid code just happened to work and, now that you've tightened up some of the page handling, that is no longer true.

I think that returning the "nothing found" invalid location and letting the Python glue set an exception might be the right move here. Any Python files with invalid API calls will not be shown in the Python File component list, the error will be logged in the Python log, and the plugin won't crash.

That's still a change in behavior, but it's not a catastrophic change like crashing. Hopefully it will encourage correctness moving forward.

dgelessus: okay, I changed it to raise a Python exception with an explanatory message instead of silently "succeeding". This gives better feedback than a generic "Could not find any page" error, and avoids having to handle this special case inside `plKeyFinder`.

Hoikas: I was also hoping to see the check against "is the res manager" alive not be here. We typically don't see these kinds of calls into the engine in the glue code itself. OTOH, it looks like this whole function calls directly into the engine anyway, I suppose it's fine. The Python API is pretty sloppy.

Hoikas (approved):

Sources/Plasma/FeatureLib/pfPython/cyMiscGlue2.cpp
dgelessus: Admittedly, there were fewer engine calls in these two glue functions before I made my changes :D It was just the easiest way to get proper error reporting here, and I didn't know that we try to keep the glue files free of lower-level engine calls.

PR #1883, opened 2026-06-16
Document PythonFileMod callback methods

dgelessus:
While looking into #1874, I couldn't find any documentation about all the different callback methods that can be defined in a PythonFileMod script. So I wrote some, based on my understanding of the C++ and Python code.

This is meant as a reference, not a tutorial. Not everything is documented in the same amount of detail, but I wrote at least a minimal explanation for every callback and all parameters.

For now, I put everything in a "loose" Markdown file. These docs would also fit well into Python method docstrings, but there's no good place to put them. I don't want to define all callback methods on the `ptModifier` base class, because I think that would break the detection of which PythonFileMod wants which callbacks (which would lead to many unnecessary calls into Python). Also, with docstrings, there's no way to build custom overview tables like I did here.

I'm not an age builder, so I don't have a good feeling for how these callbacks are used in practice. If any "script writers" could check if this looks right, that would be appreciated 🙂

### DIFF
--- a/Docs/PythonFileMod callback methods.md
+++ b/Docs/PythonFileMod callback methods.md
@@ -1,0 +1,687 @@
+# PythonFileMod callback methods
+
+Every PythonFileMod's script must define a class with the same name as the
+script. This class must subclass `ptModifier` or one of its subclasses,
+`ptMultiModifier` or `ptResponder` (all from the `PlasmaTypes` module).
+
+This modifier class can define callback methods with specific fixed names, which
+will be called by the engine in response to various events.
+
+## Overview
+
+Lifecycle callbacks:
+
+| Name | Parameters | C++ message class | Called when |
+|---|---|---|---|
+| [`__init__`](#__init__) | `(self)` | (none) | Python object created (in-game or in 3ds Max) |
+| [`OnInit`](#oninit) | `(self)` | (none) | `plPythonFileMod` loaded in-game |
+| [`OnFirstUpdate`](#onfirstupdate) | `(self)` | first `plEvalMsg` | before first [`OnUpdate`](#onupdate) |
+| [`OnServerInitComplete`](#onserverinitcomplete) | `(self)` | `plInitialAgeStateLoadedMsg` | age loading about to end (all SDL states received) |
+| [`BeginAgeUnLoad`](#beginageunload) | `(self, avatar: ptSceneobject)` | `plAgeBeginLoadingMsg` | age about to unload |
+
+Event handlers:
+
+| Name | Parameters | C++ message class | Called when |
+|---|---|---|---|
+| [`AvatarPage`](#avatarpage) | `(self, avatar: ptSceneobject, loading: bool, lastOut: bool)` | `plPlayerPageMsg` | avatar links in/out |
+| [`Load`](#load) | `(self)` | `plSDLModifierStateMsg` | `self.SDL` was updated from the network |
+| [`OnAIMsg`](#onaimsg) | `(self, brain: ..., msgType: int, userStr: str, args: ...)` | `plAIMsg` | "critter brain" NPC (Ahnonay quab) events |
+| [`OnAvatarSpawn`](#onavatarspawn) | `(self, arg: bool)` | `plAvatarSpawnNotifyMsg` | armature spawned or warped |
+| [`OnBackdoorMsg`](#onbackdoormsg) | `(self, target: str, param: str)` | `pfBackdoorMsg` | `Python.Backdoor` console command |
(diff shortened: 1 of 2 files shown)


Docs/PythonFileMod callback methods.md
```diff
@@ -0,0 +1,687 @@
...
+  resetting age SDL via the console.
+* `tag`: `str` - `plSDLNotificationMsg::fHintString`. The "tag string" provided
+  via `ptSDL.setTagString` by the code that changed the SDL variable, or empty
+  if no tag string was set (which is the usual case).
+
+## `OnOwnershipChanged`
+
+Corresponds to `plNetOwnershipMsg`. Called when the client becomes, or stops
+being, the owner of network-synchronized objects in the current age instance, as
+indicated by `ptSceneobject.isLocallyOwned`.
```
dpogue: All objects in the Age or only the one that the modifier is attached to?

IIRC for objects with physics (kickables), there was some nuance around the last client to touch them becoming the owner

dgelessus:
For a more readable preview than the PR diff view: https://github.com/dgelessus/Plasma/blob/document_PythonFileMod_callbacks/Docs/PythonFileMod%20callback%20methods.md

Docs/PythonFileMod callback methods.md
```diff
@@ -0,0 +1,687 @@
...
+intended only for Cyan's internal CCR clients, which would have received CCR
+requests from players.
+
+### Parameters
+
+* `type`: `plCCRCommunicationMsg::Type` as `int` - `plCCRCommunicationMsg::GetType()`.
+* `message`: `str` - `plCCRCommunicationMsg::GetMessageText()`.
+* `playerID`: `int` - `plCCRCommunicationMsg::GetCCRPlayerID()`.
+
+## `OnServerInitComplete`
```
dpogue, line 413: Might be worth noting that it's not safe/valid to do any SDL logic until after this has happened (i.e., when `isInitialStateLoaded` is true)

Docs/PythonFileMod callback methods.md
```diff
@@ -0,0 +1,687 @@
...
+  avatar in question, or 0 for events not related to an avatar.
+
+## `gotPublicAgeList`
+
+Corresponds to `plNetCommPublicAgeListMsg`. Called when the server replies to a
+`PtGetPublicAgeList` query.
+
+Currently, all Python modifiers receive `gotPublicAgeList` callbacks for all
+`PtGetPublicAgeList` calls - there is no way to direct one call's callback to
+one particular Python modifier.
```
dpogue, line 614: This is only intended to be used by the Nexus, right?

Docs/PythonFileMod callback methods.md
dgelessus: Not entirely sure - the whole ownership and "net group" stuff is quite complex. This code in the network message handling gave me the impression that everything is owned by the same client:

```cxx
    // Simplified object ownership model means that one client owns everything.
    hsLogEntry(nc->DebugMsg("<RCV> plNetMsgGroupOwner, isOwner={}", m->IsOwner()));
    nc->SetObjectOwner(m->IsOwner());

    plNetOwnershipMsg* netOwnMsg = new plNetOwnershipMsg();
    netOwnMsg->Send();
```

But if this other comment in `plNetClientMgr` is accurate, then physicals are indeed owned by the last player that touched them, and only non-physicals use the simplified ownership model:

```cxx
    // simplification of object ownership...one player owns all non-physical objects in the world
    // physical objects are owned by whoever touched them most recently (or the "owner" if nobody
    // has touched it yet)
    bool fIsOwner;
```

Regardless, `OnOwnershipChanged` is only called when the client receives a `plNetMsgGroupOwner` from the server. I don't know in which situations exactly the server sends that (only when the "everything owner" changes, or also when the owner of a single physical changes?), so I can't precisely document when this callback is called! 😓

Docs/PythonFileMod callback methods.md
dgelessus: At the moment, `PtGetPublicAgeList`/`gotPublicAgeList` are only really used by the Nexus. (There's also another use in Personal.py for looking up a public Ae'gura instance, but that might be dead code.) I didn't put it into the "internal KI/GUI stuff" category though, because it *could* reasonably be used by something else too, and it's not tied to internals like some of the other KI/GUI callbacks.

Still worth writing down that it's mainly for the Nexus though, yeah.

Hoikas (commented):
This LGTM, deferring final approval to @dpogue.

dpogue (approved):

PR #1884, opened 2026-06-17
Improve xAgeSDL logic gate scripts.

Hoikas:
An Age creator recently needed a way to perform a logical AND of multiple SDL variables. The built-in `xAgeSDLBoolAndSet` only supports logical ANDs of two variables, so I built an expanded script that accepts comma separated list of SDL variables. While I was here, I've also added another script that performs a logical OR to support an SDL logic gate node in Korman.

The two scripts could probably be merged to reduce the LOC, but I think having a clear separation of concerns might be a good idea, especially for backwards porting to versions of Python that should frankly be in the grave.

### DIFF
--- a/Scripts/Python/xAgeSDLBoolGateSet.py
+++ b/Scripts/Python/xAgeSDLBoolGateSet.py
@@ -60,9 +60,8 @@ stringSDLVariableOutput = ptAttribString(2, "AgeSDL Output Variable")
 stringLogicOp = ptAttribDropDownList(3, "Logic operation", ["AND", "OR", "NAND", "XOR"])
 
 class xAgeSDLBoolGateSetBase:
-    if TYPE_CHECKING:
-        key: ptKey = ...
-        sceneobject: ptSceneobject = ...
+    key: ptKey
+    sceneobject: ptSceneobject
 
     def OnServerInitComplete(self):
         inputVariables, outputVariable, op = self.inputVariables, self.outputVariable, self.logicOp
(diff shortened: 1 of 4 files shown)


dpogue:
an "invert output" option would allow it to work as a NAND operator, which means we could build an 8-bit adder CPU entirely in SDL booleans

(not necessarily encouraging anyone to do that, but, y'know, cool that it would be possible...)

Hoikas:
You can has AND, NAND, OR, and XOR. All in one script now!

dpogue:
oooh, XOR has a lot of interesting potentials for puzzles

dpogue (approved):

Scripts/Python/xAgeSDLBoolGateSet.py
```diff
@@ -0,0 +1,146 @@
...
+from PlasmaTypes import *
+
+stringSDLVariableInput = ptAttribString(1, "AgeSDL Input Variables (comma separated)")
+stringSDLVariableOutput = ptAttribString(2, "AgeSDL Output Variable")
+stringLogicOp = ptAttribDropDownList(3, "Logic operation", ["AND", "OR", "NAND", "XOR"])
+
+class xAgeSDLBoolGateSetBase:
+    if TYPE_CHECKING:
+        key: ptKey = ...
+        sceneobject: ptSceneobject = ...
```
dgelessus: This should work just the same, right?

```suggestion
    key: ptKey
    sceneobject: ptSceneobject
```

PR #1889, opened 2026-06-28
Make glue.py a separate module, don't magically load it into each PythonFileMod module's namespace

dgelessus:
Previously, when a PythonFileMod was loaded, its Python file magically had the special file plasma/glue.py appended to it, and the two shared the same namespace for global variables. This is *really weird* and there seems to be no good reason for it - you can put the glue functions in their own module and pass everything as parameters instead of implicitly via globals. So this PR does that.

This makes plPythonPack much simpler, because Python modifier modules no longer need the special treatment of having glue.py appended to them. plPythonPack also no longer needs to import the modules before compiling them, because it no longer cares whether something is a modifier module or a regular module.

In the future, the loading of the modifier modules in the client and 3ds Max plugin could also be changed to a regular Python import. Not yet though, because then multiple instances of the same Python modifier would share the same globals, and we currently have too many PythonFileMods that store per-instance state in globals.

I renamed glue.py to PlasmaGlue.py to reduce confusion when old and new code are accidentally used together. The glue functions all no longer need the `glue_` name prefix, because `PlasmaGlue` is a true separate module. The other attributes still have the prefix, because they're still set on each modifier module and not on `PlasmaGlue`.

---

Draft, because this is a risky change. I did some quick tests with loose .py files and with a python.pak, and it *seems* to work, but this really needs more gameplay testing to check that I didn't forget any weird cases. Someone also needs to test the Max plugin, which I can't do myself, as usual.


Hoikas:
I know branan had hoped to see the glue rewritten in C++ at some point, but this probably gives us the "best of both worlds" in terms of being largely backwards compatible but still de-duplicating the glue bytecode. We will also need to update any tools that attempt to build python.paks with knowledge of the glue. Korman only does this for closed source version of Uru, so it shouldn't need any changes. UruManifest, OTOH, will need to be taught when the glue doesn't need to be bolted on.

dgelessus:
> I know branan had hoped to see the glue rewritten in C++ at some point

I would agree with that actually, at least for the simple functions like `getVersion` and `isMultiModifier`. It would also be nice to replace the remaining module globals with local/instance variables on the C++ side. Will probably look into that later.

> UruManifest, OTOH, will need to be taught when the glue doesn't need to be bolted on.

Ah, does UruManifest build .pak files itself, without calling Plasma's plPythonPack tool?

This PR already renames the glue file, so I guess UruManifest could check whether the old plasma/glue.py exists or not, and do the magic glue appending only if that file exists.

Hoikas:
> Ah, does UruManifest build .pak files itself, without calling Plasma's plPythonPack tool?
> 
> This PR already renames the glue file, so I guess UruManifest could check whether the old plasma/glue.py exists or not, and do the magic glue appending only if that file exists.

Yes, it does. Now that I'm thinking more carefully about it, I think that UruManfest already works this way - it just considers the absence of the glue.py file to be a warning situation rather than something that's expected. So maybe there isn't as much to be done there as I thought.

Scripts/Python/plasma/PlasmaGlue.py
```diff
@@ -0,0 +1,271 @@
...
+            params[obj.id] = obj
+    elif isinstance(obj, list):
+        for o in obj:
+            findAndAddAttribs(o, params, verbose=verbose)
+    elif isinstance(obj, dict):
+        for o in obj.values():
+            findAndAddAttribs(o, params, verbose=verbose)
+    elif isinstance(obj, tuple):
+        for o in obj:
+            findAndAddAttribs(o, params, verbose=verbose)
```
Hoikas: I would be tempted to merge the list and tuple handling and add some sanity at the end.


```suggestion
    elif isinstance(obj, (list, tuple)):
        for o in obj:
            findAndAddAttribs(o, params, verbose=verbose)
    elif isinstance(obj, dict):
        for o in obj.values():
            findAndAddAttribs(o, params, verbose=verbose)
    else:
        raise TypeError(obj.__class__.__name__)
```

Scripts/Python/plasma/PlasmaGlue.py
```diff
@@ -0,0 +1,271 @@
...
+    pd = getParamDict(module)
+    if pd is not None:
+        try:
+            if isinstance(pd[id], ptAttribNamedActivator):
+                return 1
+            if isinstance(pd[id], ptAttribNamedResponder):
+                return 2
+        except KeyError:
+            if _isVerbose(module):
+                PtDebugPrint(f"Could not find id={id} attribute")
```
Hoikas:
```suggestion
                PtDebugPrint(f"Could not find {id=} attribute")
```

Scripts/Python/plasma/PlasmaGlue.py
```diff
@@ -0,0 +1,271 @@
...
+def isMultiModifier(module):
+    inst = getInstance(module)
+    return isinstance(inst, ptMultiModifier)
+
+def getVisInfo(module, number):
+    pd = getParamDict(module)
+    if pd is not None:
+        # see if there is a paramKey list
+        if isinstance(module.glue_paramKeys, list):
+            if number in range(len(module.glue_paramKeys)):
```
Hoikas: Hmm, checking for presence in an iteratable isn't amazing when this could just be a simple comparison.

```suggestion
            if number < len(module.glue_paramKeys):
```

Probably want to guard against `number` being < 0 at the start of the function as well.

Scripts/Python/plasma/PlasmaGlue.py
```diff
@@ -0,0 +1,271 @@
...
+                return pd[module.glue_paramKeys[number]].getVisInfo()
+            else:
+                PtDebugPrint(f"PlasmaGlue.getVisInfo: Error! {number} out of range of attribute list")
+        else:
+            pl = list(pd.values())
+            if number in range(len(pl)):
```
Hoikas:
```suggestion
            if number < len(pl):
```

Scripts/Python/plasma/PlasmaGlue.py
```diff
@@ -0,0 +1,271 @@
...
+                            pd[id].value = [] # make sure that the value starts as an empty list
+                    except AttributeError:
+                        pd[id].value = [] # or if value hasn't been defined yet, then do it now
+
+                    pd[id].value.append(value) # add in new value to list
+                else:
+                    pd[id].value = value
+        else:
+            if _isVerbose(module):
+                PtDebugPrint("PlasmaGlue.setParam: can't find id=", id)
```
Hoikas:
```suggestion
                PtDebugPrint(f"PlasmaGlue.setParam: can't find {id=}")
```

Hoikas (commented):
A few more suggestions for making PlasmaGlue more Pythonic. I haven't reviewed the C++ side of this yet.

Scripts/Python/plasma/PlasmaGlue.py
dgelessus: Merging the `list` and `tuple` cases makes sense. We can't raise an exception for other types though - this function is called for every attribute of the modifier module being inspected, so it's expected to safely ignore random objects of unknown types.

(I wonder if the `list`/`dict`/`tuple` cases are needed at all... I don't remember ever seeing any modifier that defines its attributes inside a collection.)

Sources/MaxPlugin/MaxMain/plPythonMgr.cpp
```diff
@@ -252,83 +238,69 @@ void IExtractVisInfo(PyObject* tuple, int* id, std::unordered_set<ST::string>& v
...
-            plAutoUIBlock *autoUI = new plAutoUIBlock(PythonFile::GetClassDesc(), blockID, std::move(className), version);
+            plAutoUIBlock *autoUI = new plAutoUIBlock(PythonFile::GetClassDesc(), blockID, fileName, version);
             // test to see if it is a multi-modifier type class
             if (isMulti)
                 autoUI->SetMultiModifierFlag(true);
 
             for (int i = numParams-1; i >= 0; i--)
             {
-                PyObject* ret = plPython::CallObject(getParamFunc, i).Release();
+                // plPython::CallObject steals the argument refs...
```
Hoikas, line 292: I wonder if this really shouldn't be the case. The trivial `plPython::ConvertFrom()` always return a *new* reference representing whatever we passed in. Those are then unreffed somewhere deep inside `plPython::CallObject`'s implementation, and we probably don't want to change that.

It feels wrong with that in mind to return stolen references. Ideally, the user of `ConvertFrom` would dispose of the converted object, instead of being in an unclear ownership case as we are now. The `pyObjectRef&&` being a steal makes sense, but, otherwise, I wonder if we should be returning new references from `plPython::ConvertFrom()` so this rather silly looking workaround isn't needed.

I'd prefer to see any changes in this direction in a separate PR, FWIW. I seem to recall having difficulty with a pyObjectRef const reference overload for ConvertFrom, but I don't remember what that was exactly.

Sources/MaxPlugin/MaxMain/plPythonMgr.cpp
```diff
@@ -252,83 +238,69 @@ void IExtractVisInfo(PyObject* tuple, int* id, std::unordered_set<ST::string>& v
...
+                pyObjectRef modifierModuleExtraRef = modifierModule;
+                PyObject* ret = plPython::CallObject(getParamFunc, std::move(modifierModuleExtraRef), i).Release();
```
Hoikas, line 294: Hmm... Is the return value being leaked here?

Sources/MaxPlugin/MaxMain/plPythonMgr.cpp
Hoikas: Ok, next day, and I've talked myself off the ledge I was on. The ownership model in `plPythonConvert` is all correct, as-is. We just need to add two more overloads to plPythonConvert.h so you don't have to make a temporary copy of the pyObjectRef and move it.

```C++
inline PyObject* ConvertFrom(pyObjectRef& value)
{
    Py_XINCREF(value.Get());
    return value.Get();
}

inline PyObject* ConvertFrom(const pyObjectRef& value)
{
    Py_XINCREF(value.Get());
    return value.Get();
}
```

This basically the same thing I had to do for `ST::string`, so you should be able pass in pyObjectRef l-values:
```cpp
pyObjectRef call = IGetCallable(module);
pyObjectRef foo = IGetFoo(module);
plPython::CallObject(call, foo);
```

Sources/MaxPlugin/MaxMain/plPythonMgr.cpp
dgelessus: There's a `Py_DECREF(ret)` much further down (after all the tuple processing logic) that always gets called if `ret` is non-null. That looks correct to me.

Hoikas: You're right - it's way down there, but it's there.

dgelessus:
> I know branan had hoped to see the glue rewritten in C++ at some point

For reference, the very old issue where that was discussed: #9

PR #1893, opened 2026-07-20
PFMs IDs *must* fit in the signed 32-bit range.

Hoikas:
Imagine my surprise when 3ds Max crashed trying to get xLinkEventTrigger's id. I'm adding a test for this case to make sure no one else runs afoul.


Hoikas:
Test failed - expected

Hoikas:
Test passes after changing ID.

dpogue (approved):
Good to have tests to catch these things!

Issue #1896, opened 2026-07-25
Internal `PtGetAgeSDL` change broke clftImager.py (and possibly other scripts)


dgelessus:
@Filtik and some TOC players recently reported that the cleft imager stopped working. I can reproduce this with upstream H'uru Plasma and moul-assets.

After some debugging, the root cause turned out to be a change from #1866: calling `PtGetAgeSDL` before `OnServerInitComplete` now raises an error instead of returning the not-yet-initialized age SDL object. clftImager.py's `OnNotify` unconditionally calls `PtGetAgeSDL`, which now causes problems, because apparently some of the notifications happen before `OnServerInitComplete`. (The code for those notifications doesn't actually read from the age SDL object, so this worked without issues before.)

This specific case is easy to fix - just remove the `PtGetAgeSDL` call from the code paths that run before `OnServerInitComplete` (implemented in #1897). The problem is that many other scripts also call `PtGetAgeSDL` unconditionally in `OnNotify`, and it's difficult to tell if those have the same issue. I tested a few at random and didn't see any errors, but that's not solid confirmation.

I could change `PtGetAgeSDL` so it can be called before `OnServerInitComplete` again, which would make all scripts work as before. On the other hand, the new error is useful for development/debugging - trying to access age SDL before it's initialized is almost always a programming error - so I'm not convinced that removing the error again is the best solution.

dgelessus:
Correction: `OnServerInitComplete` has nothing to do with it - I got something mixed up here. `PtGetAgeSDL` generally still works before `OnServerInitComplete`. The problem seems to be something else related to loading order.

The new implementation of `PtGetAgeSDL` gets the AgeSDLHook object from `plNetClientMgr`, which *should* receive the AgeSDLHook key very early in the age loading process, before any pages get loaded. Somehow `clftImager.OnNotify` is getting called even earlier than that, so when it calls `PtGetAgeSDL`, it can't find the AgeSDLHook yet and errors out. I'm not sure if this is caused by something specific in the Cleft age, or if other ages might still be affected. I'll need to debug this further.

The good news is that the fix/workaround (?) in #1897 does work, even if I misdiagnosed the cause.

Hoikas:
An additional question I would ask is should `OnNotify()` even be running before the initial state sync happens? It feels like it shouldn't, but I suppose it depends on what kind of notifications are being triggered and why.

dgelessus:
[...]

PR #1897, opened 2026-07-25
Fix broken cleft imager due to calling `PtGetAgeSDL` too early

dgelessus:
See #1896. This only fixes the specific issue for the cleft imager, not the general problem that may affect other scripts too.

The actual fix is the oneliner commit 9dcc4b2ed101b87caf01e8a9f51ee64c1a56b311 in the middle. The commit before it fixes a double log message, and the commits after it clean up clftImager.py's strange pattern of storing the age SDL object as an attribute (which was also repeated in some other scripts).

### DIFF
--- a/Scripts/Python/GardenBugs.py
+++ b/Scripts/Python/GardenBugs.py
@@ -97,9 +97,9 @@ class GardenBugs(ptResponder):
         return 0 # no chronicle var
 
     def OnServerInitComplete(self):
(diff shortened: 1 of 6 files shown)


Hoikas (approved):
This is the story that never ends. Something that shouldn't be done is disallowed, then we suddenly realize someone is actually doing that bad thing. :rofl:

PR #1898, opened 2026-07-26
Fix Metal alpha blends

colincornaby:
This bug was exposed by the Descent preview. It was causing the reflection frame buffer to store incorrect alpha values. The blending operation code is some of the earliest code in the Metal renderer - I don't think I understood the D3D modes when I wrote it. Fixing with the benefit of knowing more.

D3D doesn’t blend the alpha channel at all - it only does RGB blend. Set the alpha blend to preserve the state of the alpha channel.


dpogue (approved):
This seems to line up with what I've got implemented for plGLPipeline for blending modes

PR #1901, opened 2026-08-02
feat: add vulkan support and introduce linux support

pollend:
These changes are a bit of a mess but this adds vulkan support. some of this code was written with codex/claude bits and pieces were referenced from some previous rendering work I did. I need to look over this a bit more before this gets to a state that i would be happy to merge in. This is playable at a minimum would be good to test this on windows to see if this works. 

image

image


pollend:
player shadows working

image

Hoikas:
Impressive results so far! We take pride in our fork of Plasma in that we perform human code review and use that as an opportunity for every active contributor to gain knowledge about what other contributors are doing and the thought process behind their work. This is a task that we enjoy doing - this engine and game has been a passion project for several of us maintainers for 20 years, from before Cyan even released the source code. We're glad you're interested in diving into this process. I've done a very brief review of this PR, and I think there are some things that we, the maintainers, are going to have to do, and some other things that you, the contributor are going to have to do to get this to a place where we'd be comfortable accepting and merging.

I think this PR as it stands isn't something we're going to be able to proceed with. The sheer volume of changes is very large for the human-driven review process I mentioned. I always mentally cringe at large PRs because they're pretty intimidating. And, at 20,000 lines of code changed, this is definitely a large PR. That's not a blocker, exactly, but it does mean that any kind of review would be a superficial "LGTM" rubber-stamp. We do have a history of reviewing and accepting large PRs where the work cannot be broken down into smaller units. The problem with this PR is that it's conflating several different categories of changes into not just one PR but rather one monolithic commit (and two smaller follow on commits). We prefer to see PRs broken down into more small scoped units with atomic commits that focus on one aspect of the change at hand. As best I can tell, there are three main topics in this PR: the vulkan pipeline, the "unix" client, and some miscellaneous cleanups, particularly in the Metal pipeline.

As a first step, I would suggest breaking your changes down into those topics, or even more granular ones. For example, every individual fix you have for the Metal pipeline should probably be a new PR. This way, reviewers have a bite sized chunk that can be looked at and its impact evaluated. @colincornaby, the author of the Metal pipeline, has taken great care in ensuring that the rendering output of the Metal pipeline matches that of the Direct3D9 pipeline that Cyan wrote. It's important for us to identify and specifically note any deficiencies in the Metal pipeline so we can improve our knowledge and better support macOS users when the mac client is actually distributed to end users. Also, #1825, #1860, and #1877 are open involving Metal, and it would be good to ensure the changes don't all conflict with each other.

Further, you should verify if any of the topics you came up with above have PRs that already relate to them. For example, we already have a Linux client PR open in #1325. I know that one of the goals @dpogue and @colincornaby are working toward is identifying where and how code can be shared among the different rendering pipelines and windowing systems. We have generally been staying away from libsdl... Initially because the engine already has a system named "SDL," but we additionally have concerns about it being a heavy dependency and not meeting our needs for all platforms, particularly macOS.

On the maintainers' side, the acceptance of any LLM generated code will require some discussion on our end. I appreciate your disclosure of Claude/Codex usage. I think that from the volume of changes and the comment style in the code, we would have inferred it without your disclosure. That's not a complaint, though. Disclosing it upfront means that we can skip the "was AI used here?" internal discussion. I don't think any of us maintainers are categorically against LLM usage, and I know that more than one of us has experimented with LLMs. That is to say, LLM usage is not disqualifying. We would be most comfortable with LLMs being used in the review process or in the planning phase, however.

As I said before, this is a passion project for us. Many of us honed our software engineering skills with Uru, and many of our contributors learn the engine and stick around to maintain what they contribute. If they don't, the changes are done in small enough batches and with enough back and forth that we gain the experience with that contribution to continue the maintenance ourselves. With LLM generated code, we lose the ability to say, well, here is our expert on that topic and question them about the process and decisions. The expert is the LLM, and the LLM forgets everything when the context is cleared. So we're left holding the bag on something that, while very useful, we don't actually have any knowledge about. And that's a bad situation for us to be in - we have users who expect for us to be able to answer their questions.

There are also real licensing questions around LLM-generated code that we need to work through as maintainers. There have been incidents in the past involving code sourced from the Bink SDK documentation that have put shard operators and Cyan itself in some amount of legal jeopardy, so we have to be careful with these kinds of questions. As a result, that discussion needs to happen among us rather than in this PR, and we'll follow up once it has. Moving forward, it would be best to use LLM generated code lightly until we have a written policy available in [CONTRIBUTING.md](https://github.com/H-uru/Plasma/blob/master/CONTRIBUTING.md).

TL;DR: We're glad you're having fun with the engine - we do too. There are a few organizational items on your end and some process items on our end that need to be figured out :)

PR #1903, opened 2026-08-02
feature: xeGTAO global illuminiation

pollend:
base: https://github.com/H-uru/Plasma/pull/1901

probably need to be careful with a change like this i had to re-order the draw so solid meshes went first then translucency. the GTAO pass get placed right in the middle of these two draw chains. I really like how it improve the look of the scene i think this can be done as a small improvement. 

image
image


Hoikas:
The challenge with this kind of change is that it impacts the way that the content Cyan authored is drawn. Cyan very carefully baked their effects to achieve their desired artistic direction, and, at times, configured the draw order manually. By adding AO, we're basically saying that we think Cyan's art could be better in this way. That's not unreasonable, but, as mentioned in our [assets contributor's guide](https://github.com/H-uru/moul-assets/blob/master/CONTRIBUTING.md), we prefer to take any changes that affect content on a case-by-case basis, especially with regard to subjective changes like this one. I know you like the results, but any kind of change to the base game's content can be polarizing in the community, and we have to walk a very tight rope as a result. That means this should either be an opt-in feature such that its impact can be observed enable/disabled as needed *or* we need to look at every single Age and carefully evaluate the impact.

In my opinion, this kind of feature would be tremendously helpful for artists building new Ages in that it would reduce the need to paint certain shadows by hand with vertex coloring or decals. Everyone actually playing the game right now is using the Direct3D9 renderer, but this is largely tied up with the Vulkan pipeline that you're proposing. And, as best I can tell, the vulkan pipeline will only function on the "unix" client that you've added. Since this depends on #1901, this is blocked behind everything we're working through there. So, this feature is largely out of reach for everyone but you as proposed.

It seems to me that we need to take a step back. One of the things we find that is more important than features and code is design and fit. I'd encourage you to check out the [getting involved](https://github.com/H-uru/Plasma/blob/master/CONTRIBUTING.md#getting-involved) section of the contributor's guide and discuss your design goals before implementing large changes. For example, could this feature be implemented into the engine as it exists on master?

pollend:
> The challenge with this kind of change is that it impacts the way that the content Cyan authored is drawn. Cyan very carefully baked their effects to achieve their desired artistic direction, and, at times, configured the draw order manually. By adding AO, we're basically saying that we think Cyan's art could be better in this way. That's not unreasonable, but, as mentioned in our [assets contributor's guide](https://github.com/H-uru/moul-assets/blob/master/CONTRIBUTING.md), we prefer to take any changes that affect content on a case-by-case basis, especially with regard to subjective changes like this one. I know you like the results, but any kind of change to the base game's content can be polarizing in the community, and we have to walk a very tight rope as a result. That means this should either be an opt-in feature such that its impact can be observed enable/disabled as needed _or_ we need to look at every single Age and carefully evaluate the impact.
> 
> In my opinion, this kind of feature would be tremendously helpful for artists building new Ages in that it would reduce the need to paint certain shadows by hand with vertex coloring or decals. Everyone actually playing the game right now is using the Direct3D9 renderer, but this is largely tied up with the Vulkan pipeline that you're proposing. And, as best I can tell, the vulkan pipeline will only function on the "unix" client that you've added. Since this depends on #1901, this is blocked behind everything we're working through there. So, this feature is largely out of reach for everyone but you as proposed.
> 
> It seems to me that we need to take a step back. One of the things we find that is more important than features and code is design and fit. I'd encourage you to check out the [getting involved](https://github.com/H-uru/Plasma/blob/master/CONTRIBUTING.md#getting-involved) section of the contributor's guide and discuss your design goals before implementing large changes. For example, could this feature be implemented into the engine as it exists on master?

this was more of an experiment to see if its possible not to actually propose this as a change. This breaks in a lot of ways that I haven't shown in these images. 

A lot of the in game assets are low poly and rendering is flat so those crevices will just get darkened with AO as a black crease.  I think the side effect is because the normals at the moment are derived from the depth buffer so the quality is a bit rough, would need to introduce an additional normal buffer i can't recall if the assets have normals. 

I don't like how the draw order has to be re-order just for this feature its a fundamental shift that I think is kind of undesirable and has a lot of side effects that are hard to predict.  the kernel size is kind of large so the effects is fairly expensive I think the intention with the original algorithm was to smooth out the AO using some kind of temporal pass with a camera jitter. It's probably too much of a change for something that only has a marginal improvement. The game already authors a lot of masks to fake the AO so it only helps in a couple specific cases.

PR #1907, opened 2026-08-04
CPU in crashhandler UNKNOWN fix

Filtik:
Fix HAVE_CPUID feature check always failing on MSVC


Hoikas:
Always failing? What is the mode of failure? While you are correct that `__cpuid` expects an `int[4]`, not an `unsigned int[4]`, the code as-is has been in place and worked from our perspective ~~for over a decade~~. I'd like to know what broke.

(Edit: I misread the diff - I thought this was from the CPU dispatcher code, but it's the compile time test, which is newer)

Hoikas:
Ignore the previous comment - I see that HAVE_CPUID is false in the CMakeCache

Hoikas (approved):

PR #1908, opened 2026-08-04
Add `try_compile()` debug information to cmake config output.

Hoikas:
Follow up to #1907. One of the annoyances with this issue is that the only way to diagnose it that you need to actually open the CMakeCache to see what the result of the `try_compile` test is. So, I've wrapped our `try_compile` calls in a macro that prints the result to the config log. The message formats were pretty shamelessly stolen from the `CheckSourceCompiles` builtin module.

On Windows, the output should look something like:
```
Performing Test HAVE_SYSCTL
Performing Test HAVE_SYSCTL - Failed
Performing Test HAVE_BUILTIN_AVAILABLE
Performing Test HAVE_BUILTIN_AVAILABLE - Failed
Performing Test HAVE_CPUID
Performing Test HAVE_CPUID - Success
```


dpogue (approved):

cmake/CompilerChecks.cmake
```diff
@@ -1,3 +1,19 @@
...
+    if(NOT DEFINED "${ARGV0}")
+        if(NOT CMAKE_REQUIRED_QUIET)
+            message(CHECK_START "Performing Test ${ARGV0}")
+        endif()
+        try_compile(${ARGV})
+        if(NOT CMAKE_REQUIRED_QUIET)
+            if("${${ARGV0}}")
+                message(CHECK_PASS "Success")
+            else()
+                message(CHECK_FAIL "Failed")
```
zrax: I wonder if it makes sense to use different words (e.g. "Found" and "Not Found") to avoid confusing messages about "tests" failing

Hoikas: I've changed the verbiage to remove the references to "tests" and success/failure

zrax (approved):

PR #1909, opened 2026-08-04
CI builds for arm64 Windows/Linux

dpogue:
I don't know if we _want_ this because it makes our build matrix even bigger (and slower), but as a proof of concept it's nice to see these builds working.

If we don't merge this as-is, I'll probably make a separate PR to address some of the other updates in here like bumping to Ubuntu 26.04 and using the Qt Install action for macOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,8 @@ jobs:
       matrix:
         platform:
           - { generator: Visual Studio 17 2022, max-version: 7, arch: Win32, qt-arch: win32_msvc2019, qt-version: 5.15.2, runner: "windows-2022", str: windows-x86 }
-          - { generator: Visual Studio 18 2026, max-version: 2022, arch: x64, qt-arch: win64_msvc2022_64, qt-version: 6.9.1, runner: "windows-2025-vs2026", str: windows-x64 }
-          - { generator: Visual Studio 18 2026, max-version: 2022, arch: arm64, qt-arch: win64_msvc2022_arm64, qt-version: 6.9.1, runner: "windows-11-vs2026-arm", str: windows-arm64 }
+          - { generator: Visual Studio 18 2026, max-version: 2022, arch: x64, qt-arch: win64_msvc2022_64, qt-version: 6.11.1, runner: "windows-2025-vs2026", str: windows-x64 }
+          - { generator: Visual Studio 18 2026, max-version: 2022, arch: arm64, qt-arch: win64_msvc2022_arm64, qt-version: 6.11.1, runner: "windows-11-vs2026-arm", str: windows-arm64 }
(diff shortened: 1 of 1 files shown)


Deledrius:
I don't have anything to test this on, but these do seem like platforms of increasing relevance.

.github/workflows/ci.yml
```diff
@@ -181,13 +181,89 @@ jobs:
...
           path: build/install
 
+  # This has to be separate because we don't use a 3DS Max SDK with arm64 support
+  windows_arm:
+    runs-on: ${{ matrix.platform.runner }}
+    name: ${{ matrix.platform.str }}-${{ matrix.cfg.str }}
+    strategy:
+      matrix:
+        platform:
+          - { generator: Visual Studio 18 2026, max-version: 2022, arch: arm64, qt-arch: win64_msvc2022_arm64, qt-version: 6.9.1, runner: "windows-11-vs2026-arm", str: windows-arm64 }
```
Hoikas:
```suggestion
          - { generator: Visual Studio 18 2026, arch: arm64, qt-arch: win64_msvc2022_arm64, qt-version: 6.9.1, runner: "windows-11-vs2026-arm", str: windows-arm64 }
```

I don't think the max version is relevant here?

colincornaby:
ARM Windows is always of interest. It would be nice to support, there is a growing array of devices out there. But it does feel like we're a little early - no games really support it yet. Still an interesting platform though.

I wonder what would be required for an ARM native installer.

.github/workflows/ci.yml
```diff
@@ -182,6 +181,81 @@ jobs:
...
+        uses: jurplel/install-qt-action@v4
+        with:
+          version: ${{ matrix.platform.qt-version }}
+          arch: ${{ matrix.platform.qt-arch }}
+          dir: ${{ github.workspace }}/qt
+
+      - name: Setup CMake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '4.2.7'
```
Hoikas: Since this is basically the latest version of CMake, do we need to have this step at all? Is the GHA supplied CMake insufficient?

Hoikas (commented):
I wonder if, instead of duplicating the Windows workflow, if we could instead add an on/off switch for max in the matrix?

.github/workflows/ci.yml
dpogue: We had issues previously where we needed to force a specific CMake version to work around some issues, and it seems like it's probably a good practice to make sure all the builds use the same version even if we could probably rely on the system-provided one.

At the very least, it's probably worth pulling this version out into a top-level variable so we don't need to repeat it a bunch of times.

Hoikas (approved):

PR #1910, opened 2026-08-04
Unify checksum code.

Hoikas:
In the asset download work I'm proposing in #737, we will want to be able to use SHA-512. Currently, the checksum code is a bit of a mess in that every checksum is reimplemented as a separate class, and most of them are just thin wrappers over OpenSSL EVP. This PR removes the old, unused `plChecksum` and unifies all hashes using a PIMPL pattern. This allows us to de-duplicate code and more trivially add new algorithms. I went ahead and added SHA-256 and SHA-512, for an example. This will help with the asset download overhaul because I can just use `plChecksum` instead of having to do something like `std::variant<plMD5Checksum, plSHA512Checksum>` to support both MD5 and SHA-512.

It may be best to review [plChecksum.cpp itself](https://github.com/Hoikas/Plasma/blob/c1c2254bbcb425d180f61a3d347413f2610a2a0e/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp), rather than trying to read the diff.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -119,14 +119,14 @@ struct pfPatcherQueuedFile
     Type fType;
     plFileName fClientPath;
     plFileName fServerPath;
-    plMD5Checksum fChecksum;
+    plChecksum fChecksum;
     uint32_t fFileSize;
     uint32_t fZipSize;
     uint32_t fFlags;
 
     pfPatcherQueuedFile(Type t, const NetCliFileManifestEntry& file)
-    : fType(t), fClientPath(plFileName(ST::string::from_utf16(file.clientName)).Normalize()),
-          fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(),
+        : fType(t), fClientPath(plFileName(ST::string::from_utf16(file.clientName)).Normalize()),
+          fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(plChecksum::Type::kMD5),
           fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
     {
         ST::string temp(file.md5, std::size(file.md5));
@@ -134,7 +134,8 @@ struct pfPatcherQueuedFile
     }
 
     pfPatcherQueuedFile(Type t, plFileName path, uint32_t flags=0)
-        : fType(t), fClientPath(std::move(path)), fChecksum(), fFileSize(), fZipSize(), fFlags(flags)
+        : fType(t), fClientPath(std::move(path)), fChecksum(plChecksum::Type::kMD5),
+          fFileSize(), fZipSize(), fFlags(flags)
     { }
 
     pfPatcherQueuedFile(const pfPatcherQueuedFile& copy) = delete;
@@ -575,7 +576,7 @@ void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
     }
     plFileInfo mine(clientPathForComparison);
     if (mine.FileSize() == file.fFileSize) {
-        plMD5Checksum cliMD5(clientPathForComparison);
+        plChecksum cliMD5(plChecksum::Type::kMD5, clientPathForComparison);
         if (cliMD5 == file.fChecksum) {
             WhitelistFile(file.fClientPath, false);
             return;
--- a/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
+++ b/Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
@@ -119,22 +119,23 @@ struct pfPatcherQueuedFile
     Type fType;
     plFileName fClientPath;
     plFileName fServerPath;
-    plMD5Checksum fChecksum;
+    plChecksum fChecksum;
     uint32_t fFileSize;
     uint32_t fZipSize;
     uint32_t fFlags;
 
     pfPatcherQueuedFile(Type t, const NetCliFileManifestEntry& file)
     : fType(t), fClientPath(plFileName(ST::string::from_utf16(file.clientName)).Normalize()),
-          fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(),
-          fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
+      fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(plChecksum::Type::kMD5),
+      fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
     {
         ST::string temp(file.md5, std::size(file.md5));
         fChecksum.SetFromHexString(temp.c_str());
     }
 
     pfPatcherQueuedFile(Type t, plFileName path, uint32_t flags=0)
-        : fType(t), fClientPath(std::move(path)), fChecksum(), fFileSize(), fZipSize(), fFlags(flags)
+        : fType(t), fClientPath(std::move(path)), fChecksum(plChecksum::Type::kMD5),
+          fFileSize(), fZipSize(), fFlags(flags)
     { }
 
     pfPatcherQueuedFile(const pfPatcherQueuedFile& copy) = delete;
@@ -575,7 +576,7 @@ void pfPatcherWorker::IHashFile(pfPatcherQueuedFile& file)
     }
     plFileInfo mine(clientPathForComparison);
     if (mine.FileSize() == file.fFileSize) {
-        plMD5Checksum cliMD5(clientPathForComparison);
+        plChecksum cliMD5(plChecksum::Type::kMD5, clientPathForComparison);
         if (cliMD5 == file.fChecksum) {
             WhitelistFile(file.fClientPath, false);
             return;
--- a/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
+++ b/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
@@ -373,9 +373,9 @@ bool plChecksum::operator==(const plChecksum& rhs) const
 
     // In progress and invalid checksums can't be equal. They're in an
     // indeterminant state..
-    if (!IsValid())
+    if (!IsFinished())
         return false;
-    if (!rhs.IsValid())
+    if (!rhs.IsFinished())
         return false;
 
     if (fType != rhs.fType)
--- a/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
+++ b/Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
@@ -307,7 +307,9 @@ void plChecksum::CalcFromStream(hsStream* stream)
 void plChecksum::SetFromHexString(const char* string)
 {
     size_t stringLen = strlen(string);
-    if (fStatus != Status::kReady)
+    if (fStatus == Status::kInvalid)
+        throw plChecksumException("Checksum invalid");
+    if (fStatus == Status::kStarted)
         throw plChecksumException("Checksum in use");
     if (stringLen != 2 * GetSize())
         throw plChecksumException("Invalid string in ISetFromHexString");
@@ -361,8 +363,8 @@ void plChecksum::Finish()
 
 size_t plChecksum::GetSize() const
 {
-    if (fStatus != Status::kFinished)
-        throw plChecksumException("Checksum not finished");
+    if (fStatus == Status::kInvalid)
+        throw plChecksumException("Checksum is invalid");
     return fImpl->GetSize();
 }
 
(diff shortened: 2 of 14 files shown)


Sources/Plasma/FeatureLib/pfPatcher/pfPatcher.cpp
```diff
@@ -119,22 +119,23 @@ struct pfPatcherQueuedFile
...
     uint32_t fFileSize;
     uint32_t fZipSize;
     uint32_t fFlags;
 
     pfPatcherQueuedFile(Type t, const NetCliFileManifestEntry& file)
     : fType(t), fClientPath(plFileName(ST::string::from_utf16(file.clientName)).Normalize()),
-          fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(),
-          fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
+      fServerPath(ST::string::from_utf16(file.downloadName)), fChecksum(plChecksum::Type::kMD5),
+      fFileSize(file.fileSize), fZipSize(file.zipSize), fFlags(file.flags)
```
dgelessus: nit: if we're going to fix the indent here, it should be by indenting the first line, not by dedenting the other two - right?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+void plChecksum::IInit(plChecksum::Type type)
+{
+    switch (type) {
+        case Type::kMD5:
+            fImpl = std::make_unique<plEVPChecksum>(EVP_md5());
+            break;
+        case Type::kSHA0: {
+            // SHA-0 is not supported by OpenSSL these days, so we may have to
+            // use our own implementation.
+            const EVP_MD* md = EVP_get_digestbyname("sha");
```
dgelessus: Is there any advantage to using OpenSSL's SHA-0 implementation anymore? We already have to trust that our own implementation is correct, because that's the only option when building against current OpenSSL. Should we use this opportunity to switch to our implementation unconditionally?

On the other hand, with this new code structure, it takes almost zero effort to keep supporting the OpenSSL code path. So as long as it works reliably (where available), I suppose there's no need to remove it.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+        : fCtx(), fMd(md)
+    {
+        // This could have been passed in as a template paraemter, which could
+        // be used to statically allocate the buffer. That would be more performant,
+        // but it would also mean generating a different class in the executable
+        // for almost every checksum type, which seems a little silly. This is
+        // not a performace-critical path, so we'll just allocate the buffer dynamically.
+        size_t size = EVP_MD_size(md);
+        fValue = std::make_unique<uint8_t[]>(size);
+        memset(fValue.get(), 0, size);
```
dgelessus: Strictly speaking, you don't need to explicitly zero the array - `std::make_unique` already does that by default. (Technically, it's "value-initialization", but for primitives, that's the same as zero-initialization.)

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
-    if (fContext)
-        EVP_MD_CTX_destroy(fContext);
-    fContext = nullptr;
-    memset(fChecksum, 0, sizeof(fChecksum));
-    fValid = false;
-}
+    ~plEVPChecksum()
+    {
+        if (fCtx)
+            EVP_MD_CTX_destroy(fCtx);
```
dgelessus, line 149: Could this be a `std::unique_ptr` with a custom deleter?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+    plSha0 fCtx;
+    ShaDigest fValue;
 
-    size_t out_size = EVP_MD_size(md);
-    hsAssert(out_size == sizeof(fChecksum), "Incorrect output size for MD5");
+public:
+    plSHA0Checksum()
+    {
+        memset(fValue, 0, sizeof(fValue));
+    }
```
dgelessus: Can be simplified like this I think, because `fValue` is an array. Another use case for "value-initialization"!

```suggestion
    plSHA0Checksum() : fValue() {}
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+    IInit(type);
 }
 
-void plMD5Checksum::SetValue(uint8_t* checksum)
+plChecksum::plChecksum(plChecksum::Type type, const plFileName& fileName)
+    : fStatus(Status::kReady), fType(type)
 {
-    fValid = true;
-    memcpy(fChecksum, checksum, sizeof(fChecksum));
+    IInit(type);
```
dgelessus: If you make these three constructors delegate to the basic 1-arg constructor, I think you can inline away `IInit`:

```suggestion
    : plChecksum(type)
{
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+    CalcFromFile(fileName);
 }
 
-bool plMD5Checksum::operator==(const plMD5Checksum& rhs) const
+plChecksum::plChecksum(plChecksum::Type type, hsStream* stream)
+    : fStatus(Status::kReady), fType(type)
 {
-    return (fValid && rhs.fValid && memcmp(fChecksum, rhs.fChecksum, sizeof(fChecksum)) == 0);
+    IInit(type);
```
dgelessus:
```suggestion
    : plChecksum(type)
{
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
+    CalcFromStream(stream);
 }
 
-//============================================================================
-plSHAChecksum::plSHAChecksum(size_t size, const uint8_t* buffer)
-    : fValid(), fOpenSSLContext()
+plChecksum::plChecksum(plChecksum::Type type, size_t size, const uint8_t* buffer)
+    : fStatus(Status::kReady), fType(type)
 {
+    IInit(type);
```
dgelessus:
```suggestion
    : plChecksum(type)
{
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,313 @@ static uint8_t IHexCharToInt(char c)
...
 
-void plSHA1Checksum::SetFromHexString(const char* string)
+bool plChecksum::operator==(const plChecksum& rhs) const
 {
-    const char* ptr;
-    int         i;
+    if (!IsValid())
+        return false;
+    if (!rhs.IsValid())
+        return false;
```
dgelessus: Do we want "NaN behavior" here? Would it be better to compare unfinished `plChecksum`s by identity (i. e. by comparing the `fImpl` pointers)?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
```diff
@@ -43,146 +43,87 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        kSHA512,
+    };
 
-        bool operator==(const plMD5Checksum& rhs) const;
-        bool operator!=(const plMD5Checksum& rhs) const { return !operator==(rhs); }
-};
+private:
+    enum class Status
+    {
+        kInvalid,
```
dgelessus, line 73: The invalid status is only used for moved-from `plChecksum`s and in no other situation, correct?

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
```diff
@@ -43,146 +43,87 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
 
-        bool IsValid() const { return fValid; }
-        void Clear();
+    ~plChecksum();
 
-        void CalcFromFile(const plFileName& fileName);
-        void CalcFromStream(hsStream* stream);
+    size_t GetSize() const;
+    const uint8_t* GetValue() const;
+    bool IsValid() const { return fStatus == Status::kFinished; }
```
dgelessus: This name is a bit misleading - before seeing the implementation, I expected `IsValid` to check whether the status is not `Status::kInvalid`. I think one of the two names should be changed to make it clear that they don't mean "valid" in the same sense.

dgelessus (commented):
Sounds like a good idea to me. Relatively minor comments/suggestions below.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
dgelessus: GitHub's "Suggested change" diff seems to be a broken here - just ignore the red parts...

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
Hoikas: This is exactly what I was thinking. The downside to allowing OpenSSL's implementation to come through is that our SHA-0 implementation isn't tested, so I think there's an argument for just removing this branch. I didn't want to change that behavior without feedback, though.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.h
Hoikas: Correct. It felt like a good touch for debugging.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
Hoikas: It could be. I started making that change, but, as I usually find, `unique_ptr` with custom deleters that wrap C free functions are verbose. While this requires some discipline in the destructor, I think it makes the code cleaner and clearer than an `std::unique_ptr`.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,303 @@ static uint8_t IHexCharToInt(char c)
...
+        case Type::kSHA1:
+            fImpl = std::make_unique<plEVPChecksum>(EVP_sha1());
+            break;
+        case Type::kSHA256:
+            fImpl = std::make_unique<plEVPChecksum>(EVP_sha256());
+            break;
+        case Type::kSHA512:
+            fImpl = std::make_unique<plEVPChecksum>(EVP_sha512());
+            break;
+            DEFAULT_FATAL("checksumType");
```
dgelessus: Indent oops in f750aefc0671844534285dfabdad6a1979ed7ebf

```suggestion
        DEFAULT_FATAL("checksumType");
```

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
```diff
@@ -97,427 +98,303 @@ static uint8_t IHexCharToInt(char c)
...
+//============================================================================
 
-void plSHA1Checksum::SetFromHexString(const char* string)
+bool plChecksum::operator==(const plChecksum& rhs) const
 {
-    const char* ptr;
-    int         i;
+    // Early bail out: if the pointers are equal, it's obviously the same checksum.
+    if (fImpl && rhs.fImpl && fImpl == rhs.fImpl)
+        return true;
```
dgelessus: This still returns false when comparing a moved-from `plChecksum` with itself (I think that's the only case where `fImpl` can be null). The following should be safe and cover such cases (it considers all moved-from checksums equal, but I think that makes sense):

```suggestion
    if (fStatus == rhs.fStatus && fImpl == rhs.fImpl)
        return true;
```

Hoikas: Considering all moved-from checksums to be equal makes me a little uncomfortable. I wonder if it would make more sense to do `if (this == &rhs)`. Checksums can't be copied or share their internal state, so we can just compare the instance itself.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
Hoikas: VS's auto formatting continues to get more and more annoying every day.

Sources/Plasma/NucleusLib/pnEncryption/plChecksum.cpp
dgelessus: That may be the simplest/safest solution. I understand not wanting two checksums to become equal by accident. I just want to avoid `x != x` unless there's a *really* good reason for it.

dgelessus (approved):

dpogue (approved):

PR #1913, opened 2026-08-08
Fix `PtGetAgeSDL` when called during age-global Python initialization

### LINKED ISSUE #1896: Internal `PtGetAgeSDL` change broke clftImager.py (and possibly other scripts)

dgelessus:
Fixes #1896 in a more robust way than the workaround in #1897, in case other age scripts also accidentally need this corner case to work.

This is intentionally branched off the last commit before #1897 was merged, to allow confirming that this fix works even without that workaround.

While here, I also removed some unneeded special cases from the logic for calculating the AgeSDLHook's UOID.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plAgeLoader/plAgeLoader.cpp
+++ b/Sources/Plasma/PubUtilLib/plAgeLoader/plAgeLoader.cpp
@@ -254,7 +254,7 @@ tl::expected<tl::monostate, ST::string> plAgeLoader::LoadAge(const ST::string& a
     //
     // Load the Age's SDL Hook object (and it's python modifier)
     //  
-    plUoid oid=nc->GetAgeSDLObjectUoid(fAgeName);
+    plUoid oid = plNetClientMgr::GetAgeSDLObjectUoidForAge(ad);
     plKey ageSDLObjectKey = hsgResMgr::ResMgr()->FindKey(oid);
     // Loading the AgeSDLHook also loads the VeryVerySpecialPythonFileMod,
     // so the create notification only gets sent after the age Python has finished initializing.
--- a/Sources/Plasma/PubUtilLib/plAgeLoader/plAgeLoader.cpp
+++ b/Sources/Plasma/PubUtilLib/plAgeLoader/plAgeLoader.cpp
@@ -188,7 +188,7 @@ void plAgeLoader::NotifyAgeLoaded( bool loaded )
 
 tl::expected<tl::monostate, ST::string> plAgeLoader::LoadAge(const ST::string& ageName)
 {
-    plNetClientApp* nc = plNetClientApp::GetInstance();
+    plNetClientMgr* nc = plNetClientMgr::GetInstance();
     ASSERT(!nc->GetFlagsBit(plNetClientApp::kPlayingGame));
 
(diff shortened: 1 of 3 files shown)


Sources/Plasma/NucleusLib/pnNetCommon/plNetApp.h
```diff
@@ -207,7 +207,6 @@ class plNetClientApp : public plNetApp
     virtual plNetGroupId GetEffectiveNetGroup(const plSynchedObject* obj) const { hsAssert(false, "stub"); return plNetGroup::kNetGroupUnknown; }
     virtual void Update(double secs) {}
     virtual const char* GetServerLogTimeAsString(ST::string& ts) const { hsAssert(false, "stub"); return nullptr; }
-    virtual plUoid GetAgeSDLObjectUoid(const ST::string& ageName) const { hsAssert(false, "stub"); return plUoid(); }
```
Hoikas, line 210: I know this can be static, but it would be nice if we could avoid changing the vftable for ModDLL purposes 😄. Maybe leave this and rename your static version to `FindAgeSDLObjectUoid`.

dgelessus: If we only care about the vtable layout, is it okay to leave the old method unimplemented in `plNetClientMgr`, i. e. have it default to the stub in the base class?

Hoikas: I'd prefer to see the old method still be implemented so any ModDLL that wants to use it wouldn't have to be adjusted.

dgelessus: Hmm, I'll have to see what to leave as the implementation for the old method. Nothing seems to call it outside of the one place in `plAgeLoader`, so I was hoping to get rid of the existing code with all the different fallback cases that never get used. I guess I could have the old method parse the .age file from disk (which is the final fallback in the existing code) and then delegate to the new method.

Hoikas (approved):

PR #1917, opened 2026-08-09
Add word movement and deletion to several elements

Mystler:
This adds support for using Ctrl+Left and Ctrl+Right to move the cursor by whole words, as well as Ctrl+Backspace to delete the last word and Ctrl+Delete to delete the next word to the following elements:
- EditBox (single line text boxes like the KI chat input line or Avatar naming)
- MultiLineEdit (most notably KI notes)
- The Developer Console (to make it easy to quickly delete parts of a command, in combination with using up/down to browse historical commands)

The changes to each elements were done for ease of review in a style that should match what is already there. This does mean using some old c-style strings in parts that would likely benefit from an update to use ST more widely, but changing that felt out of scope for this.

As an aside, I noticed and fixed an issue in plFont when rendering clipped strings. It would advance the string rendering pointer past the first value if an empty string is passed, thus rendering garbage that may be in the buffer behind the null terminator. In this particular case, the issue would surface when deleting the solely remaining last word on clipped edit boxes like the KI Chat input or the avatar naming field.

Demo:

https://github.com/user-attachments/assets/2ebbd2a1-6928-47a8-a6d6-75d0f2dffaca

### DIFF
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -390,19 +390,6 @@ bool    pfConsole::MsgReceive( plMessage *msg )
 
 //// IHandleKey //////////////////////////////////////////////////////////////
 
-static inline bool IIsWordBoundary(const char c)
-{
-    if (c == ' ')
-        return true;
-    if (c == '_')
-        return true;
-    if (c == '.')
-        return true;
-    if (c == '\0')
-        return true;
-    return false;
-}
-
 void    pfConsole::IHandleKey( plKeyEventMsg *msg )
 {
     char            *c;
@@ -517,44 +504,27 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
     }
     else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_LEFT)
     {
-        // Move one word back
-        if (fWorkingCursor > 0) {
-            do {
-                fWorkingCursor--;
-            } while (fWorkingCursor > 0 && !IIsWordBoundary(fWorkingLine[fWorkingCursor - 1]));
-        }
+        IAdvanceWordFromPos(false, fWorkingCursor);
     }
-    else if( msg->GetKeyCode() == KEY_LEFT )
+    else if (msg->GetKeyCode() == KEY_LEFT)
     {
-        if( fWorkingCursor > 0 )
-            fWorkingCursor--;
+        if (fWorkingCursor > 0)
+            IAdvanceChar(false, fWorkingCursor);
     }
     else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_RIGHT)
     {
-        // Move one word forward
-        size_t end = strlen(fWorkingLine);
-        if (fWorkingCursor < end) {
-            do {
-                fWorkingCursor++;
-            } while (fWorkingCursor < end && !IIsWordBoundary(fWorkingLine[fWorkingCursor - 1]));
-        }
+        IAdvanceWordFromPos(true, fWorkingCursor);
     }
     else if( msg->GetKeyCode() == KEY_RIGHT )
     {
-        if( fWorkingCursor < strlen( fWorkingLine ) )
-            fWorkingCursor++;
+        if (fWorkingCursor < strlen(fWorkingLine))
+            IAdvanceChar(true, fWorkingCursor);
     }
     else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_BACKSPACE)
     {
         // Delete last word
         if (fWorkingCursor > 0) {
-            uint32_t removed = 0;
-            do {
-                fWorkingCursor--;
-                removed++;
-            } while (fWorkingCursor > 0 && !IIsWordBoundary(fWorkingLine[fWorkingCursor - 1]));
-            memmove(&fWorkingLine[fWorkingCursor], &fWorkingLine[fWorkingCursor + removed], (strlen(&fWorkingLine[fWorkingCursor + removed]) + 1) * sizeof(char));
-
+            IDeleteWord(false);
             findAgain = false;
             findCounter = 0;
         } else if (fHelpMode)
@@ -566,15 +536,7 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
     {
         if( fWorkingCursor > 0 )
         {
-            fWorkingCursor--;
-
-            c = &fWorkingLine[ fWorkingCursor ];
-            do
-            {
-                *c = *( c + 1 );
-                c++;
-            } while( *c != 0 );
-
+            IDeleteChar(false);
             findAgain = false;
             findCounter = 0;
         }
@@ -588,12 +550,7 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
         // Delete next word
         size_t end = strlen(fWorkingLine);
         if (fWorkingCursor < end) {
-            uint32_t removed = 0;
-            do {
-                removed++;
-            } while (fWorkingCursor + removed < end && !IIsWordBoundary(fWorkingLine[fWorkingCursor + removed - 1]));
-            memmove(&fWorkingLine[fWorkingCursor], &fWorkingLine[fWorkingCursor + removed], (strlen(&fWorkingLine[fWorkingCursor + removed]) + 1) * sizeof(char));
-
+            IDeleteWord(true);
             findAgain = false;
             findCounter = 0;
             IUpdateTooltip();
@@ -601,13 +558,7 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
     }
     else if( msg->GetKeyCode() == KEY_DELETE )
     {
-        c = &fWorkingLine[ fWorkingCursor ];
-        do
-        {
-            *c = *( c + 1 );
-            c++;
-        } while( *c != 0 );
-
+        IDeleteChar(true);
         findAgain = false;
         findCounter = 0;
         IUpdateTooltip();
@@ -740,6 +691,82 @@ void    pfConsole::IHandleCharacter(const char c)
     for (i += 1; i > fWorkingCursor; --i)
         fWorkingLine[i] = fWorkingLine[i - 1];
     fWorkingLine[fWorkingCursor++] = c;
+}
+
+static inline bool IIsWordBreaker(const char c)
+{
+    if (c == ' ')
+        return true;
+    if (c == '_')
+        return true;
+    if (c == '.')
+        return true;
+    if (c == '\0')
+        return true;
+    return false;
+}
+
+/**
+ * Check the next/previous character and adjust our position output variable accordingly.
+ */
+pfConsole::CharType pfConsole::IAdvanceChar(bool next, uint32_t& pos) const
+{
+    hsAssert(pos >= 0, "out of range");
+    hsAssert(pos <= strlen(fWorkingLine), "out of range");
+    hsAssert(next || pos > 0, "advanced left at start of string");
+
+    char c = next ? fWorkingLine[pos] : fWorkingLine[pos - 1];
+    if (next)
+        pos++;
+    else
+        pos--;
+    if (IIsWordBreaker(c))
+        return CharType::kWordBreaker;
+    return CharType::kNormal;
+}
+
+/**
+ * Advance the supplied position by one word left or right.
+ */
+void pfConsole::IAdvanceWordFromPos(bool next, uint32_t& pos) const
+{
+    // If we're moving left, we want to move into the previous word first.
+    while (!next && pos > 0) {
+        if (IAdvanceChar(next, pos) == CharType::kNormal)
+            break;
+    }
+    // Now, keep moving until we advanced past a word breaker
+    while (next && pos < strlen(fWorkingLine) || !next && pos > 0) {
+        if (IAdvanceChar(next, pos) == CharType::kWordBreaker) {
+            // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
+            if (!next)
+                IAdvanceChar(true, pos);
+            break;
+        }
+    }
+}
+
+void pfConsole::IDeleteChar(bool next)
+{
+    if (!next && fWorkingCursor > 0 || next && fWorkingCursor < strlen(fWorkingLine)) {
+        if (!next)
+            IAdvanceChar(false, fWorkingCursor);
+        memmove(&fWorkingLine[fWorkingCursor], &fWorkingLine[fWorkingCursor + 1], (strlen(&fWorkingLine[fWorkingCursor + 1]) + 1) * sizeof(char));
+    }
+}
+
+void pfConsole::IDeleteWord(bool next)
+{
+    uint32_t oldCursor, newCursor;
+    oldCursor = newCursor = fWorkingCursor;
+    IAdvanceWordFromPos(next, newCursor);
+
+    if (oldCursor != newCursor) {
+        if (!next)
+            fWorkingCursor = newCursor;
+        int32_t count = next ? newCursor - oldCursor : oldCursor - newCursor;
+        memmove(&fWorkingLine[fWorkingCursor], &fWorkingLine[fWorkingCursor + count], (strlen(&fWorkingLine[fWorkingCursor + count]) + 1) * sizeof(char));
+    }
 }
 
 //// IAddLineCallback ////////////////////////////////////////////////////////
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -390,7 +390,7 @@ bool    pfConsole::MsgReceive( plMessage *msg )
 
 //// IHandleKey //////////////////////////////////////////////////////////////
 
-inline bool IIsWordBoundary(const char& c)
+static inline bool IIsWordBoundary(const char c)
 {
     return strchr(" _.", c) != nullptr;
 }
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -730,8 +730,8 @@ pfConsole::CharType pfConsole::IAdvanceChar(bool next, uint32_t& pos) const
  */
 void pfConsole::IAdvanceWordFromPos(bool next, uint32_t& pos) const
 {
-    // If we're moving left, we want to move into the previous word first.
-    while (!next && pos > 0) {
+    // We want to move into the next/previous word first.
+    while (next && pos < strlen(fWorkingLine) || !next && pos > 0) {
         if (IAdvanceChar(next, pos) == CharType::kNormal)
             break;
     }
@@ -741,6 +741,13 @@ void pfConsole::IAdvanceWordFromPos(bool next, uint32_t& pos) const
             // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
             if (!next)
                 IAdvanceChar(true, pos);
+            break;
+        }
+    }
+    // Lastly, if moving right, we also wanna go to the end of the sequence if we have multiple word breakers in a row.
+    while (next && pos < strlen(fWorkingLine)) {
+        if (IAdvanceChar(next, pos) != CharType::kWordBreaker) {
+            IAdvanceChar(false, pos); // One back
             break;
         }
     }
@@ -757,14 +764,14 @@ void pfConsole::IDeleteChar(bool next)
 
 void pfConsole::IDeleteWord(bool next)
 {
-    uint32_t oldCursor, newCursor;
-    oldCursor = newCursor = fWorkingCursor;
-    IAdvanceWordFromPos(next, newCursor);
+    uint32_t cursor, deletePos;
+    cursor = deletePos = fWorkingCursor;
+    IAdvanceWordFromPos(next, deletePos);
 
-    if (oldCursor != newCursor) {
+    if (cursor != deletePos) {
         if (!next)
-            fWorkingCursor = newCursor;
-        int32_t count = next ? newCursor - oldCursor : oldCursor - newCursor;
+            fWorkingCursor = deletePos;
+        int32_t count = next ? deletePos - cursor : cursor - deletePos;
         memmove(&fWorkingLine[fWorkingCursor], &fWorkingLine[fWorkingCursor + count], (strlen(&fWorkingLine[fWorkingCursor + count]) + 1) * sizeof(char));
     }
 }
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -392,7 +392,15 @@ bool    pfConsole::MsgReceive( plMessage *msg )
 
 static inline bool IIsWordBoundary(const char c)
 {
-    return strchr(" _.", c) != nullptr;
+    if (c == ' ')
+        return true;
+    if (c == '_')
+        return true;
+    if (c == '.')
+        return true;
+    if (c == '\0')
+        return true;
+    return false;
 }
 
 void    pfConsole::IHandleKey( plKeyEventMsg *msg )
@@ -511,7 +519,6 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
     {
         // Move one word back
         if (fWorkingCursor > 0) {
-            fWorkingCursor--;
             do {
                 fWorkingCursor--;
             } while (fWorkingCursor > 0 && !IIsWordBoundary(fWorkingLine[fWorkingCursor - 1]));
--- a/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
+++ b/Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
@@ -711,7 +711,6 @@ static inline bool IIsWordBreaker(const char c)
  */
 pfConsole::CharType pfConsole::IAdvanceChar(bool next, uint32_t& pos) const
 {
-    hsAssert(pos >= 0, "out of range");
     hsAssert(pos <= strlen(fWorkingLine), "out of range");
     hsAssert(next || pos > 0, "advanced left at start of string");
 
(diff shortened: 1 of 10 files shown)


Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -390,6 +390,11 @@ bool    pfConsole::MsgReceive( plMessage *msg )
 
 //// IHandleKey //////////////////////////////////////////////////////////////
 
+inline bool IIsWordBoundary(const char& c)
+{
+    return strchr(" _.", c) != nullptr;
```
Hoikas: I would be tempted to just inline this logic

```c++
if (c == ' ')
     return true;
if (c == '_')
    return true;
if (c == '.')
    return true;
if (c == '\0')
    return true;
return false;
```

I left in the check for the NUL-terminator that `strchr` does implicitly.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -390,6 +390,11 @@ bool    pfConsole::MsgReceive( plMessage *msg )
 
 //// IHandleKey //////////////////////////////////////////////////////////////
 
+inline bool IIsWordBoundary(const char& c)
```
Hoikas:
```suggestion
static inline bool IIsWordBoundary(const char c)
```

A reference is all the overhead of a pointer for no benefit here.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -502,16 +507,54 @@ void    pfConsole::IHandleKey( plKeyEventMsg *msg )
             IUpdateTooltip();
         }
     }
+    else if (msg->GetCtrlKeyDown() && msg->GetKeyCode() == KEY_LEFT)
+    {
+        // Move one word back
+        if (fWorkingCursor > 0) {
+            fWorkingCursor--;
+            do {
+                fWorkingCursor--;
```
Hoikas: This is a potential buffer underflow. `fWorkingCursor` could be 1 on line 513, be decremented to 0 on line 514, then be decremented to 0xFFFFFFFF here, which is almost certainly going to be out of bounds on line 517 and crash. Can probably verify this by typing a single character into the console, then pressing Ctrl+Left.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1500,6 +1512,43 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
     }
 }
 
+void pfGUIMultiLineEditCtrl::DeleteNextWord()
+{
+    if (fBuffer[fCursorPos] != L'\0') {
+        int32_t removed = 0;
+        bool forceUpdate = false;
+        do {
+            forceUpdate = forceUpdate || IIsCodeChar(fBuffer[fCursorPos + removed]);
```
Hoikas:
```suggestion
            forceUpdate |= IIsCodeChar(fBuffer[fCursorPos + removed]);
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1500,6 +1512,43 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
...
+            removed++;
+        } while (fBuffer[fCursorPos + removed] != L'\0' && !IIsWordBreaker(fBuffer[fCursorPos + removed - 1]));
```
Hoikas: There is a slight chance that this could trip in a way that makes life really bad. The buffer can contain code characters that have arguments after them. Those are used for things like links and color changes. If those arguments happen to have the same value as a word breaker, then the delete can happen in the middle of a code sequence. The consequences of that are probably not good. I think you'll have to ensure that no code sequence is "active" before breaking. See `IOffsetToNextChar`

There is actually already an `IMoveCursor` with a `kOneWordForward` and `kOneWordBack`. I wonder if it would make sense to use those to find the range of characters to delete and fix that code character scan bug there. I see it's also vulnerable to the same bug.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1500,6 +1512,43 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
...
+void pfGUIMultiLineEditCtrl::DeletePreviousWord()
+{
+    if (fCursorPos > 0) {
+        int32_t removed = 0;
+        bool forceUpdate = false;
+        do {
+            fCursorPos--;
+            removed++;
+            forceUpdate = forceUpdate || IIsCodeChar(fBuffer[fCursorPos]);
+        } while (fCursorPos > 0 && !IIsWordBreaker(fBuffer[fCursorPos - 1]));
```
Hoikas: Same thing here, except we're scanning backward for the code sequence. `IOffsetToNextChar` should still handle it because the code character is repeated at the end of the control sequence.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1500,6 +1512,43 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
...
+
+void pfGUIMultiLineEditCtrl::DeletePreviousWord()
+{
+    if (fCursorPos > 0) {
+        int32_t removed = 0;
+        bool forceUpdate = false;
+        do {
+            fCursorPos--;
+            removed++;
+            forceUpdate = forceUpdate || IIsCodeChar(fBuffer[fCursorPos]);
```
Hoikas:
```suggestion
            forceUpdate |= IIsCodeChar(fBuffer[fCursorPos]);
```

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
```diff
@@ -668,7 +668,7 @@ void    plFont::IRenderString( plMipmap *mip, uint16_t x, uint16_t y, const wcha
                 prevX = fRenderInfo.fX;
                 IRenderLoop( string, 1 );
             }
-            while( fRenderInfo.fX <= fRenderInfo.fClipRect.fX && *++string != 0 );
+            while( fRenderInfo.fX <= fRenderInfo.fClipRect.fX && *string != 0 && *++string != 0 );
```
Hoikas: The while condition is getting a little out of hand. It might be clearer to write:

```suggestion
            while (*string != 0) {
                prevX = fRenderInfo.fX;
                IRenderLoop(string, 1);
                if (fRenderInfo.fX > fRenderInfo.fClipRect.fX)
                    break;
                ++string;
            }
```

GitHub's suggestion view seems a little messed up here. I'm proposing replacing the whole do... while loop.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1500,6 +1512,43 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
...
+    }
+}
+
+void pfGUIMultiLineEditCtrl::DeletePreviousWord()
+{
+    if (fCursorPos > 0) {
+        int32_t removed = 0;
+        bool forceUpdate = false;
+        do {
+            fCursorPos--;
```
Hoikas: There's a lot of book-keeping that needs to get updated when the cursor position changes. I think you should probably not change the cursor position here and instead offset by the number of removed characters, like in `DeleteNextWord`. Then, be sure to call `IMoveCursorTo` to update the `fCursorPos` at the end of the function.

Although, tbh, I think I would prefer to see you use `IMoveCursor` with `kOneWordForeward`/`kOneWordBackward` instead so the deletion behavior is the same as the word moving behavior.

Hoikas (commented):
Really like the feature. I think this will be very helpful, especially combined with the URL support. I would, generally, really like to see the word scrolling code de-duplicated a bit and less nested. There's generally a pattern of
```c++
if (foo) {
     do {
         // work
    } while (foo && bar)
}
```
It would be nice to pull all of that out into something like `pfGUIMultiLineEditCtrl`'s `IMoveCursor` where it uses two for loops to do the word movement:

```c++
case kOneWordBack:
    if( cursor > 0 )
    {
        for( ; cursor > 0 && IIsWordBreaker( fBuffer[ cursor - 1 ] ); cursor-- );
        for( ; cursor > 0 && !IIsWordBreaker( fBuffer[ cursor - 1 ] ); cursor-- );
    }
    break;

case kOneWordForward:
    if (cursor < (int32_t)fBuffer.size() - 1)
    {
        for (; cursor < (int32_t)fBuffer.size() - 1 && !IIsWordBreaker(fBuffer[cursor]); cursor++);
        for (; cursor < (int32_t)fBuffer.size() - 1 && IIsWordBreaker(fBuffer[cursor]); cursor++);
    }
    break;
```

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
Mystler: Wowies, that first decrement line has no purpose in there and must have accidentally slipped through, thanks for catching that.

Mystler:
Updated based on what we discussed. There's definitely still some duplication of code between the different elements happening unless we scope creep into externalizing this kind of stuff, but it should be better within the code for each of them now.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1247,6 +1263,57 @@ void    pfGUIMultiLineEditCtrl::ISetCursor( int32_t newPosition )
...
+            break;
+    }
+    // Now, keep moving until we advanced past a word breaker
+    while (next && pos < (int32_t)fBuffer.size() - 1 || !next && pos > 0) {
+        CharType charType = IAdvanceChar(next, pos);
+        hitCtrlCode |= charType == CharType::kControlCode;
+        if (charType == CharType::kWordBreaker) {
+            // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
+            if (!next)
+              IAdvanceChar(true, pos);
```
Hoikas: Indentation nitpick here

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1247,6 +1263,57 @@ void    pfGUIMultiLineEditCtrl::ISetCursor( int32_t newPosition )
...
+
+/**
+ * Advance the supplied position by one word left or right.
+ */
+bool pfGUIMultiLineEditCtrl::IAdvanceWordFromPos(bool next, int32_t& pos) const
+{
+    bool hitCtrlCode = false;
+
+    // If we're moving left, we want to move into the previous word first.
+    while (!next && pos > 0) {
```
Hoikas: I wonder if we should also do this when moving to the right? If I understand the code correctly, the loop after this one will put the cursor at the end of the next word it sees, and we'll want to consume that on the next Ctrl+Right so we can move to the end of the next word and so on. Or maybe I'm not following correctly?

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1514,39 +1562,23 @@ void    pfGUIMultiLineEditCtrl::DeleteChar()
...
-            fCursorPos--;
-            removed++;
-            forceUpdate |= IIsCodeChar(fBuffer[fCursorPos]);
-        } while (fCursorPos > 0 && !IIsWordBreaker(fBuffer[fCursorPos - 1]));
+    int32_t oldCursor, newCursor;
+    oldCursor = newCursor = fCursorPos;
+    bool forceUpdate = IAdvanceWordFromPos(next, newCursor);
 
+    if (oldCursor != newCursor) {
+        if (!next)
```
Hoikas, line 1586: Out of curiosity, is there a case where `oldCursor != newCursor && next`? It feels like the cursor will only be different if we're deleting the previous word.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
Mystler: It only makes a difference when there's multiple word breakers in sequence, but you're right after giving this another spin, in fact to make this even more intuitive, I do need to add a third loop on top. Give me a sec.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
Mystler: Yes, oldcursor and newcursor are not movements of the actual cursor that is rendered but the cursor pointing to our processed position in the buffer. Maybe i should rename them to avoid the confusion that they're ultimately just representing the range that we delete? That check is equivalent to "do we need to delete anything at all?" which does  nothing at the end/start of the string.

Mystler:
Updated

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -292,28 +371,35 @@ bool    pfGUIEditBoxMod::HandleKeyEvent( pfGameGUIMgr::EventType event, plKeyDef
             {
                 SetCursorToEnd();
             }
+            else if (key == KEY_LEFT && modifiers & pfGameGUIMgr::kCtrlDown) {
+                IAdvanceWordFromPos(false, fCursorPos);
+            }
             else if( key == KEY_LEFT )
             {
                 if( fCursorPos > 0 )
```
Hoikas, line 379:
```suggestion
```

This condition is handled in `IAdvanceChar` now

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.h
```diff
@@ -145,6 +145,16 @@ class pfGUIMultiLineEditCtrl : public pfGUIControlMod
 
         int     fTopMargin,fLeftMargin,fBottomMargin,fRightMargin;
 
+        enum class CharType
+        {
+          kNormal,
+          kWordBreaker,
+          kControlCode,
```
Hoikas: Indentation nitpick here

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.h
```diff
@@ -145,6 +145,16 @@ class pfGUIMultiLineEditCtrl : public pfGUIControlMod
...
+        };
+        
+        CharType IAdvanceChar(bool next, int32_t &pos) const;
+        bool IAdvanceWordFromPos(bool next, int32_t &pos) const;
```
Hoikas:
```suggestion
        CharType IAdvanceChar(bool next, int32_t& pos) const;
        bool IAdvanceWordFromPos(bool next, int32_t& pos) const;
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1262,37 +1352,21 @@ void    pfGUIMultiLineEditCtrl::IMoveCursor( pfGUIMultiLineEditCtrl::Direction d
...
 
         case kOneBack:
-            if( cursor > 0 )
-            {
-                cursor--;
-                while (cursor > 0 && IIsCodeChar(fBuffer[cursor]))
-                    cursor -= IOffsetToNextChar( fBuffer[ cursor ] );
-            }
+            if (cursor > 0)
+                IAdvanceChar(false, cursor);
```
Hoikas: There is a behavior change here. Previously, this would move the cursor back to the previous printable character, skipping over all code sequences. Now, it only skips over one printable character or code sequence. Should probably do something like

```suggestion
            while (cursor > 0 && IAdvanceChar(false, cursor) == CharType::kControlCode);
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1262,37 +1352,21 @@ void    pfGUIMultiLineEditCtrl::IMoveCursor( pfGUIMultiLineEditCtrl::Direction d
...
             break;
 
         case kOneForward:
             if (cursor < (int32_t)fBuffer.size() - 1)
-            {
-                cursor++;
-                while (cursor < (int32_t)fBuffer.size() - 1 && IIsCodeChar(fBuffer[cursor]))
-                    cursor += IOffsetToNextChar( fBuffer[ cursor ] );
-            }
+                IAdvanceChar(true, cursor);
```
Hoikas: This has the same weakness as `kOneBack`. I think we'd want to do something like:
```suggestion
            while (cursor < (int32_t)fBuffer.size() - 1 && IAdvanceChar(true, cursor) == CharType::kControlCode);
```

The suggestion box breaks when I try to select the deleted lines. You can also remove the entire `if` block.

Sources/Plasma/FeatureLib/pfConsole/pfConsole.cpp
```diff
@@ -676,6 +693,89 @@ void    pfConsole::IHandleCharacter(const char c)
...
+        return true;
+    return false;
+}
+
+/**
+ * Check the next/previous character and adjust our position output variable accordingly.
+ */
+pfConsole::CharType pfConsole::IAdvanceChar(bool next, uint32_t& pos) const
+{
+    hsAssert(pos >= 0, "out of range");
```
Hoikas:
```suggestion
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -168,6 +168,85 @@ void    pfGUIEditBoxMod::IUpdate()
...
+        return true;
+    return false;
+}
+
+/**
+ * Check the next/previous character and adjust our position output variable accordingly.
+ */
+pfGUIEditBoxMod::CharType pfGUIEditBoxMod::IAdvanceChar(bool next, uint32_t& pos) const
+{
+    hsAssert(pos >= 0, "out of range");
```
Hoikas:
```suggestion
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -168,6 +168,85 @@ void    pfGUIEditBoxMod::IUpdate()
...
+    return CharType::kNormal;
+}
+
+/**
+ * Advance the supplied position by one word left or right.
+ */
+void pfGUIEditBoxMod::IAdvanceWordFromPos(bool next, uint32_t& pos) const
+{
+    // We want to move into the next/previous word first.
+    while (next && pos < wcslen(fBuffer.c_str()) || !next && pos > 0) {
```
Hoikas: Just in case the compiler doesn't realize actually calling `wcslen` every loop invocation is unnecessary.

```suggestion
    size_t len = wcslen(fBuffer.c_str());
    while (next && pos < len || !next && pos > 0) {
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -168,6 +168,85 @@ void    pfGUIEditBoxMod::IUpdate()
...
+        if (IAdvanceChar(next, pos) == CharType::kNormal)
+            break;
+    }
+    // Now, keep moving until we advanced past a word breaker
+    while (next && pos < wcslen(fBuffer.c_str()) || !next && pos > 0) {
```
Hoikas:
```suggestion
    while (next && pos < len || !next && pos > 0) {
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
```diff
@@ -168,6 +168,85 @@ void    pfGUIEditBoxMod::IUpdate()
...
+        if (IAdvanceChar(next, pos) == CharType::kWordBreaker) {
+            // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
+            if (!next)
+                IAdvanceChar(true, pos);
+            break;
+        }
+    }
+    // Lastly, if moving right, we also wanna go to the end of the sequence if we have multiple word breakers in a row.
+    while (next && pos < wcslen(fBuffer.c_str())) {
```
Hoikas:
```suggestion
    while (next && pos < len) {
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1233,6 +1263,66 @@ void    pfGUIMultiLineEditCtrl::ISetCursor( int32_t newPosition )
...
+            // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
+            if (!next)
+                IAdvanceChar(true, pos);
+            break;
+        }
+    }
+    // Lastly, if moving right, we also wanna go to the end of the sequence if we have multiple word breakers in a row.
+    while (next && pos < (int32_t)fBuffer.size() - 1) {
+        CharType charType = IAdvanceChar(next, pos);
+        hitCtrlCode |= charType == CharType::kControlCode;
```
Hoikas: Hmmm... But on the next line, we regress to before the control character like we never hit it at all, so `hitCtrlCode` is not entirely accurate.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1233,6 +1263,66 @@ void    pfGUIMultiLineEditCtrl::ISetCursor( int32_t newPosition )
...
+        if (charType == CharType::kNormal)
+            break;
+    }
+    // Now, keep moving until we advanced past a word breaker
+    while (next && pos < (int32_t)fBuffer.size() - 1 || !next && pos > 0) {
+        CharType charType = IAdvanceChar(next, pos);
+        hitCtrlCode |= charType == CharType::kControlCode;
+        if (charType == CharType::kWordBreaker) {
+            // If we're moving left, we now want to stop before a word breaker, so go one back when we moved past one.
```
Hoikas, line 1309: Hmmm... But on the next lines, we could advance past the control character like we never hit it at all, so `hitCtrlCode` may not be accurate.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIMultiLineEditCtrl.cpp
```diff
@@ -1233,6 +1263,66 @@ void    pfGUIMultiLineEditCtrl::ISetCursor( int32_t newPosition )
...
+        CharType charType = IAdvanceChar(next, pos);
+        hitCtrlCode |= charType == CharType::kControlCode;
+        if (charType != CharType::kWordBreaker) {
+            IAdvanceChar(false, pos); // One back
+            break;
+        }
+    }
+
+    return hitCtrlCode;
+}
```
Hoikas, line 1328: Something about the logic here doesn't feel right, but I'm still trying to put my finger on it. I'll re-read this later.

Sources/Plasma/PubUtilLib/plGImage/plFont.cpp
```diff
@@ -662,13 +662,14 @@ void    plFont::IRenderString( plMipmap *mip, uint16_t x, uint16_t y, const wcha
...
             fRenderInfo.fRenderFunc = &plFont::IRenderCharNull;
-            int16_t prevX;
-            do
-            {
+            int16_t prevX = fRenderInfo.fX;
+            while (*string != 0) {
+                IRenderLoop(string, 1);
+                if (fRenderInfo.fX > fRenderInfo.fClipRect.fX)
+                    break;
                 prevX = fRenderInfo.fX;
```
Hoikas: There's a very slight behavior change in when the `prevX` value is set. This all feeds into how text is rendered, so I'd prefer to retain the old behavior to prevent any surprises later.

```suggestion
                prevX = fRenderInfo.fX;
                IRenderLoop(string, 1);
                if (fRenderInfo.fX > fRenderInfo.fClipRect.fX)
                    break;
```

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
Mystler: Huh?

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
Mystler: No, it's not. I can move it there if you prefer but now it's needed here and asking to advance will advance. I didn't wanna duplicate nested checks for this when advancing words.

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
Hoikas: The `pos` here is an unsigned integer, so it can't be less than zero 😉

Sources/Plasma/FeatureLib/pfGameGUIMgr/pfGUIEditBoxMod.cpp
Hoikas: Oh, whoops. I was probably looking at `IAdvanceWordFromPos`.

Mystler:
Thanks! Updated again!

Mystler:
Updated this with some basic additional unit tests.

PR #1921, opened 2026-08-13
Migrate FileSrv/AuthSrv manifest parsing to `ST::string`

dgelessus:
More of the usual, working on getting rid of all the manual C string manipulation.

I cleaned up the manifest parsing logic a little bit, but didn't want to change too much, to keep the diff readable. I also added proper bounds checking everywhere.

Because I want to use my new hsEndian.h functions for converting UTF-16-LE to `ST::string`, I changed where the byte order conversion happens for the FileSrv manifests. Instead of converting the entire manifest char by char to native byte order before parsing, like before, each field of the manifest is now converted individually as it's read. (The AuthSrv manifest parsing had no byte order handling at all - that's fixed now too.)

I wrote unit tests for parsing both kinds of manifests, based on the original code, to try to ensure that the parsing behavior doesn't change. (These should also be useful when working towards a new manifest format as discussed in #737.) I did some quick local testing of the AuthSrv manifests, but not the FileSrv ones, because my local test server doesn't have that set up. So I'm marking this as draft until I (or someone else) confirms that the patcher still works.

(One commit here has a trivial conflict with #1910 - I'll resolve that once that PR is merged.)

### DIFF
--- a/Sources/Plasma/PubUtilLib/plNetGameLib/plNglAuth.cpp
+++ b/Sources/Plasma/PubUtilLib/plNetGameLib/plNglAuth.cpp
@@ -2877,12 +2877,12 @@ bool IReceiveFileList(const Auth2Cli_FileListReply& reply, std::vector<NetCliAut
 
             // read in the filename
             size_t filenameConsumed;
-            ST::string filename = hsSTStringFromTerminatedUTF16LE(curChar, kNetDefaultStringSize - 1, filenameConsumed);
+            ST::string filename = hsSTStringFromTerminatedUTF16LE(curChar, wcharCount * sizeof(char16_t), filenameConsumed);
 
             unsigned filenameLen = filenameConsumed / sizeof(char16_t);
             curChar += filenameLen; // advance the pointer
             wcharCount -= filenameLen; // keep track of the amount remaining
-            if ((*curChar != L'\0') || (wcharCount <= 0))
+            if (wcharCount == 0 || *curChar != L'\0')
                 return false; // something is screwy, abort and disconnect
 
             curChar++; // point it at the first part of the size value (format: 0xHHHHLLLL)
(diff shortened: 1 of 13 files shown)


Sources/Plasma/PubUtilLib/plNetGameLib/plNglAuth.cpp
```diff
@@ -2856,69 +2856,71 @@ void FileListRequestTrans::Post () {
...
+    } else {
         // fileData format: "filename\0size\0filename\0size\0...\0\0"
-        bool done = false;
-        while (!done) {
-            if (wcharCount == 0)
-            {
-                done = true;
-                break;
+        while (wcharCount > 0) {
+            if (*curChar == L'\0') {
```
dpogue: Should this be `0` instead of `L'\0'`? It's _probably_ fine because it's all 0, but I'm concerned about `wchar` being 2-bytes on Windows and 4-bytes on non-Windows and potential widening of `*curChar`

dgelessus: Oh I think you're right, I didn't notice that despite staring at it for so long :D I guess all of the `L'\0'` literals weren't adjusted when the type was changed from `wchar_t` (from Cyan's original code) to `char16_t`. Technically `u'\0'` would be a `char16_t` literal I think, but just `0` is simpler probably.

Sources/Plasma/PubUtilLib/plNetGameLib/plNglFile.cpp
```diff
@@ -886,82 +868,58 @@ void ManifestRequestTrans::Post () {
...
-    if (!(*length)) {
-        size_t maxlen = FIXME_u16snlen(curMsgPtr, kNetDefaultStringSize - 1);   // Hacky sack
-        (*length) = maxlen;
-        destPtr[maxlen] = 0;    // Don't do this on fixed length, because there's no room for it
-    }
-    memcpy(destPtr, curMsgPtr, *length * sizeof(char16_t));
+ST::string ReadStringFromMsg(const char16_t* curMsgPtr, size_t msgChar16Count, size_t* length)
+{
+    size_t consumedSize;
+    ST::string res = hsSTStringFromTerminatedUTF16LE(curMsgPtr, msgChar16Count * sizeof(char16_t), consumedSize);
```
dpogue, line 875: Not suggesting this get changed, but just thought I'd mention that this looked weird at first glance because normally I'd expect an out variable to be passed in as a pointer (i.e., `&consumedSize`).

PR #1922, opened 2026-08-13
Move some wiki pages to Markdown files under Docs

dgelessus:
As proposed in #1916, this converts the "[CMake Flags](https://github.com/H-uru/Plasma/wiki/CMake-Flags)" and "[Preferred Code Style](https://github.com/H-uru/Plasma/wiki/Preferred-Code-Style)" wiki pages to Markdown files in the Docs folder, because the information in these pages is useful to have in the repo itself.

To preserve attribution, I set the commit author to the person who wrote each wiki page, with myself only as the committer. Is that okay? Or should I use `Co-authored-by` instead to avoid confusion?

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -7,6 +7,9 @@ bcaade13ddc0c9728e87f625fe755d4a2d0cc10b
 c7724ed256aead5371ed5f47ec285cef8b6f8e68
 7b1debfabdebea3bf8bab395ed894c740900b36d
 
+# Line length rewrapping
+85692737847d1c6614215d22ee0c967f7a14aebe
+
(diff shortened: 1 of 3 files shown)


Hoikas (commented):
> To preserve attribution, I set the commit author to the person who wrote each wiki page, with myself only as the committer. Is that okay?

That seems fair to me. I would be tempted to either add 85692737847d1c6614215d22ee0c967f7a14aebe to `.git-blame-ignore-revs` *or* folding the 80 column reformat into the commits themselves and adding yourself as a co-author to avoid your 80 column commit being blamed for every line of those files, though.

Hoikas (approved):

dpogue (approved):

PR #1923, opened 2026-08-14
Clean up plShader a bit, decouple from plShaderTable

dpogue:
Rather than the plShader grabbing its declaraction data from the table, it can instead just store its shader ID, and the pipeline (or whatever wants the declaration data) can worry about fetching it from the table.

This moves us towards it being possible for each pipeline to have its own shader table with different declarations for each of the shader IDs. I haven't figured out exactly what that will look like yet, so I'm opening this as the groundwork for that eventual next step.

**Note:** I changed the stream format of plShader because it didn't read or write the flags and had some other obsolete values. I am 99% sure plShaders have never been streamed and are all created locally on the client, so this _should_ have no effect, but to be safe I did bump the minor version for that class (which should also have no effect).

### DIFF
--- a/Sources/Plasma/PubUtilLib/plResMgr/plVersion.h
+++ b/Sources/Plasma/PubUtilLib/plResMgr/plVersion.h
@@ -50,7 +50,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 // If you change the minor, see GetChangedCreateables in plVersion.cpp
 
 #define PLASMA2_MAJOR_VERSION   70  // Major Version...every file will need to be reexported
-#define PLASMA2_MINOR_VERSION   2   // Minor Version...only files with the specified creatables
+#define PLASMA2_MINOR_VERSION   3   // Minor Version...only files with the specified creatables
                                     // will need to be reexported
 
 // Don't modify this, it's automatically updated when branches are made
@@ -288,6 +288,7 @@ public:
     0   02/12/07 bob    Reset for major version change
     1   03/29/07 jeff   Changed plLoadAvatarMsg and plArmatureMod to be more flexible
     2   06/28/07 jeff   Changed plAvBrainHuman format to store whether it's an actor or not
+    3   08/13/26 dpogue Changed plShader format (which I'm pretty sure was never used anywhere)
 */
 
 #endif // plVersion_h_inc
--- a/Sources/Plasma/PubUtilLib/plSurface/plShader.h
+++ b/Sources/Plasma/PubUtilLib/plSurface/plShader.h
@@ -45,7 +45,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 
 #include <vector>
 
-#include "plShaderTable.h"
+#include "plShaderID.h"
 
 #include "hsGeometry3.h"
 #include "hsMatrix44.h"
@@ -174,7 +174,7 @@ public:
     plPipeConst(Type t, uint16_t r) : fType(t), fReg(r) { }
 
     Type        fType;
-    uint16_t              fReg;
+    uint16_t    fReg;
 };
 
 typedef plPipeConst::Type plPipeConstType;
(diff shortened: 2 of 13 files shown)


Sources/Plasma/PubUtilLib/plResMgr/plVersion.h
```diff
@@ -50,7 +50,7 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
 // If you change the minor, see GetChangedCreateables in plVersion.cpp
 
 #define PLASMA2_MAJOR_VERSION   70  // Major Version...every file will need to be reexported
-#define PLASMA2_MINOR_VERSION   2   // Minor Version...only files with the specified creatables
```
dgelessus, line 53: This seems to be the first time that the minor version is being increased since Cyan's original open-source release. Are we sure that this won't cause problems anywhere?

If I understand correctly, it's not an issue at the PRP level, because MOUL PRPs track the minor version per class (based on the info in `GetChangedCreatables`) and so only PRPs containing the specific changed class would be affected (i. e. almost certainly none).

But does anything else rely on the minor version constant? I know that libHSPlasma uses the full Plasma version numbers for a lot of conditional logic. Would that be affected in any way if this constant is changed?

Sources/Plasma/PubUtilLib/plSurface/plShaderID.h
```diff
@@ -0,0 +1,98 @@
...
+
+// When adding to the compiled table, make sure
+// you add the include in plShaderTable.cpp, or you'll
+// compile fine but have a nil shader (FFP) at runtime.
+namespace plShaderID
+{
+    enum ID
+    {
+        Unregistered = 0,
+            vs_WaveFixedFin6,   //OBSOLETE
```
dgelessus: Indent slipped?

colincornaby:
I will take a look. I'll need to refresh myself on how the D3D9 shaders are managed by these types in since that was almost entirely abstracted away for Metal.

Sources/Plasma/PubUtilLib/plSurface/plShader.h
```diff
@@ -245,32 +242,24 @@ class plShader : public hsKeyedObject
...
 
     const void*             GetConstBasePtr() const { return !fConsts.empty() ? &fConsts[0] : nullptr; }
 
     void                    CopyConsts(const plShader* src) { fConsts = src->fConsts; }
 
-    void                    SetInputFormat(uint8_t format) { fInput = format; }
-    void                    SetOutputFormat(uint8_t format) { fOutput = format; }
-
-    uint8_t                   GetInputFormat() const { return fInput; }
-    uint8_t                   GetOutputFormat() const { return fOutput; }
```
colincornaby, line 272: What is/was these getters/setters for?

dpogue: They were seemingly unused. It looks like the input format was usually set to the number of UV channels, but neither of these were ever passed to any DirectX functions

colincornaby:
I'm not going to approve yet because I have similar questions around the creatable and version change. But I don't know enough about that system to understand the why's. Other than that I would be fine with these changes.

The only wrinkle I see is that this system - so far - would be inadequate for a Metal (or Vulkan/D3D12) pipeline. Those pipelines would have additional variables like the vertex format. And there would be more complications for managing the shader states that emulate the fixed pipeline.

I'm confident that can be worked out. But it may be more complicated than just having these shader IDs. Maybe each pipeline having it's own kind of table would help. But I'd like to consolidate as much as possible.

dpogue:
yeah, the part that seems most problematic to me is that the plPipeConst variables (which effectively map to uniforms) have an int fReg, which is meant to be the shader register that holds that value. That seems like it's not going to translate well between pipeline types because how GL handles uniform locations probably doesn't align with how DX handles shader registers and how Metal handles passing data to shaders.

That said, so far this doesn't actually change anything about how shaders work, it just moves where the plShaderTable lookup happens for the DirectX compiled shader code.

colincornaby:
You could use them in GL by hardcoding the uniform ids to match. But I think explicit uniform ids are only in GLSL 3.3 and higher?

Metal actually just sends over the uniform buffer as one giant struct - straight from the value in memory. So it doesn't use the uniform ids, but also doesn't need to interact with that system.

colincornaby:
Was thinking about this PR today - one other realization I had is that the separate vertex/fragment shader IDs work in Direct3D 9 but they probably won't work in GL or Metal. Both GL and Metal require a linked single program which would need to be stored someplace. I think GL4 has a function that emulates the way D3D9 does things. But we're targeting GL2.

Not a blocker but we'll need to keep iterating on the design.

PR #1201, opened 2022-06-27
[Very very unfinished] OpenGL rendering pipeline

dpogue:
This allows running with the OpenGL pipeline on Windows by running with the `/Renderer=gl` command-line flag.

There is a lot of stuff broken, or missing, or causing crashes! This is definitely not ready for actual use, but I do want to start a conversation about how to keep this more closely in line with the work on master so I can avoid future rebasing headaches after major refactorings.

Note: All the X11/Linux client stuff has been stripped out here, this is just the OpenGL rendering pipeline.

PR #1924, opened 2026-08-14
Foundation for OpenGL Pipeline

dpogue:
This is an initial first pass at getting the OpenGL pipeline to the point of rendering solid geometry without textures. It has been lightly tested with both the Windows client and the Linux client, this is by no means intended to be playable.

The goal is to support OpenGL 2.1 and newer. There's built-in fallback behaviour, as well as a console command that can be used in graphics.ini to force using a particular OpenGL version.

The goal here is to get an initial basic implementation merged and to start deduplicating pipeline code across DirectX, Metal, and OpenGL.

Some of the GL code here will need to change as more features are added (particularly the handling of dynamic vertex buffers, and the hardcoded temporary shaders). But this should be solid enough (and simple enough) that it could be reviewed and merged as is.

This supersedes the long-standing and far-too-large #1201 as a starting point for the GL pipeline.


dpogue:
Note, this does _not_ hook up OpenGL rendering on the macOS client, but I'm hoping that merging this will make it easier to test and work out with Colin the best way to set that up.

colincornaby:
> Note, this does _not_ hook up OpenGL rendering on the macOS client, but I'm hoping that merging this will make it easier to test and work out with Colin the best way to set that up.

I think this would be on my immediate list of things to do.

Sources/Plasma/FeatureLib/pfConsole/pfConsoleCommands.cpp
```diff
@@ -1017,6 +1017,22 @@ PF_CONSOLE_CMD( Graphics, ForceGeForce2Quality, "", "Forces higher-level hardwar
...
+
+    extern int plGLVersionOverride;
+    plGLVersionOverride = version;
+
+    if (version)
+        PrintString(ST::format("Forced OpenGL to version: {.2}", 0.1f * version));
+    else
+        PrintString("Using default system OpenGL version");
+}
+#endif // PLASMA_PIPELINE_GL
```
colincornaby, line 1035: Few questions here:

- OpenGL 1/2 and 3/4 are completely not compatible - at least in the core profiles. Are we planning on forking the renderer for both generations? (Guessing with the PowerPC goal - the answer might be yes.)
- Because of ^ there's no concept of a system default GL version. Just like there wouldn't be a default version of Direct3D on Windows.
- Do we need OpenGL 2 and OpenGL 4 render constants for renderer selection similar to what happened with Metal? I'm pretty sure that got hooks - but I can't remember if that was a argument or a CMake setting.

colincornaby: Ok, went off and did some checking. There is kArgRenderer as a startup arg. I'd maybe consider using that in since there is an existing pattern in the code base.

I might also consider adopting a similar set of specific GL versions that we support, and we need to pick a default in since the system won't pick one. But that's (sigh) complicated. OpenGL 4.1 is as high as the Mac goes. Some early Intel Macs only do GL3. Only the late model PowerPC Macs support GL2.

The Myst 5 port to Mac supports GL1. But GLSL support didn't come until GL2.

At this point in development we might consider picking either GL2 or GL4 as our initial target and putting that in the renderer list. And we can expand as we go.

dpogue: This mostly refers to the EGL backend: By "system default", I mean it should be grabbing a context with the Core Profile for the highest supported OpenGL version. On Linux for me, that is 4.6. There is a fallback to grab a 2.1 context if a newer one cannot be retrieved.

On macOS, we ask for a Core profile on OS X 10.7 and higher, otherwise we get whatever the system gives us (which on 10.5 should be 2.1).

I don't especially want to deal with OpenGL 1 and all the pushing and popping of matrices. That sounds so much worse to deal with than just binding shader uniforms.

Right now, the code for GL 2 and 3+ is mostly the same. There are some conditionals when creating buffers so that we only use VAOs on 3+, and if we're on 4.5+ we can use the DSA APIs for avoid binding until we need to. My plan with the shaders is to dynamically set the `#version` line based on the GL context version, and then use `__VERSION__` ifdefs within the shaders to do whatever conditionalizing we need. It might not be pretty in a few spots, but the bulk of the code should be the same across all versions.

colincornaby: I want to get this PR through so I don't want to get too wound up in making huge changes - but I'm wondering if the GL renderer should template on a GL version (eventually). That would compile multiple versions of the renderer and reduce any GL version checks to compile time checks. The renderer switch that Metal is using should support that sort of thing. Metal doesn't do anything that elaborate though. It just checks the device type. Metal doesn't do that many checks so it's not a huge performance concern. But if GL is doing a lot of forking of behavior that would eliminate the runtime cost and even allow multiple specialized implementations of functions.

Your plan seems good. By avoiding the GL2 built ins you should be able to get code that is compatible with GL3 with both the shaders and functions.

I see the Epoxy GL include. Does that roll GL3/4 and GL2 together into the same header, or are they separate headers?

I think on macOS the default is _still_ 2.1. Which is why I'm a little leery about the idea of a default.

dpogue: I should mention that the reason for supporting GL 2.1 here is primarily because of what [dgelessus found on lower-end Linux laptops](https://github.com/dpogue/Plasma/pull/20) and not the expectation that this is immediately going to work on OS X 10.5 or anything like that.

If we set this up so that it works on GL 2.1 on Linux (for which we know a working test environment exists), then making it work on Leopard will probably be easier down the line.

Hoikas: Yeah, I feel like this should be folded into the argument to the executable for now.

Sources/Plasma/FeatureLib/pfDisplayHelpers/plWinDisplayHelper.cpp
```diff
@@ -83,5 +85,11 @@ std::vector<plDisplayMode> plWinDisplayHelper::GetSupportedDisplayModes(
...
 hsDisplayHndl plWinDisplayHelper::DefaultDisplay() const
 {
-    return GetActiveWindow();
+    if (fCurrentDisplay == HDC_INVALID_HANDLE_VALUE) {
+        HWND hWnd = GetActiveWindow();
+        HDC hDC = GetDC(hWnd);
+        SetCurrentScreen(hDC);
+    }
+
+    return fCurrentDisplay;
```
Hoikas, line 94: I'm guessing it's ok to simply return an invalid handle?

Sources/Plasma/PubUtilLib/plPipeline/pl3DPipeline.h
```diff
@@ -851,10 +919,58 @@ class pl3DPipeline : public plPipeline
...
+     */
+    bool ISoftwareVertexBlend(plDrawableSpans* drawable, const std::vector<int16_t>& visList);
+
+    /**
+     * Given a pointer into a buffer of verts that have blending data in the
+     * plVertCoder format, blends them into the destination buffer given
+     * without the blending info.
+     */
+    void IBlendVertsIntoBuffer(const plSpan* span, hsMatrix44* matrixPalette, int numMatrices, const uint8_t* src, uint8_t format, uint32_t srcStride, uint8_t* dest, uint32_t destStride, uint32_t count, uint16_t localUVWChans) {
+        fBlendVertBuffer.call(span, matrixPalette, numMatrices, src, format, srcStride, dest, destStride, count, localUVWChans);
```
Hoikas, line 968: I was actually intending to, in the not too distant future, pull the software skinning code out of plDXPipeline and into a completely different helper class. This way, it could be both tested and benchmarked independently of any client/renderer code. One of the problems that I have with the software implementation is that improvements are hard to quantify - `Stats.Show PipeT` is more of a vibe check than anything in this regard.

Hoikas (commented):
I still need to read the GL meat and potatoes, but it would be nice if we could slim this PR down even further by pulling things like ef58e2ac8fb176860f39f4820e026a4f3b8bc11d out into separate PRs. Like I said, that's something I already intended to do in a slightly different way - adding tests and benchmarks isn't a requirement, IMO, but having it outside of the pipeline for testing purposes might be nice, if you're interested. Or maybe that's a good use for the null pipeline.

Sources/Plasma/FeatureLib/pfGLPipeline/plGLDevice.cpp
```diff
@@ -40,29 +40,751 @@ You can contact Cyan Worlds, Inc. by email legal@cyan.com
...
+        CGLReleaseContext(ctx);
+    }
+
+    if (pix)
+        CGLReleasePixelFormat(pix);
+
+    IGNORE_WARNINGS_END
+}
+#endif // HS_BUILD_FOR_MACOS
+#pragma endregion CGL_Init
```
colincornaby, line 295: I'm going through this code after the discussions this week.

The different render loops are a problem but they probably are surmountable. Each platform might have a different render loop - and we'll have to leave it at that. Even SDL is transitioning to a callback based render system instead of a loop based one. Render loops are the topic of this specific code - but are related to the distribution of responsibilities.

What I'm stuck on here is that the above code is all a series of client specific code. but in the GL device instead of the client. Understanding why would be helpful. I can understand a hesitation to put GL specific code into something like the Windows client. We haven't had Direct3D code in the Windows client itself either. But the Windows/Direct3D pairing never had to deal with this sort of complexity - so they got to make a few decisions that benefit from that simplicity.

I'm still thinking through options. The obvious thing here for Mac is this code isn't compatible with the layer system - or the GL callback on macOS.

One option I'm thinking is that the client window abstraction could have an interface for requesting a service. The pipeline requests a surface and the client does whatever work to make sure that surface is present. Could be as simple as supplying an HWND for Direct3D - or for GL the client could do the work of setting up GL.

The other tweak I'd like to see here - either here or as a following on - is more robust capability selection. Like Metal - we should be able to switch between GL2 and GL4 for testing. That matches up with feedback elsewhere in the PR. There also needs to be some sort of record keeping on what GL profile was actually selected so we know which GL functions to call.

Sources/Plasma/Apps/plClient/Mac-Cocoa/main.mm
```diff
@@ -42,9 +42,6 @@
 
 // System Frameworks
 #import <Cocoa/Cocoa.h>
-#ifdef PLASMA_PIPELINE_GL
-#import <OpenGL/gl.h>
-#endif
```
colincornaby, line 47: With everything discussed - this might end up being too early of a change. I don't know if GL can be removed from the Mac client.

PR #1925, opened 2026-08-15
CMake/Metal support fixes

dpogue:
Bring in some fixes from https://github.com/dpogue/CMake-MetalShaderSupport/ for detecting the Metal compiler (because Apple insists on making this unreasonably difficult)

Also make sure the pfMetalPipelineShadersMTL30 library is set up properly so it builds as a dependency when it should, and doesn't try to get bundled when it shouldn't.

Is there some better option for detecting Metal 3 support? Maybe based on the Metal compiler version?


colincornaby:
> Is there some better option for detecting Metal 3 support? Maybe based on the Metal compiler version?

There is likely a toolchain version check. But I don't know if the toolchain versions are listed anywhere by language support. The toolchain is also decoupled from Xcode these days. But I think the added check is fine.

colincornaby (approved):

PR #1926, opened 2026-08-15
Fix Rainy Cleft page-ins

Mystler:
These were forgotten to be removed from the page in list and are part of GlobalAnimations. Having two invalid pages in the list caused the entire page in process to fail so we would have a broken partially-loaded rainy Cleft version.

(Edited as a second look made me realize this was before some more animations were moved over. Idk for how long this was broken and why it's working on live. Maybe some page-in code regressed?)


Hoikas (approved):
Good catch. There have recently been some changes around asking for invalid pages and producing proper errors/failures in that case.

PR #1928, opened 2026-08-17
Add simple /timestamps toggle for the chat

Mystler:
I tend to dislike chats without timestamps in MMOs. Especially in those that hide chat due to overlays or fade outs so when you see a message later, you wonder for how long you've missed it. In URU, this can happen a lot as well with how various circumstances hide the chat area, like using the Relto book or bookshelf. Simply being tabbed out for a moment is of course too a factor.

This adds a new /timestamps command that has two optional params
- Displayed time zone: Supported options are `ki`, `utc`, and `local`. `ki` is default.
- Displayed format: Allows a custom format for strftime. Default is `%H:%M`

image

### DIFF
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1670,7 +1670,6 @@ class xKI(ptModifier):
 
     ## Sets the KI Text Color from the Chronicle.
     def DetermineShowTimestamps(self):
-        self.chatMgr.timestamps = None
         if entry := ptVault().findChronicleEntry(kChron.ShowTimestamps):
             if value := entry.getValue():
                 self.chatMgr.timestamps = xKIChat.ChatTimestampSettings.from_string(value)
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1670,12 +1670,22 @@ class xKI(ptModifier):
 
     ## Sets the KI Text Color from the Chronicle.
     def DetermineShowTimestamps(self):
+        self.chatMgr.timestamps = None
         vault = ptVault()
         entry = vault.findChronicleEntry(kChron.ShowTimestamps)
-        if entry is not None and entry.getValue() == "yes":
-            self.chatMgr.timestamps = True
-            return
-        self.chatMgr.timestamps = False
+        if entry is not None:
+            if value := entry.getValue():
+                parts = value.strip().split(";", 1)
+                timeZone = "ki"
+                if parts[0]:
+                    timeZone = parts[0]
+                timeFormat = "%H:%M"
+                if len(parts) > 1 and parts[1]:
+                    timeFormat = parts[1]
+                self.chatMgr.timestamps = {
+                    "zone": timeZone,
+                    "format": timeFormat
+                }
 
 
     #~~~~~~~~~~#
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1337,6 +1337,7 @@ class xKI(ptModifier):
         self.DetermineKIFlags()
         self.DetermineGZ()
         self.DetermineTextColor()
+        self.DetermineShowTimestamps()
 
         # Hide all dialogs first.
         KIMicroBlackbar.dialog.hide()
@@ -1662,6 +1663,19 @@ class xKI(ptModifier):
         self.chatMgr.chatTextColor = None
         PtDebugPrint("xKI.DetermineTextColor(): KI Text Color is not overridden.", level=kWarningLevel)
 
+
+    #~~~~~~~~~~~~~~~~~~~~~~#
+    #  KI Chat Timestamps  #
+    #~~~~~~~~~~~~~~~~~~~~~~#
+
+    ## Sets the KI Text Color from the Chronicle.
+    def DetermineShowTimestamps(self):
+        vault = ptVault()
+        entry = vault.findChronicleEntry(kChron.ShowTimestamps)
+        if entry is not None and entry.getValue() == "yes":
+            self.chatMgr.timestamps = True
+            return
+        self.chatMgr.timestamps = False
 
 
     #~~~~~~~~~~#
--- a/Scripts/Python/ki/__init__.py
+++ b/Scripts/Python/ki/__init__.py
@@ -1671,21 +1671,9 @@ class xKI(ptModifier):
     ## Sets the KI Text Color from the Chronicle.
     def DetermineShowTimestamps(self):
         self.chatMgr.timestamps = None
-        vault = ptVault()
-        entry = vault.findChronicleEntry(kChron.ShowTimestamps)
-        if entry is not None:
+        if entry := ptVault().findChronicleEntry(kChron.ShowTimestamps):
             if value := entry.getValue():
-                parts = value.strip().split(";", 1)
-                timeZone = "ki"
-                if parts[0]:
-                    timeZone = parts[0]
-                timeFormat = "%H:%M"
-                if len(parts) > 1 and parts[1]:
-                    timeFormat = parts[1]
-                self.chatMgr.timestamps = {
-                    "zone": timeZone,
-                    "format": timeFormat
-                }
+                self.chatMgr.timestamps = xKIChat.ChatTimestampSettings.from_string(value)
 
 
     #~~~~~~~~~~#
--- a/Scripts/Python/ki/xKIChat.py
+++ b/Scripts/Python/ki/xKIChat.py
@@ -44,6 +44,8 @@ import re
 import time
 import random
 import inspect
+from dataclasses import dataclass
+from typing import Literal
 
 # Plasma Engine.
 from Plasma import *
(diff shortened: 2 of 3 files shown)


DamnBriggsy:
I would find this very helpful

Mystler:
After some discussion elsewhere, I'm currently working on expanding this slightly into allowing (optionally) selecting between KI-Time, UTC, and Local time as well as specify a custom time format.

Mystler:
Updated

image

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
                 self.chatMgr.AddChatLine(None, PtGetLocalizedString("KI.Errors.MalformedChatSetTextColorCmd"), kChat.SystemMessage)
                 return
 
+    def ToggleChatTimestamps(self, arg1 = "", arg2 = "%H:%M"):
+        # Docs on invalid params
+        if arg1 and arg1 not in ["ki", "utc", "local"]:
```
Hoikas:
```suggestion
        if arg1 and arg1 not in {"ki", "utc", "local"}:
```

Prefer in set membership test over in list test:

```
>>> timeit.timeit("'local' in {'ki', 'utc', 'local'}")
0.011551900010090321
>>> timeit.timeit("'local' in ['ki', 'utc', 'local']")
0.01831549999769777
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+            self.chatMgr.AddChatLine(None, "Usage: /timestamps <ki|utc|local> <format>. The first parameter defaults to 'ki' and the second to '%H:%M'. Use without parameters to toggle timestamps off again.", kChat.SystemMessage)
+            return
+
+        # Toggle off if currently on and no params given
+        if self.chatMgr.timestamps is not None and not arg1:
+            self.chatMgr.timestamps = None
+            PtDebugPrint(f"xKIChat.ToggleChatTimestamps(): Setting KI Timestamps chronicle to empty", level=kWarningLevel)
+            ptVault().addChronicleEntry(kChron.ShowTimestamps, kChron.ShowTimestampsType, "")
+            self.chatMgr.AddChatLine(None, f"Chat timestamps disabled", 0)
```
Hoikas:
```suggestion
            self.chatMgr.AddChatLine(None, "Chat timestamps disabled", 0)
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+            return
+
+        timeZone = arg1 if arg1 else "ki"
```
Hoikas:
```suggestion
        timeZone = arg1 or "ki"
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+        self.chatMgr.timestamps = { "zone": timeZone, "format": arg2 }
+        chronVal = f"{timeZone};{arg2}";
+        PtDebugPrint(f"xKIChat.ToggleChatTimestamps(): Setting KI Timestamps chronicle to: \"{chronVal}\".", level=kWarningLevel)
+        ptVault().addChronicleEntry(kChron.ShowTimestamps, kChron.ShowTimestampsType, chronVal)
+        self.chatMgr.AddChatLine(None, f"Chat timestamps enabled", 0)
```
Hoikas:
```suggestion
        self.chatMgr.AddChatLine(None, "Chat timestamps enabled", 0)
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
                 self.chatMgr.AddChatLine(None, PtGetLocalizedString("KI.Errors.MalformedChatSetTextColorCmd"), kChat.SystemMessage)
                 return
 
+    def ToggleChatTimestamps(self, arg1 = "", arg2 = "%H:%M"):
```
Hoikas: It would be nice if the function's arguments were more descriptive than `arg1` and `arg2`.

```suggestion
    def ToggleChatTimestamps(self, zone="", format="%H:%M"):
```

Also, PEP8 says to not use spaces for default arguments if no type hint is given.

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+        # Toggle off if currently on and no params given
+        if self.chatMgr.timestamps is not None and not arg1:
+            self.chatMgr.timestamps = None
+            PtDebugPrint(f"xKIChat.ToggleChatTimestamps(): Setting KI Timestamps chronicle to empty", level=kWarningLevel)
+            ptVault().addChronicleEntry(kChron.ShowTimestamps, kChron.ShowTimestampsType, "")
+            self.chatMgr.AddChatLine(None, f"Chat timestamps disabled", 0)
+            return
+
+        timeZone = arg1 if arg1 else "ki"
+        self.chatMgr.timestamps = { "zone": timeZone, "format": arg2 }
```
Hoikas: I would be tempted to use a [`dataclass`](https://docs.python.org/3.8/library/dataclasses.html) to have a more strongly typed settings object instead of a dict. You could move the load and serialize code to the dataclass. Just as an example, that could look something like:

```python
@dataclass
class ChatTimestampSettings:
    enabled: bool = False
    zone: Literal["ki", "local", "utc"] = "ki"
    format: str = "%H:%M"

    @classmethod
    def from_string(cls, value: str):
        if not value:
            return cls()
        return cls(True, *value.split(';', 1))

    def __str__(self) -> str:
        if not self.enabled:
            return ""
        return f"{self.zone};{self.format}"
```

Using an enabled bool might be nicer than comparing against `None` to determine if timestamps should be shown.

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+        chronVal = f"{timeZone};{arg2}";
```
Hoikas:
```suggestion
        chronVal = f"{timeZone};{arg2}"
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -1334,6 +1347,27 @@ def SetTextColor(self, arg1, arg2=None, arg3=None):
...
+    def ToggleChatTimestamps(self, arg1 = "", arg2 = "%H:%M"):
+        # Docs on invalid params
+        if arg1 and arg1 not in ["ki", "utc", "local"]:
+            self.chatMgr.AddChatLine(None, "Usage: /timestamps <ki|utc|local> <format>. The first parameter defaults to 'ki' and the second to '%H:%M'. Use without parameters to toggle timestamps off again.", kChat.SystemMessage)
+            return
+
+        # Toggle off if currently on and no params given
+        if self.chatMgr.timestamps is not None and not arg1:
+            self.chatMgr.timestamps = None
+            PtDebugPrint(f"xKIChat.ToggleChatTimestamps(): Setting KI Timestamps chronicle to empty", level=kWarningLevel)
```
Hoikas:
```suggestion
            PtDebugPrint("xKIChat.ToggleChatTimestamps(): Setting KI Timestamps chronicle to empty", level=kWarningLevel)
```

Scripts/Python/ki/xKIChat.py
```diff
@@ -622,13 +623,25 @@ def AddChatLine(self, player, message, cFlags, forceKI=True):
...
 
+        timestamp = ""
+        if self.timestamps is not None:
+            timeZone = self.timestamps.get("zone");
+            if timeZone == "local":
+                timeStruct = time.localtime(PtGetServerTime())
+            elif timeZone == "utc":
+                timeStruct = time.gmtime(PtGetServerTime())
+            else:
+                timeStruct = time.gmtime(PtGetDniTime())
```
Hoikas: I would prefer to see us validate that we're getting the expected value. So,

```suggestion
            elif timeZone == "ki":
                timeStruct = time.gmtime(PtGetDniTime())
            else:
                raise ValueError(timeZone)
```

Scripts/Python/ki/__init__.py
```diff
@@ -1663,6 +1664,29 @@ def DetermineTextColor(self):
...
+    #~~~~~~~~~~~~~~~~~~~~~~#
+    #  KI Chat Timestamps  #
+    #~~~~~~~~~~~~~~~~~~~~~~#
+
+    ## Sets the KI Text Color from the Chronicle.
+    def DetermineShowTimestamps(self):
+        self.chatMgr.timestamps = None
+        vault = ptVault()
+        entry = vault.findChronicleEntry(kChron.ShowTimestamps)
+        if entry is not None:
```
Hoikas:
```suggestion
        if entry := vault.findChronicleEntry(kChron.ShowTimestamps):
```

Scripts/Python/ki/xKIChat.py
Mystler: My thinking here was that this is the code that translates it into what to do. So if the chronicle is unexpected garbage it will still work on fallbacks rather than crash the python. The validation for normal users is happening in the slash command.

Mystler:
Updated. Dataclass sure helped make this a lot more appealing to look at, thanks!

Hoikas (approved):

PR #1931, opened 2026-08-21
Cleaning up the WaveDecEnv Metal shader

colincornaby:
Cleanup of WaveDecEnv. Has not gone through alignment with the other shaders yet.

### DIFF
--- a/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveDecEnv.metal
+++ b/Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveDecEnv.metal
@@ -87,11 +87,11 @@ typedef struct
 typedef struct
 {
     float4 position [[position]];
-    float4 c1;
+    float4 environmentTint;
     float4 texCoord0;
-    float4 texCoord1;
-    float4 texCoord2;
-    float4 texCoord3;
+    float4 basisRowX;
+    float4 basisRowY;
+    float4 basisRowZ;
     float fog;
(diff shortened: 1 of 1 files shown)


Sources/Plasma/FeatureLib/pfMetalPipeline/ShaderSrc/WaveDecEnv.metal
```diff
@@ -328,39 +327,38 @@ vertex vs_WaveDecEnv7InOut vs_WaveDecEnv_7(Vertex in                        [[ s
...
 
-    r10.x = rsqrt(dot(r3.xyz, r3.xyz));
-    out.texCoord2 = r3 * r10.xxxw;
+    out.basisRowZ.xyz = normalize(basisRowY.xyz);
+    out.basisRowZ.w = basisRowY.w * uniforms.NumericConsts.z;
 
-    r10.x = rsqrt(dot(r2.xyz, r2.xyz));
-    out.texCoord3 = r2 * r10.xxxw;
+    out.basisRowY.xyz = normalize(basisRowZ.xyz);
+    out.basisRowY.w = basisRowZ.w * uniforms.NumericConsts.z;
```
colincornaby, line 358: Swap here is intentional. See comment above. This was a bug in the older version.

I've verified the swap is necessary in Metal - but Metal also loads the environment maps in a way that's different - specifically for this orientation. It's possible if I loaded up the environment maps to the proper locations for manual - this swap would not be necessary.

Something to watch for a GLSL port of this shader code. I'm continue to dig into it.

PR #1933, opened 2026-08-21
Pull software vertex blending into its own helper

dpogue:
This includes a fix (https://github.com/H-uru/Plasma/pull/947 ported from DX to the macOS skinning implementation) for the avatar UV map not getting updated after morphs were changed in the closet.

### DIFF
--- a/Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.cpp
+++ b/Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.cpp
@@ -1,0 +1,368 @@
+/*==LICENSE==*
+
+CyanWorlds.com Engine - MMOG client, server and tools
+Copyright (C) 2011  Cyan Worlds, Inc.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Additional permissions under GNU GPL version 3 section 7
+
+If you modify this Program, or any covered work, by linking or
+combining it with any of RAD Game Tools Bink SDK, Autodesk 3ds Max SDK,
+NVIDIA PhysX SDK, Microsoft DirectX SDK, OpenSSL library, Independent
+JPEG Group JPEG library, Microsoft Windows Media SDK, or Apple QuickTime SDK
+(or a modified version of those libraries),
+containing parts covered by the terms of the Bink SDK EULA, 3ds Max EULA,
+PhysX SDK EULA, DirectX SDK EULA, OpenSSL and SSLeay licenses, IJG
+JPEG Library README, Windows Media SDK EULA, or QuickTime SDK EULA, the
+licensors of this Program grant you additional
+permission to convey the resulting work. Corresponding Source for a
+non-source form of such a combination shall include the source code for
+the parts of OpenSSL and IJG JPEG Library used as well as that of the covered
+work.
+
+You can contact Cyan Worlds, Inc. by email legal@cyan.com
+ or by snail mail at:
+      Cyan Worlds, Inc.
+      14617 N Newport Hwy
+      Mead, WA   99021
+
+*==LICENSE==*/
+
+#include "plSoftwareMeshSkinner.h"
+
+#include "hsMatrix44.h"
+#include "plDrawable/plGBufferGroup.h"
+#include "plDrawable/plSpanTypes.h"
+
+#if !defined(__has_include)
+#   define __has_include(x) 0
+#endif
+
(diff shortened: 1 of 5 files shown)


Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.cpp
```diff
@@ -0,0 +1,368 @@
...
+            simd_float4 destNorm_buf = { 0.f, 0.f, 0.f, 0.f };
+            simd_float4 destPt_buf = { 0.f, 0.f, 0.f, 1.f };
+
+            // Blend
+            for (uint32_t j = 0; j < numWeights + 1; ++j) {
+                float weight = weights[j];
+                if (weight) {
+                    const simd_float4x4& simdMatrix = hsMatrix2SIMD(matrixPalette[indices & 0xFF]);
+
+                    // Note: This uses Accelerate.framework so this is also accelerated on ARM through NEON or maybe even the Neural Engine.
```
colincornaby, line 180: We can probably cut the note on the Neural Engine here. I don't think Apple has ever pushed this sort of matrix math off of NEON and onto the Neural Engine. It doesn't seem like it's what the Neural Engine is meant for.

Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.h
```diff
@@ -0,0 +1,64 @@
...
+class hsMatrix44;
+
+namespace plSoftwareMeshSkinner
+{
+    /**
+     * Given a pointer into a buffer of verts that have blending data in the
+     * plVertCoder format, blends them into the destination buffer given
+     * without the blending info.
+     */
+    void BlendVertBuffer(const plSpan* span,
```
Hoikas, line 57: Can we also expose the individual implementations, perhaps in a `_detail` namespace so they can be individually tested and benched?

Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.cpp
```diff
@@ -0,0 +1,368 @@
...
+        &IBlendVertBuffer<ISkinVertexSSE3>
+    };
+
+    void plSoftwareMeshSkinner::BlendVertBuffer(
+            const plSpan* span, hsMatrix44* matrixPalette, int numMatrices,
+            const uint8_t* src, uint8_t format, uint32_t srcStride,
+            uint8_t* dest, uint32_t destStride, uint32_t count,
+            uint16_t localUVWChans)
+    {
+        blend_vert_buffer.call(span, matrixPalette, numMatrices, src, format, srcStride, dest, destStride, count, localUVWChans);
```
Hoikas, line 365: We probably want this in the header to improve the chances of this being inlined in case someone turns off LTCG.

Sources/Plasma/PubUtilLib/plPipeline/plSoftwareMeshSkinner.cpp
```diff
@@ -0,0 +1,368 @@
...
+        __m128 mc1 = _mm_load_ps(xfm.fMap[1]);
+        __m128 mc2 = _mm_load_ps(xfm.fMap[2]);
+        __m128 mwt = _mm_set_ps1(wgt);
+
+        SPLIT_SKIN_BUFFER(srcBuf, pt_src, vec_src);
+        SPLIT_SKIN_BUFFER(dstBuf, pt_dst, vec_dst);
+
+        ISkinDpSSE3(pt_src, pt_dst, mc0, mc1, mc2, mwt);
+        ISkinDpSSE3(vec_src, vec_dst, mc0, mc1, mc2, mwt);
+#   endif // HAVE_SSE3
```
Hoikas, line 288: This needs to be in a separate file with the SSE3 source set on just that file. Otherwise, gcc and clang will be free to emit SSE3 instructions in the FPU code path, potentially causing an invalid instruction crash on x86 CPUs without SSE3 support.

... Does anyone even play with a CPU that doesn't support SSE3 these days? I feel like Windows won't even boot without SSE4.

colincornaby: SSE3 looks like it was late revision Pentium 4s?

Hoikas: In the past, we had a user blow up on an invalid instruction with the SSE3 skinning code. I think that user might have been on AMD hardware, but it's been a *very* long time since then. Still, the point is, we don't need to have an FPU codepath that could potentially have SSE instructions in it.

dgelessus: Windows 10 still works on (32-bit) CPUs with only SSE2. The last time we discussed requiring some kind of vector instructions (#1795), we considered requiring some more recent instruction sets for x86_64 builds, but keeping the 32-bit x86 builds compatible with SSE2 (so that players on ancient x86_64 CPUs can use a 32-bit build as a fallback).

colincornaby:
For my project today I'm looking at adding NEON optimizations - at least for Windows. I can commit it after this PR to minimize churn. Just an FYI I'm looking at stacking work on top of this.

It'll also probably be an open question if we maintain the Accelerate path after that. But let's keep it for now.

colincornaby:
I do have NEON intrinsics ready to go. But I'll wait until this PR is in and then I'll open another one.

PR #1941, opened 2026-08-27
Add confirmation dialog to xTakableClothing

Mystler:
I remember this existing before, probably on TPotS based clients, but I am missing this in the modern game.

Ask if we want to wear the item or only unlock it for the wardrobe.
This includes:
- A fix for a bug that probably came with the conversion to Python 3 where `wornnamelist = [worn[0] for i in worn]` caused a repeating array of the same item instead of the intended `wornnamelist = [i[0] for i in worn]`.
- Special handling for guild shirts. If we are currently wearing a guild shirt and select no on another one, we will enforce a swap, so players cannot cheat the intended single guild mechanic.

Comes with a localization file PR to moul-assets at https://github.com/H-uru/moul-assets/pull/295.

### DIFF
--- a/Scripts/Python/xTakableClothing.py
+++ b/Scripts/Python/xTakableClothing.py
@@ -362,11 +362,9 @@ class xTakableClothing(ptModifier):
         wornnamelist = [i[0] for i in worn]
         changeGuildShirt = False
         if base.find('Torso_GuildBlue') != -1 or base.find('Torso_GuildGreen') != -1 or base.find('Torso_GuildRed') != -1 or base.find('Torso_GuildYellow') != -1 or base.find('Torso_GuildWhite') !=  -1:
-            for guildShirt in guildSDLValues.keys():
-                if guildShirt in wornnamelist:
-                    PtDebugPrint(f"xTakableClothing: Found worn guild short, enforcing swap", level=kWarningLevel)
-                    changeGuildShirt = True
-                    break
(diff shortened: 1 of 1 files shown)


Scripts/Python/xTakableClothing.py
```diff
@@ -334,63 +334,78 @@ def OnNotify(self,state,id,events):
...
-                    avatar.avatar.tintClothingItem(matchingItem[0], color1, 0)
+        worn = avatar.avatar.getAvatarClothingList()
+        wornnamelist = [i[0] for i in worn]
+        changeGuildShirt = False
+        if base.find('Torso_GuildBlue') != -1 or base.find('Torso_GuildGreen') != -1 or base.find('Torso_GuildRed') != -1 or base.find('Torso_GuildYellow') != -1 or base.find('Torso_GuildWhite') !=  -1:
+            for guildShirt in guildSDLValues.keys():
+                if guildShirt in wornnamelist:
+                    PtDebugPrint(f"xTakableClothing: Found worn guild short, enforcing swap", level=kWarningLevel)
+                    changeGuildShirt = True
+                    break
```
Hoikas:
```suggestion
            changeGuildShirt = any((guildShirt in wornnamelist for guildShirt in guildSDLValues))
            if changeGuildShirt:
                    PtDebugPrint("xTakableClothing: Found worn guild short, enforcing swap", level=kWarningLevel)
```

This looks like new code (correct?), so we could make it a little more Pythonic.

Hoikas (approved):

PR #1942, opened 2026-09-03
Set Ahnonay fog values by sphere

Mystler:
A simple approach that sets the fog values by sphere based on the TPotS values.
This effectively only sets them when linking into spheres 1, 2, and 4. Sphere 3 uses the same values as the default fni already.

Related to https://github.com/H-uru/moul-assets/issues/115, see below for comparison screenshots.

### DIFF
--- a/Scripts/Python/Ahnonay.py
+++ b/Scripts/Python/Ahnonay.py
@@ -211,6 +211,7 @@ class Ahnonay(ptResponder):
             PtPageInNode("MaintRoom04")
             PtPageInNode("EngineerHut")
             PtPageInNode("ahnySphere04")
+        self.SetFogToSphere(newSphere)
 
     ###########################
     def OnPageLoad(self,what,who):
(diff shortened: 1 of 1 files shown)


dpogue:
This is definitely an improvement over the current behaviour, but it's been a while since I played through Ahnonay... is it possible to end up in a situation where you go from Sphere 1 to Sphere 2 or to the Vortex, or is access always gated through Sphere 4?

Mystler:
It's gated through that. If you expand the diff above my fog function call you'll see it paging in the spheres conditionally. The Vortex is Sphere 4 only.

Mystler:
Updated after noticing the colors were interpreted wrong thanks to @EhrenCG. Including Sphere 4 as well now.

Posting comparison here now.

Sphere 1:
image
image

Sphere 2:
image
image

Sphere 4:
image
image

Hoikas (approved):

Deledrius (approved):

PR #1943, opened 2026-09-05
Change log date format to ISO with years

Mystler:
Is this necessary for anything? No.
Is this free of opinions? No.
Does it bother me to have weird dates in logfiles? Yes. 😜

Shout if this would actually break anyone's workflow of parsing these, but I have my doubts anyone would parse them when they don't even include a year. If anything i expect this change to make potential parsing a lot easier for everyone.


dpogue:
This is definitely an improvement for all the non-Americans who are tired of wondering which is the month and which is the day with that ambiguous slash syntax.

Hoikas (approved):

PR #1945, opened 2026-09-10
Fix ChatFlags masks discarding flags above bit 15

Stone-:
kRTChatFlagMask is 65535 and the channel setter masks with kRTChatNoChannel (255). Both are used as "clear this bit, keep the rest", but they silently discard every flag above their width.

No current flag is affected — all of them live in the low byte — but any new flag above bit 15 would be cleared on assignment, and above bit 7 whenever channel is set. There's no log output when it happens.

Also changes the channel decode from / 256 to // 256, which was producing a float.


Hoikas (approved):

Hoikas:
This is a pretty straightforward set of bugfixes. I'm inlcined to merge this PR when CI passes regardless of what happens with the other two PRs.

Scripts/Python/plasma/PlasmaKITypes.py
```diff
@@ -209,7 +209,9 @@ def PtDetermineKIMarkerLevel():
 kRTChatLocKeyMsg = 0x80
 
 # flags channel mask
-kRTChatFlagMask = 65535
+# Must cover the full uint32 that pfKIMsg serializes. ChatFlags uses this as
+# "flags &= mask ^ bit", so a narrower mask silently clears every flag above it.
+kRTChatFlagMask = 0xFFFFFFFF
```
dgelessus: I wonder, what's the point of having `kRTChatFlagMask` if it includes all bits? If we don't want to mask anything away, it would be better to remove the mask entirely.

Scripts/Python/plasma/PlasmaKITypes.py
```diff
@@ -209,7 +209,9 @@ def PtDetermineKIMarkerLevel():
...
 kRTChatChannelMask = 65280
 kRTChatNoChannel = 255
```
dgelessus: The `kRTChatNoChannel` mask would need to be widened to `0xffff00ff` to match. But as it's only used in a single place, which was changed above to use `~kRTChatChannelMask` instead, I think we can simply remove `kRTChatNoChannel` as well.

Stone-: dropped kRTChatFlagMask and the now-unused kRTChatNoChannel

PR #1947, opened 2026-09-10
Add D'ni font rendering for KI chat

Stone-:
Depends on #1945 and #1946.

Adds a /dni message prefix and a kDniFontMsg chat flag so a message renders in a D'ni typeface for both sender and recipients.

On security: the wire carries one bit, not a font name. The face string comes from a local constant, so there's no path from network input to SetFont(). The existing control-character sanitizer is unchanged, so face codes can't be injected into the buffer remotely. Worst case from a hostile client is cosmetic.

Requires Dnifont-*.p2f in the client's dat folder; without them plFontCache falls back to Arial and logs a missing-font message.

Screenshot_2026-08-18_at_6 21 52_PM


Mystler:
Looks very cool. Hopefully we can get #1917 in soon so this can be rebased on top of that.

PR #1948, opened 2026-09-11
Add debug backdoor to xLinkingBookGUIPopup

Mystler:
Using `Python.Backdoor xlinkingbookgui debugall` in the console, timers in 1s offsets will be started to open all age entries in xLinkingBookDefs, display them, and take a screenshot to the client folder. This is a simple helper to facilitate checking the textures if changes to PRPs like BkBookImages are made.

Inspired by discussion and testing for https://github.com/H-uru/moul-assets/pull/289.

### DIFF
--- a/Scripts/Python/xLinkingBookGUIPopup.py
+++ b/Scripts/Python/xLinkingBookGUIPopup.py
@@ -63,8 +63,8 @@ from PlasmaKITypes import *
 from PlasmaNetConstants import *
 import xLinkingBookDefs
 from xPsnlVaultSDL import *
-import time
 import re
+from contextlib import contextmanager
 
 
 # define the attributes that will be entered in max
@@ -1020,10 +1020,8 @@ class xLinkingBookGUIPopup(ptModifier):
             bkIdx = id - kDebugTimerStartIdx
             defs = [x for x in xLinkingBookDefs.xAgeLinkingBooks]
             age = defs[bkIdx]
-            TargetAge.value = age
-            global stringAgeRequested
-            stringAgeRequested = age
-            self.IShowBookNoTreasure()
+            with self.OverrideTargetAge(age):
+                self.IShowBookNoTreasure()
             PtSaveScreenShot(f"{age}.jpg", PtGetDesktopWidth(), PtGetDesktopHeight())
 
     def GetOwnedAgeLink(self, age):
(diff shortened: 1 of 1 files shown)


Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -1013,6 +1014,17 @@ def OnTimer(self,id):
...
+        # Debug Timers for our backdoor
+        elif id >= kDebugTimerStartIdx:
+            self.HideBook(0)
+            bkIdx = id - kDebugTimerStartIdx
+            defs = [x for x in xLinkingBookDefs.xAgeLinkingBooks]
+            age = defs[bkIdx]
+            TargetAge.value = age
+            global stringAgeRequested
+            stringAgeRequested = age
+            self.IShowBookNoTreasure()
```
Hoikas: This will permanently change the target age for the module. That might lead to surprising results if we're in some non-Relto age that uses `TargetAge.value` instead of the bookshelf's `stringAgeRequested`. It might be worth using helper to override the age name. For example:

```python
from contextlib import contextmanager

@contextmanager
def OverrideTargetAge(self, age: str):
    global stringAgeRequested
    prevTarget, prevRequested = TargetAge.value, stringAgeRequested
    TargetAge.value, stringAgeRequested = age, age
    try:
        yield
    finally:
        TargetAge.value, stringAgeRequested = prevTarget, prevRequested
```

Then, this can become


```suggestion
            with self.OverrideTargetAge(age):
                self.IShowBookNoTreasure()
```

Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -1204,4 +1216,8 @@ def LinkToPelletCave(self,spawnPt):
         linkMgr = ptNetLinkingMgr()
         linkMgr.linkToAge(als)
 
-
+    def OnBackdoorMsg(self, target, param):
+        if target == "xlinkingbookgui" and param == "debugall":
+            # Start 1s interval timers to show every linking book and take a screenshot of it
+            for idx, age in enumerate(xLinkingBookDefs.xAgeLinkingBooks):
+                PtAtTimeCallback(self.key, 1 + idx, kDebugTimerStartIdx + idx)
```
Hoikas, line 1223: If the age has multiple linking books, for example Neighborhood or city, this will run for every linking book in the age. Unfortunately, I don't think we can do anything about it aside from note the behavior as a future fixme if/when it ever becomes possible to share state among the different instances of a script. Maybe #1889 can help with that.

Scripts/Python/xLinkingBookGUIPopup.py
Mystler: This looks clean and teaches me another bit of special python fanciness. Awesome. I'll refactor that later.

Mystler:
Updated with the changes, however, I just noticed that when using this in other ages (I tested in spyroom) it can auto-fire linking with the book, which is definitely not intended.

Mystler:
Updated again to deal with various edge cases. Kids, please do not do this at home. This is not a PR to learn from.

The helper did actually reveal multiple interesting edge cases or bugs with linking panels that we may or may not want to deal with in follow up PRs. I will investigate these cases separately.

Hoikas (approved):

Issue #1875, opened 2026-05-19
Feature request: KI command to reset kickables location


TheScarFr:
A KI command to reset the kickables location, at least in the private Ages and in the Neighborhood if we're member of it, would be nice.

dgelessus:
There is already a console command `Physics.ResetKickables` in internal clients. Does that work for you, or do you have a use case where it really needs to be a KI command?

TheScarFr:
On MOUL(a) with the default client, to unstuck Relto's firemarbles/logs/stones/crystals, and to unsick Minkata's soccerball and Eder Gira's baskets, for example.

DamnBriggsy:
There is the potential for abuse in shared spaces in opening this feature up to all clients. Though the stakes are admittedly rather low, there is a nonzero chance that someone could be upset by repeated resets of kickables in neighborhoods, for example. 

If this were to be implemented, it would require more nuance than just adding the existing command to ki chat. I imagine there would be some debate over how exactly this would work.

dpogue:
The other issue is that there's no way to do this through a KI command in an IC-compliant way. The KI is a communication device, it cannot affect the physics of random objects in the world

TheScarFr:
What's the IC thing for the `/[start/stop/dump]log` or `/[load/save]columns` commands?
The "IC" shouldn't disallow to add things to the gameplay... Don't forget that it's a game.
So, if it isn't done via a KI command, it could be done differently, but I don't see how... A KI command would be the easier/better way, IMO.
Regarding the potential abuse, I agree, it's why I mentioned "at least in the private Ages and in the Neighborhood if we're member of it", but if it's only in the private Ages, it would be already great.

dgelessus:
Both the Jalak controls and text chat are KI functions, so it's logical to have KI commands related to those - I don't see any issue with explaining those IC. Besides, even if there are gameplay features that already break immersion in some way, that doesn't mean that we should just ignore immersion when adding new features.

There's really no way to explain a magic button that teleports all physics objects back to their spawn points, no matter if you make it a KI command or something else. Even if it's a button in the settings and thus not directly part of the game world, it still has very visible unexplainable effects on the game world.

I also worry that this feature could encourage fan ages to have easily breakable physics puzzles, because there's an easy way out by resetting all kickables.

[...]

PR #1950, opened 2026-09-17
Add resetSyncState method to reset Relto physicals

Stone-:
Create Function ResetSyncState: restore each receiver’s physical to its default recipe pose and clear its velocities. This matches the console command Physics.ResetKickables but limits the effect to the current object.  

It is used in psnlYeeshaPageChanges.py when YeeshaPage07 (kickable) or YeeshaPage16 (firemarbles) is disabled, returning the physical to its default pose before suppression so it remains correct if re-enabled.


TheScarFr:
Possibly related: https://github.com/H-uru/Plasma/issues/1875

PR #1951, opened 2026-09-20
Add persistent option to xRandomBoolChange

Mystler:
As it is and was intended, the xRandomBoolChange Python File Mod used by things like the Neighborhood Relto Page or the linking stones in Ae'Gura will simply change the visibility state when the chance roll is a success. When setting randomization to enabled and chance to 100, that means with every person linking into the age the chance roll is a success and the state is flipped from visible to invisible and invisible to visible.

While this is fine, intended, and not touched by this PR, shard admins lack an intended option to say that they always want it to be on. This PR treats a chance value of 255 (maximum BYTE value) as "always on" (or always off in the case that a disabled state was set up to mean it should be visible).

Side note: The Yeesha Page in the Neighborhood actually has a PRP file bug that causes this Python File Mod to trigger twice, which caused an unintended side effect that made having it always on already possible.


Mystler:
Additional thought, I'm noticing that another _theoretical_ way of achieving persistence would be the combination of
Enabled = 1
Chance = 0
Vis = 1 (via globals)

However, the random roll is currently an inclusive [0,100] roll so even at 0 chance, ironically, that would mean there is a 0.99% chance that the visibility state will flip, which would be bad when it happens. I could change the roll range for this approach instead, if it is preferred, however, that change would not be backwards compatible and people using old python scripts may destroy the intended state. This is less of a problem with a 255 chance that, on old scripts, will just retain predictable flip flop behavior.

PR #1952, opened 2026-09-21
Make share seek optional for Linking Books

dpogue:
If none is provided, the sharee should just link immediately rather than stalling waiting for a callback on a behaviour that doesn't exist.


Scripts/Python/xLinkingBookGUIPopup.py
```diff
@@ -313,15 +313,24 @@ def OnNotify(self,state,id,events):
...
-                                #walk to the book and use it
-                                avatar=PtGetLocalAvatar()
-                                avatar.avatar.setReplyKey(self.key)
-                                shareBookSeek.run(avatar)
                                 OffereeWalking = True
                                 avID = PtGetClientIDFromAvatarKey(BookOfferer.getKey())
                                 PtNotifyOffererLinkAccepted(avID)
                                 ClosedBookToShare = 1
-                                PtDebugPrint("xLinkingBookGUIPopup: seeking avatar to use book offered")
+                                if shareBookSeek:
```
Hoikas:
```suggestion
                                if shareBookSeek.value:
```

`shareBookSeek` is always truthy.

Hoikas (changes requested):

Traveler263:
Python File Fix was tested tonight to success. Share-ee avvie was whisked away with a vwroom and fade out, but no hand reaching out for a phantom book or warping to a seek point. (Yippie!)

Hoikas (approved):

